PR #2, opened 2012-02-22
Feature/vumigo 21 conversation closing

hodgestar:
Conversation closing ready for review I think.

### DIFF
--- a/go/conversation/migrations/0012_auto__add_field_conversation_end_date__add_field_conversation_end_time.py
+++ b/go/conversation/migrations/0012_auto__add_field_conversation_end_date__add_field_conversation_end_time.py
@@ -1,0 +1,114 @@
+# encoding: utf-8
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        
+        # Adding field 'Conversation.end_date'
+        db.add_column('conversation_conversation', 'end_date', self.gf('django.db.models.fields.DateField')(null=True), keep_default=False)
+
+        # Adding field 'Conversation.end_time'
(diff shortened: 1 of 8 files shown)


jerith:
General comment: Why do we have separate date and time fields, and what do we do about timezone?

hodgestar:
I realized I still have a little UI work to do for the conversation closing. I need to not display the close conversation button if a conversation has ended. Hopefully that's a quick template fix.

hodgestar:
@jerith I'm not sure why the date and time fields are separate. Maybe ask @smn? I think times in the database should be UTC so I should probably fix that. If it also affects the existing date and time fields I propose we cut a ticket and leave it for a separate branch (since having two different kinds of times around will be confusing for the upcoming demo).

smn:
@hodgestar @jerith date and time fields are separate for historical reasons, happy to join that in a single field. I think the reason was that in order to do that well in the UI I needed to write a Django form widget specifically for that (iirc) and I didn't have the time available.

jerith:
Other than the timezone comment (which applies more generally than this PR), looks good to go.

hodgestar:
@smn @jerith let's leave the date and time joining for another branch. Also, I need a migration for this branch.

smn:
Agreed. Also, the time zone issue is a tricky one as you have to possibly deal with multiple time zones. We're in GMT+2 but want to launch the same campaign in countries with GMT+3 and GMT+5, scheduling that will require some pretty good UI and goes way beyond this PR. In the discussions I've had with Gustav I touched on this as well, decision was to default to UTC for now and let the user make sure he/she schedules things appropriately.

hodgestar:
I think it's ready to land now.

smn:
:+1:

PR #4, opened 2012-03-02
Feature/vumigo 32 factor out tag pool manager

hodgestar:
Tagpools have now been factored out into their own module. Ready for review.


hodgestar:
Actually landing this branch will need some care because someone will have to update the QA instance.

hodgestar:
See also https://github.com/praekelt/vumi/pull/204.

PR #9, opened 2012-05-03
Feature/vumigo 63 fix go app router

smn:
This removes the Django dependencies and uses the vumi.persist stuff.


go/vumitools/api_worker.py
```diff
@@ -85,38 +87,56 @@ def setup_routing(self):
...
+                    account_key)
+                account_submanager = conversation_store.manager
+                conversations = yield batch.backlinks.conversations(
+                                                        account_submanager)
+                if conversations:
+                    if len(conversations) > 1:
+                        conv_keys = [c.key for c in conversations]
+                        log.warning('Multiple conversations found '
+                            'going with most recent: %r' % (conv_keys,))
+                    conversation = sorted(conversations, reverse=True,
```
hodgestar, line 113: Might be better to filter out conversations that have ended first? Ideally we should be able to combine the backlinks search and the end_time index search into one Riak M/R query but that isn't supported by vumi.persist yet. For now just filtering out conversations with end_time != None should do it.

go/vumitools/api_worker.py
```diff
@@ -85,38 +87,56 @@ def setup_routing(self):
...
+                        key=lambda c: c.start_timestamp)[0]
+                    returnValue(conversation)
+                log.error('No conversations found for %r' % (batch,))
```
hodgestar, line 116: Won't both of these error messages get logged if don't find a conversation because there's no else between them?

go/vumitools/api_worker.py
```diff
@@ -85,38 +87,56 @@ def setup_routing(self):
...
+        if tag:
+            conversation = yield self.get_conversation_for_tag(tag)
+            if conversation:
+                metadata = msg['helper_metadata']
+                conv_metadata = metadata.setdefault('conversations', {})
+                conv_metadata.update({
+                    'conversation_key': conversation.key,
+                    'conversation_type': conversation.conversation_type,
+                })
+                conv_type = conv_metadata.get('conversation_type')
```
hodgestar: Maybe 'conv_type = conversation.conversation_type' just below 'if conversation' and then use it in both places?

go/vumitools/tests/test_api_worker.py
```diff
@@ -101,29 +80,55 @@ def setUp(self):
                 'app_2',
             ],
             'conversation_mappings': {
-                'type_1': 'app_1',
-                'type_2': 'app_2',
+                'bulk_message': 'app_1',
+                'survey': 'app_2',
             }
         })
+        self.r_server = FakeRedis()
```
hodgestar: Since the GoApplicationRouter uses get_redis(config) you can avoid some of this monkey patching by putting "'redis_cls': lambda: self.r_server" into the config you pass to the dispatcher and setting self.r_server = FakeRedis() at the top of setUp.

go/vumitools/tests/test_api_worker.py
```diff
@@ -101,29 +80,55 @@ def setUp(self):
...
 
     @inlineCallbacks
     def test_tag_retrieval_and_dispatching(self):
         msg = self.mkmsg_in(transport_type='xmpp',
-                                transport_name='xmpp_transport')
-        TaggingMiddleware.add_tag_to_msg(msg, ('xmpp', 'test1@xmpp.org'))
-        yield self.dispatch(msg, self.transport_name)
+                                transport_name=self.transport_name)
+
+        tag = self.tpm.acquire_tag('xmpp')
```
hodgestar: You can probably actually just skip tag acquisition and declaration and make up a tag name here since the message store doesn't care about the tag pool.

go/vumitools/api_worker.py
smn, line 113: Good one, thanks.

go/vumitools/api_worker.py
smn, line 116: Thanks.

jerith:
Looks reasonable.

PR #11, opened 2012-05-21
Feature/vumigo 76 update dynamic fields to dict like interface

hodgestar:
The littlest PR is ready for review.

### DIFF
--- a/config/poll.yaml
+++ b/config/poll.yaml
@@ -4,7 +4,7 @@ batch_size: 2
 
 message_store:
   store_prefix: vumigo.
-riak:
+riak_manager:
   bucket_prefix: vumigo.
 vxpolls:
-  prefix: vumigo.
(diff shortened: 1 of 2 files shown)


config/poll.yaml
```diff
@@ -4,7 +4,7 @@ batch_size: 2
 
 message_store:
   store_prefix: vumigo.
-riak:
+riak_manager:
   bucket_prefix: vumigo.
 vxpolls:
-  prefix: vumigo.
\ No newline at end of file
+  prefix: poll_manager
```
hodgestar, line 10: What are these changes doing here? They should be in develop.

hodgestar, line 10: These changes have been pushed into develop and merged into the branch so the diff looks sane now.

smn:
:+1:

PR #12, opened 2012-05-24
app permissions management command and UI changes

smn:


go/base/management/commands/go_manage_application.py
```diff
@@ -0,0 +1,92 @@
...
+                    '--disable, not both.')
+
+            user = User.objects.get(username=email_address)
+            account = user.get_profile().get_user_account()
+            existing_applications = [permission.application for permission in
+                                        account.applications.get_all()]
+
+            if disable:
+                if application_module in existing_applications:
+                    permission = [permission for permission in
```
hodgestar, line 63: Maybe just call disable_application() on all of them rather than assuming there's only one?

go/templates/base.html
```diff
@@ -24,68 +25,69 @@
 </head>
 <body>
 
-	<!--[if lt IE 7]><p class=chromeframe>Your browser is <em>ancient!</em> <a href="http://browsehappy.com/">Upgrade to a different browser</a> or <a href="http://www.google.com/chromeframe/?redirect=true">install Google Chrome Frame</a> to experience this site.</p><![endif]-->
```
hodgestar, line 27: Is all the reindenting here intentional? If so I'm happy for it to stay.

go/vumitools/api.py
```diff
@@ -364,6 +366,18 @@ def tagpools(self):
                          for pool in pool_names)
         returnValue(TagpoolSet(pool_data))
 
+    @Manager.calls_manager
+    def applications(self):
+        account_store = self.api.account_store
```
hodgestar, line 371: This can probably all be short-circuited into a single Riak call. I just need to look up how quickly.

go/base/management/commands/go_manage_application.py
```diff
@@ -0,0 +1,92 @@
...
+
+        except User.DoesNotExist, e:
+            raise CommandError(e)
+
+    def disable_application(self, app_permission, account):
+        account.applications.remove(app_permission)
+        account.save()
+
+    def enable_application(self, user, account, application_module):
+        user_api = vumi_api_for_user(user)
```
hodgestar, line 84: All this hopping around to get the user_api should be avoidable since we have the account object. There is similar hopping around in the tag pool management command but that should be avoidable too. Maybe a method on the account object is the way to go (I'm not entirely sure).

go/vumitools/api.py
hodgestar, line 371: Actually I'm wrong about this but only because .backlinks doesn't support it yet (so ignore the above comment for now :).

hodgestar:
I'm :+1: on landing. Other than the indentation question, all of my comments are more just thoughts about how we can do model stuff better in future.

jerith:
Wouldn't it be easier to have separate "enable app" and "disable app" commands that called into some backend logic rather than building everything into one big management command?

I assume we're going to want a web UI on this stuff at some point...

go/templates/base.html
smn, line 27: It was a mess of tabs & spaces making template editing painful.

smn:
Agreed on the backend logic, we need that - but I need this feature before I have the time to build the backend logic. There's a ticket for that though -- http://praekelt.onjira.com/browse/VUMIGO-64

jerith:
I'm +0 on the technical debt of not having the underlying logic in the backend, but if the need this now I'm not against landing it.

PR #13, opened 2012-05-24
Feature/vumigo 50 pick shortcodes

hodgestar:
This branch isn't complete yet but I wouldn't mind a sanity check of the approach being taken before continuing.

Outstanding things needed are:
- Tests for bulk_message:new and survey:new using a tagpool that has user_selects_tag set in its metadata.
- Some manual testing of the UI by humans to make sure they like it.
- Some thought about where this bit of UI is going in the long term and whether that affects the current ticket (i.e. are people meant to be able to select multiple delivery classes and/or multiple tag pools within each delivery class? will tag pools eventually be autoselected?)

### DIFF
--- a/config/tagpools.yaml
+++ b/config/tagpools.yaml
@@ -42,3 +42,21 @@ pools:
       server_initiated: true
       msg_options:
         transport_name: gtalk_vumigo
+  mama1:
+    tags:
+      - '*120*646*MAMA1'
+    metadata:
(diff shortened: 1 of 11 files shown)


hodgestar:
I'm much happier with the state of things now. I think the UI is in its final form (modulo reviewer comments). Now to just write the promised tests.

smn:
:+1: on landing and have some :cake: with :coffee:

PR #14, opened 2012-05-25
Cache people calls.

hodgestar:
I don't know whether this should go in but I'm starting the PR so we don't forget about the idea. Feel free to comment anyway.


hodgestar:
I'm closing this because @smn fixed it by calling .people() less often from the templates.

PR #16, opened 2012-05-29
some optimizations to the conversations page.

smn:
- remove the total from the delivery report stuff
- removed some context manager calls what weren't necessary


hodgestar:
:+1:

PR #20, opened 2012-06-07
make CSV upload two step & allow for free-form CSV

smn:
does somewhat intelligent guessing about whether or not a CSV file has a
header row. If it does it respects it, if it doesn't the user will need
to specify what's what.

Added a `extra` Dynamic field to Contact model to allow us to store the
keys that aren't specified explicitly on the model. The Contact page now
shows these fields are read-only text input-fields. That'll change
later.

### DIFF
--- a/go/base/templates/base/includes/messages.html
+++ b/go/base/templates/base/includes/messages.html
@@ -7,7 +7,7 @@
             show
             {% endspaceless %}">
             <a class="close" data-dismiss="alert">×</a>
-            {{message}}
+            {{message|safe}}
         </div>
         {% if error_form.errors %}
         <div class="control-group error">
--- a/go/base/templates/base/includes/messages.html
+++ b/go/base/templates/base/includes/messages.html
@@ -7,7 +7,7 @@
             show
             {% endspaceless %}">
             <a class="close" data-dismiss="alert">×</a>
-            {{message|safe}}
+            {{message}}
         </div>
         {% if error_form.errors %}
         <div class="control-group error">
--- a/go/contacts/templates/contacts/group.html
+++ b/go/contacts/templates/contacts/group.html
@@ -94,6 +94,7 @@
         </div>
     </div>
     {% include "base/includes/pagination.html" %}
+
 {% endblock %}
 
 {% block extra_js %}
@@ -101,6 +102,11 @@
       $(document).ready(function(){
           $('#frm-new-group').validate();
           $('#frm-upload-contacts').validate();
+
+        {% if csv_data_row %}
+            $('#recMatchFrm').modal('show');
+        {% endif %}
+
       });
     </script>
     <link type="text/css" href="{{ STATIC_URL }}css/smoothness/jquery-ui-1.8.11.custom.css" rel="stylesheet" />
--- a/go/contacts/templates/contacts/group.html
+++ b/go/contacts/templates/contacts/group.html
@@ -102,11 +102,7 @@
       $(document).ready(function(){
           $('#frm-new-group').validate();
           $('#frm-upload-contacts').validate();
-
-        {% if csv_data_row %}
-            $('#recMatchFrm').modal('show');
-        {% endif %}
-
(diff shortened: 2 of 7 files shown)


go/base/templates/base/includes/messages.html
```diff
@@ -7,7 +7,7 @@
             show
             {% endspaceless %}">
             <a class="close" data-dismiss="alert">×</a>
-            {{message}}
+            {{message|safe}}
```
hodgestar: What if we have some error messages that contain data from users?

smn: Right, will remove.

go/contacts/templates/contacts/group.html
```diff
@@ -101,6 +102,11 @@
       $(document).ready(function(){
           $('#frm-new-group').validate();
           $('#frm-upload-contacts').validate();
+
+        {% if csv_data_row %}
```
hodgestar: Does this need an "csv_data_headers and" in addition to "csv_data_row"? Since jQuery returns an empty object set if nothing matches '#recMatchFrm' do we need the 'if' at all?

smn: Right, not needed & removed.

go/contacts/views.py
```diff
@@ -100,6 +106,132 @@ def groups(request):
...
+        'dob': 'Date of Birth',
+        'facebook_id': 'Facebook ID',
+        'twitter_handle': 'Twitter handle',
+        'email_address': 'Email address',
+    }
+
+    if _is_header_row(first_row):
+        for column in first_row:
+            sample_row = SortedDict(zip(second_row, first_row))
+            default_headers.setdefault(column, column)
```
hodgestar, line 132: I guess this should be outside the for loop? I also expected first_row to be the first argument to zip and second_row the second?

go/contacts/views.py
```diff
@@ -100,6 +106,132 @@ def groups(request):
...
+        return True, default_headers, sample_row
```
hodgestar, line 133: This means the user's headers never override anything in default_headers?

go/contacts/views.py
```diff
@@ -100,6 +106,132 @@ def groups(request):
...
+    return (False, default_headers,
+        SortedDict([(column, None) for column in first_row]))
+
+
+def _get_file_hints(file_object):
+    # Save the file object temporarily so we can present
+    # some UI to help the user figure out which columns are
+    # what of what type.
+    content_file = ContentFile(file_object.read())
+    temp_file_name = uuid.uuid4().hex
```
hodgestar, line 142: It looks like this reads the entire file into memory? Not sure if this is an issue. I guess file_object.chunks() could be used to avoid this. Maybe this just needs a TODO for later?

go/contacts/views.py
smn, line 132: Further down below, for columns that don't have a header type I use the value `None` and the value of the column as the `key`. If I swamped it around everything would be keyed under `None` and would overwrite each other. Did it like this to keep the two consistent.

go/contacts/views.py
smn, line 133: It means that if they have a column called 'msisdn' we still show 'Contact Number' instead of 'msisdn'. If they have a column called 'opted-in' (for which we have no display-value then we'll just display 'opted-in' as is.

go/contacts/views.py
smn, line 142: hm, good one. Let me see if Django will let me work with a normal file object instead.

smn, line 142: Works & fixed.

hodgestar:
I think I'm :+1: on this landing. I'm a bit sleepy and I haven't reviewed the template stuff that well and the `group` view has probably become a bit complicated but it should be easy enough to split up later.

smn:
Agreed on the `group` view being quite complicated. I'd prefer to leave things as they are now until we've sorted out the celery side of things, I'd like to defer some of that complexity to a Celery task.

PR #21, opened 2012-06-08
Write results of surveys to a contact as dynamic attributes

smn:
Does this on session end. When a new message arrives, any previous
dynamic attributes are saved to the participant before the poll starts
working on that state. Allows for things like global per-account
variables for contacts for things like opt-ins. Does not protect the
user from stomping on previously set variables.

### DIFF
--- a/go/apps/surveys/vumi_app.py
+++ b/go/apps/surveys/vumi_app.py
@@ -1,8 +1,9 @@
 # -*- test-case-name: go.apps.surveys.tests.test_vumi_app -*-
 
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import inlineCallbacks, returnValue
 from go.vumitools.api import VumiApiCommand, get_redis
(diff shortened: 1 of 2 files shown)


go/apps/surveys/vumi_app.py
```diff
@@ -47,11 +48,42 @@ def teardown_application(self):
             yield self.control_consumer.stop()
             self.control_consumer = None
 
+    @inlineCallbacks
+    def get_contact_for_message(self, message):
+        helper_metadata = message.get('helper_metadata', {})
```
hodgestar, line 53: Should we add helper methods for getting and setting the conversation key and account key? We're duplicating this code in a bunch of places already.

hodgestar:
I think we probably need to restrict which labels go into and out of the extras contact field somehow (either by convention or by explicitly marking them) but otherwise I think the branch looks fairly sane.

PR #22, opened 2012-06-12
Allow for contacts to be deleted.

smn:
Added the delete code to the view and covers it with a test. Also adds
404.html and 500.html templates to prevent Django falling over while
already falling over.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from go.vumitools.api import VumiUserApi
 
 
-@task
+@task(ignore_result=True)
 def delete_group(account_key, group_key):
     api = VumiUserApi(account_key, settings.VUMI_API_CONFIG)
     contact_store = api.contact_store
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -7,6 +7,12 @@ from go.vumitools.api import VumiUserApi
 
 @task(ignore_result=True)
 def delete_group(account_key, group_key):
+    # NOTE: There is a small chance that this can break when running in
+    #       production if the load is high and the queues have backed up.
+    #       What could happen is that while contacts are being removed from
+    #       the group, new contacts could have been added before the group
+    #       has been deleted. If this happens those contacts will have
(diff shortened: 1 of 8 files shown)


smn:
Accidentally also did the work for VUMIGO-103 in this branch.

go/contacts/tasks.py
```diff
@@ -0,0 +1,16 @@
+from celery.task import task
+
+from django.conf import settings
+
+from go.vumitools.api import VumiUserApi
+
+
+@task
```
hodgestar: I think this needs an ignore result flag so we don't build up result queues.

hodgestar:
Hooray for 404 and 500 templates. :)

hodgestar:
:+1: on landing after the ignore result flag has been added.

go/contacts/tasks.py
```diff
@@ -0,0 +1,16 @@
...
+def delete_group(account_key, group_key):
+    api = VumiUserApi(account_key, settings.VUMI_API_CONFIG)
+    contact_store = api.contact_store
+    group = contact_store.get_group(group_key)
+    for contact in group.backlinks.contacts():
+        contact.groups.remove(group)
+        contact.save()
+    group.delete()
```
jerith, line 22: If this takes a long time, it's possible for new contacts to be added to the group before the group gets deleted. Will stuff explode if this happens?

hodgestar, line 22: It might be possible to move the group.delete() before the contact deletion and then do group.backlinks.contacts() afterwards. That leaves a brief interval where contacts have links to a non-existent group. Not sure whether that's likely to be an issue.

jerith, line 22: That guarantees an interval during which there are contacts in a deleted group. I'd rather do a second group removal pass afterwards if we go that route. My concern is that a contact in a non-existent group could break other stuff. We need to make sure we don't have that assumption baked into anything.

smn, line 22: We possibly introduced that assumption by chosing a db with an eventual consistency model. As a result, should we perhaps resolve this at read time?

PR #24, opened 2012-06-19
Opt-out middleware and GoAppRouter refactor

smn:
A lot of the functionality provided by the GoApplicationRouter has now
been moved to middlewares so they can be moved around our stack easier.

An opt-out middleware has been added and the GoAppRouter routes to the
optout_transport if it detects a message with an optout flag set.

I'm pretty sure there's something I can do better in the
LookupAccountMiddleware when given a batch_id but I'm not seeing it at
the moment. Both the LookupAccountMiddleware and the
LookupBatchMiddleware first get the current_tag stuff which seems
redundant.

### DIFF
--- a/config/application_message_dispatcher.yaml
+++ b/config/application_message_dispatcher.yaml
@@ -1,13 +1,42 @@
-message_store:
+message_store: &MESSAGE_STORE
     store_prefix: vumigo.
+
 transport_names:
   - vumigo_router
 exposed_names:
   - bulk_message_transport
   - survey_transport
   - multi_survey_transport
+  - account_opt_out_transport
+
 router_class: go.vumitools.api_worker.GoApplicationRouter
 conversation_mappings:
     survey: survey_transport
     multi_survey: multi_survey_transport
     bulk_message: bulk_message_transport
+
+# NOTE: the order is important
+middleware:
+  - tagging_middleware: vumi.middleware.TaggingMiddleware
+  - account_middleware: go.vumitools.middleware.LookupAccountMiddleware
+  - batch_middleware: go.vumitools.middleware.LookupBatchMiddleware
+  - conversation_middleware: go.vumitools.middleware.LookupConversationMiddleware
+  - debit_middleware: go.vumitools.middleware.DebitAccountMiddleware
+  - opt_out_middleware: go.vumitools.middleware.OptOutMiddleware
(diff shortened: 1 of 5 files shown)


config/application_message_dispatcher.yaml
```diff
@@ -1,13 +1,47 @@
-message_store:
+message_store: &MESSAGE_STORE
```
hodgestar, line 1: Funky use of YAML references. :)

go/vumitools/middleware.py
```diff
@@ -42,6 +50,215 @@ class InsufficientCredit(DebitAccountError):
...
+    this from the message tag's info.
+
+    *NOTE*  This requires the `TaggingMiddleware` to be configured and placed
+            before this middleware for this to work as it expects certain
+            values to be set in the `helper_metadata`
+    """
+
+    @inlineCallbacks
+    def find_account_key_for_message(self, message):
+        # NOTE: there is probably a better way of doing this when given a
```
hodgestar, line 122: This is probably the best that one can do right now because the batch model stores the tags as a list of tag values and I don't think there's an index on that so one has to get the record for the tag first and get the batch id from there.

hodgestar, line 122: If you already have the batch key from LookupBatchMiddleware, you can just use that value of the batch id though (yield self.message_store.get_batch(batch_id) instead of yield current_tag.current_batch.get() -- you can get the batch key from current_tag using current_tag.current_batch.key if needed).

go/vumitools/middleware.py
```diff
@@ -42,6 +50,215 @@ class InsufficientCredit(DebitAccountError):
...
+        yield self.add_metadata_to_message(message)
+        returnValue(message)
+
+    @inlineCallbacks
+    def handle_event(self, event, endpoint):
+        yield self.add_metadata_to_message(event)
+        returnValue(event)
+
+    @inlineCallbacks
+    def handle_outbound(self, message, endpoint):
```
hodgestar, line 105: I think currently this will do a lot of unnecessary lookup on outbound messages (which I think have a tag given to them by the application worker).

hodgestar:
This is looking pretty cool. Some of my comments might be tricky to deal with in a hurry. I'll try think up solutions in the morning.

PR #25, opened 2012-06-20
Display opt-out status on contact page.

smn:
Implemented via a `contact_has_opted_out(contact)` function on the
ContactStore.

### DIFF
--- a/go/contacts/views.py
+++ b/go/contacts/views.py
@@ -423,6 +423,9 @@ def person(request, person_key):
             'groups': [group.key for group in contact.groups.get_all()],
         })
 
+    if contact_store.contact_has_opted_out(contact):
+        messages.error(request, 'This contact has opted out.')
+
     return render(request, 'contacts/person.html', {
         'contact': contact,
         'contact_extra_items': contact.extra.items(),
--- a/go/vumitools/contact.py
+++ b/go/vumitools/contact.py
@@ -136,6 +136,22 @@ class ContactStore(PerAccountStore):
         returnValue(sorted(groups, key=lambda group: group.name))
 
     @Manager.calls_manager
+    def contact_has_opted_out(self, contact):
+        # FIXME:    Need to import this to make sure the backlinks are created
+        #           even though it isn't used directly.
+        from go.vumitools.opt_out import OptOutStore
+
+        # FIXME:    opt-outs are currently had coded to only work for msisdns
+        if not contact.msisdn:
+            return
+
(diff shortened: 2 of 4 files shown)


go/contacts/views.py
```diff
@@ -423,6 +423,9 @@ def person(request, person_key):
             'groups': [group.key for group in contact.groups.get_all()],
         })
 
+    if contact_store.contact_has_opted_out(contact):
+        messages.error(request, 'This contact has opted out.')
+
```
smn, line 428: `messages.error` isn't exactly what we want but it's what we currently have styling for. It displays a red alert bar across the type, same as our confirmation messages but then red.

go/vumitools/contact.py
```diff
@@ -136,6 +136,22 @@ def list_groups(self):
...
+        # FIXME:    Need to import this to make sure the backlinks are created
+        #           even though it isn't used directly.
+        from go.vumitools.opt_out import OptOutStore
+
+        # FIXME:    opt-outs are currently had coded to only work for msisdns
+        if not contact.msisdn:
+            return
+
+        user_account = yield self.get_user_account()
+        optouts = yield user_account.backlinks.optouts(manager=contact.manager)
```
dmaclay: if you're checking just one contact, surely it makes more sense to just retrieve one opt_out like opt_out_store.get_opt_out("msisdn", "12345")

go/vumitools/contact.py
```diff
@@ -10,6 +10,7 @@
                                     Dynamic)
 
 from go.vumitools.account import UserAccount, PerAccountStore
+from go.vumitools.opt_out import OptOutStore
```
dmaclay, line 13: perhaps this should be contact_has_opted_out(self, contact, addr_type="msisdn"):
to allow for future specification of conext to opt out ?

go/vumitools/contact.py
```diff
@@ -136,6 +137,17 @@ def list_groups(self):
         returnValue(sorted(groups, key=lambda group: group.name))
```
dmaclay, line 138: and perhaps something like if nor contact.get(addr_type)

jerith, line 138: We probably need to do this per address rather than per contact. I may be opted out on my phone, but opted in on gtalk or something.

smn, line 138: I'm inclined to think of opt-ing out as and end to all communications, not an opting out of a specific transport. I'd rather leave it as is now until we figure out how this should actually work.

PR #26, opened 2012-06-21
VUMIGO-113: Opt out of nothing

jerith:


hodgestar:
:+1:

dmaclay:
cool +1 (i had the benefit of an in person explanation complete with hand waving ;)

go/apps/opt_out/vumi_app.py
```diff
@@ -68,6 +68,12 @@ def consume_user_message(self, message):
         helper_metadata = message.get('helper_metadata', {})
         go_metadata = helper_metadata.get('go', {})
         account_key = go_metadata.get('user_account', None)
+
+        if account_key is None:
+            # We don't have an account to opt out of.
+            # Therefore, just throw the opt-out on the floor.
+            return
+
         opt_out_store = OptOutStore(self.manager, account_key)
```
smn, line 81: I would prefer if we could provide a configurable reply to send back, helps people differentiate between 'something went wrong with the opt-out' and 'the message just didn't arrive'.

jerith, line 81: I thought about that. The configuration would have to be global, since there's no account to tie it to.

I went with dropping it on the floor, because that's standard behaviour for any message to an endpoint that has no application attached to it.

smn:
:+1:

PR #27, opened 2012-06-22
Feature/vumigo 110 opt out worker that creates or updates a contact with an opted out tag

dmaclay:


config/application_message_dispatcher.yaml
```diff
@@ -19,6 +19,7 @@ conversation_mappings:
     survey: survey_transport
     multi_survey: multi_survey_transport
     bulk_message: bulk_message_transport
+    opt_out: opt_out_transport
 
 # NOTE: the order is important
 middleware:
```
smn, line 24: This stuff is managed by the praekelt-machines repository & Puppet.

config/pcm_transport.yaml
```diff
@@ -0,0 +1,33 @@
...
+    tagname_pattern: '.*'
+    msg_template: {}
+
+normalize_msisdns_mw:
+  country_code: '27'
+
+storing_mw:
+  riak:
+    bucket_prefix: vumigo.
+
```
smn: This stuff is managed by the praekelt-machines repository & Puppet.

config/tagpools.yaml
```diff
@@ -60,3 +60,14 @@ pools:
...
+    tags:
+      - '0711667783'
+    metadata:
+      display_name: "Cell Life PCM"
+      delivery_class: http_api
+      transport_type: http_api
+      client_initiated: true
+      user_selects_tag: true
+      msg_options:
+        transport_name: pcm_transport
```
smn: also managed by the praekelt-machines repository & Puppet.

config/vumigo_router.yaml
```diff
@@ -25,3 +26,5 @@ route_mappings:
     - vumigo_router
   afropinions_smpp_transport:
     - vumigo_router
+  pcm_transport:
+    - vumigo_router
```
smn: also praekelt-machines

smn:
I'm a bit worried that we're doing a little too much shoe-horning this to make an opt-out fit into a conversation, gut feel is that we're missing something in Vumi Go on a conceptual level. Do we need to start exposing UI to applications for things that aren't necessarily related to a specific conversation or even the conversation model?

jerith:
This definitely needs to be an account-level thing rather than a conversation (at the UI level, at least). I'm happy to land this now so we can at least use it and figure out what the account-level stuff needs to look like later, for "very soon" values of "later".

jerith:
I'm :+1: on the code and templates, :-1: on the config stuff living in this repo. We need to do something about the config situation more generally as a matter of urgency, though.

dmaclay:
I'm agreed on this should be account level - but all message routing currently requires a conversation

smn:
Can we find a way of making this more account level and less conversation-level? Or at least make a start at it so we have something to work off? My gut feeling is that when we land this it'll stay longer than we'd all like it to. If there are other tickets that've been created to address this issue then that's fine, as long we're working on a clear strategy to properly solve the problem that this feature highlights.

config/haproxy.cfg
```diff
@@ -38,4 +40,7 @@ backend http-backend
     server http_backend4 127.0.0.1:8074 weight 1 check maxconn 500
 
 backend safaricom_ussd
-    server safaricom_ussd0 127.0.0.1:8075 weight 1 check maxconn 200
\ No newline at end of file
+    server safaricom_ussd0 127.0.0.1:8075 weight 1 check maxconn 200
+
+backend pcm
+    server pmc0 127.0.0.1:8076 weight 1 check maxconn 200
```
smn: tyop

smn:
While this is still fresh can we start creating tickets that address the problems that this PR highlights?

PR #28, opened 2012-06-25
Feature/vumigo 111 opt out tech debt payment plan

jerith:


go/contacts/views.py
```diff
@@ -179,10 +179,9 @@ def _has_uncompleted_csv_import(request):
 
 
 def _import_csv_file(group, csv_path, field_names, has_header):
-    full_path = os.path.join(settings.MEDIA_ROOT, csv_path)
     # open in Universal mode to allow us to reed files with Windows,
     # MacOS9 & Unix line-endings
-    csv_file = open(full_path, 'rU')
+    csv_file = default_storage.open(csv_path, 'rU')
```
hodgestar, line 184: This seems sensible but doesn't appear related to the issue title? Just checking it's meant to be here.

jerith, line 184: I cleaned up a couple of unrelated (minor) things while I was in there.

go/vumitools/api_worker.py
```diff
@@ -69,6 +73,163 @@ def consume_control_command(self, cmd):
...
+        if tag_info:
+            batch = yield tag_info.current_batch.get()
+            self._store_objects['batch'] = batch
+            self._go_metadata['batch_key'] = batch.key
+            returnValue(batch)
+
+    @inlineCallbacks
+    def get_batch_key(self):
+        if 'batch_key' not in self._go_metadata:
+            yield self._find_batch()
```
hodgestar, line 147: Doesn't _find_batch() return the batch and not the key?

jerith, line 147: Yes, but we throw away the return value and rely on the side effect of `_find_batch()` putting the batch key in the metadata if it can find one. I'll add a comment.

hodgestar:
I'm :+1: on this landing (modulo existing comments) but I think @dmaclay and @smn need to look too just to make sure that everyone is onboard with the approach (i.e. converting middleware to helper classes that are re-usable elsewhere).

My one concern is that the current version of this approach looks a bit like ConversationWrapper which has bitten us a lot (because we end up with wrapped and unwrapper objects everywhere and dealing with that is a pain).

jerith:
This is quite different from `ConversationWrapper` in that it's providing higher-level access to metadata on a message rather than presenting a message interface with extra functionality.

hodgestar:
I think ConversationWrapper type problems still exist -- there are two levels of access to something and in various places you won't always have the one you want. This doesn't pretend to proxy for the message though, which helps.

Another point -- we need to support getting and setting metadata both on message objects and on message _payloads_. There are various places where we have incomplete message payloads (i.e. dicts with some message values in them) than nevertheless need to have account keys and such set on them and later retrieved.

jerith:
The `ConversationWrapper`-type problems still exist only to the extent that we need to operate on things that are external to the message object. I don't see any way around this that doesn't involve putting all that stuff in the message object all the time.

We do need to support setting the metadata and operating on payloads, but that's out of scope for this branch.

smn:
:+1: thanks for cleaning this up

PR #32, opened 2012-07-11
VUMIGO-122: txredising all the things

jerith:
This is necessary for us to be able to land praekelt/vumi#281

Currently, everything vxpolls-related still uses sync-redis.


jerith:
@hodgestar: Can we revert 26cca12? Most of the middlewares are already gone in #28, which just needs config changes before it can land.

Apart from that, I think we're in a good position to merge both this and praekelt/vumi#281 tomorrow morning and _finally_ make the big divergence go away. The vxpolls stuff will have to be fixed as well, but that's pretty isolated and can happen in its own branch when the vxpolls code supports txredis.

go/apps/multi_surveys/vumi_app.py
```diff
@@ -1,15 +1,26 @@
...
 from vumi.persist.txriak_manager import TxRiakManager
+from vumi.persist.txredis_manager import TxRedisManager
 from vumi.message import TransportUserMessage
 from vumi import log
 
+from go.vumitools.api import VumiApiCommand
+from go.vumitools.conversation import ConversationStore
+
+
+def hacky_hack_hack(config):
```
hodgestar, line 13: I guess this needs a TODO so that it'll definitely be picked up by later greps.

go/base/tests/utils.py
```diff
@@ -54,12 +55,20 @@ def get_riak_manager(self, config=None):
...
             # We don't purge here, because fixtures put stuff in riak.
 
+    def set_up_redis(self):
+        self.redis = RedisManager.from_config('FAKE_REDIS')
+        vumi_config = settings.VUMI_API_CONFIG.copy()
+        vumi_config['redis'] = self.redis._client
+        self.patch_settings(VUMI_API_CONFIG=vumi_config)
+
     def tearDown(self):
+        self.redis._close()
```
hodgestar: I need to look at why we're calling an underscored method here and in similar places elsewhere.

jerith: Probably because "close" is already a redis command or something. This is a foible of the redis manager stuff, and I don't recall the details offhand.

go/apps/multi_surveys/vumi_app.py
hodgestar, line 13: Actually, there are probably enough related TODOs already that this won't be missed.

go/base/tests/utils.py
hodgestar: It doesn't look like 'close' is a command on the redis managers so we could just implement it on the base manager as a call to _close (or rename _close to close).

hodgestar:
Other than the note about ._close(), :+1:.

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -112,6 +68,7 @@ def test_start(self):
         [batch_id] = conversation.batches.keys()
 
         # check commands made it through to the dispatcher and the vumi_app
+        # yield self._amqp.wait_messages('vumi', 'vumi.api', 1)
```
smn, line 69: Why is this commented?

go/apps/surveys/views.py
```diff
@@ -16,7 +17,7 @@
 from vxpolls.manager import PollManager
 
 
-redis = redis.Redis(**settings.VXPOLLS_REDIS_CONFIG)
+redis = RedisManager.from_config(settings.VXPOLLS_REDIS_CONFIG)
```
smn, line 20: We're going to need to add the `settings.VXPOLLS_PREFIX` to this when `vxpolls` is fixed to use the `RedisManager`. Also, it currently uses `:` as a separator and we have live data in Redis for which we need to rename keys. Left a comment in praekelt/vxpolls#20 that addresses this.

go/apps/bulk_message/tests/test_vumi_app.py
jerith, line 69: Because I was fiddling with a bunch of stuff to see if it was actually necessary and forgot to remove it completely when it turned out not to be.

go/apps/surveys/views.py
jerith, line 20: We could do it with a `sub_manager()` call if necessary. There's still a bunch of work that needs to be done around that stuff before it's where I'd like it to be.

PR #33, opened 2012-07-24
Feature/vumigo 120 122 one more try

dmaclay:
this replaces the pull request for feature/VUMIGO-120-events


go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -129,3 +129,43 @@ def test_close_session(self):
...
+            'from_addr': '666666',
+            'transport_type': 'sphex',
+            "helper_metadata": {
+                "go": {
+                    "user_account": user_account_key
+                },
+                'tag': {
+                    'tag': ['pool', 'tag1']
+                },
+                'transport_type': 'sphex'
```
hodgestar: Is there really meant to be a transport_type in helper_metadata?

go/apps/multi_surveys/vumi_app.py
```diff
@@ -31,6 +31,7 @@ class MamaPollApplication(MultiPollApplication):
 class MultiSurveyApplication(MamaPollApplication, GoApplicationMixin):
 
     worker_name = None
+    SEND_TO_TAGS = frozenset(['default'])
```
hodgestar: This looks like it's accidentally left over from

```
commit cd9a7f1d16ebe36ace51cdb6a4693678e69844cc
Author: dmaclay <dmaclay@gmail.com>
Date:   Wed Jul 11 01:30:10 2012 +0200

    accept send message commands on multi survey - mostly for testing
```

And I think it can just be dropped.

go/vumitools/api_worker.py
```diff
@@ -231,6 +232,78 @@ def get_conversation_info(self):
                      self._go_metadata['conversation_type']))
 
 
+class EventDispatcher(ApplicationWorker):
+    """
+    An application worker that forwards event arriving on the Vumi Api Event
+    queue to the relevant handlers.
+
+    FIXME: The configuration is currently static.
```
hodgestar, line 244: Is this FIXME still relevant? There is configuration attached to the user account -- is that non-static enough?

go/vumitools/api_worker.py
```diff
@@ -231,6 +232,78 @@ def get_conversation_info(self):
...
+        A mapping from handler name to fully-qualified class name.
+    """
+
+    def validate_config(self):
+        self.api_routing_config = VumiApiEvent.default_routing_config()
+        self.api_routing_config.update(self.config.get('api_routing', {}))
+        self.api_consumer = None
+        self.handler_config = self.config.get('event_handlers', {})
+        self.account_handler_configs = self.config.get(
+            'account_handler_configs', {})
```
hodgestar, line 266: Should this still be here now that the user accounts have account handler config? If so, it should probably be documented in the configuration parameters.

hodgestar, line 266: Could someone add a comment with a description of what's meant to be in the event_handler_config dictionary somewhere? I was kind of thrown by the `tuple(k)` down below.

go/vumitools/api_worker.py
```diff
@@ -231,6 +232,78 @@ def get_conversation_info(self):
...
+
+    @inlineCallbacks
+    def setup_application(self):
+        self.handlers = {}
+        setup_deferreds = []
+        for name, handler_class in self.handler_config.items():
+            cls = load_class_by_string(handler_class)
+            self.handlers[name] = cls(self, self.config.get(name, {}))
+            yield setup_deferreds.append(self.handlers[name].setup_handler())
```
hodgestar: This `yield` looks like it doesn't do anything because `.append` returns `None`. Maybe there just needs to be a yield on a DeferredList after the loop or instead?

go/vumitools/api_worker.py
```diff
@@ -231,6 +232,78 @@ def get_conversation_info(self):
...
+
+        self.api_command_publisher = yield self.publish_to('vumi.api')
```
hodgestar, line 276: Does the dispatcher actually publish commands or can this be dropped?

go/vumitools/api_worker.py
```diff
@@ -231,6 +232,78 @@ def get_conversation_info(self):
...
+        Dictionary describing where to consume API commands.
+    :type handler_names: list
+    :param event_handlers:
+        A mapping from handler name to fully-qualified class name.
+    """
+
+    def validate_config(self):
+        self.api_routing_config = VumiApiEvent.default_routing_config()
+        self.api_routing_config.update(self.config.get('api_routing', {}))
+        self.api_consumer = None
```
hodgestar, line 263: I think this needs to be `self.api_event_consumer = None` otherwise the check in `teardown_application` isn't useful?

go/vumitools/api_worker.py
```diff
@@ -231,6 +232,78 @@ def get_conversation_info(self):
...
+
+    @inlineCallbacks
+    def teardown_application(self):
+        if self.api_event_consumer:
+            yield self.api_event_consumer.stop()
+            self.api_event_consumer = None
+
+    @inlineCallbacks
+    def get_account_config(self, account_key):
+        if account_key not in self.account_config:
```
hodgestar, line 322: So if the riak configuration changes we need to re-start this worker before the change will take effect?

go/vumitools/api_worker.py
hodgestar, line 276: Ah. I see handlers might want to publish commands. Maybe there should be a method wrapping command publishing or a comment here or something?

go/vumitools/handler.py
```diff
@@ -0,0 +1,107 @@
...
+
+        tag_info = yield user_api.api.tpm.get_metadata(tag[0])
+
+        command_data = event.payload['content']
+        command_data['batch_id'] = batch_id
+        command_data['msg_options'] = {
+                'helper_metadata': {
+                    'go': {'user_account': event.payload['account_key']},
+                    'tag': {'tag': tag},
+                    'transport_type': tag_info['transport_type'],
```
hodgestar: Does `helper_metadata` really need to include the `transport_type`?

go/vumitools/handler.py
```diff
@@ -0,0 +1,107 @@
...
+                },
+                'transport_type': tag_info['transport_type'],
+                'transport_name': tag_info['msg_options']['transport_name'],
+                'from_addr': tag[1],
+        }
```
hodgestar, line 98: It would probably be a good idea to check that `tag_info` has the keys you expected because it's not guaranteed to have anything -- it's just metadata.

go/vumitools/handler.py
```diff
@@ -0,0 +1,107 @@
...
+
+        sm_cmd = VumiApiCommand.command(
+                handler_config['worker_name'],
+                "send_message",
```
hodgestar, line 102: If one just made the `send_message` here not hard coded (and go it, say, from the event content) we could have a completely generic VumiApiCommand sending event handler?

hodgestar:
Commenting done for now. Awaiting feedback.

go/apps/bulk_message/tests/test_vumi_app.py
dmaclay: It might be superfluous, I built the message structure up by copying live messages.
However, it works, tested on a real deploy, and until I can retest under the same condition I don't want to just rip it out.

go/vumitools/handler.py
dmaclay: It might be superfluous, I built the message structure up by copying live messages.
However, it works, tested on a real deploy, and until I can retest under the same condition I don't want to just rip it out.

go/apps/bulk_message/tests/test_vumi_app.py
hodgestar: That just makes me want to delete it more quickly. :P

go/vumitools/api_worker.py
jerith, line 244: We don't have a UI to configure this yet, so it's still too static.

go/vumitools/api_worker.py
jerith, line 263: Good point.

go/vumitools/api_worker.py
jerith, line 266: This is a stopgap until we have a UI to manage it. It will go away at that point.

jerith, line 266: The real config (as opposed to the current static one) lives in Riak, and is stored as JSON. Because of this, we can't use dicts with compound keys and have to use a list of k/v pairs instead. The `tuple(k)` is to turn the compound key-which-is-a-list into a tuple so we can reconstruct the dict we actually want here.

This stuff could definitely use better commentary.

go/apps/multi_surveys/vumi_app.py
jerith: Good point.

go/vumitools/api_worker.py
jerith, line 263: It does. Good catch.

go/vumitools/api_worker.py
jerith: Probably.

go/vumitools/api_worker.py
jerith, line 322: Currently, yes. I'll add a TODO for a command to the worker that triggers a cache invalidation when the config is updated through the UI.

go/vumitools/api_worker.py
jerith, line 263: Looks like this has already been fixed, actually.

go/vumitools/api_worker.py
jerith, line 276: I'll add a TODO.

go/vumitools/handler.py
jerith, line 98: Probably. I think we need a wrapper object around this or something, though.

go/vumitools/handler.py
jerith, line 102: Yes, but then all the configuration would need to be more generic. I'll add a TODO.

go/vumitools/api_worker.py
hodgestar, line 244: Maybe the FIXME needs to actually describe the problem? I.e. that there is no UI? :)

go/vumitools/api_worker.py
hodgestar, line 263: It still says `self.api_consumer` in my copy of the branch and in this diff -- maybe something wasn't pushed?

go/vumitools/handler.py
smn: Removed this.

go/apps/bulk_message/tests/test_vumi_app.py
smn: Deleting this for the same reasons as above.

go/vumitools/handler.py
smn, line 98: I'm happy to keep this as is, if we don't have those values then things should blow up as I think they should be required.

smn:
I think I'm :+1: on this, but it's gotten to the point where I'm going to do some live testing to see what breaks.

PR #34, opened 2012-07-25
Allowing for public sign-up with django-registration

smn:
Just uses existing logic, adds a few templates. Needs to be able to send
an email so unless you set that up locally then signups will fail for
you. Otherwise working great.


hodgestar:
:+1: and woot on compiling the JS-less stuff. Although we should let @calston know that node.js is now / will shortly be a dependency for Go.

dmaclay:
:+1:

PR #35, opened 2012-08-01
New templates for password resets

smn:
Also changed settings.py to send emails from hello@vumi.org and to use
the Console email backend for development.


hodgestar:
I'm not the most confident Django template reviewer, but :+1: from my side.

PR #36, opened 2012-08-14
The old VUMIGO-115 approach has died, this ports it to the new events setup.

smn:


go/apps/afropinions/handlers.py
```diff
@@ -0,0 +1,67 @@
...
+from vumi import log
+
+from go.vumitools.handler import EventHandler
+from go.vumitools.api import VumiUserApi
+
+
+class YoPaymentHandler(EventHandler):
+
+    def get_user_api(self, account_key):
+        return VumiUserApi(self.dispatcher.vumi_api, account_key)
```
hodgestar: This seems like a really common method. Maybe we should plonk it on the base EventHandler?

go/apps/afropinions/handlers.py
```diff
@@ -0,0 +1,67 @@
...
+        Expects 'content' to be a dict with the following keys and values:
+
+        :param from_addr:
+            The address from which the message was received and which
+            should be topped up with airtime.
+
+        """
+        if not self.url:
+            log.error('No URL configured for YoPaymentHandler')
+            return
```
hodgestar: Should this error perhaps not be thrown during setup_handler()? Seems a pity not to fail early.

hodgestar:
Just two minor comments. Otherwise :+1: on landing.

go/apps/afropinions/handlers.py
```diff
@@ -0,0 +1,67 @@
...
+            'Authorization': 'Basic %s' % (credentials.strip(),)
+        }
+
+    @inlineCallbacks
+    def handle_event(self, event, handler_config):
+        """
+
+        Hit the Yo payment gateway when a vxpoll is completed.
+
+        Expects 'content' to be a dict with the following keys and values:
```
dmaclay, line 35: we need to have a better name for this dict - i support either event_data or event_dict to go wth the event_type field

but this is not Yo! specific issue

go/apps/sna/handlers.py
```diff
@@ -0,0 +1,65 @@
...
+
+        NOTE:   perhaps all we need here is the message_id and lookup the
+                rest of the required data from the message in the message store
+
+        :param from_addr:
+            the from_addr of the user interaction that triggered the event
+        :param message_id:
+            the id of the message that triggered the event, used to keep an
+            audit trail of opt-outs and which messages they were triggered by.
+        :param transport_type:
```
dmaclay, line 38: shouldn't a link to the original message via message_id get us transport_type ?

dmaclay, line 38: ah - we need it now - not for audit purposes

dmaclay:
also :+1: ... but this has highlighted how the event 'content' dict really needs renaming

go/apps/afropinions/handlers.py
smn: Yes, but it turns out I wasn't using it anywhere so I'm removing it.

go/apps/afropinions/handlers.py
smn: the config already did a a `config['url']` lookup that would blow up if not specified so removing this entirely.

hodgestar:
Other than comment about event content, :+1:.

PR #40, opened 2012-08-22
Cherry picking patches from 115 that was in prod

smn:
Mostly related to:

1) allowing deduping of stuff pre-bulk send (was already reviewed)
2) fixing issues with reading CSV files that are \r\n line delimited

### DIFF
--- a/go/settings.py
+++ b/go/settings.py
@@ -253,11 +253,6 @@ VXPOLLS_POLL_ID = VXPOLLS_CONFIG.get('poll_id')
 VXPOLLS_TRANSPORT_NAME = 'vxpolls_transport'
 VXPOLLS_WORKER_NAME = 'vxpolls_worker'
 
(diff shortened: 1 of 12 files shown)


go/settings.py
```diff
@@ -253,6 +253,11 @@ def abspath(*args):
 VXPOLLS_TRANSPORT_NAME = 'vxpolls_transport'
 VXPOLLS_WORKER_NAME = 'vxpolls_worker'
 
+FILE_UPLOAD_HANDLERS = (
+    # "django.core.files.uploadhandler.MemoryFileUploadHandler",
```
hodgestar: Should the commented out line still be here? Maybe it just needs a comment saying when one would use the commented out line?

smn: Thanks, found a way of working around this change.

hodgestar:
Oh, that was fixed in a later commit.

:+1: :)

PR #42, opened 2012-08-30
that's enough to get the clinic sing-ups live again better rewrite in a ...

dmaclay:
...diff issue


smn:
Have a look at `go/apps/surveys.py` since that fixes this problem properly, rather switch over fully to using the `vumi.persist.redis_manager.RedisManager` than extend the life of this hack?

hodgestar:
I'm +1 on landing this now as a minimal fix for production. This should all switch over to TxRedisManager but that needs more testing.

PR #41, opened 2012-08-30
Assorted test fixes

jerith:


smn:
tests now all pass. :+1: from me but perhaps @hodgestar or @dmaclay can still review as well.

hodgestar:
:+1:

PR #44, opened 2012-08-30
Fixing VUMIGO-130 contact footer

smn:
Also removing a stray include file and adding some Javascript to
automagically select the right tab when the page is loaded.

### DIFF
--- a/go/templates/app.html
+++ b/go/templates/app.html
@@ -59,6 +59,6 @@
 	<hr>
 	<footer class="footer">
 	    <p class="pull-right"><a href="#">Back to top</a></p>
-	    <p>&copy; 2012 <a href="http://praekeltfoundation.org">Praekelt Foundation</a>. <a href="mailto:hello@vumi.org">Contact</a>. <a href="https://github.com/praekelt/vumi">Github</a></p>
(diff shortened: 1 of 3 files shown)


go/templates/app.html
```diff
@@ -59,6 +59,6 @@
 	<hr>
 	<footer class="footer">
 	    <p class="pull-right"><a href="#">Back to top</a></p>
-	    <p>&copy; 2012 <a href="http://praekeltfoundation.org">Praekelt Foundation</a>. <a href="mailto:hello@vumi.org">Contact</a>. <a href="https://github.com/praekelt/vumi">Github</a></p>
+	    <p>&copy; 2012 <a href="http://praekeltfoundation.org">Praekelt Foundation</a>. <a href="{% url account:index %}#6">Contact</a>. <a href="https://github.com/praekelt/vumi">Github</a></p>
```
hodgestar: What is `#6`?

hodgestar: By which I mean, is there a more understandable way to reference the tab that that link points to? Because I have no clue what tab six is. :)

jerith:
:+1:, modulo @hodgestar's comment about `#6`.

smn:
The `#6` is the `<a href="#6">Contact Us</a>` -- which, I agree, can be named better.

PR #45, opened 2012-08-30
Tiny UI change

smn:
the CreditManager returns `None` if there's no set credit. Which is
think is correct since `None` is not the same as `0` credit.

The change is in the context_processor to display 0 nonetheless.


hodgestar:
:+1:

PR #46, opened 2012-08-31
Update the survey_completed event to include participant data

smn:
Participant data is passed along with the event. The number of
interactions (and other information) might be relevant to the event
handler. This allows us to update the Afrop handler to refund users with
airtime for the number of messages sent.

### DIFF
--- a/go/apps/surveys/tests/test_vumi_app.py
+++ b/go/apps/surveys/tests/test_vumi_app.py
@@ -191,14 +191,31 @@ class TestSurveyApplication(AppWorkerTestCase):
             'Thanks for completing the survey')
         self.assertEqual(last_msg['session_event'],
             TransportUserMessage.SESSION_CLOSE)
+
+        poll_id = 'poll-%s' % (self.conversation.key,)
+
         [app_event] = self.get_dispatched_app_events()
+
+        # The poll has been completed and so the results have been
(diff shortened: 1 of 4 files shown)


go/apps/surveys/tests/test_vumi_app.py
```diff
@@ -191,14 +191,31 @@ def complete_survey(self, questions, start_at=0):
...
-            'from_addr': last_sent_msg['from_addr'],
-            'transport_type': last_sent_msg['transport_type'],
-            'message_id': last_sent_msg['message_id'],
-        })
+        self.assertEqual(app_event['content']['from_addr'],
+            last_sent_msg['from_addr'])
+        self.assertEqual(app_event['content']['transport_type'],
+            last_sent_msg['transport_type'])
+        self.assertEqual(app_event['content']['message_id'],
+            last_sent_msg['message_id'])
```
hodgestar: I would keep this as a dict equality check and just add the check on the participant.interactions value separately (maybe with a comment). The dict equality check also guards against extraneous keys creeping in, which I think is valuable.

hodgestar:
Apart from minor comment about dict equality check, :+1:.

dmaclay:
:+1:

hodgestar:
:+1:

PR #47, opened 2012-08-31
Remove scheduling UI

smn:
Keeping it in is misleading since it doesn't do anything. Commented it
out until we actually have a proper scheduler.


jerith:
:+1:, but we're going to be needing it again fairly soon. ;-)

hodgestar:
Seconding jerith's comment. :)

smn:
Yes, that's why I only commented it out.

PR #49, opened 2012-09-03
VUMIGO-164 refactor vumitools

jerith:


go/vumitools/api.py
```diff
@@ -504,9 +235,19 @@ def batch_done(self, batch_id):
...
+        """Create a VumiApiCommand and send it.
+
+        :param str worker_name: Name of worker to send command to.
+        :param str command: Type of command to send.
+        :param *args: Positional args for command.
+        :param **kwargs: Keyword args for command.
+        """
+        self.mapi.send_command(
+            VumiApiCommand.command(worker_name, command, *args, **kwargs))
+
```
smn, line 248: Can we make a note about this being required due to circular imports?

smn:
I think we can find a better home for the `ConversationException` but that will likely happen over time as we grow more exceptions - nothing to fix now.

:+1: except for a bit more docs on why we have `send_new_command()` now alongside `send_command()`

jerith:
Turns out we're not actually calling `send_command()` from anywhere outside what used to be `api.py`, so I just replaced it.

smn:
:+1:

PR #51, opened 2012-09-04
Removing old configs

smn:
A lot of the old configs had old keys & prefixes for stuff that has now
been fixed properly with the managers and submanagers. I want to avoid
getting the config files confused and resulting in debugging stuff that
relates to using wrong prefixes.

All needed config files for deploying Go is in the praekelt-machines
repo and is managed by puppet.


hodgestar:
:+1:

We need to have a standard demo / dev config at some point but that can wait.

smn:
At some point we'll need to rewrite history and completely purge the `./config` dir from Git's tracked history. This needs to be done on all branches that haven't landed in `develop` yet, so preferably at a point where there are no outstanding feature branches. After that we can add a demo config with a new set of files in a location that doesn't conflict with our configs as managed by puppet.

.gitignore
```diff
@@ -16,3 +16,6 @@ requirements.pip.md5
 .pydevproject
 distribute-0.6.10.tar.gz
 .DS_Store
+/go/media/
+/go/static/
```
jerith: We _really_ shouldn't be writing this stuff to the middle of our code dirs. I'd rather leave it visible and annoying until we fix it.

jerith:
Aside from my `.gitignore` comment, :+1:

jerith:
Thanks. :+1:

dmaclay:
:+1: & not scary :)

PR #53, opened 2012-09-04
Removing go/settings.py reliance on poll.yaml

smn:
poll.yaml was being read by the Django webapp but it is not longer
necesary. This removes that.


hodgestar:
:+1:

dmaclay:
:+1:

PR #55, opened 2012-09-10
`yield` missing in a bunch of places, causes problems with real Redis.

smn:
Surfaced in SNA with Deferreds firing in the wrong order. All tests now
pass against real redis with the exception of `test_start` for
`surveys/vumi_app.py` which fails with 'Connection was closed cleanly`
which is a bit odd.

```
go.apps.surveys.tests.test_vumi_app
  TestSurveyApplication
    test_clearing_old_survey_data ...
[OK]
    test_ensure_participant_cleared_after_archiving ...
[OK]
    test_start ...
[ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
                                                     [ERROR]
    test_survey_completion ...
[OK]

===============================================================================
[ERROR]
Traceback (most recent call last):
Failure: twisted.internet.error.ConnectionDone: Connection was closed
cleanly.

go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
go.apps.surveys.tests.test_vumi_app.TestSurveyApplication.test_start
-------------------------------------------------------------------------------
Ran 4 tests in 9.925s

FAILED (errors=12, successes=3)
```

@Hodgestar @jerith @dmaclay please review

PR #56, opened 2012-09-11
surveys/vumi_app.py is missing yields to super() calls

smn:
The result is Deferreds firing out of order. This only happens when
runnin against real Redis (which PR #55 deals with). This is only the
fix for the SNA issue (which is already running in production)


jerith:
:+1:

PR #57, opened 2012-09-11
Cleanup contacts/views.py and offload CSV processing to Celery

smn:
Allows us to return the HTTP request quicker.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -3,8 +3,13 @@ import os.path
 from celery.task import task
 
 from django.conf import settings
+from django.core.mail import send_mail
+from django.template.loader import render_to_string
+
+from vumi.persist.fields import ValidationError
 
 from go.vumitools.api import VumiUserApi
+from go.base.models import UserProfile
 from go.contacts.utils import parse_contacts_csv_file
 
 
@@ -26,7 +31,6 @@ def delete_group(account_key, group_key):
 
 @task(ignore_result=True)
 def import_csv_file(account_key, group_key, csv_path, field_names, has_header):
-    print 'importing csv file'
     api = VumiUserApi.from_config(account_key, settings.VUMI_API_CONFIG)
     contact_store = api.contact_store
     group = contact_store.get_group(group_key)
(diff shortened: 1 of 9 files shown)


go/contacts/utils.py
```diff
@@ -0,0 +1,165 @@
...
+    We need to know what we cannot set to avoid a CSV import overwriting
+    things like account details. Excluded attributes is a list of contact
+    attributes that are to be ignored.
+
+    """
+    data_dictionaries = read_data_from_csv_file(csv_file, field_names)
+
+    known_attributes = [attribute
+        for attribute in Contact.field_descriptors.keys()
+        if attribute not in excluded_attributes]
```
hodgestar: This will be a _lot_ faster if you make this a set.

go/contacts/tasks.py
```diff
@@ -20,3 +28,58 @@ def delete_group(account_key, group_key):
...
+    written_contacts = []
+
+    try:
+        with open(full_path, 'rU') as csv_file:
+            for csv_data in parse_contacts_csv_file(csv_file, field_names,
+                                                has_header):
+                [count, contact_dictionary] = csv_data
+
+                # Make sure we set this group they're being uploaded in to
+                groups = contact_dictionary.setdefault('groups', [])
```
hodgestar: Can `contact_dictionary['groups']` actually be usefully populated from the CSV file? It seems unlikely it can end up with a value that is a list? Maybe `groups` just needs to be set to `[group.key]` and possibly excluded from the valid fields in the CSV file?

go/contacts/views.py
```diff
@@ -105,137 +82,8 @@ def groups(request):
...
-        # If we've determined that the first line of the file is
-        # a header then skip it.
-        if has_header and counter == 0:
-            continue
-
-        # Make sure we set this group they're being uploaded in to
-        groups = data_dictionary.setdefault('groups', [])
-        groups.append(group.key)
-
-        # Make sure we normalize the msisdn before saving in the db
```
hodgestar, line 215: I can't see the normalization step anywhere now? Did I miss it? Seems like a strange feature to drop.

smn, line 215: Pulled out the normalization since it was site-wide and hard coded in the `settings.py` to SA - which is worse than not having it and telling clients to normalize the MSISDNs themselves. Reopening ticket VUMIGO-31 to address this.

hodgestar:
:+1:

PR #58, opened 2012-09-11
Filter contacts on surname using mapreduce and not in-memory python.

smn:
Adds a `filter_contacts_on_surname()` function which in time could be
made more generic to allow filtering on other fields. Resisting the
temptation to do that now because a) we don't need it yet and b) it
comes awfully close to what Riak search gives us anyway with the work
being done in VUMIGO-67


hodgestar:
:+1:

PR #59, opened 2012-09-12
Actually send to Celery

smn:
Stuff was blocking becausing calling the task straight does exactly
that. `.delay(..)` is what we wanted.


jerith:
:+1:

PR #60, opened 2012-09-12
Fix for when surname is null/undefined

smn:


go/vumitools/contact/models.py
```diff
@@ -136,12 +136,13 @@ def filter_contacts_on_surname(self, letter, group=None):
...
         js_function = """function(value, keyData, arg){
             var data = Riak.mapValuesJson(value)[0];
-            var surname = data.surname.toLowerCase();
-            if(surname && surname[0] === arg){
-                return [[value.key, value[0]]];
-            } else {
-                return [];
+            if(data.surname) {
+                var surname = data.surname.toLowerCase();
+                if(surname && surname[0] === arg){
```
jerith: Do you need to check for the existence of `surname` here now that you're checking `data.surname` above?

smn: surname can still be an empty string.

smn: Ah right. Removing.

peterp:
+1

PR #61, opened 2012-09-12
VUMIGO-107 app view overhaul (part 1)

jerith:
This is the first part of the app view overhaul stuff. There's a lot we still need to do, but this gives us a reasonable starting point for new work.


go/apps/bulk_message/urls.py
```diff
@@ -1,10 +1,4 @@
...
+from go.apps.bulk_message.views import BulkSendConversationViews
 
-urlpatterns = patterns('',
-    url(r'^new/', views.new, name='new'),
-    url(r'^(?P<conversation_key>\w+)/people/', views.people, name='people'),
-    url(r'^(?P<conversation_key>\w+)/send/', views.send, name='send'),
-    url(r'^(?P<conversation_key>\w+)/end/', views.end, name='end'),
-    url(r'^(?P<conversation_key>\w+)/', views.show, name='show'),
-)
+urlpatterns = patterns('', *BulkSendConversationViews().urls())
```
smn: You can do `urlpatterns += other_url_patterns` -- which might make more sense here.

jerith: Good point. I may as well just move the whole thing into `ConversationViews.urls()`, though.

smn: Actually, why not change `def urls()` to `def get_urls()` and have it return a `patterns()` to begin with and assign that to the `urlpatterns` var?

go/conversation/base.py
```diff
@@ -0,0 +1,245 @@
...
+class NewConversationView(ConversationView):
+    template_name = 'new'
+
+    def request_setup(self, request):
+        return (), {}
+
+    def get(self, request):
+        # We used to put the time in the form, but that's a silly thing to do.
+        # Especially since we now ignore it anyway.
+        form = self.make_conversation_form(request.user_api, initial={})
```
smn, line 64: Where is `self.conversation_form` being set for this?

jerith, line 64: In the view constructor, via keyword params passed in by `ConversationViews.mkview()`. I'll add a comment somewhere.

go/conversation/templates/generic/includes/nosend.html
```diff
@@ -0,0 +1,4 @@
+<div class="span8">
```
smn: I suspect this needs to be a `span4`.

jerith: Perhaps. It was a span12 before I included the first column (which is a span4) in both `send` and `nosend` versions.

go/conversation/base.py
jerith, line 64: I actually wrote a big docstring.

go/conversation/templates/generic/includes/nosend.html
smn: Yeah, it was on a separate page earlier, which was an ugly hack but probably for a reason. I think it was for client initiated survey stuff (which this doesn't touch)

smn:
:+1:

jerith:
Since github's hiding the diff commentary, I'll bring this out here.

We still have separate pages for "new" and "start", and both bulk_send and opt_out were written like that previously. We need to take a good look at what our UI flows currently are and what they should become, but for now I'm mostly keeping things looking the way they looked before.

PR #62, opened 2012-09-13
Tiny change in settings.py

smn:
Our dev setup assumes '/develop' and our production setup uses '/go' --
which is a good distinction. Our `go/settings.py` had '/go' for celery,
I've moved that to `production_settings.py` in praekelt-machines and
defaulted it to `/develop` instead.


hodgestar:
:+1:

PR #63, opened 2012-09-13
VUMIGO-58 scheduled message delivery

jerith:
Lots of stuff in here. I'd like to land it sooner rather than later.


go/apps/sequential_send/tests/test_vumi_app.py
```diff
@@ -0,0 +1,265 @@
...
+        scheduled_conversations = yield self.app._get_scheduled_conversations()
+
+        def get_conversations(conv_pointers):
+            self.assertEqual(sorted(conv_pointers), sorted(expected))
+            return list(convs)
+        self.app.get_conversations = get_conversations
+
+        self.app._get_last_poll_time = lambda: poll_times[-1]
+        self.app._set_last_poll_time = lambda t: poll_times.append(str(t))
+        self.app._get_scheduled_conversations = lambda: scheduled_conversations
```
smn, line 127: Perhaps using `self.patch()` for this would be cleaner.

go/apps/sequential_send/vumi_app.py
```diff
@@ -0,0 +1,185 @@
...
+    def validate_config(self):
+        super(SequentialSendApplication, self).validate_config()
+        self.poll_interval = self.config.get('poll_interval', 60)
+
+    def _setup_poller(self):
+        self.poller = LoopingCall(self.poll_conversations)
+        if self._clock is not None:
+            # So we can stub in a fake clock.
+            self.poller.clock = self._clock
+        self.poller.start(self.poll_interval, now=False)
```
smn, line 59: I think this can be solved with `self.patch()` in the test case as well?

jerith, line 59: It can't, because we need to do the replacement after we've created the `LoopingCall`, but before we start it.

go/apps/sequential_send/tests/test_vumi_app.py
jerith, line 127: Actually, it wouldn't. The main reason to use `self.patch()` is to undo the patching during teardown, and since we're patching a specific instance that's going away during teardown anyway that's unnecessary.

go/apps/sequential_send/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+MessageFormSet = formsets.formset_factory(
+    MessageForm, can_delete=True, extra=1)
+
+
+class EditSequentialSendConversationView(EditConversationView):
+    def make_form(self, conversation):
+        metadata = conversation.get_metadata()
+        print metadata
```
hodgestar: Stray print.

go/apps/sequential_send/vumi_app.py
hodgestar, line 59: I think you can avoid this by something like `self.patch('vumi_app.LoopingCall', MyLoopingCall)` and have MyLoopingCall do the clock fiddling?

go/apps/sequential_send/views.py
jerith: Good catch.

hodgestar:
Other than making an attempt to remove the `_clock` testing stub using my LoopingCall patch suggestions, :+1: on landing.

go/apps/sequential_send/vumi_app.py
jerith, line 59: I spent a few minutes trying to get this right, and it got uglier than the current solution before I managed to make it work. :-/

smn:
:+1:

go/conversation/base.py
```diff
@@ -171,18 +184,88 @@ class ShowConversationView(ConversationView):
...
+
+    def make_form(self, key, form, metadata):
+        data = metadata.get(key, {})
+        if hasattr(form, 'initial_from_metadata'):
+            data = form.initial_from_metadata(data)
+        return form(prefix=key, initial=data)
+
+    def make_forms(self, conversation):
+        metadata = conversation.get_metadata() or {}
+        return [self.make_form(key, edit_form, metadata)
```
smn, line 243: `get_metadata()` allows for a `default={}` kwarg that does basically this.

go/conversation/base.py
```diff
@@ -171,18 +184,88 @@ class ShowConversationView(ConversationView):
...
+        return form.cleaned_data
+
+    def process_forms(self, request, conversation):
+        metadata = conversation.get_metadata()
+        for key, edit_form in self.edit_conversation_forms:
+            form = edit_form(request.POST, prefix=key)
+            # Is this a good idea?
+            if not form.is_valid():
+                return self.get(request, conversation)
+            metadata[key] = self.process_form(form)
```
smn, line 258: For now yes, don't see a better way.

smn:
:+1: other than the tiny suggestion that I can live without.

PR #64, opened 2012-09-13
Make Contact importing code more generic & pluggable.

smn:
First implementation is the ported CSV parser, XLS parser coming
shortly.


go/contacts/parsers/base.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        column_set = set([column.lower() for column in columns])
+        hint_set = set(['phone', 'contact', 'msisdn', 'number'])
+        return hint_set.intersection(column_set)
+
+    @classmethod
+    def get_file_extension(cls, file_name):
+        name, extension = os.path.splitext(file_name)
+        separator, suffix = extension.rsplit('.', 1)
+        return suffix
```
jerith: Isn't `os.path.splitext(file_name)[1][1:]` sufficient here?

go/contacts/parsers/base.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # We're expecting a generator so loop over it and save as contacts
+        # in the contact_store, normalizing anything we need to
+        data_dictionaries = self.read_data_from_file(file_path, field_names)
+        for counter, data_dictionary in enumerate(data_dictionaries):
+
+            # If we've determined that the first line of the file is
+            # a header then skip it.
+            if has_header and counter == 0:
+                continue
```
jerith: Rather than mucking about with counter values all over the place, maybe do something like this:

``` python
if has_header:
    next(data_dictionaries)
for counter, data_dictionary in enumerate(data_dictionaries, 1):
    # do stuff
    yield (counter, contact_dictionary)
```

go/contacts/parsers/base.py
hodgestar: I'd keep the first assignment for clarity but skip the second split as @jerith suggests.

go/contacts/parsers/base.py
hodgestar: I think `read_data_from_file` should skip the header itself.

go/contacts/tasks.py
```diff
@@ -46,17 +43,23 @@ def import_csv_file(account_key, group_key, csv_path, field_names, has_header):
     count = 0
     written_contacts = []
 
+    extension, parser = ContactFileParser.get_parser(file_name)
+
+    if parser is None:
+        logging.warn('No file parser available for: %s. Stopping' % (
+            extension,))
+        return
```
hodgestar: Shouldn't this also email the user an error message?

smn: Fixed thanks.

go/contacts/parsers/base.py
smn: Having `read_data_from_file` skip the header itself. Took out the counters from all this, added it to the tasks instead.

go/contacts/parsers/base.py
```diff
@@ -0,0 +1,110 @@
...
+        'bbm_pin': 'BBM Pin',
+        'msisdn': 'Contact Number',
+        'gtalk_id': 'GTalk (or XMPP) address',
+        'dob': 'Date of Birth',
+        'facebook_id': 'Facebook ID',
+        'twitter_handle': 'Twitter handle',
+        'email_address': 'Email address',
+    }
+
+    EXCLUDED_ATTRIBUTES = ['user_account', 'created_at', 'extra', 'groups']
```
jerith: Maybe this should be a whitelist instead of a blacklist? Or even better, some kind of metadata on the Contact model?

go/contacts/parsers/csv_parser.py
```diff
@@ -0,0 +1,66 @@
...
+            csvfile.seek(0)
+            reader = csv.DictReader(csvfile, field_names, dialect=dialect)
+            if has_header:
+                reader.next()
+            for row in reader:
+                # Only process rows that actually have data
+                if any([column for column in row]):
+                    # Our Riak client requires unicode for all keys & values stored.
+                    unicoded_row = dict([(key, unicode(value, 'utf-8'))
+                                            for key, value in row.items()])
```
jerith, line 24: How sure are we that the files will have UTF-8 in them? I suspect Latin-1 or CP-1252 are more likely to show up in the real world.

go/contacts/parsers/base.py
smn: In a way this is a whitelist, it's just a list of what's excluded from the Contact model's field descriptors.

go/contacts/parsers/csv_parser.py
smn, line 24: We don't but I'd like to tell people that they need to supply double-quoted UTF-8 encoded CSV files for importing otherwise it will fail. CSV is a minefield and I don't see a really solid way of working around that without putting in some contraints as to what we will tolerate as input.

go/contacts/parsers/base.py
jerith: I meant whitelist as opposed to blacklist. As we add new fields to the Contact model, this will get out of date and might let badness slip through.

smn: cool, happy to change this.

PR #65, opened 2012-09-14
pep8/pyflakes cleanup

jerith:
There's a bunch of this stuff, and I'd like to do it in its own branch to avoid adding noise to other reviews.


smn:
:+1: and :cake:

PR #66, opened 2012-09-14
Dynamic Groups

smn:
Called "Smart Groups" in the UI.


go/contacts/templates/contacts/groups.html
```diff
@@ -22,7 +22,7 @@
       			    <li class="active"><a href="{% url contacts:groups %}">Groups</a></li>
       			</ul>
       			<form id="search-filter" name="search-filter" action="" method="get" class="well form-search">
-                <input type="text" id="search-filter-input" name="q" value="{{query|default_if_none:"Search..."}}" class="input-large search-query">
+                <input type="text" id="search-filter-input" name="q" value="{{query|default:""}}" placeholder="Search..." class="input-large search-query">
```
hodgestar: I'm a bit surprised all those double-quotes work. Can they be single quotes instead?

go/vumitools/contact/models.py
```diff
@@ -124,6 +133,44 @@ def search_group(self, group, return_keys=False, **kw):
...
+        static_groups = [group for group in all_groups if group.query is None]
+        dynamic_groups = [group for group in all_groups if group.query]
+
+        # Grab all contacts we can find
+        contacts = []
+        for group in static_groups:
+            contacts.extend((yield self.get_static_contacts_for_group(group)))
+
+        for group in dynamic_groups:
+            contacts.extend((yield self.get_dynamic_contacts_for_group(group)))
```
hodgestar: Should these contact list be de-duplicated for the case of overlapping groups?

hodgestar:
Other than my two comments, :+1:.

go/contacts/templates/contacts/groups.html
smn: Can do.

go/vumitools/contact/models.py
smn: I'm leaving that the responsibility of the application that actually does the sending.

hodgestar: I'm not sure how much sense that makes. It's not like the application will lose information if we just remove contacts that have the same key?

hodgestar:
I think it's just deleting the commented out `people` method and de-duping the contact objects left to go?

go/vumitools/contact/models.py
smn: you're right, will do.

PR #67, opened 2012-09-16
VUMIGO-118 campaign subscription management

jerith:


go/apps/subscription/tests/test_views.py
```diff
@@ -0,0 +1,136 @@
...
+from go.vumitools.tests.utils import VumiApiCommand
+from go.apps.tests.base import DjangoGoApplicationTestCase
+from go.base.utils import vumi_api_for_user
+
+
+TEST_GROUP_NAME = u"Test Group"
+TEST_CONTACT_NAME = u"Name"
+TEST_CONTACT_SURNAME = u"Surname"
+TEST_SUBJECT = u"Test Conversation"
+
```
smn, line 6: These are also class attributes on `DjangoGoApplicationTestCase`

go/apps/subscription/vumi_app.py
```diff
@@ -0,0 +1,85 @@
...
+    def consume_user_message(self, message):
+        gmd = self.get_go_metadata(message)
+        user_api = self.get_user_api((yield gmd.get_account_key()))
+        conv = user_api.wrap_conversation((yield gmd.get_conversation()))
+
+        contact = yield user_api.contact_store.contact_for_addr(
+            conv.delivery_class, message['from_addr'])
+        if contact is None:
+            print "No contact. :-("
+            return
```
smn: `vumi.log.warn` perhaps?

go/apps/subscription/vumi_app.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def consume_ack(self, event):
+        return self.vumi_api.mdb.add_event(event)
+
+    def consume_delivery_report(self, event):
+        return self.vumi_api.mdb.add_event(event)
+
+    def close_session(self, msg):
+        # TODO: Update
+        tag = TaggingMiddleware.map_msg_to_tag(msg)
```
smn: What are you wanting up update here?

go/vumitools/subscription/handlers.py
```diff
@@ -0,0 +1,49 @@
...
+        user_api = self.get_user_api(event.payload['account_key'])
+        fields = event.payload['content']
+
+        contact = yield user_api.contact_store.get_contact_by_key(
+            fields['contact_id'])
+        contact.subscription[fields['campaign_name']] = {
+            'subscribe': u'subscribed',
+            'unsubscribe': u'unsubscribed',
+            }[fields['operation']]
+        yield contact.save()
```
smn, line 49: Would it make sense to store the timestamps here instead of the `subscribed` / `unsubscribed` strings?

go/apps/subscription/vumi_app.py
jerith: Actually, we create the contact if we don't already have one. That was leftover debuggery.

go/apps/subscription/vumi_app.py
jerith: I tagged all the methods with those TODOs, then removed them once I was done. Must have missed this one.

go/vumitools/subscription/handlers.py
jerith, line 49: The strings are to tell you the status. If we care about tracking subscription history, we'd be better off building something that does that specifically.

smn:
:+1: other than the timestamp question and some stray `TODO` statements that either need to be removed or need some more context.

go/vumitools/subscription/handlers.py
smn, line 49: Agreed, but just wondering if we're doing anything specific with the 'subscribed' value of that `dict` entry or not. If we're not then for debugging purposes having the timestamp in there makes sense, even if we don't track subscription history. I'm ±1, really just asking.

jerith, line 49: The dict is just a lookup from `operation` to the value of the field. We're storing a string there, not a dict. I'll rewrite that section to make it clearer.

smn, line 49: Ah, thanks - missed that

smn:
:+1:

PR #71, opened 2012-09-18
Add Wikipedia application - G wants to demo it to Airtel on 19/9/2012.

smn:
What this does:
1. Creates two new conversation types, 'wikipedia_sms' and 'wikipedia_ussd' of which the first is only a dummy conversation to claim a tag
2. When creating a `wikipedia_ussd` conversation you can fill in the `tagpool` and the `tag` you want to use to send and receive messages from, this is stored in the conversation metadata.
3. When the USSD Wikipedia App wants to send an SMS it first looks up the conversation it is part of, looks up the tagpool information in the metadata. Populates the message with that data and then sends it off.

The `wikipedia.yaml` config looks like this:

``` yaml
worker_name: wikipedia_ussd_application
transport_name: wikipedia_ussd_transport
sms_transport: vumigo_router
incoming_sms_transport: wikipedia_sms_transport
redis_manager:
  key_prefix: "vumigo"
  db: 1
riak_manager:
  bucket_prefix: vumigo.
```

And the addition to `application_message_dispatcher.yaml` are:

In the `exposed_names` block:

``` yaml
  - wikipedia_ussd_transport
  - wikipedia_sms_transport
```

In the `conversation_mappings` block:

``` yaml
    wikipedia_ussd: wikipedia_ussd_transport
    wikipedia_sms: wikipedia_sms_transport
```

Which results in the SMS and USSD being handled by the same application worker but because we're looking up the tagpool in the conversation metadata we're sending SMSs through a different transport back.

Since the `SessionManager` uses the `msisdn` for the key, sessions are shared between all Wikipedia conversations. I don't see this as a big issue for a demo where G is the only account it's been hooked up to.


hodgestar:
It looks surprisingly sane. :D

jerith:
Not quite what I was getting at last night, but it doesn't look too horrible. :+1:

PR #74, opened 2012-09-20
Make survey editable.

smn:
That is all.


jerith:
+0.5 -- I'd prefer to switch over to the new views properly, but we can land this if there's a tight deadline.

smn:
So would I but that touches on a number of this that I don't want to do as part of this PR:
1. The ConversationViews assumes that `conversation_initiator` is either `server` or `client` when for surveys it can be both.
2. Changing the the initiator stuff also touches on stuff in the tagpools.yaml file which probably needs to be addressed as part of this.

VUMIGO-182 addresses this.

PR #75, opened 2012-09-21
App Msg Dispatcher should also route events

smn:
It was already trying to but was failing because it was looking for tags
in event messages. This changes looks up the message for the event and
then uses that message to route the event to the appropriate
application.

### DIFF
--- a/go/apps/bulk_message/vumi_app.py
+++ b/go/apps/bulk_message/vumi_app.py
@@ -35,6 +35,10 @@ class BulkMessageApplication(GoApplicationWorker):
             return
 
         conv = yield self.get_conversation(batch_id, conversation_key)
+        if conv is None:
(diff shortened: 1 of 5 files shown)


dmaclay:
If we need to find an application for an event, shouldn't the event be populated with the required info, rather than going via the original message.  Aside from the performance overhead, what if we have events without triggering messages ?

smn:
These aren't application events, they are transport events, acks & delivery-reports arriving via `transport_name.event`. The name is a bit confusing but they're different from application events which are all routed through `vumi.events`.

dmaclay:
Ah - right - I was confusing vumi.event events and transport_name.event events

smn:
Yeah. We might want to think of coming up with better names at some point.

dmaclay:
:+1:

PR #76, opened 2012-09-25
Middleware for generating metrics for Vumi Go

smn:
It inspects all inbound & outbound messages, maintains counters and
calculates response times for messages.

Not quite sure what to do about events and failures at this point since
they're not guaranteed to have a transport_name.

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -10,6 +10,7 @@ from vumi.utils import normalize_msisdn
 from vumi.components.tagpool import TagpoolManager
 from vumi.blinkenlights.metrics import MetricManager, Count, Metric
 from vumi.persist.txredis_manager import TxRedisManager
+from vumi.errors import ConfigError
 
 from go.vumitools.credit import CreditManager
 
@@ -178,13 +179,26 @@ class MetricsMiddleware(BaseMiddleware):
         metric is published.
     :param dict redis_manager:
         Connection configuration details for Redis.
+    :param str inspection_mode:
+        What mode to operate in, options are `passive` or `active`.
+        Defaults to passive.
+        *passive*:  assumes the middleware endpoints are to be used as the names
+                    for metrics publishing.
+        *active*:   assumes that the individual messages are to be inspected
+                    for their `transport_name` values.
     """
+
+    KNOWN_MODES = frozenset(['active', 'passive'])
 
     def validate_config(self):
         self.manager_name = self.config['manager_name']
         self.count_suffix = self.config.get('count_suffix', 'count')
         self.response_time_suffix = self.config.get('response_time_suffix',
             'response_time')
+        self.inspection_mode = self.config.get('inspection_mode', 'passive')
+        if self.inspection_mode not in self.KNOWN_MODES:
+            raise ConfigError('Unknown inspection_mode: %s' % (
+                self.inspection_mode,))
 
     @inlineCallbacks
     def setup_middleware(self):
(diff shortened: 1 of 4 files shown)


smn:
This relies on work being done in praekelt/vumi#361

go/vumitools/middleware.py
```diff
@@ -151,3 +156,110 @@ def handle_outbound(self, msg, endpoint):
...
+
+    @inlineCallbacks
+    def compare_timestamps(self, transport_name, message):
+        timestamp = yield self.get_outbound_timestamp(transport_name, message)
+        if timestamp:
+            self.set_response_time(transport_name, time.time() - timestamp)
+
+    @inlineCallbacks
+    def handle_inbound(self, message, endpoint):
+        self.increment_counter(message['transport_name'], 'inbound')
```
hodgestar: Would it not make more sense to use `endpoint` here instead of `message['transport_name']`? That would allow a more detailed recording of message flow? I guess potentially both are useful under different circumstances.

go/vumitools/middleware.py
```diff
@@ -151,3 +156,110 @@ def handle_outbound(self, msg, endpoint):
...
+        returnValue(message)
+
+    @inlineCallbacks
+    def handle_outbound(self, message, endpoint):
+        self.increment_counter(message['transport_name'], 'outbound')
+        yield self.compare_timestamps(message['transport_name'], message)
+        returnValue(message)
+
+    def handle_event(self, event, endpoint):
+        self.increment_counter(event['transport_name'], 'event')
```
hodgestar: Is it worthwhile also splitting events out by type and maybe delivery status?

go/vumitools/middleware.py
```diff
@@ -151,3 +156,110 @@ def handle_outbound(self, msg, endpoint):
...
+        return event
+
+    def handle_failure(self, failure, endpoint):
+        self.increment_counter(failure['transport_name'], 'failure')
```
hodgestar: Here we could potentially also split out temporary vs permanent failures.

hodgestar:
:+1: -- should the metric middleware itself maybe go into Vumi a some point?

smn:
Yes, happy to have that happen at some point. Would like to run it in Go first and then move.

hodgestar:
Sounds good.

go/vumitools/tests/test_api_worker.py
```diff
@@ -391,15 +391,6 @@ def setUp(self):
                 u'bulk_message', u'subject', u'message'))
 
     @inlineCallbacks
-    def tearDown(self):
-        # These aren't ready until we use the dispatcher, so we add them here
-        # instead of in setUp()
-        vumi_api = yield self.dispatcher._router.get_vumi_api()
-        self._persist_riak_managers.append(vumi_api.manager)
-        self._persist_redis_managers.append(vumi_api.redis)
```
jerith, line 399: Shouldn't we move these to `setUp()` instead of just removing them?

go/vumitools/tests/test_middleware.py
```diff
@@ -102,3 +114,174 @@ def test_case_sensitivity(self):
...
+    @inlineCallbacks
+    def get_middleware(self, config=None, mw_class=MetricsMiddleware):
+        default_config = self.default_config.copy()
+        default_config.update({
+            'manager_name': 'metrics_manager',
+            'count_suffix': 'counter',
+            'response_time_suffix': 'timer',
+            'redis_manager': {
+                'FAKE_REDIS': 'yes please',
+            }
```
jerith: Instead of adding this redis manager config, use `mk_config()` from the persistence mixin. That does the right thing with fakeredis and also sets appropriate test prefixes to avoid stomping on other things accidentally.

go/vumitools/tests/test_middleware.py
```diff
@@ -102,3 +114,174 @@ def test_case_sensitivity(self):
...
+            default_config.update(config)
+        mw = yield self.create_middleware(MetricsMiddleware,
+            config=default_config)
+        self._middlewares.append(mw)
+        returnValue(mw)
+
+    @inlineCallbacks
+    def tearDown(self):
+        for mw in self._middlewares:
+            yield maybeDeferred(mw.teardown_middleware)
```
jerith: We're using redis here, so we should call the parent `tearDown()` to clean it up.

go/vumitools/tests/test_api_worker.py
smn, line 399: This is already handled by `get_application()` in `AppWorkerTestCase`

go/vumitools/tests/test_middleware.py
smn: thanks, fixed.

go/vumitools/tests/test_middleware.py
smn: Thanks, done.

go/vumitools/tests/test_api_worker.py
jerith, line 399: Ah, right. I forgot we were using that.

jerith:
:+1:

PR #77, opened 2012-09-27
WindowManager for allowing partial sends of big batches

smn:
The idea is that acks will pop stuff out of the inflight window which
will allow for more stuff to be added to be sent out.

next() should return a deferred that fires when the next set is
available.

This is still a work in progress.

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -93,6 +93,21 @@ class TestBulkMessageApplication(AppWorkerTestCase):
         msgs.sort(key=lambda msg: msg['to_addr'])
         [msg1, msg2] = msgs
 
+        # Create acks for messages
+        ack1 = self.mkmsg_ack(user_message_id=msg1['message_id'],
+            sent_message_id=msg1['message_id'])
+        ack2 = self.mkmsg_ack(user_message_id=msg2['message_id'],
+            sent_message_id=msg2['message_id'])
+
+        yield self.dispatch(ack1, rkey='%s.event' % (self.transport_name,))
+        yield self.dispatch(ack2, rkey='%s.event' % (self.transport_name,))
+
+        # Assert that the window's now empty because acks have been received
+        self.assertEqual(
+            (yield self.app.window_manager.count_in_flight(window_id)), 0)
+        self.assertEqual(
+            (yield self.app.window_manager.count_waiting(window_id)), 0)
+
         # check that the right to_addr & from_addr are set and that the content
         # of the message equals conversation.message
         self.assertEqual(msg1['to_addr'], contact1.msisdn)
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -7,7 +7,7 @@ from twisted.internet.defer import inlineCallbacks
 from vumi.message import TransportUserMessage
 
 from go.vumitools.api_worker import CommandDispatcher
-from go.vumitools.api import VumiUserApi, VumiApiCommand
+from go.vumitools.api import VumiUserApi
 from go.vumitools.tests.utils import AppWorkerTestCase
 from go.apps.bulk_message.vumi_app import BulkMessageApplication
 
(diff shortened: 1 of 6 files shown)


go/apps/bulk_message/vumi_app.py
```diff
@@ -11,12 +15,89 @@
...
+            if window_size == 0:
+                raise StopIteration()
+
+            flight_size = yield self.redis.scard(flight_key)
+            room_available = window_size - flight_size
+            if room_available:
+                next_keys = self.redis.zrange(window_key, 0, room_available)
+                for key in next_keys:
+                    yield self.redis.sadd(flight_key, key)
+                d.callback(next_keys)
```
smn: Possible race condition here if there are multiple `WindowManagers` active and looking at the same `window_id`

go/vumitools/tests/test_window_manager.py
```diff
@@ -0,0 +1,56 @@
...
+
+class WindowManagerTestCase(TestCase):
+
+    timeout = 1
+
+    @inlineCallbacks
+    def setUp(self):
+        redis = yield TxRedisManager.from_config({
+            'FAKE_REDIS': 'yes please',
+        })
```
jerith: Rather pull in the persistence stuff and use that. We'd like to clean up after ourselves when we run this against real redis.

go/vumitools/tests/test_window_manager.py
```diff
@@ -0,0 +1,56 @@
...
+        self.window_id = 'window_id'
+        self.wm = WindowManager(redis, window_size=10)
+        self.wm.create(self.window_id)
```
jerith: You need to yield here.

go/vumitools/tests/test_window_manager.py
```diff
@@ -0,0 +1,56 @@
...
+        self.redis = self.wm.redis
+
+    @inlineCallbacks
+    def test_windows(self):
+        windows = yield self.wm.get_windows()
+        self.assertTrue(self.window_id in windows)
+
+    def test_window_recreation(self):
+        self.assertFailure(self.wm.create(self.window_id), WindowException)
```
jerith: `assertFailure()` returns a deferred, so you need to return it here.

go/vumitools/window_manager.py
```diff
@@ -0,0 +1,146 @@
...
+        return self.window_key(self.FLIGHT_KEY, *keys)
+
+    def stats_key(self, *keys):
+        return self.window_key(self.FLIGHT_STATS_KEY, *keys)
+
+    def get_clocktime(self):
+        return self.gc.clock.seconds()
+
+    @inlineCallbacks
+    def create(self, window_id):
```
jerith: I don't like the inconsistency between `create()` and `remove_window()`. Can we either rename this to `create_window()` or rename the other one to just `remove()`? (I'd prefer the former.)

jerith:
Can we add underscore prefixes to a bunch of the utility methods that aren't really part of the public API? (I'm thinking about methods like `set_timestamp()` here.)

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -169,42 +186,3 @@ def test_close_session(self):
         yield self.dispatch(msg)
         dbmsg = yield self.vumi_api.mdb.get_inbound_message(msg['message_id'])
         self.assertEqual(dbmsg, msg)
-
-    @inlineCallbacks
-    def test_send_message_command(self):
```
smn, line 174: Deleted because nothing uses this.

go/vumitools/window_manager.py
```diff
@@ -0,0 +1,215 @@
...
+            yield self.redis.delete(self.map_key(window_id, 'internal',
+                external_id))
+
+    def monitor(self, key_callback, interval=10, cleanup=True,
+        cleanup_callback=None):
+        monitor = LoopingCall(lambda: self._monitor_windows(
+            key_callback, cleanup, cleanup_callback))
+        monitor.clock = self.get_clock()
+        monitor.start(interval)
+        self.window_monitors.append(monitor)
```
smn: Not happy with this as we can potentially get a bunch of `LoopingCall` all of which are competing for the same set of windows.

go/vumitools/window_manager.py
```diff
@@ -0,0 +1,215 @@
...
+    @inlineCallbacks
+    def add(self, window_id, data, key=None):
+        key = key or uuid.uuid4().get_hex()
+        yield self.redis.lpush(self.window_key(window_id), key)
+        yield self.redis.set(self.window_key(window_id, key),
+            json.dumps(data))
+        returnValue(key)
+
+    @inlineCallbacks
+    def get_next_key(self, window_id):
```
smn, line 97: Renamed from `next()` to `get_next_key()` to prevent confusing this thing for an iterator, it isn't.

go/apps/bulk_message/vumi_app.py
```diff
@@ -17,13 +17,56 @@ class BulkMessageApplication(GoApplicationWorker):
...
+    max_ack_window = 100
+    max_ack_wait = 10
+    max_ack_retries = 10
+    monitor_interval = 10
+    monitor_window_cleanup = True
+
+    @inlineCallbacks
+    def setup_application(self):
+        yield super(BulkMessageApplication, self).setup_application()
+        self.window_manager = WindowManager(self.redis,
```
jerith: Do we want a submanager here, perhaps?

go/apps/bulk_message/vumi_app.py
```diff
@@ -17,13 +17,56 @@ class BulkMessageApplication(GoApplicationWorker):
...
+            window_size=self.max_ack_window,
+            flight_lifetime=self.max_ack_wait,
+            max_flight_retries=self.max_ack_retries)
+        self.window_manager.monitor(self.on_window_key_ready,
+            interval=self.monitor_interval,
+            cleanup=self.monitor_window_cleanup,
+            cleanup_callback=self.on_window_cleanup)
+
+    def teardown_application(self):
+        self.window_manager.stop()
```
jerith, line 43: Do we need to call super()'s teardown?

jerith:
:+1:

PR #79, opened 2012-10-01
Add normalization on import

smn:
Currently normalizes for str, int, float and msisdns for countries we
are currently active in.

### DIFF
--- a/go/contacts/parsers/base.py
+++ b/go/contacts/parsers/base.py
@@ -33,6 +33,8 @@ class FieldNormalizer(object):
             ('msisdn_gh', 'Ghanaian contact number (+233)'),
             ('msisdn_cm', 'Cameroonian contact number (+237)'),
             ('msisdn_ng', 'Nigerian contact number (+234)'),
+            ('msisdn_int',
+                'Contact number (already prefixed with country code)'),
         ]
 
     def __iter__(self):
@@ -83,6 +85,11 @@ class FieldNormalizer(object):
 
     def normalize_msisdn_ng(self, value):
         return self.do_msisdn(value, '234')
+
+    def normalize_msisdn_int(self, value):
+        str_value = self.normalize_string(value).lstrip('+')
+        str_value = str_value.lstrip('00')
+        return '+%s' % (str_value,)
 
 class ContactFileParser(object):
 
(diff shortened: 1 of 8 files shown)


go/contacts/parsers/base.py
```diff
@@ -1,11 +1,74 @@
...
+            ('', 'Leave as is'),
+            ('string', 'Plain Text'),
+            ('integer', 'Whole Numbers (0, 10, 245, ... )'),
+            ('float', 'Numbers (0.1, 3.14, 4.165, ...)'),
+            ('msisdn_za', 'South African contact number (+27)'),
+            ('msisdn_ke', 'Kenyan contact number (+254)'),
+            ('msisdn_ug', 'Ugandan contact number (+256)'),
+            ('msisdn_gh', 'Ghanaian contact number (+233)'),
+            ('msisdn_cm', 'Cameroonian contact number (+237)'),
+            ('msisdn_ng', 'Nigerian contact number (+234)'),
```
hodgestar, line 35: Do we maybe want a generic msisdn one too for internationalized numbers?

smn, line 35: I'm not sure what that does and why the `Leave as is` option wouldn't work for that?

go/contacts/tests.py
```diff
@@ -283,6 +297,9 @@ def test_contact_parsing_failure(self):
             'column-0': 'name',
             'column-1': 'surname',
             'column-2': 'msisdn',
+            'normalize-0': '',
+            'normalize-1': '',
```
jerith, line 301: Do you need this assignment if you're not using the value anywhere? pyflakes will complain...

smn, line 301: Why will pyflakes compain? With sending a blank string as a dict value I'm telling the backend that no normalization should take place and it leaves the value as is.

jerith, line 301: Um. I must have been commenting on an outdated diff. The line I commented on was `response = <thing that was there before>`, which seems to be gone now. Sorry about that.

jerith:
If we're not doing any _validation_ on the normalized fields, we should probably be a little more explicit about that. Definitely in the code (docstring or comment or something), and probably a little note in the UI as well.

go/contacts/tests.py
smn, line 301: Yeah removed that late last night.

jerith:
Perhaps we shouldn't push things through strings and then turn them into unicode afterwards? At the very least, we should probably specify character encodings when we do conversions.

jerith:
:+1:

PR #81, opened 2012-10-02
SortedDict isn't safe to put on the wire as JSON

smn:
When decoded it is in a different order. This change the relevant code
to use [(k, v), ...] instead


jerith:
After-the-fact :+1:.

PR #82, opened 2012-10-03
Make sure we unicode all normalized field values

smn:


go/contacts/parsers/base.py
```diff
@@ -46,10 +48,14 @@ def normalize(self, name, value):
 
     def normalize_string(self, value):
         if value is not None:
-            return str(value)
+            if isinstance(value, str):
+                return unicode(value, self.encoding, self.encoding_errors)
+            else:
+                return unicode(value)
```
jerith, line 54: Maybe it would be better to do something like:

``` python
try:
    return unicode(value, self.encoding, self.encoding_errors)
except TypeError:
    return unicode(value)
```

go/contacts/parsers/base.py
```diff
@@ -46,10 +48,14 @@ def normalize(self, name, value):
...
 
     def is_numeric(self, value):
-        return str(value).replace('.', '').isdigit()
+        str_value = self.normalize_string(value)
+        return str_value.replace('.', '').isdigit()
```
jerith: We probably want `.replace('.', '', 1)` there. `1.2.3` isn't a valid number.

jerith:
:+1:

PR #84, opened 2012-10-06
Deal with MSISDNs in float notation somewhat sanely.

smn:
We can get floats from Excel or floats as strings from CSV.
It makes sure the float is really an integer, otherwise it raises and
error. This'll break for masked shortcodes (like GUINNESS) obviously
but I'm fine with that since those aren't likely to be uploaded into the
contact database.

### DIFF
--- a/go/contacts/parsers/base.py
+++ b/go/contacts/parsers/base.py
@@ -61,7 +61,8 @@ class FieldNormalizer(object):
     def is_numeric(self, value):
         str_value = self.normalize_string(value)
         try:
-            return float(str_value)
+            float(str_value)
+            return True
         except ValueError:
             return False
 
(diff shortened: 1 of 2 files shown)


go/contacts/parsers/base.py
```diff
@@ -55,7 +58,10 @@ def normalize_string(self, value):
 
     def is_numeric(self, value):
         str_value = self.normalize_string(value)
-        return str_value.replace('.', '', 1).isdigit()
+        try:
+            return float(str_value)
+        except ValueError:
+            return False
```
jerith, line 67: This will return `False` for `"0"`, which is problematic. Maybe we want this:

``` python
try:
    float(str_value)
    return True
except ValueError:
    return False
```

go/contacts/parsers/base.py
```diff
@@ -67,14 +73,24 @@ def normalize_float(self, value):
             return float(value)
         return value
 
+    def lchop(self, chops, string):
```
jerith: I think it makes slightly more sense to have these params the other way around. (That's aesthetic more than anything else, though.)

jerith:
:+1:

PR #85, opened 2012-10-13
SNA Event handler crashes, causing event backlog.

smn:
Because of how the contact import worked at one point, there are some
values in the `contact.extra` dict that are empty strings. When trying
to `int()` that value it spews an error.


hodgestar:
:+1:

PR #87, opened 2012-10-15
WindowManager shouldn't retry acks.

smn:
That was a bad idea since we cannot put any guarantees on within what
time frame an ack should come back.

### DIFF
--- a/go/vumitools/window_manager.py
+++ b/go/vumitools/window_manager.py
@@ -18,14 +18,13 @@ class WindowManager(object):
     FLIGHT_STATS_KEY = 'flightstats'
     MAP_KEY = 'keymap'
 
-    def __init__(self, redis, window_size=100,
-        flight_lifetime=10, max_flight_retries=10, gc_interval=10):
+    def __init__(self, redis, window_size=100, flight_lifetime=10,
+        gc_interval=10):
         self.window_size = window_size
         self.flight_lifetime = flight_lifetime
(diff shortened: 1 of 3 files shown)


jerith:
I'd prefer a much longer default ack timeout. If we're dealing with a heavily rate-limited transport, we'll end up timing out on all the acks and shoving everything into the queue anyway, which is pretty much what we're trying to avoid here.

jerith:
:+1:

go/vumitools/window_manager.py
```diff
@@ -141,19 +131,12 @@ def get_expired_flight_keys(self, window_id):
...
             for key in expired_keys:
-                tries = yield self._increment_tries(window_id, key)
-                window_key = self.window_key(window_id)
-                inflight_key = self.flight_key(window_id)
-                if tries < self.max_flight_retries:
-                    yield self.redis.rpoplpush(inflight_key, window_key)
-                    yield self._clear_timestamp(window_id, key)
-                else:
-                    yield self.remove_key(window_id, key)
+                yield self.redis.lrem(self.flight_key(window_id), key, 1)
```
hodgestar, line 139: I found it a bit surprising that the call to `self.remove_key` was replaced with a direct call to `self.redis.lrem`. Just checking that this was intended?

smn, line 139: Yes, since we're just removing them from the flight since we might still receive an ack at _some_ point.

PR #89, opened 2012-10-16
Django management command to pull out simple stats per conversation.

smn:
```
List command options:
  ./go-admin.sh go_account_stats 

List conversations (append active to exclude inactives):
  ./go-admin.sh go_account_stats <email-address> list_conversations 

Get stats:
  ./go-admin.sh go_account_status <email-address> stats <conv-key>
```


jerith:
:+1: on condition that we replace it with a better solution in the next week or two.

smn:
@jerith yes that's the idea. This buys us some time and gives others at least _something_ to work with while we build proper tools.

PR #91, opened 2012-10-17
VUMIGO-200 application metrics infrastructure

jerith:
This adds the infrastructure for app workers to collect metrics and a prototype metrics collector for the sequential_send app to make sure that it works.

### DIFF
--- a/go/base/management/commands/go_manage_metrics.py
+++ b/go/base/management/commands/go_manage_metrics.py
@@ -1,0 +1,62 @@
+from optparse import make_option
+
+from django.core.management.base import BaseCommand, CommandError
+from django.contrib.auth.models import User
+
+from go.base.utils import vumi_api_for_user
+
+
+class Command(BaseCommand):
+    help = "Enable or disable metrics for a Vumi Go user"
+
+    LOCAL_OPTIONS = [
+        make_option('--email-address',
+            dest='email-address',
+            help='Email address for the Vumi Go user'),
+        make_option('--enable',
+            dest='enable',
+            action='store_true',
+            default=False,
+            help='Give access to this application'),
+        make_option('--disable',
+            dest='disable',
+            action='store_true',
+            default=False,
+            help='Revoke access to this application'),
+    ]
+    option_list = BaseCommand.option_list + tuple(LOCAL_OPTIONS)
+
+    def handle(self, *args, **options):
+        options = options.copy()
+        for opt in self.LOCAL_OPTIONS:
+            if options.get(opt.dest) is None:
+                value = raw_input("%s: " % (opt.help,))
+                if value:
+                    options[opt.dest] = value
+                else:
+                    raise CommandError('Please provide %s:' % (opt.dest,))
+
+        self.handle_validated(*args, **options)
+
+    def handle_validated(self, *args, **options):
+        try:
+            email_address = options['email-address']
+            enable = options['enable']
+            disable = options['disable']
+
+            if enable and disable:
+                raise CommandError('Please provide specify --enable or '
+                    '--disable, not both.')
+
+            user = User.objects.get(username=email_address)
+            user_api = vumi_api_for_user(user)
+            user_account_key = user_api.user_account_key
+            redis = user_api.api.redis
+
+            if enable:
+                redis.sadd('metrics_accounts', user_account_key)
+            elif disable:
+                redis.srem('metrics_accounts', user_account_key)
+
+        except User.DoesNotExist, e:
+            raise CommandError(e)
--- a/go/base/management/commands/go_manage_metrics.py
+++ b/go/base/management/commands/go_manage_metrics.py
@@ -44,7 +44,7 @@ class Command(BaseCommand):
             enable = options['enable']
             disable = options['disable']
 
-            if enable and disable:
+            if (enable and disable) or not (enable or disable):
                 raise CommandError('Please provide specify --enable or '
                     '--disable, not both.')
 
--- a/go/base/management/commands/go_manage_metrics.py
+++ b/go/base/management/commands/go_manage_metrics.py
@@ -45,8 +45,8 @@ class Command(BaseCommand):
             disable = options['disable']
 
             if (enable and disable) or not (enable or disable):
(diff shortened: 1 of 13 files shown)


go/vumitools/app_worker.py
```diff
@@ -17,6 +28,11 @@ def _go_validate_config(self):
         self.app_event_routing_config.update(
             self.config.get('app_event_routing', {}))
         self.control_consumer = None
+        # TODO: Make this less hacky?
+        self.metrics_prefix = self.config.get('metrics_prefix', None)
+        if self.metrics_prefix is None:
+            self.metrics_prefix = self.config['redis_manager']['key_prefix']
+        self.metrics_prefix += '.'
```
smn: Let's bite the bullet and work on a deployment wide configuration file that specifies sensible defaults for this stuff. There is a ticket for this somewhere.

go/vumitools/metrics_worker.py
```diff
@@ -0,0 +1,73 @@
...
+            conversations.extend(convs)
+        log.info(
+            "Processing metrics for %s conversations owned by %s users." % (
+                len(conversations), len(account_keys)))
+        for conversation in conversations:
+            yield self.send_metrics_command(conversation)
+
+    def find_account_keys(self):
+        return self.redis.smembers('metrics_accounts')
+
```
smn, line 63: We'd probably want to write a command line tool so we can add people to this set.

go/vumitools/conversation/models.py
```diff
@@ -46,6 +46,10 @@ class Conversation(Model):
     def ended(self):
         return self.end_timestamp is not None
 
+    def running(self):
+        # TODO: Better way to tell if we're running than looking for batches.
+        return self.batches.keys() and not self.ended()
```
hodgestar: If I can bikeshed a bit, I think it would be cool to have `.started()` and then `.running()` can be `self.started() and not self.ended()`.

go/vumitools/app_worker.py
```diff
@@ -63,7 +83,12 @@ def consume_control_command(self, command_message):
         if cmd_method:
             return cmd_method(*args, **kwargs)
         else:
-            return self.process_unknown_cmd(cmd_method_name, )
+            return self.process_unknown_cmd(cmd_method_name, *args, **kwargs)
+
+    def process_command_collect_metrics(self, conversation_key,
+                                        user_account_key):
+        # By default, we don't collect metrics.
+        pass
```
smn: We need to introduce something to prevent a metric aggregator build-up backlog that we will never be able to complete since we could be polling for metrics faster than we can aggregate them.

hodgestar: I think we can implement this reasonably straight-fowardly by having `.process_command_collector_metrics` call `.collect_metrics` and put some logic in `.process_command_collector_metrics` that checks to see whether a `.collect_metrics` is still in flight.

go/vumitools/metrics_worker.py
jerith, line 63: Ah, right. Your other scripts already landed, so this shouldn't be too hard to do.

go/vumitools/app_worker.py
```diff
@@ -2,11 +2,22 @@
...
 from vumi import log
 from vumi.application import ApplicationWorker
+from vumi.blinkenlights.metrics import MetricManager, Metric, MAX
 
 from go.vumitools.api import VumiApiCommand, VumiApi, VumiUserApi, VumiApiEvent
 from go.vumitools.api_worker import GoMessageMetadata
 
 
+class OneShotMetricManager(MetricManager):
+    def _clear_metrics(self):
```
dmaclay, line 15: maybe this should be in the parent class ?

jerith:
@dmaclay: Not without groveling through some other config. I originally stole it from redis.

go/vumitools/app_worker.py
jerith, line 15: I have a better solution which involves clearing just the metrics for a given conversation when the conversation ends, but that requires more infrastructure that we haven't built yet.

go/vumitools/metrics_worker.py
jerith, line 63: Done.

go/vumitools/tests/utils.py
```diff
@@ -247,3 +260,16 @@ def get_application(self, *args, **kw):
...
+    def poll_metrics(self, assert_prefix=None, app=None):
+        if app is None:
+            app = self.app
+        values = {}
+        if assert_prefix is not None:
+            assert_prefix += '.'
+        for name, metric in app.metrics._metrics_lookup.items():
+            if assert_prefix is not None:
+                _, sep, name = name.partition(assert_prefix)
+                self.assertEqual(sep, assert_prefix)
```
hodgestar: This looks like it's equivalent to `self.assertTrue(assert_prefix in name)`?

go/base/management/commands/go_manage_metrics.py
```diff
@@ -0,0 +1,62 @@
...
+    def handle_validated(self, *args, **options):
+        try:
+            email_address = options['email-address']
+            enable = options['enable']
+            disable = options['disable']
+
+            if enable and disable:
+                raise CommandError('Please provide specify --enable or '
+                    '--disable, not both.')
+
```
hodgestar, line 50: We should probably complain if neither `--enable` nor `--disable` are set too.

go/vumitools/app_worker.py
hodgestar, line 15: Can we have a TODO here to fix this class once we have things in Vumi?

go/vumitools/tests/utils.py
jerith: Pretty much, but the `.partition()` call also strips the prefix if there is one.

hodgestar:
:+1: once three minor comments are handled.

go/base/management/commands/go_manage_metrics.py
jerith, line 50: Probably. I stole and edited one of the other management commands.

hodgestar:
Otherwise, :+1:.

hodgestar:
:+1:

PR #92, opened 2012-10-17
VUMIGO-205 less celery

jerith:
This was actually easier to fix than I had expected. We now differentiate between sync and async message senders and require an AMQP client object to be passed into `VumiApi.from_config_async()` for the latter.

### DIFF
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -411,6 +411,7 @@ class AsyncMessageSender(object):
         self.publisher.publish_message(command)
 
     def make_publisher(self):
+        "Build a Publisher class with the right attributes on it."
         return type("VumiApiCommandPublisher", (Publisher,),
                     self.publisher_config.copy())
 
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -180,31 +180,30 @@ class VumiApi(object):
 
         # message sending API
         if sender is None:
-            sender = MessageSender({})
+            sender = MessageSender()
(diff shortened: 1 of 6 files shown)


hodgestar:
Could you explain what is being fixed?

jerith:
celery is unnecessary in Twisted-land, and it sets up a bunch of infrastructure when it's imported. I'd prefer not to have that extra stuff hanging around in our worker processes, especially if we accidentally use it and end up doing blocking network I/O to AMQP. (Also, I don't like the "config not found" warnings when running the non-Django tests.)

hodgestar:
The adhoc VumiApiCommands that have crept in and that were revealed by the deletion of the now-unused `.send()` and `.batch_send()` methods need sorting out but I'm happy for this PR to land. :+1:

dmaclay:
:+1: totally support replacing celery with std amqp

go/vumitools/api.py
```diff
@@ -414,15 +387,33 @@ def purge_pool(self, pool):
...
 
-class MessageSender(object):
-    def __init__(self, config):
+class SyncMessageSender(object):
+    def __init__(self):
+        self.publisher_config = VumiApiCommand.default_routing_config()
         from go.vumitools import api_celery
-        self.config = config
-        self.sender_api = api_celery
+        self.api_celery = api_celery
```
dmaclay, line 394: is this working towards eventual removal of celery ?

PR #94, opened 2012-10-19
Vxpolls improved exporting of user data.

smn:
This makes use of that and removes vxpolls.urls which was a hacky way of
providing the same features but with messier data.

This requires praekelt/vxpolls#37 to land first.

### DIFF
--- a/go/urls.py
+++ b/go/urls.py
@@ -37,7 +37,6 @@ urlpatterns = patterns('',
         include('go.apps.urls')),
     url(r'^contacts/', include('go.contacts.urls', namespace='contacts')),
     url(r'^account/', include('go.account.urls', namespace='account')),
-    url(r'^', include('vxpolls.urls')),
 )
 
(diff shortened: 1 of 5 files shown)


go/urls.py
```diff
@@ -37,7 +37,6 @@ def health(request):
         include('go.apps.urls')),
     url(r'^contacts/', include('go.contacts.urls', namespace='contacts')),
     url(r'^account/', include('go.account.urls', namespace='account')),
-    url(r'^', include('vxpolls.urls')),
```
dmaclay, line 40: what was this line doing ?

dmaclay:
looks ok :+1:

go/urls.py
smn, line 40: We were using the download views from vxpolls in the Go UI. This PR removes the need for that.

smn:
This kills the `export_results` URL since the CSV it generated was pretty much useless and generated more questions than it was useful.

PR #95, opened 2012-10-23
Fix stats script errors

smn:
The stats script is making Riak unhappy because we're pulling out all
objects. This change has the stats script use the new `by_index_count`
stuff to get the totals and then pulling all keys one by one to get the
per date breakdown. It's slower but unlikely to create problems for
Riak.

### DIFF
--- a/go/base/management/commands/go_account_stats.py
+++ b/go/base/management/commands/go_account_stats.py
@@ -117,9 +117,7 @@ class Command(BaseCommand):
             self.err(u'Provide a conversation key')
             return
         conv_key = options[0]
-        raw_conv = api.conversation_store.get_conversation_by_key(conv_key)
-
-        conversation = api.wrap_conversation(raw_conv)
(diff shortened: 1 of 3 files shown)


go/base/management/commands/go_account_stats.py
```diff
@@ -124,18 +118,34 @@ def handle_stats(self, user, api, options):
             return
         conv_key = options[0]
         raw_conv = api.conversation_store.get_conversation_by_key(conv_key)
-        conversation = api.wrap_conversation(raw_conv)
 
+        conversation = api.wrap_conversation(raw_conv)
```
jerith: You can use `api.get_wrapped_conversation(conv_key)` here.

hodgestar:
:+1:

jerith:
:+1:, although I'd prefer to avoid the mkmsg madness.

smn:
So would I, but not sure how to do so at the moment? Open for ideas.

dmaclay:
:+1:

PR #96, opened 2012-10-24
vumi-397 use riak less

jerith:
This is the vumi-go side of praekelt/vumi#397.


smn:
Do we have tests somewhere that test the celery stuff you ripped out as part of this PR?

smn:
:+1: in general though

jerith:
The celery stuff I ripped out actually got moved to the API tests, because we still send a message through celery to test the sync command sending. It's no longer used anywhere else in the Twisted side of things.

go/base/management/commands/go_account_stats.py
```diff
@@ -58,12 +58,14 @@ def handle(self, *args, **options):
...
+            handler = getattr(self, 'handle_%s' % (command,),
+                self.unknown_command)
+            handler(user, api, args[2:])
+
         except User.DoesNotExist:
             self.err(u'Account does not exist\n')
 
-        handler = getattr(self, 'handle_%s' % (command,), self.unknown_command)
-        handler(user, api, args[2:])
-
```
hodgestar, line 66: Was there a reason this moved?

jerith, line 66: Yes. The tests were failing and @smn fixed it. :-)

go/vumitools/contact/models.py
```diff
@@ -191,12 +191,18 @@ def filter_contacts_on_surname(self, letter, group=None):
         mr.add_bucket(bucket)
         if group is not None:
             mr.index(bucket, 'groups_bin', group.key)
+        # We need to filter out deleted values here. (Mostly for tests.)
+        # TODO: Make this a general thing?
```
hodgestar, line 197: Could you describe this issue to me a little more? When does Riak returned deleted values?

go/vumitools/tests/test_api.py
```diff
@@ -11,14 +13,24 @@
...
+from go.vumitools.tests.utils import (
+    AppWorkerTestCase, CeleryTestMixIn, DummyConsumerFactory)
+
+
+def dummy_consumer_factory_factory_factory(publish_func):
+    def dummy_consumer_factory_factory():
+        dummy_consumer_factory = DummyConsumerFactory()
+        dummy_consumer_factory.publish = publish_func
+        return dummy_consumer_factory
+    return dummy_consumer_factory_factory
```
hodgestar: I know the naming is kind of amusing but why don't we make `_celery_consumer_factory(self)` a method on the TestTxVumiApi test case? The we don't have to have quite as much "what the heck is this thing"? :)

go/vumitools/tests/test_contact.py
```diff
@@ -12,15 +12,17 @@
...
         self._persist_setUp()
         self.manager = self.get_riak_manager()
-        # yield self.manager.purge_all()
         self.account_store = AccountStore(self.manager)
-        self.account = yield self.account_store.new_user(u'user')
-        self.account_alt = yield self.account_store.new_user(u'other_user')
+        # We pass `self` in as the VumiApi object here, because mk_user() just
+        # grabs .account_store off it.
+        self.account = yield self.mk_user(self, u'user')
+        self.account_alt = yield self.mk_user(self, u'other_user')
```
hodgestar, line 25: Passing the test case in as the Django user (at least that's what I think it's supposed to be?) seems rather WAT. Isn't there a good utility method for this? If there isn't, maybe we need one? I'm not sure it's worth saving 2 lines (or 0 if you count that this has to be commented).

hodgestar, line 25: It's a VumiApi that's needed not a user? So it's only one extra line to avoid the commenting?

go/vumitools/tests/test_api.py
jerith: Good point. It was a little more generic in the past.

go/vumitools/tests/test_contact.py
jerith, line 25: We use this in a lot of places where we have a VumiApi instance to give it and grabbing the AccountStore off it is extra boilerplate and leads to line splitting. I think there's only one other place where we pass in something that isn't an AccountStore, versus almost every worker-level test that has a VumiApi object handy.

hodgestar:
:+1: I'm not wild about the tombstone issue but we're not going to fix that now unfortunately.

smn:
:+1:

PR #97, opened 2012-10-26
The WindowManager should understand `nack` messages

smn:
The WindowManager pops messages out of the flight window when an ack is
received. It should do the same for `nack` messages.

Wrote a `handle_event()` method at which we can throw any type of event,
would've liked to have called it `consume_event` but that's already
being used by the base class.

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -95,8 +95,8 @@ class TestBulkMessageApplication(AppWorkerTestCase):
         nack = self.mkmsg_nack(user_message_id=msg2['message_id'],
             nack_reason='unknown')
 
-        yield self.dispatch(ack, rkey='%s.event' % (self.transport_name,))
-        yield self.dispatch(nack, rkey='%s.event' % (self.transport_name,))
+        yield self.dispatch_event(ack)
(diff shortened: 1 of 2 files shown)


dmaclay:
:+1:

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -89,14 +89,14 @@ def test_start(self):
...
             sent_message_id=msg1['message_id'])
-        ack2 = self.mkmsg_ack(user_message_id=msg2['message_id'],
-            sent_message_id=msg2['message_id'])
+        nack = self.mkmsg_nack(user_message_id=msg2['message_id'],
+            nack_reason='unknown')
 
-        yield self.dispatch(ack1, rkey='%s.event' % (self.transport_name,))
-        yield self.dispatch(ack2, rkey='%s.event' % (self.transport_name,))
+        yield self.dispatch(ack, rkey='%s.event' % (self.transport_name,))
+        yield self.dispatch(nack, rkey='%s.event' % (self.transport_name,))
```
jerith: You can probably use `dispatch_event()` here and avoid mucking about with the rkey stuff.

jerith:
:+1:

PR #98, opened 2012-10-26
Use the new message caches in the Go UI.

smn:
Not yet ready for review, creating a pull request so I can get an
overview of what's still needing to be done.

First of all, tests.


smn:
Two tests remaining to be implemented, they're tested elsewhere as part of some other test but would prefer to have them tested here in isolation as well. Still deciding on how to go about those, will tackle them tomorrow. The rest is ready for an initial review before this grows much bigger.

Also, I need to add an option in the view somewhere where we can schedule a Celery task for now that somewhere triggers a cache reconciliation. My thoughts on this are, either:

a) straight in celery (potentially slower)
b) have celery send a VumiApiCommand to the application which then triggers an event which gets handled by the EventHandler so we can do it in Twisted land.

Thoughts @jerith  @hodgestar @dmaclay ?

jerith:
:+1:, with a slight concern about the Django version bump. (We'll need to update virtualenvs and stuff.)

We can defer the cache recon to another branch, because my riak keys changes will probably impact how that can best be done.

PR #99, opened 2012-10-29
Smallish cleanup pass

jerith:
There's a bunch of inconsistency in naming things and a bunch of stuff that might benefit from some little utility methods. This is a branch for those things that I've come across recently. I'll try keep each individual thing in its own commit.

### DIFF
--- a/go/vumitools/tests/utils.py
+++ b/go/vumitools/tests/utils.py
@@ -212,19 +212,23 @@ class GoPersistenceMixin(PersistenceMixin):
 class AppWorkerTestCase(GoPersistenceMixin, ApplicationTestCase):
 
     use_riak = True
-    worker_name = None
+
+    def _worker_name(self):
+        return self.application_class.worker_name
+
+    def _command_rkey(self):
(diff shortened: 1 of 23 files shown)


smn:
:+1:

jerith:
I think now's a good time to land this. I've cleaned up the things that were annoying me the most in my other branch and I can always do another of these later if necessary.

go/vumitools/tests/utils.py
```diff
@@ -212,19 +212,28 @@ def mk_user(self, vumi_api, username):
...
     use_riak = True
-    worker_name = None
+
+    def _worker_name(self):
+        return self.application_class.worker_name
+
+    def _conversation_type(self):
+        # This is a guess based on worker_name.
+        # We need a better way to do this.
+        return self._worker_name().rpartition('_')[0].decode('utf-8')
```
smn, line 222: Can we add a `FIXME` / `TODO` here so we can grep for it later?

smn:
:+1: but a `TODO` or something similar on the `_conversation_type()` function would be good.

PR #102, opened 2012-11-05
vumi-389 new mapreduce api

jerith:
This depends on praekelt/vumi#389.


go/apps/sequential_send/vumi_app.py
```diff
@@ -137,36 +135,13 @@ def send_scheduled_messages(self, conv):
                 continue
 
             yield self.send_message(
-                batch_id, to_addr, messages[message_index], message_options)
+                batch_id, contact.addr_for(conv.delivery_class),
+                messages[message_index], message_options)
```
smn: What if `contact.addr_for(conv.delivery_class)` cannot find a suitable addr?

go/vumitools/contact/models.py
```diff
@@ -162,26 +146,24 @@ def get_contacts_for_conversation(self, conversation):
...
-        returnValue(contacts)
+        return self.contacts.raw_search(group.query).get_keys()
 
     @Manager.calls_manager
     def filter_contacts_on_surname(self, letter, group=None):
+        # FIXME: This does a mapreduce over a bucket, which means hitting every
+        #        key in riak.
         # TODO: vumi.persist needs to have better ways of supporting
         #       generic map reduce functions. There's a bunch of boilerplate
         #       around getting bucket names and indexes that I'm doing
```
smn, line 166: Do we want to use the `filter_not_found()` for the Tombstones here?

go/vumitools/conversation/utils.py
```diff
@@ -182,6 +182,7 @@ def get_latest_batch_key(self):
                 using Redis' zunionstore to provide a temporary cached
                 view on all keys for a set of batch_ids)
         """
+        # TODO: Use the message cache for this?
         batch_keys = self.get_batch_keys()
```
smn, line 191: What are your thoughts here?

go/vumitools/conversation/utils.py
```diff
@@ -375,14 +374,49 @@ def get_opted_in_addresses(self):
...
+            contact_keys = contact_keys[100:]
+            current_contacts = yield contact_store.load_all_from_keys(
+                contact_store.contacts, current_keys)
+            addresses = self.get_contacts_addresses(current_contacts)
+            opt_out_keys = yield opt_out_store.opt_outs_for_addresses(
+                address_type, addresses)
+            opted_out_addrs = set(key.split(':')[1] for key in opt_out_keys)
+            opted_in_addrs.extend(addr for addr in addresses
+                                  if addr not in opted_out_addrs)
         returnValue(opted_in_addrs)
```
smn: Is there a way where we can prevent this duplication? It seems like if we can get the opted in contacts, we should be able to get their addresses without having to duplicate all that code since the contact already has the addr?

go/apps/sequential_send/vumi_app.py
jerith: Good point.

go/vumitools/contact/models.py
jerith, line 166: Not really, because that would add an extra stage to the mapreduce job -- better to do it all at once. We need to completely rewrite this anyway, so we can fix it up then.

go/vumitools/conversation/utils.py
jerith, line 191: I don't really know. This stuff needs a rewrite as part of the conversation/routing overhaul anyway, so I'm happy to worry about it then.

go/vumitools/conversation/utils.py
jerith: I want to replace this with the new tools for handling large numbers of objects from Riak. `get_opted_in_contacts()` is currently much more memory-efficient, which is necessary for large bulk-sends. Once we can more easily batch up this kind of thing, the efficiency issues go away and we can be more flexible without breaking stuff.

hodgestar:
:+1: once Vumi 389 is in.

PR #103, opened 2012-11-06
Use bytestrings when listing accounts with go_list_accounts

smn:
Kotzé was causing ascii errors.


hodgestar:
+1. In theory for stdouts connected to terminals `sys.stdout.encoding` is the sensible default encoding and for pipes it's `locale.getpreferredencoding()`. The whole area is a minefield though (there's also the PYTHONIOENCODING environment variable).

PR #106, opened 2012-11-08
Reduce GoApplicationRouter verbose logging.

smn:
Fixes some small issues:
1. The code didn't return from a situation where it couldn't continue
   anyway and as a result a second error was logged.
2. When we received an unroutable message we now log a warning and not
   an error since this almost exclusively happens with USSD.


hodgestar:
:+1:

PR #107, opened 2012-11-08
contact_for_addr fixes for twitter and tests

smn:
contact_for_addr was updated to work with twitter addresses &
delivery_classes. The merge didn't apply the stuff that now works with
keys and so things broke. Tests didn't tell us this because there were
no tests.


hodgestar:
:+1:

PR #108, opened 2012-11-08
VumiUserApi.applications is broken

smn:
It checks UserAppPermissions against a list of application names and as
a result failes returning back an empty dict.

Until this lands no one can start a new conversation on go.


hodgestar:
:+1: -- there was a lot of broken stuff our reviews didn't catch. :/

smn:
yeah, too much :(

PR #109, opened 2012-11-09
Shaving some deprecation warnings

smn:
In Django 1.5 the behaviour of {% url ... %} will change, since Django
1.3 it gives a deprecation warning. This changes loads the new url tag
from future and gets us ready for 1.5 (and removes the deprecation
warning).

There is still a warning about the function based generic views being
deprecated. That is caused by a 3rd party library (django-registration) 
using the deprecated `django.views.generic.simple`.

### DIFF
--- a/go/apps/multi_surveys/templates/multi_surveys/surveys.html
+++ b/go/apps/multi_surveys/templates/multi_surveys/surveys.html
@@ -47,7 +47,7 @@
                     {% for friendly_name, poll_id in polls.items %}
                         <tr>
                             <td>
-                                <a href="{% url "multi_survey:survey" conversation_key=conversation.key poll_name=friendly_name %}">
+                                <a href="{% url 'multi_survey:survey' conversation_key=conversation.key poll_name=friendly_name %}">
                                     {{conversation.subject}}, survey: {{friendly_name}}
                                 </a>
                             </td>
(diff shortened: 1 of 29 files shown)


dmaclay:
:+1:   that's a lot of " " 's

go/apps/multi_surveys/templates/multi_surveys/surveys.html
```diff
@@ -46,7 +47,7 @@
                     {% for friendly_name, poll_id in polls.items %}
                         <tr>
                             <td>
-                                <a href="{% url multi_survey:survey conversation_key=conversation.key,poll_name=friendly_name %}">
+                                <a href="{% url "multi_survey:survey" conversation_key=conversation.key poll_name=friendly_name %}">
```
hodgestar: I still find the nested unescaped double-quotes weird. I guess it works because the inner ones go away before anyone who cares gets to see them. Maybe I can just get used to it. :)

go/templates/registration/activation_email.txt
```diff
@@ -1,3 +1,3 @@
-Welcome to Vumi Go!
+{% load url from future %}Welcome to Vumi Go!
```
hodgestar, line 1: Can these be on separate lines or does that break something?

hodgestar:
:+1: with some minor comments.

go/templates/registration/activation_email.txt
smn, line 1: It'll create an empty newline after the tag, putting the welcome messages on the second line.

PR #110, opened 2012-11-16
Add MessageStore search to Vumi Go by using the API.

smn:
Ready for review.
1. Interfaces with the new MessageStore API for searching & grabbing results.
2. Results are grabbed asynchronously with Javascript & cached in Redis.
3. At some point we'll probably want a generic "run this map reduce and give me a token for the results" type of thing built into the API but not as part of this PR.


go/apps/bulk_message/tests/test_views.py
```diff
@@ -183,3 +183,13 @@ def test_show_cached_message_overview(self):
...
+
+    def test_message_search(self):
+        self.put_sample_messages_in_conversation(self.user_api,
+            self.conv_key, 10,
+            content_generator=('hello world %s' % (i,) for i in range(10)))
+        response = self.client.get('%s?q=%s' % (
+            reverse('bulk_message:show', kwargs={
+                'conversation_key': self.conv_key,
+            }), 'hello world 1'))
+        # print response.content
```
jerith: Debug print?

go/apps/tests/base.py
```diff
@@ -204,14 +204,17 @@ def get_api_commands_sent(self):
...
         conversation = user_api.get_wrapped_conversation(conversation_key)
         conversation.start()
         batch_key = conversation.get_latest_batch_key()
 
         for i in range(message_count):
             msg_in = self.mkmsg_in(from_addr='from-%s' % (i,),
-                message_id=TransportUserMessage.generate_id())
+                message_id=TransportUserMessage.generate_id(),
+                content=(content_generator.next() if content_generator
+                            else 'hello'))
```
jerith: This would probably be more readable as a local variable above the mkmsg call.

go/conversation/templatetags/conversation_tags.py
```diff
@@ -45,3 +67,22 @@ def show_conversation_messages(conversation, direction=None, page=None,
...
+
+    :param ConversationWrapper conversation:
+        The conversation who's inbound messages are to be searched.
+    :param str batch_id:
+        The batch_id used to find messages with for this conversation.
+    :param str query:
+        The search term.
+    """
+    matching_messages = conversation.match_inbound_messages(query)
+    print matching_messages
```
jerith: Debug print?

go/base/message_store_client.py
```diff
@@ -0,0 +1,111 @@
...
+
+class Client(object):
+
+    def __init__(self, base_url):
+        self.base_url = base_url
+
+    def match(self, batch_id, direction, query):
+        url = '%sbatch/%s/%s/match/' % (self.base_url, batch_id, direction)
+        response = requests.post(url, data=json.dumps(query))
+        return response.headers['x-vms-result-token']
```
hodgestar, line 29: Would it not make sense to just return MatchResult from here and mostly hide the token from the code using this?

go/conversation/base.py
```diff
@@ -274,6 +277,49 @@ def post(self, request, conversation):
         return self.redirect_to('show', conversation_key=conversation.key)
 
 
+class QueryResultsConversationView(ConversationView):
```
hodgestar: QueryResults seems like an overly generic name. Maybe MessagesConversationView?

go/conversation/base.py
```diff
@@ -339,6 +385,7 @@ class ConversationViews(object):
     show_conversation_view = ShowConversationView
     edit_conversation_view = EditConversationView
     end_conversation_view = EndConversationView
+    results_conversation_view = QueryResultsConversationView
```
hodgestar: I think `results_` is too generic here too. Maybe `messages_conversation_view`?

go/base/message_store_client.py
hodgestar, line 29: Nm. I see the token is passed around to requests.

go/base/message_store_client.py
```diff
@@ -0,0 +1,111 @@
...
+        cache_key = self._cache_key(start, stop)
+        cache_hit = self._cache.get(cache_key)
+        if cache_hit is not None:
+            return cache_hit
+
+        response = requests.get(self.url, params={
+            'token': self.token,
+            'start': start,
+            'stop': stop,
+        })
```
hodgestar: I think I'd be happier if all the requests happened via the client class (i.e. in one place in the code).

go/base/message_store_client.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def mk_message(self, dictionary):
+        return TransportUserMessage(_process_fields=False, **dictionary)
+
+    def count(self):
+        # Django Paginator calls this.
+        if self._total_count is None:
+            raise ClientException(
+                'count() called before page load.')
+        return self._total_count
```
hodgestar, line 100: What if we make page and page size mandatory and then we can move the result grabbing out of the constructor and just have them fetched when `.count()` is called if they're not available already?

hodgestar:
I made a bunch of comments on how we might structure this better, but I'm +1 on landing it and trying it out. It would be really cool if the client could be automatically constructed from the API -- in theory if we mark up the API / server side properly we can do proper parameter checking and get free client construction.

go/conversation/base.py
smn: I had something longer earlier but it got very long, it was: `MessageSearchResultConversationView` which makes sense since that's really all this does. Happy to name it to that if it's better.

go/conversation/base.py
smn: Updating but it's running over the 79 chars.

hodgestar:
:+1:

PR #111, opened 2012-11-19
VUMIGO-238 New Conversation model

jerith:
Full details in http://praekelt.onjira.com/browse/VUMIGO-238

This isn't yet ready to land, but early reviews would be appreciated.


go/vumitools/conversation/models.py
```diff
@@ -20,38 +21,89 @@
...
+            status = CONVERSATION_FINISHED
+        mdata.set_value('status', status, index='status_bin')
+
+        # Add indexes for fields with new (or updated) indexes
+        mdata.add_index('end_timestamp_bin', mdata.new_data['end_timestamp'])
+        mdata.add_index(
+            'start_timestamp_bin', mdata.new_data['start_timestamp'])
+        mdata.add_index('created_at_bin', mdata.new_data['created_at'])
+
+        return mdata
```
hodgestar: I think this should go into its own module.

go/vumitools/conversation/models.py
```diff
@@ -20,38 +21,89 @@
...
+CONVERSATION_DRAFT = u'draft'
+CONVERSATION_RUNNING = u'running'
+CONVERSATION_FINISHED = u'finished'
+
+
+class ConversationV1Migrator(ModelMigrator):
+    def migrate_from_unversioned(self, mdata):
+        # Migrator assertions
+        assert self.data_version is None
+        assert self.model_class.VERSION == 1
```
hodgestar: I think this assertion is wrong -- we might be doing a None -> 1 upgrade on a model class that is currently VERSION 2. The data_version is None assertion can probably be dropped too -- it's the migration infrastructures job to ensure this doesn't happen (it's just a distraction here).

go/vumitools/conversation/models.py
```diff
@@ -20,38 +21,89 @@
...
+
+        # Data assertions
+        assert mdata.old_data['VERSION'] is None
+        assert set(mdata.old_data.keys()) == set([
+            'VERSION',
+            'end_timestamp', 'conversation_type', 'start_timestamp',
+            'created_at', 'subject', 'metadata', 'message',
+            'delivery_class', 'delivery_tag', 'delivery_tag_pool',
+            ])
```
hodgestar: I'm -1 on asserting this stuff here. If we really care we should try build it into the migration infrastructure somehow.

go/vumitools/conversation/models.py
```diff
@@ -20,38 +21,89 @@
...
+        config = (mdata.old_data['metadata'] or {}).copy()
+        config['content'] = mdata.old_data['message']
+        mdata.set_value('config', config)
+
+        status = CONVERSATION_DRAFT
+        if mdata.new_index['batches_bin']:
+            # ^^^ This kind of hackery is part of the reason for the migration.
+            status = CONVERSATION_RUNNING
+        if mdata.new_data['end_timestamp'] is not None:
+            status = CONVERSATION_FINISHED
```
hodgestar: These constants may change in future versions which is why this migrator needs to be in it's own module. :)

go/vumitools/conversation/models.py
jerith: Good point. These were originally sanity checks while all the infrastructure was being built and I forgot to remove them.

go/vumitools/conversation/models.py
jerith: Yeah. I forgot to remove it after I built the copy stuff just below it.

I'd like to build some tools that assert that the data has the right "shape" both before and after the migration, but that's more work than I have time for at the moment.

go/vumitools/conversation/old_models.py
```diff
@@ -0,0 +1,80 @@
+# -*- test-case-name: go.vumitools.tests.test_conversation -*-
```
hodgestar: What are old models being kept around for?

jerith: They're used in tests and it's good to be able to refer back to them if it turns out our migrations are broken somehow.

hodgestar: We have version control for referring back. What are the tests doing with them? Testing migrations?

go/vumitools/conversation/models.py
```diff
@@ -121,3 +127,43 @@ def new_conversation(self, conversation_type, subject, message,
...
+        keys = yield self.conversations.index_lookup(
+            'end_timestamp', None).get_keys()
+        # NOTE: This assumes that we don't have very large numbers of active
+        #       conversations.
+        filtered_keys = []
+        for convs_bunch in self.load_all_bunches(keys):
+            for conv in (yield convs_bunch):
+                if conv.running:
+                    filtered_keys = conv.key
+        returnValue(filtered_keys)
```
hodgestar, line 153: Blerg. This is a case that really isn't handled well by our migrations currently. I think this means we need to add the migration tool to Vumi ASAP (i.e. ideally before trying to land all this stuff) or come up with a better plan (no good ideas yet).

go/vumitools/conversation/old_models.py
jerith: Yes. Constructing old versions that we push through the migration process and validate.

go/vumitools/conversation/models.py
```diff
@@ -101,10 +109,9 @@ def get_conversation_by_key(self, key):
         return self.conversations.load(key)
 
     @Manager.calls_manager
-    def new_conversation(self, conversation_type, subject, message,
-                         start_timestamp=None, **fields):
+    def new_conversation(self, conversation_type, name, config, **fields):
         conversation_id = uuid4().get_hex()
-        start_timestamp = start_timestamp or datetime.utcnow()
+        start_timestamp = fields.pop('start_timestamp', datetime.utcnow())
```
hodgestar: This minor change in behaviour seems unrelated to this PR and I'd rather keep it out unless there's actually a good need to make it now (especially since the new behaviour seems a bit obscure and there's suddenly no way to explicitly pass in a value that gives the behaviour of leaving the start_timestamp value out).

go/vumitools/conversation/old_models.py
hodgestar: I think +1 on keeping them for testing but maybe the test module in the line above needs to be fixed and perhaps the relevant tests need to be moved into test_old_models?

go/vumitools/conversation/models.py
```diff
@@ -20,38 +21,42 @@
...
-    start_timestamp = Timestamp()
+    name = Unicode(max_length=255)
+    conversation_type = Unicode(index=True)
+    config = Json(default=dict)
+
+    created_at = Timestamp(default=datetime.utcnow, index=True)
+    start_timestamp = Timestamp(index=True)
     end_timestamp = Timestamp(null=True, index=True)
-    created_at = Timestamp(default=datetime.utcnow)
+    status = Unicode(default=CONVERSATION_DRAFT, index=True)
```
hodgestar, line 43: Would be cool (and I think pretty straight forward) to add an EnumField or ChoicesField to Vumi for this.

go/vumitools/conversation/migrators.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class ConversationV1Migrator(ModelMigrator):
+    def migrate_from_unversioned(self, mdata):
+        # Copy stuff that hasn't changed between versions
+        mdata.copy_values(
+            'conversation_type',
+            'start_timestamp', 'end_timestamp', 'created_at',
+            'delivery_class', 'delivery_tag_pool', 'delivery_tag')
+        mdata.copy_indexes('user_account_bin', 'groups_bin', 'batches_bin')
```
hodgestar, line 12: Having to explicitly copy fields that aren't changed is going to make it very easy to mess up. Not quite sure what to suggest here though.

go/vumitools/conversation/models.py
```diff
@@ -20,38 +21,42 @@
...
+CONVERSATION_DRAFT = u'draft'
+CONVERSATION_RUNNING = u'running'
+CONVERSATION_FINISHED = u'finished'
 
 
 class Conversation(Model):
     """A conversation with an audience"""
+
+    VERSION = 1
+    MIGRATOR = ConversationV1Migrator
```
hodgestar: Why the V1 in the name here?

go/vumitools/conversation/models.py
jerith: I have no idea why this changed. It probably made sense in one of the variations I tried before getting to this one.

go/vumitools/conversation/models.py
jerith, line 153: This isn't so much a migration as a way of handling the broken mechanism for finding these conversations in the past. Ideally this hackery can go away once all the data's been migrated.

go/vumitools/conversation/models.py
jerith, line 43: Possibly. Not in this branch, though. :-)

go/vumitools/conversation/migrators.py
jerith, line 12: This is where some kind of migration generator (or validator) would be useful. That's a bit too big for this branch, though.

go/vumitools/conversation/models.py
jerith: Mostly left over from when I thought we might need to keep old models around to use in the migration process, and they might all still need their own migrators. I haven't entirely discarded that idea, but the V1 can be dropped from the migrator's name.

go/vumitools/conversation/old_models.py
jerith: The relevant tests are actually testing migrations to the new model version, so they need to stay where they are. I'll drop the test-case-name line, though.

go/conversation/templates/generic/show.html
```diff
@@ -11,7 +11,7 @@
     <div class="row">
         <div class="span12">
             <div id="welcome">
-                <h1><a href="{% url 'conversations:index' %}">Conversations</a> &raquo; {{conversation.subject}}</h1>
+                <h1><a href="{% url 'conversations:index' %}">Conversations</a> &raquo; {{conversation.nsme}}</h1>
```
smn: tyop

go/apps/multi_surveys/views.py
```diff
@@ -65,15 +65,11 @@ def new(request):
...
-                'message',
-                'delivery_class',
-            ]
-
-            for key in copy_keys:
-                conversation_data[key] = form.cleaned_data[key]
+            conversation_data = {
+                'name': form.cleaned_data['subject'],
+                'delivery_class': form.cleaned_data['delivery_class'],
+                'config': {u'content': form.cleaned_data['message']},
```
smn, line 71: Can we update the Form model to also use `content` as its key?

go/apps/wikipedia/ussd/views.py
```diff
@@ -21,17 +21,16 @@ def post(self, request):
...
-            'delivery_class',
-            'delivery_tag_pool',
-            ]
-        conversation_data = dict((k, form.cleaned_data[k]) for k in copy_keys)
-        conversation_data['metadata'] = {
-            'send_from_tagpool': form.cleaned_data['send_from_tagpool'],
-            'send_from_tag': form.cleaned_data['send_from_tag']
-        }
+        conversation_data = {
+            'name': form.cleaned_data['subject'],
```
smn, line 25: Ditto on the Form's `subject` rename.

go/vumitools/conversation/models.py
smn, line 153: Which, since our migrations are done lazily at load time is potentially never. How do we go about this?

jerith, line 153: We need a tool of some kind to force a migration of all the stored conversations. Until we've built and run such a tool, this code will find both old and new conversations. (I'm -1 on forcing a migration of everything upfront, though.)

go/vumitools/conversation/models.py
```diff
@@ -121,3 +128,43 @@ def new_conversation(self, conversation_type, subject, message,
...
+        # We need to list things by both the old-style 'end_timestamp' index
+        # and the new-style 'status' index, at least until we've migrated all
+        # old-style conversations to v1.
+        keys = yield self.conversations.index_lookup(
+            'status', CONVERSATION_RUNNING).get_keys()
+        keys.extend((yield self.conversations.index_lookup(
+            'status', CONVERSATION_DRAFT).get_keys()))
+        keys.extend((yield self.conversations.index_lookup(
+            'end_timestamp', None).get_keys()))
+        returnValue(keys)
```
smn: What are the chances duplicates showing up in here?

go/vumitools/conversation/models.py
hodgestar, line 153: Part of my concern is this the index and searching issue weakens our migration model fairly strongly. Once we delete code like this (and I have a feeling there might be other index searches or MR jobs we need to hack for this migration -- e.g. conversation search) encountering unmigrated objects is an error (should we have some way to flag it as such?) which makes keeping around all the earlier migration code much less useful.

jerith, line 153: As far as I can tell, this is the only place we're searching for conversations based on the "end_timestamp" index. Unmigrated conversations wouldn't really be an error, they just wouldn't be found by the index search once we've removed (or disabled) this bit. Old versions found some other way would be migrated as usual.

We could switch to listing all conversations (by the "user_account" index) and filtering by status, but that's a bunch of extra unnecessary work. Maybe it's worth doing that and having a config switch to swap between that and the index-only method?

In the general case, there's no way to handle this kind of thing gracefully. Part of the reason for having a new `Conversation` model version in the first place is to get rid of the old index search code that relies on a bug in how we store indexes with `None` as the value so we can fix that broken behaviour.

go/vumitools/conversation/models.py
jerith: Good catch. I was assuming that only old-version data would have "None" in the end_timestamp index, but anything that saves the model post-migration would write it back to the index.

go/apps/multi_surveys/views.py
jerith, line 71: I'd prefer to leave this for the more extensive view reworking that'll come with the next bout of yak shaving.

go/apps/wikipedia/ussd/views.py
jerith, line 25: See above.

go/vumitools/conversation/models.py
hodgestar, line 153: I checked my other concern (the conversation search form, but I see the query is handled in Python rather than via Riak map reduce (so isn't an issue in this case).

go/vumitools/conversation/migrators.py
hodgestar, line 12: Agreed.

go/vumitools/conversation/models.py
hodgestar: I think the V1 is still in migrator's name?

jerith:
Obsolete, see #203 which is basically this work on top of the new routing.

PR #113, opened 2012-11-21
VUMIGO-239 day of month scheduler

jerith:


go/apps/sequential_send/views.py
```diff
@@ -10,6 +10,7 @@
 
 class ScheduleForm(BootstrapForm):
     recurring = forms.CharField()
+    days = forms.CharField(required=False)
```
smn: If you add a `help_text` value here documenting that days can either be comma or space separated that would help.

smn:
:+1:

PR #114, opened 2012-11-27
Add setting for Google Analytics account

hodgestar:
Currently Go has code for using Google Analytics in `go/templates/base.html` but no way to set the account token for it.


smn:
:+1:

PR #115, opened 2012-11-27
Feature/vumigo 124 multi survey for mama needs opt back in capability

dmaclay:


go/apps/multi_surveys/vumi_app.py
```diff
@@ -50,11 +51,33 @@ def teardown_application(self):
...
         go = helper_metadata.get('go')
         helper_metadata['poll_id'] = 'poll-%s' % (
             go.get('conversation_key'),)
+
+        # It is possible the user is opted-out
+        # For the current use case (USSD sign-ups) we can
+        # now assume that if someone is dialing back in they
+        # want to register, therefore we can delete the opt-out
+        gmt = self.get_go_metadata(message)
+        account_key = yield gmt.get_account_key()
```
hodgestar, line 66: I think having no account key here is an error since that would mean that something was wrong with the conversation key (and then all the poll data would be bad).

go/apps/multi_surveys/vumi_app.py
```diff
@@ -50,11 +51,33 @@ def teardown_application(self):
...
+        if account_key is not None:
+            # check if user is opted out
+            opt_out_store = OptOutStore(self.manager, account_key)
+            from_addr = message.get("from_addr")
+            opt_out = yield opt_out_store.get_opt_out("msisdn", from_addr)
```
hodgestar, line 72: Assuming the from_addr is an msisdn is going to be wrong in a lot of cases. Is there a way we can make this more generic?

go/apps/multi_surveys/tests/test_vumi_app.py
```diff
@@ -233,3 +234,36 @@ def test_surveys_in_succession_demo_mode(self):
...
+
+        self.contact1 = yield self.create_contact(name=u'First',
+            surname=u'Contact', msisdn=u'27831234561', groups=[self.group])
+        self.create_survey(self.conversation)
+        with LogCatcher() as log:
+            yield self.start_conversation(self.conversation)
+            self.assertEqual(log.errors, [])
+
+        # First run through to the second poll
+        [msg1] = (yield self.wait_for_dispatched_messages(1))
```
hodgestar: You don't need brackets around yields when they're used as the complete RHS of an assignment.

go/apps/multi_surveys/tests/test_vumi_app.py
```diff
@@ -233,3 +234,36 @@ def test_surveys_in_succession_demo_mode(self):
...
+        opt_out_addr = '27831234561'
+        opt_out_store = OptOutStore(self.app.manager, self.user_account.key)
+        opt_out = yield opt_out_store.get_opt_out('msisdn', opt_out_addr)
+        yield opt_out_store.new_opt_out('msisdn', opt_out_addr,
+                                        {'message_id': u'test_message_id'})
+
+        # Check that on re-entry the survey is reset and the
+        # opening copy is delivered
+        yield self.reply_to(msg3, "1")
+        [msg1, msg2, msg3, msg4] = (yield self.wait_for_dispatched_messages(4))
```
hodgestar: You can call `self.clear_dispatched_messages()` to get rid of the ones you've looked at.

go/apps/multi_surveys/tests/test_vumi_app.py
```diff
@@ -233,3 +234,36 @@ def test_surveys_in_succession_demo_mode(self):
...
+        [msg1, msg2] = (yield self.wait_for_dispatched_messages(2))
+        self.assertEqual(msg2['content'], self.end_of_survey_copy[0])
+        yield self.reply_to(msg2, "1")
+        [msg1, msg2, msg3] = (yield self.wait_for_dispatched_messages(3))
+        self.assertEqual(msg3['content'], self.default_polls[1][0]['copy'])
+
+        # Now opt the msisdn out
+        opt_out_addr = '27831234561'
+        opt_out_store = OptOutStore(self.app.manager, self.user_account.key)
+        opt_out = yield opt_out_store.get_opt_out('msisdn', opt_out_addr)
```
hodgestar: Do we need this line? I can't see anything ahppening with `opt_out`?

go/apps/multi_surveys/tests/test_vumi_app.py
```diff
@@ -233,3 +234,36 @@ def test_surveys_in_succession_demo_mode(self):
...
+        yield opt_out_store.new_opt_out('msisdn', opt_out_addr,
+                                        {'message_id': u'test_message_id'})
+
+        # Check that on re-entry the survey is reset and the
+        # opening copy is delivered
+        yield self.reply_to(msg3, "1")
+        [msg1, msg2, msg3, msg4] = (yield self.wait_for_dispatched_messages(4))
+        self.assertEqual(msg4['content'], self.default_polls[0][0]['copy'])
```
hodgestar, line 272: Maybe we should also check that the opt_out is gone?

go/apps/multi_surveys/vumi_app.py
```diff
@@ -50,11 +51,33 @@ def teardown_application(self):
...
     def consume_user_message(self, message):
         helper_metadata = message['helper_metadata']
         go = helper_metadata.get('go')
         helper_metadata['poll_id'] = 'poll-%s' % (
             go.get('conversation_key'),)
+
+        # It is possible the user is opted-out
+        # For the current use case (USSD sign-ups) we can
+        # now assume that if someone is dialing back in they
+        # want to register, therefore we can delete the opt-out
```
hodgestar, line 64: If this is just for a specific poll, maybe it should be optional or configurable via the poll somehow?

go/apps/multi_surveys/vumi_app.py
dmaclay, line 66: I don't quite understand - this is checking for an account based opt-out (diff opt-out conversation via PCM) so I need the account_key

hodgestar, line 66: I meant more the line below (apologies). The `if` block seems the wrong thing to do since no account key means something is wrong not "oh well, let's not do this bit".

dmaclay, line 66: This app is live - so anything i add i want to fail in a non-disruptive manner - but yes, some else: log error at least would be in order

go/apps/multi_surveys/vumi_app.py
dmaclay, line 72: This app is msisdn only

go/apps/multi_surveys/tests/test_vumi_app.py
dmaclay, line 272: Done

go/apps/multi_surveys/tests/test_vumi_app.py
dmaclay: moved to final check that opt-out is gone

go/apps/multi_surveys/tests/test_vumi_app.py
dmaclay: done

go/apps/multi_surveys/tests/test_vumi_app.py
dmaclay: done

go/apps/multi_surveys/vumi_app.py
dmaclay, line 64: This is just for MAMA - I would not advise re-using this app :P

go/apps/multi_surveys/vumi_app.py
```diff
@@ -50,11 +51,36 @@ def teardown_application(self):
...
+        if account_key is not None:
+            # check if user is opted out
+            opt_out_store = OptOutStore(self.manager, account_key)
+            from_addr = message.get("from_addr")
+            opt_out = yield opt_out_store.get_opt_out("msisdn", from_addr)
+            if opt_out:
+                # delete the opt-out
+                yield opt_out_store.delete_opt_out("msisdn", from_addr)
+                # archive the user record so they can start from scratch
+                scope_id = message['helper_metadata'].get('poll_id', '')
```
hodgestar: We can just do `message['helper_metadata']['poll_id']` here since we set it at the top. Or better yet, define a local variable `poll_id` and use that in both cases. What is the difference between a `poll_id` and a `scope_id`, btw?

hodgestar:
Just the `poll_id` / `scope_id` comment, otherwise I'm +1.

go/apps/multi_surveys/vumi_app.py
dmaclay: I'll prob. switch to helper_metadata["poll_id"], the scope_id is just a vxpolls naming carry over.

PR #116, opened 2012-11-28
Make GoApplicationRouter logging less verbose.

smn:
Not including the full message in the reported error will allow Sentry
to group those reports and do more useful email reporting.

We are also adding event= and message= kwargs to the log.error() call
which we can intercept at a later stage to include in the data being
sent to Sentry for logging.


hodgestar:
:+1:

PR #43, opened 2012-08-30
Feature/vumigo 142 remove hacky hack hack

jerith:
This is an extension of #42 that removes `hacky_hack_hack()` completely and replaces RedisManager with TxRedisManager.

PR #117, opened 2012-12-03
VUMIGO-142 Remove hacky_hack_hack().

jerith:
This supersedes #43 and isn't several months old.

All tests pass, but @dmaclay should probably check it out.


hodgestar:
:+1:

dmaclay:
looks ok - i think we can land it
I'm no longer setup to do a manual test on a replica local deploy

:+1:

PR #123, opened 2012-12-05
Allow for exporting of aggregated message counts per day.

smn:
Adds a view and a test. Builds on
https://github.com/praekelt/vumi/pull/431/


jerith:
I'm not sure this makes sense for all conversation types, but fixing that would touch on a lot of the views rearchitecting I'm planning to do in the next few weeks. Perhaps we need a ticket to address that at the appropriate time?

dmaclay:
wouldn't  ZRANGEBYSCORE be better as the start / finish constraints can then be timestamp scores ?

dmaclay:
another option is ZCOUNT (as opposed to ZCARD) where you can also specify min/max per withscore param.
it might be more efficient with many messages to retrieve a sep count for each day rather than retrieve all messages and then group by day

smn:
I considered pulling the stats out per day but that ends up being more complex code for now and because zrange, zcount & zrangebyscore all have the same time complexity I'm not sure it is worth the effort. We could make concurrent zcount calls with Twisted but since this is called from Django land using the @calls_manager decorater I don't see that being better.

hodgestar:
I think :+1: from my side.

dmaclay:
:+1:

jerith:
:+1: now that the tests pass. :-)

PR #124, opened 2012-12-05
Add the aggregates to the survey app.

smn:
The survey app doesn't use the ConversationViews base yet and as a
result the aggregates view needs to be added manually.


hodgestar:
:+1:

PR #125, opened 2012-12-05
Fix event routing for when a tag has been released.

smn:
Currently if a tag is released (to be re-used or returned to the pool)
has the side effect of events not being able to be routed to the
appropriate application worker any more since we are using the tag to
find the conversation.

This changes that to use the batch_id instead which doesn't change after
a conversation is closed and/or a tag is reassigned.

### DIFF
--- a/go/vumitools/api_worker.py
+++ b/go/vumitools/api_worker.py
@@ -370,21 +370,31 @@ class GoApplicationRouter(BaseDispatchRouter):
     def find_application_for_event(self, event):
         """
         Look up the application for a given event by first looking up the
-        message that the event is for and then using
-        `find_application_for_msg()` to look up the application.
+        outbound message that the event is for and then using that messages'
+        batch to find which conversation it is associated with.
         """
         user_message_id = event.get('user_message_id')
         if user_message_id is None:
             log.error('Received event without user_message_id: %s' % (event,))
             return
 
-        message = yield self.vumi_api.mdb.get_outbound_message(user_message_id)
-        if message is None:
-            log.error('Unable to find message for event: %s' % (event,))
+        mdb = self.vumi_api.mdb
+        outbound_message = yield mdb.outbound_messages.load(user_message_id)
+        if outbound_message is None:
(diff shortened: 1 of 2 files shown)


go/vumitools/api_worker.py
```diff
@@ -370,21 +370,31 @@ def find_application_for_msg(self, msg):
...
             return
 
-        application = yield self.find_application_for_msg(message)
-        returnValue(application)
+        batch = yield outbound_message.batch.get()
+        account_key = batch.metadata['user_account']
+        user_api = self.vumi_api.get_user_api(account_key)
+        conversations = user_api.conversation_store.conversations
+        mr = conversations.index_lookup('batches', batch.key)
+        [conv_key] = yield mr.get_keys()
```
hodgestar, line 393: Do we want to log an error if no or multiple keys are found or is raising an exception fine? (I know it shouldn't happen)

go/vumitools/api_worker.py
```diff
@@ -370,21 +370,31 @@ def find_application_for_msg(self, msg):
...
         user_message_id = event.get('user_message_id')
         if user_message_id is None:
             log.error('Received event without user_message_id: %s' % (event,))
             return
 
-        message = yield self.vumi_api.mdb.get_outbound_message(user_message_id)
-        if message is None:
-            log.error('Unable to find message for event: %s' % (event,))
+        mdb = self.vumi_api.mdb
+        outbound_message = yield mdb.outbound_messages.load(user_message_id)
```
hodgestar, line 382: Was the change from `get_outbound_message` to `outbound_messages.load` intentional and if so, reason?

smn, line 382: `get_outbound_message()` gives you the `outbound_message.msg` object, I need the actual `outbound_message` since that has the `batch` attached.

go/vumitools/api_worker.py
smn, line 393: I think raising the exception is fine for now, it's not really a situation we can recover from at the moment I think?

jerith:
:+1:

PR #119, opened 2012-12-06
Add ConversationConfigResource to allow the sandbox access to conversation config.

hodgestar:
The use-case driving this is making JSON objects containing translations accessible to the JS code without having to doing something like make an HTTP call.

This ticket includes adding UI for updating config values from a URL.

### DIFF
--- a/go/apps/jsbox/forms.py
+++ b/go/apps/jsbox/forms.py
@@ -72,8 +72,6 @@ class JsboxAppConfigFormset(BaseFormSet):
             metadata[submeta['key']] = submeta
             del submeta['key']
             del submeta['DELETE']  # remove formset deletion marker
-
-        print metadata
         return metadata
 
     @staticmethod
--- a/go/apps/jsbox/forms.py
+++ b/go/apps/jsbox/forms.py
@@ -7,6 +7,20 @@ from django.forms.formsets import BaseFormSet
 from bootstrap.forms import BootstrapForm
 
 
+def possibly_load_from_url(url, default_value, update):
+    """Possibly load a value from a URL.
+
+    Returns the body of the response from the URL if update is True
+    and the URL is non-empty and a request to the URL returns
+    successfully. Otherwise returns the default value.
+    """
+    if update and url:
+        response = requests.get(url)
+        if response.ok:
+            return response.text
+    return default_value
+
+
 class JavascriptField(forms.CharField):
     widget = widgets.Textarea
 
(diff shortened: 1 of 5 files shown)


go/apps/jsbox/forms.py
```diff
@@ -28,3 +29,58 @@ def to_metadata(self):
...
+        metadata = {}
+        for form in self.forms:
+            if not form.cleaned_data or form in self.deleted_forms:
+                continue
+            submeta = form.to_metadata()
+            metadata[submeta['key']] = submeta
+            del submeta['key']
+            del submeta['DELETE']  # remove formset deletion marker
+
+        print metadata
```
smn: stray print.

go/apps/jsbox/forms.py
```diff
@@ -28,3 +29,58 @@ def to_metadata(self):
...
+            return response.text
+
+    def to_metadata(self):
+        metadata = self.cleaned_data.copy()
+        if metadata['update_from_source']:
+            source = self._update_from_source(metadata['source_url'])
+            if source is not None:
+                metadata['value'] = source
+        del metadata['update_from_source']
+        return metadata
```
smn: I suspect that if you use Django's `clean_<fieldname>` stuff here you can achieve the same but then just use `form.cleaned_data` where you're now using `.to_metadata()`

smn: Possibly relevant: https://docs.djangoproject.com/en/1.4/ref/forms/validation/

go/apps/jsbox/vumi_app.py
```diff
@@ -5,13 +5,27 @@
...
+    """Resource that provides access to conversation config."""
+
+    def handle_get(self, api, command):
+        key = command.get("key")
+        if key is None:
+            return self.reply(command, success=False)
+        conversation = self.app_worker.conversation_for_api(api)
+        app_config = conversation.metadata.get("jsbox_app_config", {})
+        key_config = app_config.get(key, {})
+        value = key_config.get('value')
```
smn, line 25: Do we absolutely need to have this `"value"` key? Can we not just pass in a dictionary of keys & values?

go/apps/jsbox/forms.py
hodgestar: I could potentially use `clean()` but this all crosses multiple fields so I think `clean_<fieldname>` wouldn't be the right place. This also isn't about cleaning form input -- it's about converting the input into what should be stored in the conversation metadata.

go/apps/jsbox/vumi_app.py
hodgestar, line 25: We need `value` so that we can also store `source_url`.

smn:
:+1:

PR #126, opened 2012-12-10
VUMIGO-247 sms confirmation to start conversation

smn:
Accounts need to be able to be configured to make conversation starting
a two step process where a URL + token is SMSd to a phone in order to
have the phone holder confirm the conversation that's about to be
started.


smn:
Has commits of #127 merged in already so review that first.

smn:
A test will randomly fail until #128 lands so review that also first.

smn:
The account config stuff now has its own PR #129, please review that first.

smn:
TokenManager specifics pull into separate PR #130, please review that first.

jerith:
The token scrubbing stuff worries me. I don't think we should be post-processing all displayed messages in case they contain sensitive data. Aside from the general messiness of this approach, it's too easy to forget to do that somewhere and leak tokens.

We should probably have a different mechanism for sending the confirmation tokens that doesn't result in a message belonging to the conversation being started before we actually start it.

smn:
@jerith ready for review again.

jerith:
Can you add a test around the display of sensitive messages? Aside from that, :+1:

jerith:
:+1:

PR #127, opened 2012-12-11
Displaying groups for a conversation in show.html was broken.

smn:
This was because it was directly calling `conv.groups.get_all()` which
no longer works since we're working with keys.

This fixes that by adding `conv.get_groups()`.
Also uncovered a bit where we weren't using `@Manager.calls_manager` & `yield` where we should've been doing so.


hodgestar:
:+1:

PR #128, opened 2012-12-11
VUMIGO-249 conv.get_latest_batch_key() is broken, this makes it a bit better.

smn:
It is still hideous though. The only way to fix it properly is to use
migrations since a batch should probably just have a timestamp of when
it was created.

The work-around we're doing now is what the previous `TODO` suggested:
1. Get all known keys
2. Get the latest timestamps for those keys
3. Sort in descending order and grab the first one.
4. Return that as the latest batch_key

If for none of the batch_keys there are any values stored in the cache
(which would mean no messages have been sent on that batch yet) then we
just return the first of whatever Riak gives us (random order) and hope
for the best.

### DIFF
--- a/go/vumitools/conversation/utils.py
+++ b/go/vumitools/conversation/utils.py
@@ -178,21 +178,51 @@ class ConversationWrapper(object):
         self.c.batches.add_key(batch_id)
         yield self.c.save()
 
+    @Manager.calls_manager
     def get_latest_batch_key(self):
         """
-        We're not storing timestamps on our batches and so we have no
-        way of telling which batch was the most recent for this conversation.
+        Here be dragons.
 
-        FIXME:  add timestamps to batches or remove the need for only allowing
-                access to the cache one batch_id at the time (possibly by
-                using Redis' zunionstore to provide a temporary cached
-                view on all keys for a set of batch_ids)
+        FIXME:  We're not storing timestamps on our batches and so we have no
+                accurate way of telling which batch was most recenty acquired
+                for this conversation.
+
+                Our current work around is looking at the cache to find out
+                which batch_key had the last outbound message sent and return
+                that batch_key
         """
-        # TODO: Use the message cache for this?
         batch_keys = self.get_batch_keys()
-        if batch_keys:
-            return batch_keys[0]
 
+        if not batch_keys:
+            return
+
+        # Cache this for however long this conversation object lives
+        if hasattr(self, '_latest_batch_key'):
+            returnValue(self._latest_batch_key)
+
+        # Loop over the batch_keys and find out which one was most recently
+        # used to send out a message.
+        batch_key_timestamps = []
+        for batch_key in batch_keys:
+            if (yield self.mdb.cache.count_outbound_message_keys(batch_key)):
+                [(_, timestamp)] = (yield self.mdb.get_outbound_message_keys(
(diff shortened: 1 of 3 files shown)


hodgestar:
Why do we need `get_latest_batch_key` at all? Shouldn't most of the count functions that use it return sums over all of the batches? I assume this is being triggered by the SMS approval messages that are now in a separate batch?

smn:
We could possibly do without it, the `batch_key` is the main thing for pulling stuff out of the `MessageStore` and the cache mirrors that approach. Are you suggesting we have the cache merge keys for multiple batches into a single sorted set?

jerith:
Missed a couple of `yield` additions:

```
lantea:vumi-go jerith$ ack '[^`]get_latest_batch' go/ | grep -v 'yield' | grep -v 'def'
go/apps/tests/base.py:214:        batch_key = conversation.get_latest_batch_key()
go/conversation/templatetags/conversation_tags.py:40:    batch_id = batch_id or conversation.get_latest_batch_key()
```

hodgestar:
Options for avoiding `get_latest_batch_key`:
- Have the cache merge messages for multiple batches
- Declare that each conversation has only one batch and remove support for multiple batches
- Allow conversations to have only one _active_ batch and record that somewhere on the conversation (might require a migration)

smn:
@jerith both of those are in Django land where the sync manager is used.

smn:
@hodgestar I think having the cache merge the batches makes sense but then the message store cache becomes a conversation cache and needs to live on the conversation also I think this'll become fairly complex to maintain.

I'd be in favour of a simpler approach (limiting to a single batch perhaps) but you're probably the best to decide on whether or not that makes sense with the current MessageStore & batch_id design.

smn:
Tried writing a migrator for the batch to add a timestamp but ran into trouble: https://github.com/praekelt/vumi/commit/76679da80d966db6a6c006e68b6aecae60bd78e8#L0R19

go/vumitools/conversation/utils.py
```diff
@@ -188,21 +188,56 @@ def start(self, no_batch_tag=False, acquire_tag=True, **extra_params):
...
+                accurate way of telling which batch was most recenty acquired
+                for this conversation.
 
-        FIXME:  add timestamps to batches or remove the need for only allowing
-                access to the cache one batch_id at the time (possibly by
-                using Redis' zunionstore to provide a temporary cached
-                view on all keys for a set of batch_ids)
+                Our current work around is looking at the cache to find out
+                which batch_key had the last outbound message sent and return
+                that batch_key
```
hodgestar, line 209: I think I might word this even more strongly -- the existence of `get_latest_batch_key` is itself a bug given that conversations assume that all batches are equivalent.

go/vumitools/conversation/utils.py
```diff
@@ -188,21 +188,56 @@ def start(self, no_batch_tag=False, acquire_tag=True, **extra_params):
...
         """
-        # TODO: Use the message cache for this?
         batch_keys = self.get_batch_keys()
-        if batch_keys:
-            return batch_keys[0]
 
+        if not batch_keys:
+            return
```
hodgestar: Maybe `return None` to be explicit?

hodgestar:
Other than my comment about the docstring, I'm :+1:

PR #129, opened 2012-12-11
VUMIGO-250 adding option to account for SMS confirmations.

smn:
These are largely commits pulled out of #126 to make the reviews easier.
It adds a boolean for whether or not SMS confirmations are wanted.
If that's checked then a valid phone number is required.

Also adds some updates to the account form, requiring the old / original
password be supplied before any changes are accepted.

Also using BootstrapForm better.


go/account/forms.py
```diff
@@ -1,21 +1,83 @@
...
+        Make a best effort guess at determining whether this msisdn
+        is a valid msisdn
+        """
+        msisdn = self.cleaned_data['msisdn'].lstrip('+')
+        if not msisdn:
+            return ''
+
+        if not (len(msisdn) > 5 and
+                all(c.isdigit() for c in msisdn)):
+            raise forms.ValidationError('Please provide a valid phone number.')
```
jerith, line 70: This also disallows spaces, dashes, parens, etc. which are often handy to use when typing phone numbers. Maybe strip these out at the same time as the leading `+`?

go/account/tests.py
```diff
@@ -1,16 +1,164 @@
...
-        """
-        self.assertEqual(1 + 1, 2)
+    def setUp(self):
+        super(AccountTestCase, self).setUp()
+        self.setup_riak_fixtures()
+        self.client = Client()
+        self.client.login(username='username', password='password')
+
+    def tearDown(self):
+        pass
```
jerith: Are we explicitly overriding a superclass `tearDown()` here? If so, can we document this and explain why? If not, can we remove this?

smn:
@jerith I've addressed your review comments, can I get a re-review?

jerith:
It would probably make more sense to keep these new fields on the Account object in Riak. The Django user stuff shouldn't be much more than auth information.

jerith:
:+1:

PR #130, opened 2012-12-11
VUMIGO-251 TokenManager

smn:
TokenManager for managing once-off auto-expiring tokens to Vumi Go. Used
for SMS confirmations.

Generates a token consisting of two parts. First part is submitted to
the user, the second part is looked up serverside and then redirected
to the specified `redirect_to` URL. That URL is responsible for ensuring
that the two parts are present and valid, it is the only way we have to
ensure that a user has actually arrived via the token.

Undecided on whether or not a token needs to be deleted when a
conversation is started or whether we should just leave that to Redis to
expire.

### DIFF
--- a/go/base/token_manager.py
+++ b/go/base/token_manager.py
@@ -76,12 +76,11 @@ class TokenManager(object):
 
     def get(self, token, verify=None):
         """
-        Retrieve the data for the given token, it it doesn't exist it'll
-        return an empty dictionary.
+        Retrieve the data for the given token. If there is no match then `None`
+        will be returned.
 
         :param str verify:
-            Provide the system_token if matching needs to occur. If there
-            is no match then an empty dictionary will be returned.
+            Provide the system_token if matching needs to occur.
         """
         if not self.redis.exists(token):
             return None
--- a/go/base/token_manager.py
+++ b/go/base/token_manager.py
@@ -62,17 +62,30 @@ class TokenManager(object):
             data that can be encoded as JSON.
         """
         lifetime = lifetime or self.DEFAULT_LIFETIME
-        a = token or self.generate_token()
-        user_token, system_token = a
+        user_token, system_token = token or self.generate_token()
         extra_params = extra_params or {}
-        self.redis.hmset(user_token, {
-            'redirect_to': redirect_to,
-            'user_id': user_id or '',
-            'system_token': system_token,
-            'extra_params': json.dumps(extra_params),
-            })
-        self.redis.expire(user_token, lifetime)
-        return user_token
+
+        # This is to avoid a possible race condition which could occur in
+        # `generate_token()` if two identical user_tokens are generated before
+        # either is stored.
+        if self.redis.hsetnx(user_token, 'system_token', system_token):
+            self.redis.hmset(user_token, {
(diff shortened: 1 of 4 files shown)


go/base/token_manager.py
```diff
@@ -0,0 +1,101 @@
...
+        :param int user_token_size:
+            How big the token sent to the user should be. Defaults to 6.
+        """
+        token_taken = True
+        while token_taken:
+            full_token = uuid4().hex
+            user_token = full_token[0:user_token_size]
+            system_token = full_token[user_token_size:]
+            token_taken = self.redis.exists(user_token)
+        return (user_token, system_token)
```
jerith, line 41: There's a potential race condition here if two tokens with the same `user_token` are generated before either is stored. I don't think it's really a major concern, but it's worth noting at least.

go/base/token_manager.py
```diff
@@ -0,0 +1,101 @@
...
+        lifetime = lifetime or self.DEFAULT_LIFETIME
+        a = token or self.generate_token()
+        user_token, system_token = a
+        extra_params = extra_params or {}
+        self.redis.hmset(user_token, {
+            'redirect_to': redirect_to,
+            'user_id': user_id or '',
+            'system_token': system_token,
+            'extra_params': json.dumps(extra_params),
+            })
```
jerith: It might be worth checking for the existence of the token here to catch the race condition noted above. There doesn't seem to be a way to atomically check-and-set multiple hash fields in redis, unfortunately.

go/base/token_manager.py
```diff
@@ -0,0 +1,101 @@
...
+        self.redis.expire(user_token, lifetime)
+        return user_token
+
+    def get(self, token, verify=None):
+        """
+        Retrieve the data for the given token, it it doesn't exist it'll
```
jerith: Tyop: "if it"

go/base/token_manager.py
```diff
@@ -0,0 +1,101 @@
...
+        return an empty dictionary.
```
jerith: This is documented as returning an empty dict if "it" (the token? the data?) doesn't exist. However, `None` is actually returned in a variety of cases.

go/base/token_manager.py
```diff
@@ -0,0 +1,101 @@
...
+        return token_data
+
+    def verify_get(self, full_token):
+        """
+        Retrieve the data for a full token as supplied to the `redirect_to`
+        URL.
+        """
+        user_token_length, _, token = full_token.partition('-')
+        user_token = token[0:int(user_token_length)]
+        system_token = token[int(user_token_length):]
```
jerith, line 117: I'd be happier with something a bit more robust here. If it gets an invalid string, it's likely to throw a "can't parse int" exception.

go/base/token_manager.py
hodgestar: We could just set one field (maybe system_token?) using `hsetnx` and use that to avoid the race condition.  We can set the rest of the fields later using `hmset` as we do now.

go/base/token_manager.py
smn, line 41: Fixed, just didn't touch this code.

hodgestar:
:+1:

PR #131, opened 2012-12-13
Add delayed contact group member counting (for static and dynamic groups)

hodgestar:
Required to avoid having to report these numbers manually every day while on leave. :)


hodgestar:
I opted to do this inline for now since it's not any worse speed wise that what's being done currently in displaying the group.

jerith:
I'd prefer to have a test around the new functionality, but otherwise :+1:

hodgestar:
Test added!

PR #121, opened 2012-12-19
Add metrics resource to Go sandbox

hodgestar:
So that JS apps can publish and update metrics.

Its not quite clear what this should look like yet. One option is to expose the metrics stuff relatively directly but that's probably not that useful so there might need to be an intervening layer that does some aggregation.

### DIFF
--- a/go/apps/jsbox/metrics.py
+++ b/go/apps/jsbox/metrics.py
@@ -7,56 +7,126 @@ import time
 import json
 import math
 
-from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.defer import inlineCallbacks, returnValue, maybeDeferred
+from twisted.internet.task import LoopingCall
 
 from vumi.application.sandbox import SandboxResource
+from vumi.persist.fields import Unicode, ForeignKey, JsonField
+from vumi.persist.model import Model
+from vumi import log
+from go.vumitools.account import UserAccount, PerAccountStore
 
 
 class MetricsResource(SandboxResource):
     """Resource that provides metric storing."""
 
-    def setup_resource(self):
-        self.metrics_store = MetricsStore()
+    # default number of seconds between metric updates
+    DEFAULT_METRIC_INTERVAL = 300
 
+    def setup_resource(self):
+        metric_interval = self.config.get('metric_interval',
+                                          self.DEFAULT_METRIC_INTERVAL)
+        self.metrics_manager = MetricsStoreManager(
+            self.app_worker.vumi_api, metric_interval,
+            update_callback=self._update_holodeck)
+
+    @inlineCallbacks
     def teardown_resource(self):
+        yield self.metrics_manager.stop()
+
+    def _update_holodeck(self, user_account_key, store):
+        # TODO: implement
         pass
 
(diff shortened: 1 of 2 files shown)


go/apps/jsbox/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+    def _event_for_command(self, etype, command):
+        return MetricEvent(event=etype, store=command.get('store', 'default'),
+                           metric=command['metric'],
+                           value=float(command['value']))
+
+    @inlineCallbacks
+    def _process_event_command(self, etype, user_account_key, command):
+        try:
+            event = self._event_for_command(etype, command)
+        except Exception, e:
```
smn, line 50: Why not just a `KeyError` here?

go/apps/jsbox/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+        return self.redis.getset(last_timebucket_key, new_timebucket)
+
+    def timebucket(self, timestamp):
+        intervals = math.floor(timestamp / self.metric_interval)
+        return intervals * self.metric_interval
+
+    def push_event(self, store_id, event):
+        events_key = self.events_key(store_id)
+        timestamp = event['timestamp']
+        return self.redis.zadd(events_key, timestamp, event.to_json())
```
smn, line 163: Should we `float()` the `timestamp` here just to be safe? It can be provided manually to the `MetricEvent` and isn't checked anywhere.

go/apps/jsbox/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+        start = "(%f" % start
+        end = "%f" % end
+        raw_events = yield self.redis.zrangebyscore(events_key, start, end)
+        returnValue([MetricEvent.from_json(ev) for ev in raw_events])
+
+    def delete_events(self, store_id, start, end):
+        events_key = self.events_key(store_id)
+        start = "(%f" % start
+        end = "%f" % end
+        return self.redis.zremrangebyscore(events_key, start, end)
```
smn, line 177: Would it be possible that by the time `delete_events()` is called, new stuff has been added in between this `start` and `end` range that wasn't processed yet? Since the `timestamp` is used as the weight in theory it shouldn't be possible but I'm thinking of a scenario where our queues are backing up and we're using the timestamp value from a message that arrived late from the queue. (sorry commented on the commit earlier, copied & pasted to the PR now)

go/apps/jsbox/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+        events_key = self.events_key(store_id)
+        events = yield self.redis.zrange(events_key, 0, 1)
+        if not events:
+            returnValue(None)
+        returnValue(json.loads(events[0]))
+
+    @inlineCallbacks
+    def add_event(self, user_account_id, event):
+        store_id = self.store_id(user_account_id, event.store)
+        yield self.mark_store_updated(store_id)
```
smn, line 190: This function doesn't seem to exist?

hodgestar:
This pull-request is being deprecated in favour of https://github.com/praekelt/vumi-go/pull/134

PR #133, opened 2012-12-20
VUMIGO-247 reuse existing batch id for conversation.

smn:
After a sending an token via SMS we're re-using the already acquired
batch_id when starting the conversation after confirmation.


jerith:
:+1:

PR #134, opened 2012-12-20
Implement a metrics resource using Vumi blinkenlights metrics publishers.

hodgestar:


go/vumitools/app_worker.py
```diff
@@ -230,6 +230,16 @@ def publish_conversation_metric(self, conversation, name, value, agg=None):
             conversation.user_account.key, conversation.key, name)
         self.publish_metric(name, value, agg)
 
+    def publish_account_metric(self, user_account, store, name, value,
+                               agg=None):
+        if isinstance(user_account, basestring):
+            user_account_key = user_account
+        else:
+            user_account_key = user_account.key
```
jerith: Do we really need to accept both of these here? I'd prefer to have the caller extract the key if necessary.

hodgestar: I don't really mind -- I was just trying to fit in with `publish_conversation_metric` which takes a `conversation` but not require people to look up the account if they only have the key.

jerith:
:+1:

hodgestar:
This replaced https://github.com/praekelt/vumi-go/pull/121.

go/apps/jsbox/tests/test_metrics.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_handle_fire(self):
+        cmd = SandboxCommand(metric="foo", value=1.5, agg='sum')
+        reply = self.resource.handle_fire(self.dummy_api, cmd)
+        self.check_reply(reply, cmd, True)
+        self.check_publish('default', 'foo', 1.5, self.SUM)
+
+    def _test_error(self, cmd, expected_error):
+        with LogCatcher() as lc:
+            reply = self.resource.handle_fire(self.dummy_api, cmd)
+            self.assertEqual(lc.messages(), [expected_error])
```
smn, line 126: Why isn't this in `lc.errors` ?

hodgestar, line 126: Because we use `log.warning` to log the error. :)

PR #135, opened 2012-12-22
Add resource for sending to other tags from JS sandbox

hodgestar:
Currently the JS sandbox allows replying via the tag the incoming message was received on. Sometimes one needs to send messages out via other tags but this should only be a tag currently owned by another conversation within the same account.

Once we have the new endpoints / Go routing in place, this can allow sending out via arbitrary endpoints attached to the JS sandbox conversation.

### DIFF
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -17,8 +17,7 @@ class GoOutboundResource(OutboundResource):
         tag = (command.get('tagpool'), command.get('tag'))
         content = command.get('content')
         to_addr = command.get('to_addr')
-        if (tag[0] is None or tag[1] is None or content is None or
-                to_addr is None):
+        if None in (tag[0], tag[1], content, to_addr):
             returnValue(self.reply(
                 command, success=False,
                 reason="Tag, content or to_addr not specified"))
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -1,0 +1,26 @@
+# -*- test-case-name: go.apps.jsbox.tests.test_outbound -*-
+# -*- coding: utf-8 -*-
+
+"""Outbound message resource for JS Box sandboxes"""
+
+from vumi.application.sandbox import OutboundResource
+from vumi import log
+
+
+class GoOutboundResource(OutboundResource):
+    """Resource that provides outbound message support for Go."""
+
+    def handle_send_to_tag(self, api, command):
+        tagpool = command.get('tagpool')
+        tag = command.get('tag')
+        content = command['content']
+        to_addr = command['to_addr']
+        # TODO: check tagpool / tag is owned by an active conversation
+        # TODO: fetch tagpool metadata for tagpool
+        msg_options = {}
+        self.app_worker.send_to(to_addr, content, **msg_options)
+
+    def handle_send_to(self, api, command):
+        # Generic sending is not supported in Vumi Go sandboxes
+        return self.reply(command, success=False,
+                          reason="Generic sending not supported in Vumi Go")
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -3,6 +3,8 @@
 
 """Outbound message resource for JS Box sandboxes"""
 
+from twisted.internet.defer import inlineCallbacks, returnValue
+
 from vumi.application.sandbox import OutboundResource
 from vumi import log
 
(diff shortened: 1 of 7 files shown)


go/apps/jsbox/outbound.py
```diff
@@ -0,0 +1,42 @@
...
+class GoOutboundResource(OutboundResource):
+    """Resource that provides outbound message support for Go."""
+
+    @inlineCallbacks
+    def handle_send_to_tag(self, api, command):
+        tag = (command.get('tagpool'), command.get('tag'))
+        content = command.get('content')
+        to_addr = command.get('to_addr')
+        if (tag[0] is None or tag[1] is None or content is None or
+                to_addr is None):
```
jerith: `if None in (tag[0], tag[1], content, to_addr):` might be neater.

go/apps/jsbox/outbound.py
```diff
@@ -0,0 +1,42 @@
...
+        user_api = self.app_worker.user_api_for_api(api)
+        tags = yield user_api.list_endpoints()
+        if tag not in tags:
+            returnValue(self.reply(command, success=False,
+                                   reason="Tag %r not held by account"
+                                   % (tag,)))
+        msg_options = yield user_api.msg_options(tag)
+        yield self.app_worker.send_to(to_addr, content, **msg_options)
+
+        returnValue(self.reply(command, success=True))
```
jerith, line 36: Are there any potential failure conditions in the `.send_to()` above?

go/apps/jsbox/tests/test_outbound.py
```diff
@@ -0,0 +1,88 @@
...
+        self.app_worker.conversation_for_api = mock.Mock(
+            return_value=self.conversation)
+        self.app_worker.user_api_for_api = mock.Mock(
+            return_value=self.user_api)
+        self.user_api.list_endpoints = mock.Mock(
+            return_value=set([('pool1', '1234'), ('pool2', '1234')]))
+        self.user_api.msg_options = mock.Mock(
+            return_value={'opt1': 'bar'})
+
+    def check_reply(self, reply, cmd, success):
```
jerith: I generally name this kind of method `assert_reply` or whatever. It just makes it a little clearer that it's an assertion when we see it called in test methods.

go/apps/jsbox/tests/test_outbound.py
```diff
@@ -0,0 +1,88 @@
...
+            handler = self.resource.handle_send_to_tag
+        cmd = SandboxCommand(**cmd_args)
+        reply = yield handler(self.dummy_api, cmd)
+        self.check_reply(reply, cmd, False)
+        self.check_not_sent()
+        self.assertEqual(reply['reason'], reason)
+
+    def test_send_to_fails(self):
+        return self.check_fails("Generic sending not supported in Vumi Go",
+                                handler=self.resource.handle_send_to)
```
jerith: The lack of command args makes this test a little hard to decipher without looking at the definition of `check_fails()`. Maybe an explanatory comment would help?

go/apps/jsbox/tests/test_outbound.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_send_to_tag(self):
+        cmd = SandboxCommand(tagpool='pool1', tag='1234', to_addr='6789',
+                             content='bar')
+        with LogCatcher() as lc:
+            reply = yield self.resource.handle_send_to_tag(self.dummy_api, cmd)
+            self.assertEqual(lc.messages(), [
+                "Sending outbound message to '6789' via tag ('pool1', '1234'),"
+                " content: 'bar'"])
+        self.check_reply(reply, cmd, True)
+        self.check_sent('6789', 'bar', **{'opt1': 'bar'})
```
jerith: Is the `**{}` syntax necessary here? Can we just use a keyword arg instead?

go/vumitools/api.py
```diff
@@ -17,12 +17,14 @@
...
 from vumi.persist.redis_manager import RedisManager
 from vumi.persist.txredis_manager import TxRedisManager
+from vumi.middleware.tagger import TaggingMiddleware
 
 from go.vumitools.account import AccountStore
 from go.vumitools.contact import ContactStore
 from go.vumitools.conversation import ConversationStore
 from go.vumitools.conversation.utils import ConversationWrapper
 from go.vumitools.credit import CreditManager
+from go.vumitools.middleware import DebitAccountMiddleware
```
jerith, line 27: This looks like a potential circular dependency risk. Maybe we need to move the functionality we import this for to somewhere better?

jerith:
:+1: modulo the style comments above. My only real concern is the potential circular dependency risk and that's fairly minor.

go/apps/jsbox/outbound.py
hodgestar, line 36: There are (missing 'default' from SEND_TO_TAGS, failing to publish) but they all represent serious issues external to the sandbox resource so I think letting higher level things deal with them is fine.

go/apps/jsbox/tests/test_outbound.py
hodgestar: I was mostly just being consistent with the tests I wrote for the metrics resource but I think assert_ is clearer so I changed the ones in this branch.

go/apps/jsbox/outbound.py
jerith, line 36: Maybe a generic exception handler and a "your message wasn't sent" reply?

go/apps/jsbox/tests/test_outbound.py
hodgestar: I just thought it made it clearer which of the arguments are the msg_options (since these are commonly passed in as a dict using **). Let me actually assign them to a local variable rather.

hodgestar: I just removed the *\* from the call and the function arguments instead.

go/vumitools/api.py
hodgestar, line 27: Agreed. It's not an immediate issue though and I'm keen to leave it for when we actually deploy the credit debiting stuff.

go/apps/jsbox/outbound.py
hodgestar, line 36: I think that will just hide critical errors from us and make the reporting of them happen via users (not great).

PR #136, opened 2012-12-24
VUMIGO-254 We need a way of telling if a user_account_key is valid

smn:
Currently we assume it's valid but then only find out it's not by
crashing when we try and do something useful.

This adds a helper to verify the account key.


hodgestar:
:+1:

PR #139, opened 2013-01-07
VUMIGO-253 limits on smart groups

smn:
Since smart groups might return anything from 0 to millions of keys
without any option of providing pagination or sorting it is a good idea
to provide a sensible default that can be overriden.

Some of these changes are already live on the server as they needed to
be put in place as an emergency measure over the break.

### DIFF
--- a/go/contacts/tests.py
+++ b/go/contacts/tests.py
@@ -470,6 +470,22 @@ class GroupsTestCase(DjangoGoApplicationTestCase):
         })
         self.assertContains(response, person_url(self.contact_key))
 
+    def test_group_contact_query_limits(self):
+        default_limit = self.client.get(group_url(self.group_key), {
+            'q': TEST_CONTACT_NAME,
+        })
+        custom_limit = self.client.get(group_url(self.group_key), {
+            'q': TEST_CONTACT_NAME,
+            'limit': 10,
+        })
+        no_limit = self.client.get(group_url(self.group_key), {
+            'q': TEST_CONTACT_NAME,
+            'limit': 0,
+        })
+        self.assertContains(default_limit, 'Showing 100 random contacts')
+        self.assertContains(custom_limit, 'Showing 10 random contacts')
+        self.assertNotContains(no_limit, 'Showing 100 random contacts')
+
     def test_group_contact_filter_by_letter(self):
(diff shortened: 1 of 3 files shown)


go/contacts/tests.py
```diff
@@ -470,6 +470,22 @@ def test_group_contact_querying(self):
...
+            'q': TEST_CONTACT_NAME,
+            'limit': 10,
+        })
+        no_limit = self.client.get(group_url(self.group_key), {
+            'q': TEST_CONTACT_NAME,
+            'limit': 0,
+        })
+        self.assertContains(default_limit, 'Showing 100 random contacts')
+        self.assertContains(custom_limit, 'Showing 10 random contacts')
+        self.assertNotContains(no_limit, 'Showing 100 random contacts')
```
jerith: Can we test a "no limit" different from the "default limit"?

go/contacts/views.py
```diff
@@ -315,7 +327,12 @@ def _people(request):
     if query:
         if not ':' in query:
             query = 'name:%s' % (query,)
+        limit = int(request.GET.get('limit', 100))
         keys = contact_store.contacts.raw_search(query).get_keys()
+        if limit:
+            messages.info(request,
+                'Showing %s random contacts matching your query' % (limit,))
+            keys = keys[:limit]
```
hodgestar, line 338: Where does the limit get passed into the views from?

smn, line 338: It doesn't yet, I just wanted to have the option of specifying the limit. Generally I don't want to show more than 100 but we can show the total count though (which you added as a feature iirc).

go/contacts/tests.py
smn: Yeah, I'm not quite sure how we'd do that. We could check that the whole of Django's messages framework HTML tags don't exist but that might be a bit too rigid.

go/contacts/views.py
hodgestar, line 338: Maybe we only want to warn about the limit if the limit is < the total count?

smn, line 338: Thanks, I'll actually change the copy to say `"Show a maximum of %s random ..."`

hodgestar:
I'm :+1: on getting this in since it's a reasonable start.

jerith:
:+1:

PR #141, opened 2013-01-10
VUMIGO-258 seq send should ignore ended conversations.

smn:
An ended conversation shouldn't be allowed to send out more messages
even if the scheduled slot matches.

I'm deliberately not removing them from the set of
scheduled_conversations because of the posibility of at one point being
able to re-open a conversation. If and when that happens the scheduling
should kick in again.


jerith:
:+1: for now. I'd like to revisit this when we have better conversation lifecycle management, but this is a reasonable stopgap.

dmaclay:
:+1:  ... possibly test an ended and re-opened conv ?

smn:
@dmaclay we don't have the ability to re-open a conversation yet, the way this fix is implemented doesn't eliminate the option of conversations over being re-opened again (if we develop that).

PR #142, opened 2013-01-10
VUMIGO-260 export contact data as CSV

smn:
Provides a way of exporting contact data as a CSV file via email.
Fires a task off to celery which does the CSV writing and emailing.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -1,15 +1,17 @@
 import sys
 import traceback
+from StringIO import StringIO
 
 from celery.task import task
 
 from django.conf import settings
-from django.core.mail import send_mail
+from django.core.mail import send_mail, EmailMessage
 from django.template.loader import render_to_string
 from django.utils.safestring import mark_safe
 
 from go.vumitools.api import VumiUserApi
 from go.base.models import UserProfile
+from go.base.utils import UnicodeCSVWriter
 from go.contacts.parsers import ContactFileParser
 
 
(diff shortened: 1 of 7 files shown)


go/contacts/tasks.py
```diff
@@ -48,6 +50,72 @@ def delete_group_contacts(account_key, group_key):
...
+    group = contact_store.get_group(group_key)
+    contact_keys = contact_store.get_contacts_for_group(group)
+
+    # Get the profile for this user so we can email them when the import
+    # has been completed.
+    user_profile = UserProfile.objects.get(user_account=account_key)
+
+    io = StringIO()
+    writer = UnicodeCSVWriter(io)
+    contacts = sorted([contact_store.contacts.load(contact_key)
```
hodgestar, line 77: Would it be better to load them in bunches to reduce the number of HTTP calls to riak?

go/contacts/tasks.py
```diff
@@ -48,6 +50,72 @@ def delete_group_contacts(account_key, group_key):
...
+                        key=lambda c: c.created_at)
+
+    # collect all names that we can export
+    fields = ['name', 'surname', 'email_address', 'msisdn', 'dob',
+                    'twitter_handle', 'facebook_id', 'bbm_pin', 'gtalk_id',
+                    'created_at']
+
+    # Collect the possible field names for this set of contacts, depending
+    # the number of contacts found this could be potentially expensive.
+    extra_fields = []
```
hodgestar: May use `extra_fields = set()` and then  `extra_fields.update(contact.extra)`.

go/contacts/tasks.py
```diff
@@ -48,6 +50,72 @@ def delete_group_contacts(account_key, group_key):
...
+    if include_extra:
+        for contact in contacts:
+            for key in contact.extra.keys():
+                if key not in extra_fields:
+                    extra_fields.append('extra-%s' % (key,))
+
+    # write the header
+    writer.writerow(fields + extra_fields)
```
hodgestar, line 98: I would also sort extra_fields so the ordering is consistent.

go/contacts/tasks.py
```diff
@@ -48,6 +50,72 @@ def delete_group_contacts(account_key, group_key):
...
+
+    # loop over the contacts and create the row populated with
+    # the values of the selected fields.
+    for contact in contacts:
+        row = []
+        for field in fields:
```
hodgestar: Maybe `row = [unicode(getattr(contact, field, None) or '') field in fields]` and `row.extend(unicode(contact.extra.get(extra) or '') for extra in extra_fields)`. Assumes `extra_fields = sorted(extra_fields)` at some point if `extra_fields` becomes a set.

hodgestar:
Some comments but otherwise :+1:

go/contacts/tasks.py
smn, line 77: This happens in Django land, bunches don't make a difference here.

hodgestar, line 77: I think they do -- load_all_bunches appears to use a map-reduce to load multiple objects in a single HTTP call regardless of whether the manager is `riak_manager` or `txriak_manager`.

go/contacts/tasks.py
```diff
@@ -48,6 +50,75 @@ def delete_group_contacts(account_key, group_key):
...
+    fields = ['name', 'surname', 'email_address', 'msisdn', 'dob',
+                    'twitter_handle', 'facebook_id', 'bbm_pin', 'gtalk_id',
+                    'created_at']
+
+    # Collect the possible field names for this set of contacts, depending
+    # the number of contacts found this could be potentially expensive.
+    extra_fields = set()
+    if include_extra:
+        for contact in contacts:
+            extra_fields.update(['extra-%s' (key,)
```
hodgestar, line 91: I'm a bit surprised the `extra-` doesn't make the look-up of the key in `contact.extra` later fail? I would have thought the `extra-` would only be included in the header row maybe?

smn, line 91: This is addressed in #144

PR #144, opened 2013-01-11
VUMIGO-260 fixes for dynamic fields.

smn:
@hodgestar spotted a problem in the previous PR after I had landed it
(probably too early). This addresses those.


hodgestar:
:+1:

PR #145, opened 2013-01-11
VUMIGO-260 fix ordering of extras on export

smn:
There was a subtle bug in the export script causing the ordering of the
values from the dynamic proxy being different from the column headers.


jerith:
:+1:

PR #148, opened 2013-01-15
VUMIGO-262 Use fab for deploys

smn:
Adds some helpers for automating Go deploys & pulling in latest Vumi.
Also adds a helper for issuing random supervisorctl commands from the
command line.


jerith:
:+1:

PR #149, opened 2013-01-15
VUMIGO-263 survey conversation message search broken

smn:
The bits for search were never hooked up for surveys. This fixes that.
Also, it turns out that Django resolves a Page instance to False if that
page has no results. Might be reasonable but our code didn't work with
that. Changed to look if the paginator has any pages to show instead.

### DIFF
--- a/go/apps/surveys/templates/surveys/show.html
+++ b/go/apps/surveys/templates/surveys/show.html
@@ -27,7 +27,7 @@
 
     {% include "surveys/includes/download.html" %}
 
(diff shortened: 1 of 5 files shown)


hodgestar:
:+1:

dmaclay:
:+1: ... regarding test suggestions - obviously you're aware of the zero result possibility, but what if the search hangs/times out ? At the weird end of things to test I know, but one of those things that happens in big live datasets.

hodgestar:
:+1: on test and tiny bug fix.

PR #150, opened 2013-01-17
VUMIGO-256 HTTP streaming API.

smn:
This adds three HTTP endpoints, `messages.json`, `events.json` and `metrics.json`.
1. The messages and events endpoints allow for streaming of `vumi.message.Message` type of objects.
2. The messages and metrics endpoints allow for inserting of new objects via HTTP PUT.
3. The authentication is Basic Auth (so this needs to be run over HTTPS), the account key is used as the username and a conversation specific token is used as the password. There is a list of valid tokens stored in the conversation's `metadata` json field.
4. There is a concurrency limit that is at the moment specified per worker and set to 10. I think there's a small chance that if a worker dies without the appropriate shutdown stuff happening then the concurrent connection value stored in Redis may go out of sync.

There are a number of issues that need to be addressed:
1. It uses queues to publish messages specific to a conversation, these
   queues need to delete themselves if there are no consumers. (praekelt/vumi#451 addresses this)
2. The messages should not be persisted.

### DIFF
--- a/go/apps/http_api/tests/test_views.py
+++ b/go/apps/http_api/tests/test_views.py
@@ -1,0 +1,42 @@
+from django.test.client import Client
+from django.core.urlresolvers import reverse
+
+from go.apps.tests.base import DjangoGoApplicationTestCase
+
+
+class HttpApiTestCase(DjangoGoApplicationTestCase):
+
+    def setUp(self):
+        super(HttpApiTestCase, self).setUp()
+        self.setup_riak_fixtures()
+        self.client = Client()
+        self.client.login(username='username', password='password')
+
+    def test_new_conversation(self):
+        # render the form
+        self.assertEqual(len(self.conv_store.list_conversations()), 1)
+        response = self.client.get(reverse('http_api:new'))
+        self.assertEqual(response.status_code, 200)
+        # post the form
+        response = self.client.post(reverse('http_api:new'), {
+            'subject': 'the subject',
+            'message': 'the message',
+            'delivery_class': 'sms',
+            'delivery_tag_pool': 'longcode',
+        })
+        self.assertEqual(len(self.conv_store.list_conversations()), 2)
+        conversation = self.get_latest_conversation()
+        self.assertEqual(conversation.delivery_class, 'sms')
+        self.assertEqual(conversation.delivery_tag_pool, 'longcode')
+        self.assertEqual(conversation.delivery_tag, None)
+        self.assertEqual(conversation.metadata, None)
+        self.assertRedirects(response, reverse('http_api:people', kwargs={
+            'conversation_key': conversation.key,
+        }))
+
+    def test_show_conversation(self):
+        # render the form
+        [conversation_key] = self.conv_store.list_conversations()
+        kwargs = {'conversation_key': conversation_key}
+        response = self.client.get(reverse('http_api:show', kwargs=kwargs))
+        self.assertEqual(response.status_code, 200)
--- a/go/apps/http_api/urls.py
+++ b/go/apps/http_api/urls.py
@@ -1,0 +1,3 @@
+from go.apps.http_api.views import HttpApiConversationViews
+
+urlpatterns = HttpApiConversationViews().get_urlpatterns()
--- a/go/apps/http_api/views.py
+++ b/go/apps/http_api/views.py
@@ -1,0 +1,8 @@
+from go.conversation.base import ConversationViews
+
+
+class HttpApiConversationViews(ConversationViews):
+    conversation_type = u'http_api'
+    conversation_display_name = u'HTTP API'
+    conversation_initiator = None
+    edit_conversation_forms = None
--- a/go/apps/urls.py
+++ b/go/apps/urls.py
@@ -19,4 +19,6 @@ urlpatterns = patterns('',
         include('go.apps.wikipedia.sms.urls', namespace='wikipedia_sms')),
     url(r'^jsbox/',
         include('go.apps.jsbox.urls', namespace='jsbox')),
+    url(r'^http_api/',
+        include('go.apps.http_api.urls', namespace='http_api')),
 )
--- a/go/settings.py
+++ b/go/settings.py
@@ -130,6 +130,7 @@ TEMPLATE_DIRS = (
     abspath("apps", "sequential_send", "templates"),
     abspath("apps", "wikipedia", "ussd", "templates"),
     abspath("apps", "jsbox", "templates"),
+    abspath("apps", "http_api", "templates"),
 )
 
 INSTALLED_APPS = (
@@ -264,6 +265,10 @@ VUMI_INSTALLED_APPS = {
     'go.apps.jsbox': {
         'namespace': 'jsbox',
         'display_name': 'Javascript App',
+    },
+    'go.apps.http_api': {
+        'namespace': 'http_api',
+        'display_name': 'HTTP API',
     },
 }
 


smn:
praekelt/vumi#451 addresses the issue of auto deleting queues.

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,247 @@
...
+        request.finish()
+
+    @inlineCallbacks
+    def send_to(self, conversation, to_addr, content, **payload):
+
+        # FIXME:    At some point this needs to be done better as it makes some
+        #           assumption about how messages are routed which won't be
+        #           true for very much longer.
+        tag = (conversation.delivery_tag_pool, conversation.delivery_tag)
+        msg_options = yield conversation.make_message_options(tag)
```
hodgestar: You should just be able to call `VumiUserApi.msg_options` here. I also don't think you need to do the payload overriding below.

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,247 @@
...
+        return NOT_DONE_YET
+
+    @inlineCallbacks
+    def handle_PUT(self, request):
+        data = json.loads(request.content.read())
+        user_account = request.getUser()
+        conversation = yield self.get_conversation(user_account)
+
+        try:
+            tum = TransportUserMessage(_process_fields=True, **data)
```
hodgestar: I'm a little bit worried about starting the transport user message with `**data` -- that could create a message that might go almost anywhere inside Go and we have to blacklist stuff later. Probably better to whitelist particular parts of the message JSON?

go/apps/http_api/auth.py
```diff
@@ -0,0 +1,43 @@
...
+    @inlineCallbacks
+    def requestAvatarId(self, credentials):
+        username = credentials.username
+        token = credentials.password
+        user_exists = yield self.vumi_api.user_exists(username)
+        if user_exists:
+            user_api = self.vumi_api.get_user_api(username)
+            conversation = yield user_api.get_wrapped_conversation(
+                                                        self.conversation_key)
+            if conversation is not None:
```
smn, line 37: Per resource we are now likely to be pulling in the `conversation` object twice, once for authentication and once again when the resource itself needs it. I'm not sure how I could pass the conversation along to the resource from this `ICredentialsChecker` thing.

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,341 @@
...
+
+    @inlineCallbacks
+    def start_publishing(self, request):
+        yield self.setup_stream(request)
+        for consumer in self._consumers:
+            yield consumer.unpause()
+
+    def publish(self, request, message):
+        line = u'%s\n' % (message.to_json(),)
+        request.write(line.encode('utf-8'))
```
smn, line 85: I was expecting to be able to define a protocol for a `Resource` like and use something like `LineReceiver.sendLine()` here but I didn't find any. Open for suggestions.

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+class InvalidAggregate(errors.VumiError):
+    pass
+
+
+class EventStream(Stream):
+
+    message_class = TransportEvent
+    routing_key = '%(transport_name)s.stream.event.%(conversation_key)s'
```
smn, line 95: When we bind this queue it would need to be of type [auto-delete](http://www.rabbitmq.com/amqp-0-9-1-reference.html#queue.declare.auto-delete) which still needs to be built.

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,341 @@
...
+        tag = self.get_conversation_tag(conversation)
+        msg_options.update((yield user_api.msg_options(tag)))
+
+        content = msg_options.pop('content')
+        continue_session = (msg_options['session_event']
+                                != TransportUserMessage.SESSION_CLOSE)
+        helper_metadata = msg_options.pop('helper_metadata')
+        transport_type = msg_options.pop('transport_type')
+        session_event = msg_options.pop('session_event')
+        from_addr = msg_options.pop('from_addr')
```
smn, line 173: By now the `msg_options` may or may not be an empty dict, depending on the tag metadata.

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,341 @@
...
+
+        try:
+            metrics = self.parse_metrics(data)
+        except (ValueError, InvalidAggregate, InvalidMessage):
+            request.setResponseCode(http.BAD_REQUEST, 'Invalid Message')
+            request.finish()
+            return
+
+        for name, value, agg_class in metrics:
+            self.worker.publish_account_metric(user_account,
```
smn: This needs to be some custom string out of the conversation metadata, defaulting to 'default'.

go/apps/http_api/client.py
```diff
@@ -0,0 +1,72 @@
...
+    def lineReceived(self, line):
+        d = Deferred()
+        d.addCallback(self.callback)
+        d.addErrback(self.errback)
+        line = line.strip()
+        try:
+            data = json.loads(line)
+            d.callback(self.message_class(_process_fields=True, **data))
+        except ValueError, e:
+            d.errback(e)
```
hodgestar, line 44: This should maybe be wrapped with a Failure instance.

go/apps/http_api/vumi_app.py
```diff
@@ -0,0 +1,91 @@
...
+        if outbound_message is None:
+            log.warning('Unable to find message %s for event %s.' % (
+                event['user_message_id'], event['event_id']))
+        batch = yield outbound_message.batch.get()
+        account_key = batch.metadata['user_account']
+        user_api = self.get_user_api(account_key)
+        conversations = user_api.conversation_store.conversations
+        mr = conversations.index_lookup('batches', batch.key)
+        [conv_key] = yield mr.get_keys()
+        yield self.stream(EventStream, conv_key, event)
```
hodgestar, line 142: I think this logic should probably live near `GoMessageMetadata` somewhere?

hodgestar, line 142: We already have to look up a bunch of this information when routing the message to the application worker so if we stored it there the way GoMessageMetadata does we can avoid the double look-up. Maybe GMM can just be extended?

go/apps/http_api/resource.py
```diff
@@ -0,0 +1,348 @@
...
+        self.vumi_api = self.worker.vumi_api
+        self.user_apis = {}
+
+    def get_user_api(self, user_account):
+        if user_account in self.user_apis:
+            return self.user_apis[user_account]
+
+        user_api = self.vumi_api.get_user_api(user_account)
+        self.user_apis[user_account] = user_api
+        return user_api
```
hodgestar, line 38: Creating a `VumiUserApi` instance is a pretty cheap operation so I don't know that caching is particularly necessary (although since this is happening per message I guess it can't hurt).

go/apps/http_api/client.py
smn, line 44: Twisted docs do it like this, they pass things like `RuntimeErrror` and `ValueError` to an `errback`. Can't find anything telling me explicitly to turn it into a `Failure` instance.

go/apps/http_api/vumi_app.py
smn, line 142: Yeah, I can add it to the dispatcher and cache it in the `helper_metadata` if that's what you mean? I like that idea but I'm wary of having weird situations where our apps depend on undocumented keys need to be present in one of the `*_metadata` dicts for it to work.

go/apps/http_api/resource.py
smn, line 38: I had a `self._vumi_user_api` before but I got spooked by the idea of race conditions.

hodgestar:
:+1: once praekelt/vumi#451 is done and used here.

hodgestar:
And once the UI side has been implemented. :)

smn:
I'd like to gather some more ideas on how to prevent message loss when a) the client reconnects or b) the HTTP API restarts. Client reconnects we can solve with keeping the consumer alive for 10m after a disconnect giving the client time to reconnect. API restarts I'm not sure about since we won't have anything to actually consume the messages & store them in Redis (unless build something that guarantees redundancy & gradual restarts).

Thoughts @jerith @dmaclay @hodgestar @justinvdm ?

smn:
Also, do we want to be able to enable this for all conversation types or have a specific `http_api` conversation type? I'm inclined to got with 'enable it for all' though that would mean a bit more work on the GoApplicationDispatcher side of things further upstream.

hodgestar:
I think I'd prefer an `http_api` conversation type -- I don't think we've though through all the ramifications of having this piggyback on other apps and I'd like to keep things simple to start with (especially given that we're trying to clean-up a lot of other things right now).

Do we actually need to do anything particular special for API restarts? As long as the HTTP client queue isn't destroyed when the API restarts, messages will just start being processed again when the API comes back? Obviously the HTTP clients will be disconnected in this case so that needs to be handled too.

smn:
Ready for re-review again, now using a backlog cache of `MAX_BACKLOG_SIZE`.

smn:
This depends on praekelt/vumi#461 which adds `ltrim` to `FakeRedis` and `rpop` to the `RedisManager`

go/apps/http_api/vumi_app.py
```diff
@@ -0,0 +1,147 @@
...
+        # Set these to empty dictionaries because we're not interested
+        # in using any of the helper functions at this point.
+        self._event_handlers = {}
+        self._session_handlers = {}
+        self.vumi_api = yield VumiApi.from_config_async(self.config)
+        # We do give the publisher a key but won't actually use it
+        self.stream_publisher = yield self.publish_to(
+            '%s.stream.lost_and_found' % (self.transport_name,))
+        self.metric_publisher = yield self.start_publisher(MetricManager,
+            self.transport_name)
```
hodgestar: I think this should probably be `self.metrics_prefix`.

hodgestar:
:+1: from my side but someone else should review the views stuff.

smn:
:+1:

PR #151, opened 2013-01-18
Claim tags at account level rather than conversation level

jerith:
We currently claim tags for particular conversations and release them when the conversation ends. We should claim tags separately from conversations and then select tags held by the account in the conversation setup.

### DIFF
--- a/go/vumitools/account/old_models.py
+++ b/go/vumitools/account/old_models.py
@@ -1,0 +1,32 @@
+from datetime import datetime
+
+from vumi.persist.model import Model
+from vumi.persist.fields import (
+   Integer, Unicode, Timestamp, ManyToMany, Json,
+                                    Boolean)
+
+
+class UserTagPermissionVNone(Model):
+    """A description of a tag a user account is allowed access to."""
+    # key is uuid
+    tagpool = Unicode(max_length=255)
+    max_keys = Integer(null=True)
+
+
+class UserAppPermissionVNone(Model):
+    """An application that provides a certain conversation_type"""
+    application = Unicode(max_length=255)
+
+
+class UserAccountVNone(Model):
+    """A user account."""
+    # key is uuid
+    username = Unicode(max_length=255)
+    # TODO: tagpools can be made OneToMany once vumi.persist.fields
+    #       gains a OneToMany field
+    tagpools = ManyToMany(UserTagPermissionVNone)
+    applications = ManyToMany(UserAppPermissionVNone)
+    created_at = Timestamp(default=datetime.utcnow)
+    event_handler_config = Json(null=True)
+    msisdn = Unicode(max_length=255, null=True)
+    confirm_start_conversation = Boolean(default=False)
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -18,6 +18,7 @@ from vumi.persist.txriak_manager import TxRiakManager
 from vumi.persist.redis_manager import RedisManager
 from vumi.persist.txredis_manager import TxRedisManager
 from vumi.middleware.tagger import TaggingMiddleware
+from vumi import log
 
 from go.vumitools.account import AccountStore
 from go.vumitools.contact import ContactStore
@@ -107,6 +108,9 @@ class VumiUserApi(object):
         d = VumiApi.from_config_async(config)
         return d.addCallback(cls, user_account_key)
 
+    def get_user_account(self):
+        return self.api.get_user_account(self.user_account_key)
+
     def wrap_conversation(self, conversation):
         """Wrap a conversation with a ConversationWrapper.
 
@@ -145,7 +149,7 @@ class VumiUserApi(object):
 
     @Manager.calls_manager
     def tagpools(self):
-        user_account = yield self.api.get_user_account(self.user_account_key)
+        user_account = yield self.get_user_account()
         active_conversations = yield self.active_conversations()
 
         tp_usage = defaultdict(int)
@@ -167,7 +171,7 @@ class VumiUserApi(object):
 
     @Manager.calls_manager
     def applications(self):
-        user_account = yield self.api.get_user_account(self.user_account_key)
+        user_account = yield self.get_user_account()
         # NOTE: This assumes that we don't have very large numbers of
         #       applications.
         app_permissions = []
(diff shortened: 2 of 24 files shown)


dmaclay:
:+1:

go/vumitools/account/old_models.py
```diff
@@ -0,0 +1,32 @@
+from datetime import datetime
```
hodgestar, line 1: What are the old models here for?

go/vumitools/tests/test_api.py
```diff
@@ -253,6 +251,19 @@ def test_msg_options_with_tagpool_metadata(self):
...
+        tag1, tag2 = ("poolA", "tag1"), ("poolA", "tag2")
+        yield self.declare_tags(self.api, [tag1, tag2])
+        self.assertEqual((yield self.user_api.acquire_tag("poolA")), tag1)
+        self.assertEqual((yield self.user_api.acquire_tag("poolA")), tag2)
+        self.assertEqual((yield self.user_api.acquire_tag("poolA")), None)
+        self.assertEqual((yield self.user_api.acquire_tag("poolB")), None)
+
+        yield self.user_api.release_tag(tag2)
+        self.assertEqual((yield self.user_api.acquire_tag("poolA")), tag2)
+        self.assertEqual((yield self.user_api.acquire_tag("poolA")), None)
```
hodgestar: I think this should also check that the tags are saved to the user account?

go/vumitools/account/old_models.py
jerith, line 1: Partly documentation of what the old models looked like, mostly to support migration tests I haven't written yet.

go/vumitools/tests/utils.py
```diff
@@ -224,6 +224,43 @@ def _conversation_type(self):
...
+        yield self.declare_tags(self.vumi_api,
+                                [(u"pool", u"tag1"), (u"pool", u"tag2")])
+        yield self.set_pool_metadata(self.vumi_api, u"pool", {
+            "transport_type": self.transport_type,
+            "msg_options": {
+                "transport_name": self.transport_name,
+                },
+            })
+
+    def declare_tags(self, api, tags):
```
hodgestar: Are these three methods actually better than just `api.tpm.<method>` calls?

hodgestar: If everyone is mainly using `setup_tagpools` anyway they don't seem like much of a win.

hodgestar:
:+1: modulo minor comments.

go/vumitools/account/old_models.py
hodgestar, line 1: Maybe they need a sub-package and/or a name that includes the version so they don't have to be moved around when more are added?

jerith, line 1: I'm changing the class name to include the version number, so we can stick them all in the same module for now. I can't think of an alternative that isn't messier than that -- an explosion of subpackages doesn't really appeal to me.

go/vumitools/tests/test_api.py
jerith: It should. Good catch.

go/vumitools/tests/utils.py
jerith: Indeed. The other two methods predate `setup_tagpools()`.

go/vumitools/account/old_models.py
hodgestar, line 1: Ah. I missed the class name changes. In that case, looks good as it is.

go/vumitools/api.py
```diff
@@ -224,6 +230,68 @@ def msg_options(self, tag, tagpool_metadata=None):
...
+        Tags should be held for the duration of a conversation.
+
+        :type pool: str
+        :param pool:
+            name of the pool to retrieve tags from.
+        :rtype:
+            The tag acquired or None if no tag was available.
+        """
+        user_account = yield self.get_user_account()
+        # TODO: Check that account has access to pool.
```
smn: Is this still to be implemented or do we want to do that checking before `acquire_tag()` is called? I think it makes sense to still add this.

smn:
What are we going to do about tags that have already been acquired but have not yet been associated with an account yet (other than via the batch record)? How are we going to migrate those?

go/vumitools/api.py
jerith: Still something we need to do, but it isn't really urgent. We currently do this check when we list the available tagpools in the UI.

smn: My vote would be to implement it here, since this is the `VumiUserApi` it should raise an error if it tries to do something it's not allowed to do.

hodgestar: On Tue, Jan 22, 2013 at 3:23 PM, Simon de Haan notifications@github.comwrote:

> My vote would be to implement it here, since this is the VumiUserApi it
> should raise an error if it tries to do something it's not allowed to do.
> 
> I think I'm +1 on implementing it here too. We should probably document the
> decision in the class docstring or something.

Schiavo
Simon

go/vumitools/api.py
```diff
@@ -192,12 +194,10 @@ def new_conversation(self, *args, **kw):
...
     @Manager.calls_manager
-    def list_endpoints(self):
-        """Returns a set of end points owned by an account.
-
-        Currently this returns the list of tags in use by active
-        conversations.
+    def list_conversation_endpoints(self):
+        """Returns a set of endpoints owned by conversations in an account.
         """
+        # XXX: Do we need both this method and the following method?
```
smn, line 200: My gut feeling is that the `list_conversation_batch_tags()` is going to be more useful (and accurate) but it's hard to tell at this point. Happy to leave as is for now.

smn:
:+1:

go/vumitools/api.py
jerith, line 200: Both of these will probably go away with the new routing stuff that's coming.

hodgestar, line 200: Both of these will be useful for at least checking purposes while we transition. I think we can remove them when conversations lose the associated functionality (i.e. .delivery_tag and .delivery_tag_pool and batches with tags). I would just change this to "XXX: remove when the associated functionality is gone from conversations" or something.

go/vumitools/api.py
```diff
@@ -224,6 +251,77 @@ def msg_options(self, tag, tagpool_metadata=None):
...
+        :type pool: str
+        :param pool:
+            name of the pool to return the tag too (must be the same as
+            the name of the pool the tag came from).
+        :rtype:
+            None.
+        """
+        user_account = yield self.get_user_account()
+        yield self._populate_tags(user_account)
+        # TODO: Check that account holds tag
```
hodgestar: Does this TODO still need to be here? An exception is caught below. If we need to do something more we can do it in the exception handler?

go/vumitools/api.py
```diff
@@ -206,6 +206,33 @@ def list_endpoints(self):
...
+            conv_tags = yield self.wrap_conversation(conv).get_tags()
+            tags.update(conv_tags)
+        returnValue(tags)
+
+    @Manager.calls_manager
+    def _populate_tags(self, user_account):
+        if user_account.tags is None:
+            # We need to populate this from conversations
+            conv_tags = yield self.list_conversation_batch_tags()
+            user_account.tags = [list(tag) for tag in conv_tags]
```
hodgestar, line 225: I was a bit suprised by _populate_tags? I thought we were going with a script to do this?

go/vumitools/conversation/tests/test_utils.py
```diff
@@ -38,10 +38,10 @@ def setUp(self):
         self.conv = ConversationWrapper(raw_conv, self.user_api)
 
     @inlineCallbacks
-    def declare_tags(self, name='longcode', count=4, metadata=None):
+    def _declare_tags(self, name=u'longcode', count=4, metadata=None):
```
hodgestar: How come this got an underscore?

go/vumitools/api.py
jerith, line 225: Better to do it on-demand. We can write a script that triggers it for each account to make sure everything's been updated, but I'd really prefer to avoid mandatory update scripts as part of a deploy.

go/vumitools/api.py
jerith: Good catch. I'll remove it.

go/vumitools/conversation/tests/test_utils.py
jerith: I added a `declare_tags()` method on the superclass and had to rename this one to avoid conflicts. I'll give it a better name.

go/vumitools/api.py
jerith, line 200: Do we actually need the functionality `list_conversation_endpoints()` provides? I left it in because it was already there, but I don't think anything's using it.

hodgestar:
:+1:

PR #152, opened 2013-01-23
VUMIGO-280 Go webapp Sentry config

smn:
This adds the dependency, production config happening elsewhere.


jerith:
:+1: on the basis that it seems unlikely to break anything.

PR #153, opened 2013-01-23
VUMIGO-280 fix raven sentry mismatch

smn:
Our raven version is too new for our sentry deploy.
Need to stay < v3 until we upgrade our sentry.


hodgestar:
:+1:

PR #154, opened 2013-02-01
Ace or CodeMirror editor for Javascript.

peterp:


hodgestar:
Should we not just be including a custom uglified `codemirror-compressed.js` file (as generated by http://codemirror.net/doc/compress.html) rather than a complete release?

hodgestar:
Should we maybe incorporate the `django-codemirror-widget` directly rather than including it as dependency? It's only one file and in addition to adding a dependency, it also obfuscates how all the JS and CSS is meant to be layed out and hooked up.

hodgestar:
@peterp noticed that we're generating script tags for the CodeMirror javascript N times where N is the number of uses of the CodeMirror widget we have on a page. This isn't the fault of this branch but since CodeMirror is fairly hefty we should solve it before landing I think.

go/base/static/js/source_url.js
```diff
@@ -0,0 +1,26 @@
+function SourceUrl(elem, dest) {
+    var $input = $('#' + elem.id);
+
+    $button = $('<button><i class="icon-plus"></i> Update via URL</button>');
+    $button.on('click', function() {
+        var url = $input.val()
+        if (url.length == 0) {
+            alert('Source URL is required');
```
justinvdm: Could we maybe use the console instead of `alert()`?

go/base/widgets.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def render(self, name, value, attrs=None):
+        source_input_id = 'id_%s' % (name,)
+        # constructing the correct code field name like this isn't
+        # great but I don't have a better idea
+        code_field_name = self.code_field_name(name)
+        code_field_id = CodeMirrorTextarea.id_for_name(code_field_name)
+        output = [super(SourceUrlTextInput, self).render(name, value, attrs),
+                  '<script type="text/javascript">'
+                  'SourceUrl(document.getElementById("%s"),'
```
justinvdm: I think it may be a bit more standard if `SourceUrl()` accepted in selectors instead of elements, and then obtained the elements in `SourceUrl` by using `document.getElementById` there instead of here (or just wrap the selectors using jquery).  All we seem to need `elem` for in `SourceUrl` is to obtain its id, so getting the element may not be necessary anyway.

go/settings.py
```diff
@@ -234,7 +234,7 @@ def abspath(*args):
 #       configuration file so that configuration values aren't
 #       duplicated
 VUMI_API_CONFIG = {
-    'redis_manager': {'key_prefix': 'vumigo'},
+    'redis_manager': {'key_prefix': 'vumigo', 'db': 1},
```
smn, line 236: thanks :)

smn:
:+1: from me though a bit of docstring on the widget's `mode` param would be useful.

PR #162, opened 2013-02-05
Update to use config objects where applicable.

jerith:
This is the Go side of praekelt/vumi#460.

### DIFF
--- a/go/apps/jsbox/vumi_app.py
+++ b/go/apps/jsbox/vumi_app.py
@@ -3,10 +3,10 @@
 
 """Vumi application worker for the vumitools API."""
 
-from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.defer import inlineCallbacks
 
+from vumi.config import ConfigField
 from vumi.application.sandbox import JsSandbox, SandboxResource
-from vumi.message import TransportEvent
 from vumi import log
 
 from go.vumitools.app_worker import GoApplicationMixin
@@ -24,6 +24,11 @@ class ConversationConfigResource(SandboxResource):
         key_config = app_config.get(key, {})
         value = key_config.get('value')
         return self.reply(command, value=value, success=True)
+
+
+class JsBoxConfig(JsSandbox.CONFIG_CLASS):
+    conversation = ConfigField("Conversation object from message.")
+    javascript = ConfigField("Javascript from message.")
 
 
 class JsBoxApplication(GoApplicationMixin, JsSandbox):
@@ -44,6 +49,7 @@ class JsBoxApplication(GoApplicationMixin, JsSandbox):
     And those from :class:`vumi.application.sandbox.JsSandbox`.
     """
 
+    CONFIG_CLASS = JsBoxConfig
     SEND_TO_TAGS = frozenset(['default'])
     worker_name = 'jsbox_application'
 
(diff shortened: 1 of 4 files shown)


go/apps/jsbox/vumi_app.py
```diff
@@ -26,6 +27,17 @@ def handle_get(self, api, command):
...
+class JsBoxConfig(JsSandbox.CONFIG_CLASS, GoApplicationConfigMixin):
+    jsbox_app_config = ConfigDict(
+        "Custom configuration passed to the javascript code.", default={})
+    jsbox = ConfigDict(
+        "Must have 'javascript' field containing JavaScript code to run.")
+
+    @property
+    def javascript(self):
+        return self.jsbox['javascript']
+
```
smn, line 39: Why not have the `jsbox` config use the `javascript` field like in the Vumi repo? Legacy?

smn:
:+1: but curious as to why we use `jsbox` vs `javascript` fields

go/apps/jsbox/vumi_app.py
jerith, line 39: Because of the way we use Django forms in the conversation setup UI. We can fix that when we do the model migration work.

jerith:
I'll leave the rest of the app workers for the conversation model migration when most of the surrounding code is changing anyway.

go/vumitools/app_worker.py
```diff
@@ -72,6 +98,20 @@ def _go_teardown_application(self):
     def get_user_api(self, user_account_key):
         return self.vumi_api.get_user_api(user_account_key)
 
+    def _get_config_for_conversation(self, conversation):
+        config_data = GoApplicationConfigData(self.config, conversation)
+        return self.CONFIG_CLASS(config_data)
```
hodgestar, line 103: Given that this method is already used outside of this class both in test helper methods and in sequential send, perhaps it shouldn't be an underscore method?

go/apps/jsbox/vumi_app.py
```diff
@@ -51,46 +64,23 @@ def validate_config(self):
...
-        api = self.create_sandbox_api(self.resources)
-        api.javascript = javascript
-        api.conversation = conversation
-        protocol = self.create_sandbox_protocol(sandbox_id, api)
-        returnValue(protocol)
+        conv = self.conversation_for_api(api)
+        return self.get_user_api(conv.user_account.key)
+
+    def get_config(self, msg):
+        return self.get_message_config(msg)
```
hodgestar, line 85: Not quite sure what this method is for?

go/apps/jsbox/vumi_app.py
```diff
@@ -51,46 +64,23 @@ def validate_config(self):
...
 
-    @inlineCallbacks
     def teardown_application(self):
-        yield super(JsBoxApplication, self).teardown_application()
-        yield self._go_teardown_application()
-
-    def javascript_for_api(self, api):
-        return api.javascript
+        d = maybeDeferred(super(JsBoxApplication, self).teardown_application)
+        return d.addCallback(lambda r: self._go_teardown_application())
```
hodgestar: Would kinda like the inlineCallbacks put back. It's a pattern we use all over the place and it's pretty readable. If we're going to declare this pattern bad, I'd like it to be by group consensus.

go/vumitools/app_worker.py
jerith, line 103: Good point.

go/apps/jsbox/vumi_app.py
jerith: Okay.

go/apps/jsbox/vumi_app.py
jerith, line 85: `get_config()` gets called by the parent to get a message-specific config. `get_message_config()` comes from `GoApplicationMixin`. I didn't override `get_config()` directly in the mixin because then MRO matters and the multiple inheritance gets complicated.

hodgestar:
:+1: modulo comment about calling `setup_application` and `teardown_application`.

PR #164, opened 2013-02-06
VUMIGO-48 one-off replies

smn:
Adds the ability to send one-off replies to messages received on a
conversation that isn't exclusively client initiated.

Some UI for this already existed but did nothing, this adds the missing
bits. It also builds on the already existing functionality of the 'send_message' command.

### DIFF
--- a/go/apps/bulk_message/tests/test_views.py
+++ b/go/apps/bulk_message/tests/test_views.py
@@ -169,9 +169,10 @@ class BulkMessageTestCase(DjangoGoApplicationTestCase):
             'conversation_key': self.conv_key}))
 
         # Check pagination
-        # We should have 20 links to contacts which by default display
-        # the from_addr if a contact cannot be found.
-        self.assertContains(response, 'from-', 20)
+        # We should have 60 references to a contact, which by default display
+        # the from_addr if a contact cannot be found. (each block as 3
+        # references, one in the table listing, 2 in the reply-to modal div)
+        self.assertContains(response, 'from-', 60)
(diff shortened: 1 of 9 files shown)


jerith:
Looks good for the most part, but I have two concerns.

Why do we send `to_addr` from the browser if we're just going to ignore it?

Why do we need a celery task for this? I'd rather just send a command to the vumi app and have it do the message lookup itself. It's about the same amount of code and it moves the message store lookup from a new task in celery (which likes to silently eat exceptions) to the vumi app that has to handle the actual sending anyway.

smn:
Thanks for the review @jerith 
1. I put the `to_addr` in the form because that's the easiest way to get it displayed when using Django forms & bootstrap.
2. The reason I'm using a Celery task is that I want to have a single way to deal with scheduling background jobs from the Django UI. At the moment that is Celery. The fact that Celery schedules another job again is in my opinion a symptom of something else being wrong or being missing.

smn:
That said I would like to fix nr 2. but not part of this PR.

smn:
```
[08:12] <jeroud> sdehaan: I'd really like to get rid of celery completely instead of using it for more things.
[08:13] <jeroud> That said, I'm fine with landing that PR now (with a comment about to_addr) and worrying about it next week.
[08:48] <sdehaan> jeroud: thanks, can you leave that as a comment in the PR?
[08:51] <jeroud> sdehaan: Probably not for a while. In traffic. Feel free to quote me on the PR, though.
```

PR #169, opened 2013-02-12
VUMIGO-282 Random UI timeouts.

smn:
We're getting random timeouts on various bits of our UI. This is the
start of adding the necessary bits to try & debug what's going on.

The main thing here is:
1. Straight connection to AMQP w/ Kombu & librabbitmq (no longer hacking
   our way through Celery)
2. Middleware for publishing response times to Graphite.
3. I've added hiredis to the requirements.pip, I'll install that on the
   server after we've got some metrics to compare with. Hiredis implements
   the binary Redis protocol and py-redis automatically uses it if
   installed. Hiredis claims 15-90x speed up on range queries.

### DIFF
--- a/go/base/middleware.py
+++ b/go/base/middleware.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from go.base.utils import vumi_api_for_user
 from go.base.amqp import connection
 
-from vumi.blinkenlights.metrics import AVG, MetricMessage
+from vumi.blinkenlights.metrics import AVG
 
 logger = logging.getLogger(__name__)
 
@@ -53,10 +53,6 @@ class ResponseTimeMiddleware(object):
     def publish_metric(self, metric_name, value, timestamp=None,
                         aggregators=None):
         aggregators = aggregators or [AVG]
-        timestamp = timestamp or time.time()
-        metric_msg = MetricMessage()
         prefixed_metric_name = '%s%s' % (self.metrics_prefix, metric_name)
-        metric_msg.append((prefixed_metric_name,
-            tuple(sorted(agg.name for agg in aggregators)),
-            [(timestamp, value)]))
-        connection.publish_metric(metric_msg)
+        connection.publish_metric(prefixed_metric_name, aggregators,
+            value, timestamp)
--- a/go/base/middleware.py
+++ b/go/base/middleware.py
@@ -26,6 +26,9 @@ class ResponseTimeMiddleware(object):
     Marks the time when a request is received and when the response for that
     request is sent back again. Fires off metrics for the time taken to
     generate the response.
+
+    It sets an X-Response-Time HTTP header which can be useful for debugging
+    or logging slow resources upstream.
     """
     def __init__(self):
         self.metrics_prefix = getattr(settings, 'METRICS_PREFIX', 'go.django.')
--- a/go/base/middleware.py
+++ b/go/base/middleware.py
@@ -2,6 +2,7 @@ import time
 import logging
 
 from django.core.urlresolvers import resolve, Resolver404
+from django.conf import settings
 
 from go.base.utils import vumi_api_for_user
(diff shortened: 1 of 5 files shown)


smn:
I'm thinking that adding a differentiation on the HTTP method could potentially be useful. POSTs are likely to be quite a bit slower than GETs (which is generally fine) but I don't want those to average each other out.

vgeddes:
:+1: Looks good.

smn:
Good catch on the `amqp.connect()`, I checked the rabbitmq logs and `librabbitmq` only connects on the first message unless you explicitly tell it to verify the connection so I think we're fine for now. I'm not entirely happy with the approach though, Django is missing something of an application init type of call. Open to suggestions.

go/base/middleware.py
```diff
@@ -6,3 +17,42 @@ def process_request(self, request):
...
             request.user_api = vumi_api_for_user(request.user)
+
+
+class ResponseTimeMiddleware(object):
+    """
+    Middleware for generating metrics on page response times.
+
+
+
+    """
```
hodgestar, line 32: Seems like an abundance of spacing?

smn, line 32: Right, that was the space for the docstring I was going to write. Thanks.

go/base/middleware.py
```diff
@@ -6,3 +17,42 @@ def process_request(self, request):
...
+    def publish_metric(self, metric_name, value, timestamp=None,
+                        aggregators=None):
+        aggregators = aggregators or [AVG]
+        timestamp = timestamp or time.time()
+        metric_msg = MetricMessage()
+        prefixed_metric_name = '%s%s' % (self.metrics_prefix, metric_name)
+        metric_msg.append((prefixed_metric_name,
+            tuple(sorted(agg.name for agg in aggregators)),
+            [(timestamp, value)]))
+        connection.publish_metric(metric_msg)
```
hodgestar: Maybe more of this logic should live on the `AmqpConnection` class?

go/base/tests/test_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_calculating_response_time(self, publish, is_connected):
+        is_connected.return_value = True
+        publish.return_value = True
+        response = HttpResponse('ok')
+
+        request = self.factory.get('/accounts/login/')
+        request.start_time = time.time()
+
+        self.mw.process_response(request, response)
+        self.assertTrue(response['X-Response-Time'])
```
hodgestar: Might be good to assert that this has a numerical value.

requirements.pip
```diff
@@ -26,3 +26,7 @@ xlrd==0.8.0
 requests==0.14.2
 mock==1.0.1
 raven>=2.0,<3.0
+django-debug-toolbar==0.9.4
+kombu==2.5.6
+librabbitmq==1.0.1
+hiredis==0.1.1
```
hodgestar, line 32: What is `hiredis` for?

hodgestar:
Mostly :+1:, just a few questions to help me understand the bits and pieces.

go/base/middleware.py
smn: You mean pass in the metric key, aggregator and datapoint values to `publish_metric()` ?

hodgestar: Yep, rather than the `MetricMessage`. Maybe this whole method can just be added to where `publish_metric()` was and the other `publish_metric` can become `publish_metric_message()`?

hodgestar: Metric prefix might have to be an additional parameter or maybe this method can just construct the full name?

requirements.pip
smn, line 32: Much more efficient redis parser, wraps the C hiredis client and provides a significant speed improvement on all range queries.

hodgestar:
:+1:

PR #170, opened 2013-02-12
GO-285 crash during one-off reply handling.

smn:
We made the assumption that `send_to()` would work for replies as well.
Turns out it doesn't because it in turn called
`TransportUserMessage.send()` which sets the `in_reply_to` keyword
argument to `None`. We then supply a second kwarg with the key of a
message we're replying to and it fails miserably.

This fixes that by calling `reply_to()` when given a value for
`in_reply_to`.

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -2,6 +2,8 @@
 
 """Tests for go.vumitools.bulk_send_application"""
 
+import uuid
+
 from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.task import Clock
 
@@ -10,6 +12,7 @@ from vumi.components.window_manager import WindowManager
 from vumi.tests.utils import LogCatcher
 
 from go.vumitools.tests.utils import AppWorkerTestCase
+from go.vumitools.api import VumiApiCommand
 from go.apps.bulk_message.vumi_app import BulkMessageApplication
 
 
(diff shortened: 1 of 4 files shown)


go/apps/bulk_message/vumi_app.py
```diff
@@ -127,11 +127,17 @@ def handle_event(self, event):
...
-                command_data['content'],
-                command_data['msg_options'])
+        to_addr = command_data['to_addr']
+        content = command_data['content']
+        msg_options = command_data['msg_options']
+        in_reply_to = msg_options.pop('in_reply_to', None)
+        if in_reply_to:
+            msg = yield self.vumi_api.mdb.get_inbound_message(in_reply_to)
+            if msg:
+                yield self.reply_to(msg, content, continue_session=False)
```
hodgestar: I think continue_session should default to True so that the user can keep interacting if they like (this doesn't make much difference over channels that aren't session based obviously).

go/apps/surveys/vumi_app.py
```diff
@@ -164,4 +165,11 @@ def process_command_send_message(self, *args, **kwargs):
         to_addr = command_data['to_addr']
         content = command_data['content']
         msg_options = command_data['msg_options']
+        in_reply_to = msg_options.pop('in_reply_to', None)
+        if in_reply_to:
+            msg = yield self.vumi_api.mdb.get_inbound_message(in_reply_to)
+            if msg:
+                yield self.reply_to(msg, content, continue_session=False)
```
hodgestar: Same here.

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -181,6 +184,36 @@ def test_send_message_command(self):
...
+                        u'tag': {
+                            u'tag': [u'longcode', u'default10080']
+                        }
+                    },
+                    u'transport_name': u'smpp_transport',
+                    u'in_reply_to': msg['message_id'],
+                    u'transport_type': u'sms',
+                    u'from_addr': u'default10080',
+                }
+            })
```
hodgestar, line 209: Sending all the msg_options through in the command is probably unnecessary these days. The jsbox app has a snippet that looks like:

```
user_api = self.app_worker.user_api_for_api(api)
tags = yield user_api.list_endpoints()
if tag not in tags:
    returnValue(self.reply(command, success=False,
                           reason="Tag %r not held by account"
                           % (tag,)))
msg_options = yield user_api.msg_options(tag)
yield self.app_worker.send_to(to_addr, content, **msg_options)
```

Which populates the msg_options based on the account and the tag (which is closer to what will happen with the new endpoint stuff). Not something for now -- just mentioning.

hodgestar:
:+1: module logging invalid `in_reply_to` values rather than falling back to `send_to`.

vgeddes:
looks good :+1:

PR #171, opened 2013-02-12
VUMIGO-286 streaming api health checks.

smn:
The health check resource expects the transport to tell it whether
things are ok or not. This adds the necessary bits to make that happen.
It'll return how many streaming connections there currently are via HTTP
on the /health/ resource.

### DIFF
--- a/go/apps/http_api/tests/test_vumi_app.py
+++ b/go/apps/http_api/tests/test_vumi_app.py
@@ -46,6 +46,7 @@ class StreamingHTTPWorkerTestCase(AppWorkerTestCase):
         yield super(StreamingHTTPWorkerTestCase, self).setUp()
         self.config = self.mk_config({
             'worker_name': 'worker_name',
+            'health_path': '/health/',
             'web_path': '/foo',
             'web_port': 0,
             'metrics_prefix': 'metrics_prefix.',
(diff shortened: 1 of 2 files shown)


go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -322,3 +323,29 @@ def test_backlog_on_connect(self):
...
+        stream_url = '%s/%s/messages.json' % (self.url, self.conversation.key)
+        stream_receiver = self.client.stream(TransportUserMessage, queue.put,
+            queue.put, stream_url, Headers(self.auth_headers))
+
+        yield queue.get()
+
+        response = yield http_request_full(health_url, method='GET')
+        self.assertEqual(response.delivered_body, '1')
+
+        stream_receiver.disconnect()
```
hodgestar, line 348: I would expect one to have to yield until the `stream_receiver` is disconnected somehow?

hodgestar:
:+1: other than a slight comment on waiting for the stream receiver to disconnect in the test.

go/apps/http_api/tests/test_vumi_app.py
smn, line 348: The receiver is the client, disconnecting that is pretty much immediate as we're just closing a socket, Twisted doesn't give me anything to wait for.

PR #173, opened 2013-02-15
VUMIGO-289 send SMS on menu completion.

smn:
This is a custom event handler for SNA that will send an SMS whenever
someone completes the USSD menu.


jerith:
:+1:

PR #174, opened 2013-02-18
VUMIGO-289 vumi command & event durability

smn:
Our command & event default routing config assumes that our exchange is
set to being durable, while the call itself has it set to being False by
default, which is wrong.


hodgestar:
:+1:

PR #175, opened 2013-02-19
VUMIGO-290 confirm via email.

smn:
Some things in the Go UI need to be confirmed via a secondary channel
for security purposes. This is primarily the case for changing account
details like contact numbers and passwords.

This PR adds a generic bit to the TokenManager that allows you to link
specific callbacks to tokens. They'll get run immediately, if you want to schedule something then just pass in a Celery task and provide the `.delay` function on the task as the callback. Then the job will get deferred to Celery and will happen some time in the future.


go/account/views.py
```diff
@@ -24,22 +30,46 @@ def index(request):
...
+                    'msisdn': data['msisdn'],
+                    'confirm_start_conversation':
+                        data['confirm_start_conversation'],
+                }
+
+                site = Site.objects.get_current()
+                redis = RedisManager.from_config(
+                                    settings.VUMI_API_CONFIG['redis_manager'])
+                token_manager = DjangoTokenManager(
+                                    redis.sub_manager('token_manager'))
```
hodgestar: This seems like a lot of setup to copy around. Is there a way we can add getting a token manager to `vumi_api`?

go/account/views.py
```diff
@@ -24,22 +30,46 @@ def index(request):
...
+
+                token = token_manager.generate_task(request.path,
+                    'Your details are being updated', update_account_details,
+                    task_args=(request.user.id,),
+                    task_kwargs=params, user_id=request.user.id,
+                    immediate=True)
+
+                token_url = 'http://%s%s' % (site.domain,
+                                reverse('token', kwargs={'token': token}))
```
hodgestar: Maybe the token_url should be returned along with the token? Don't want other places that call this to have to construct their own URLs.

hodgestar:
Other than putting the token manager setup bits in a common place and the tests that you're currently writing, looks good.

go/account/views.py
smn: I've opted to re-using the existing redis from the `request.user_api`, I'd rather not add this to the `vumi_api` because it's currently designed to be used with a sync RedisManager. I'm happy to change that once we start needing it outside of Django.

smn:
All review comments addressed I think, please re-review @jerith @hodgestar @justinvdm @vgeddes

go/account/views.py
```diff
@@ -24,22 +30,38 @@ def index(request):
...
+                    'last_name': data['surname'],
+                    'new_password': data['new_password'],
+                    'email_address': data['email_address'],
+                    'msisdn': data['msisdn'],
+                    'confirm_start_conversation':
+                        data['confirm_start_conversation'],
+                }
+
+                token_manager = TokenManager(
+                                    redis.sub_manager('token_manager'))
```
hodgestar: I still have a mild objection to having to repeat the `'token_manager'` string. Could we maybe have the generic TokenManager available via the API and just wrap it with the Django-enhanced one?

hodgestar:
Other than making the `TokenManager` Twisted compatible and maybe making it available on the `vumi_api`, I'm +1.

hodgestar:
:+1:

hodgestar:
Final :+1: :)

PR #177, opened 2013-02-22
VUMIGO-234 pagination fixes for conversations.

smn:
Stuff just wasn't working. Now it does and has a test.


smn:
I'm bumping the number of conversations per page to 12 instead of 6. No one is using this on mobile atm and the new UI will change this again, temp fix for clients.

hodgestar:
:+1:

PR #178, opened 2013-02-22
VUMIGO-292 end draft conversations.

smn:
Our current UI doesn't allow us to end conversations that are stuck in
draft mode. We do need to allow this.


hodgestar:
:+1: -- we've needed this for awhile.

PR #179, opened 2013-02-23
VUMIGO-282 refactor conversation index views.

smn:
Do a lot less calls to Riak & sacrifice the warning about opt-outs not
working. Will revisit this with the new UI & end-points setup.

### DIFF
--- a/go/conversation/views.py
+++ b/go/conversation/views.py
@@ -1,7 +1,6 @@
 from django.contrib.auth.decorators import login_required
 from django.shortcuts import render
-from django.core.paginator import Paginator
-from django.contrib import messages
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 
 from go.conversation.forms import ConversationSearchForm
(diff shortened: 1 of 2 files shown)


go/conversation/views.py
```diff
@@ -11,64 +10,49 @@
...
 
-        has_active_sms_conversation = any([c.delivery_class == 'sms'
-                                           for c in active_conversations])
-        has_archived_conversations = conv_store.list_conversations()
-        if not (has_active_sms_conversation) and has_archived_conversations:
-            messages.error(request, "You do not have any active SMS "
-                "conversations. Opt-outs will not work until you do.")
+    # sort with newest first
+    conversations = sorted(conversations, key=lambda c: c.created_at,
+                            reverse=True)
```
hodgestar, line 43: If `conversations` is always a list you can just do `conversations.sort(key=...)`.

hodgestar:
:+1:

PR #180, opened 2013-02-25
VUMIGO-176 endpoint based routing

jerith:


go/vumitools/routing.py
```diff
@@ -0,0 +1,132 @@
...
+    @inlineCallbacks
+    def get_config(self, msg):
+        if isinstance(msg, TransportEvent):
+            msg = yield self.find_message_for_event(msg)
+
+        config_dict = self.config.copy()
+        user_account_key = None
+
+        # TODO: Better way to look up account for either tag or conversation.
+        tag = msg.get('helper_metadata', {}).get('tag', {}).get('tag')
```
smn: how about [MessageMetadataHelper.tag](https://github.com/praekelt/vumi-go/pull/180/files#L39R44) ?

go/vumitools/routing.py
```diff
@@ -0,0 +1,132 @@
...
+        if tag is not None:
+            tagpool_md = yield self.vumi_api.tpm.get_metadata(tag[0])
+            config_dict['tagpool_metadata'] = tagpool_md
+            config_dict['message_tag'] = tag
+            tag_info = yield self.vumi_api.mdb.get_tag_info(tuple(tag))
+            user_account_key = tag_info.metadata['user_account']
+
+        msg_mdh = self.get_metadata_helper(msg)
```
smn: Why not move this up and use the `tag` attribute?

go/vumitools/routing.py
```diff
@@ -0,0 +1,132 @@
...
+        if user_account_key is None:
+            user_account_key = yield msg_mdh.get_account_key()
```
smn: Either we're missing something or a `yield` isn't necessary here, the `MessageMetadataHelper` just pulls it out of an internal `_go_metadata` dict.

go/vumitools/routing.py
```diff
@@ -0,0 +1,132 @@
...
+
+        target = self.find_target(config, msg, ':'.join(config.message_tag))
+        if target is None:
+            log.debug("No target found for message from '%s': %s" % (
+                connector_name, msg))
+            return
+
+        target_conn, endpoint = target
+        conv_type, conv_key = target_conn.split(':', 1)
+        go_metadata = msg.get('helper_metadata', {}).setdefault('go', {})
```
smn: Can we not use the metadata helper for this as well?

go/base/management/commands/go_manage_routing_table.py
```diff
@@ -0,0 +1,154 @@
...
+        elif options['clear']:
+            return self.handle_clear(user_api, options)
+        elif options['add']:
+            return self.handle_add(user_api, options)
+        elif options['remove']:
+            return self.handle_remove(user_api, options)
+        raise NotImplementedError('Unknown command.')
+
+    def handle_show(self, user_api, options):
+        if options['no-migration']:
```
smn, line 88: I'd be more comfortable if we had this as the default with a `--migrate` option to migrate it explicitly.

go/base/management/commands/go_manage_routing_table.py
```diff
@@ -0,0 +1,154 @@
...
+        rt_helper = RoutingTableHelper(account.routing_table)
+        rt_helper.add_entry(*options['add'])
+        try:
+            user_api.validate_routing_table(account)
+        except Exception as e:
+            raise CommandError(e)
+        account.save()
+        self.stdout.write("Routing table entry added.\n")
+
+    def handle_remove(self, user_api, options):
```
smn, line 118: Can we at least have tests for `handle_add` and `handle_remove`?

go/base/management/commands/go_manage_routing_table.py
jerith, line 88: It's almost never useful to actually look at the pre-migrated version. I only added the option to make it easier to test the commands manually.

go/base/management/commands/go_manage_routing_table.py
jerith, line 118: I still need to write tests for the whole management command. I wanted to push first to get feedback before figuring out how best to write management command tests.

smn, line 118: It's fairly straight forward, see https://github.com/praekelt/vumi-go/blob/develop/go/base/tests/test_go_send_app_worker_command.py

jerith, line 118: Yup, I'm about to look at that now.

go/base/management/commands/go_manage_routing_table.py
jerith, line 88: To clarify, migration will never modify an existing routing table. It will only ever build a new one if the routing table on the account is `None`.

smn, line 88: ok, in that case I'm fine with it.

PR #182, opened 2013-02-27
VUMIGO-293 account summary email.

smn:
Send an account summary via email to an account holder's email address.


go/account/tasks.py
```diff
@@ -19,4 +22,14 @@ def update_account_details(user_id, first_name=None, last_name=None,
...
     account.save()
+
+
+@task(ignore_result=True)
+def send_scheduled_account_summary(interval):
+    users = User.objects.all()
+    for user in users:
+        user_account = user.get_profile().get_user_account()
+        if user_account.email_summary == interval:
+            send_user_account_summary(user)
```
hodgestar, line 35: Do we have a test for the task?

hodgestar, line 35: Also would be really good to get failed tasks into our logging and reporting somehow.

smn, line 35: We have a test for this task now.

hodgestar:
It might be nice to put the sentry handler code into a separate file and import it into settings.py and call it from there. We probably also want to put it after the import of production_settings.py and pass in the sentry details to the SentryHandler constructor?

hodgestar:
Task test looks good. Are the 'daily', 'weekly', etc values built into celery? Or do we set them up somewhere?

We should try simplify the sentry setup a bit using http://raven.readthedocs.org/en/latest/config/celery.html.

hodgestar:
I'm :+1: on this as soon as we sort out the sentry settings.

hodgestar:
New sentry settings have landed in production so I'm +1 on this.

PR #186, opened 2013-03-02
Stop double confirmations

smn:


go/conversation/base.py
```diff
@@ -250,15 +250,20 @@ def get(self, request, conversation):
...
         """
+        success = request.GET.get('success')
         token_manager = DjangoTokenManager(request.user_api.api.token_manager)
         token = request.GET.get('token')
         token_data = token_manager.verify_get(token)
         if not token_data:
             raise Http404
+        if token_data and success:
+            user_token, sys_token = token_manager.parse_full_token(token)
+            print 'deleting token', user_token
```
jerith: Debug print? Maybe that should be a `log.debug()`.

jerith:
:+1: modulo comment about print.

go/conversation/base.py
hodgestar: Debugging print?

go/conversation/base.py
```diff
@@ -250,15 +250,20 @@ def get(self, request, conversation):
...
+            token_manager.delete(user_token)
         return self.render_to_response({
             'form': ConfirmConversationForm(initial={'token': token}),
             'conversation': conversation,
-            'success': request.GET.get('success'),  # FIXME
+            'success': success,  # FIXME
             })
```
hodgestar: I'm a bit confused by this method? What actually triggers the bulk send?

hodgestar: Okay, so it's the post that triggers the conversation start. I would have expected the deletion of the token to go there (or maybe even in `_start_conversation_via_confirmation`) not in the confirmation get view?

hodgestar:
Should we also not start conversations that are already past the draft state? Should this be added to the token post? The conversation start method? Should we remove the token if the conversation was already started?

smn:
:+1: from me on this.

smn:
Technically the conversation is in a running state once the confirmation message has been sent out so confirmations will always start a confirmation that's technically already "running" since our definition of running is when a conversation has batch ids assigned.

PR #188, opened 2013-03-06
Fix confirm post tests and possibly protect against double-confirm race condition

hodgestar:
The post confirm test is failing (it asserts the wrong things) and there is a possibility of double-confirming a conversation if two HTTP requests are handled simultaneously.


jerith:
:+1:

PR #138, opened 2013-03-06
The Sent message list displays the contact for the `from_addr`

hodgestar:
Since this is usually the same, it's not very exciting. Sent messages should display the contact for the `to_addr` instead. This requires adding a `direction` argument to `get_contact_for_message` in `go.conversation.templatetags.conversation_tags` and then using that from the message list template.


jerith:
:+1:

PR #191, opened 2013-03-13
VUMIGO-293 split sent & received counts

smn:
The email summary lumped sent & received into one big total.
This breaks it up into separate values.


jerith:
:+1:

PR #193, opened 2013-03-15
http api needs to tell nginx to not buffer the responses.

smn:
We're streaming, buffering defeats the purpose.


jerith:
:+1:

vgeddes:
:+1: Looks good

PR #199, opened 2013-03-19
Creation of a getting started framework 

imsickofmaps:
For people new to vumi-go and comfortable with vagrant


smn:
:+1: thanks for contributing.

PR #197, opened 2013-03-19
we need a script to bootstrap a working base environment

smn:
We need to test some fairly radical changes in a local environment that mimics our production setup. We need a script to setup & teardown this environment.

What this needs to do is the following:
1. Set up at least 2 user accounts
2. Set up at least 4 tags
3. Assign 2 tags to each user account
4. Setup at least 8 conversations, half active & half ended. These conversations would be split between apps that do non-standard routing (WP, JSBox & Seq-send) & ones that do (bulk-msg & survey).
5. Provide means of automated end to end testing on a transport level once the accounts have been setup & routing is in place.


jerith:
Given that there's a bunch of UI stuff called "bootstrap", maybe we should rename this work? "setup_env" or something perhaps?

bootstrap/accounts.yaml
```diff
@@ -0,0 +1,18 @@
+user1@foo.org:
```
jerith: Maybe quote these and use "example.com" (which is explicitly reserved for this kind of thing) instead of "foo.org"?

go/base/management/commands/go_start_conversation.py
```diff
@@ -0,0 +1,64 @@
...
+        if not email_address:
+            raise CommandError('Please provide --email-address.')
+
+        conversation_key = options['conversation_key']
+        if not conversation_key:
+            raise CommandError('Please provide --conversation-key.')
+
+        conversation_type = options['conversation_type']
+        if not conversation_type:
+            raise CommandError('Please provide --conversation-type.')
```
jerith: I don't think this sanity check is really that necessary, but I'm happy to leave it in if you prefer.

smn:
up for review again @jerith

PR #204, opened 2013-03-27
Remove CONVERSATION_TYPES

jerith:
It's old and incomplete and we can get the information from other places.

### DIFF
--- a/go/conversation/forms.py
+++ b/go/conversation/forms.py
@@ -7,8 +7,6 @@ from bootstrap.forms import BootstrapForm
 
 from vumi.persist.fields import (ListProxy, ForeignKeyProxy, ManyToManyProxy)
 
-from go.base.utils import configured_conversation_types
-
 
 class VumiModelForm(BootstrapForm):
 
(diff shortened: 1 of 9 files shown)


go/conversation/forms.py
```diff
@@ -124,7 +125,8 @@ class ConversationSearchForm(BootstrapForm):
         'class': 'input-xlarge',
         }))
     conversation_type = forms.ChoiceField(required=False,
-        choices=([('', 'Type ...')] + CONVERSATION_TYPES),
+        choices=([('', 'Type ...')] +
+                 sorted(configured_conversation_types().items())),
```
smn: Side effect of this is that people can now see applications listed that they've not seen before but also don't have access to. I think we can live with that since this is just search but just highlighting it.

jerith: This was always the case, except the list was much shorter. We'll need something more dynamic to build the list based on the conversations the account has access to, but maybe that belongs in a different branch?

go/conversation/views.py
```diff
@@ -13,14 +13,18 @@
 @login_required
 def index(request):
     # grab the fields from the GET request
-    search_form = ConversationSearchForm(request.GET)
+    user_api = request.user_api
+    conversation_types = [(app['namespace'], app['display_name'])
+                          for app in user_api.applications().values()]
+    print conversation_types
```
smn: stray print

smn:
:+1:

PR #205, opened 2013-04-02
http api should respond with correct content-type

smn:
```
> GET /api/v1/go/http_api/xxxxxxxx/messages.json HTTP/1.1
> Authorization: Basic xxxxxxxxxxxxxxxxxxxxxxxx
> User-Agent: curl/7.24.0 (x86_64-apple-darwin12.0) libcurl/7.24.0 OpenSSL/0.9.8r zlib/1.2.5
> Host: go.vumi.org
> Accept: */*
> Connection: keep-alive
>
< HTTP/1.1 200 OK
< Server: nginx/0.7.65
< Date: Mon, 01 Apr 2013 08:57:39 GMT
< Content-Type: text/html
< Transfer-Encoding: chunked
< Connection: close
<
```

The `Content-Type: text/html` in the response should in this case be `Content-Type: application/javascript; charset=utf-8`

### DIFF
--- a/go/apps/http_api/client.py
+++ b/go/apps/http_api/client.py
@@ -2,7 +2,8 @@ import json
 
 from twisted.internet.defer import Deferred
 from twisted.internet import reactor
-from twisted.web.client import Agent, ResponseDone, ResponseFailed
+from twisted.web.client import (Agent, ResponseDone, ResponseFailed,
+    HTTPConnectionPool)
 from twisted.web import http
 from twisted.protocols import basic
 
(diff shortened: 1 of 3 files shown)


smn:
Also the content-type header is wrong.

smn:
This issue changed directions a few times, the "Connection: closed" stuff is handled curiously by Twisted to begin with but Twisted isn't adding that HTTP header, nginx is. This PR now fixes the published content-type & encoding values instead. Leaving the original comment for historical reasons.

go/apps/http_api/client.py
```diff
@@ -47,8 +47,8 @@ def connectionLost(self, reason):
         # the PotentialDataLoss here is because Twisted didn't receive a
         # content length header, which is normal because we're streaming.
         if (reason.check(ResponseDone, ResponseFailed, http.PotentialDataLoss)
-            and self._response is not None
-            and not self._wait_for_response.called):
+                and self._response is not None
+                and not self._wait_for_response.called):
```
smn, line 51: pep8 change

go/apps/http_api/resource.py
```diff
@@ -59,11 +61,13 @@ def __init__(self, worker, conversation_key):
         }
 
     def render_GET(self, request):
+        resp_headers = request.responseHeaders
+        request.defaultContentType = self.content_type
```
hodgestar: Is there a reason to set `defaultContentType` rather than just setting the header?

smn: Probably because I hadn't had coffee yet.

hodgestar:
:+1:

PR #208, opened 2013-04-05
"Reply To" function always replies to the most recent message.

smn:
Somehow our #reply-to anchor is gone missing.


justinvdm:
:+1:

PR #209, opened 2013-04-08
HTTP API needs to implement send_message command handler.

smn:
Currently it doesn't which leaves the UI hanging.


hodgestar:
It would be cool to make both implementing and testing these shorter but :+1: on this PR.

PR #212, opened 2013-04-16
Write a Vumi bridge transport

smn:
This would connect two Vumi instances using the streaming HTTP API available in https://github.com/praekelt/vumi-go (and which is being ported to core Vumi by @jerith in praekelt/vumi#486)

(this was originally praekelt/vumi#510)


go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -19,24 +22,7 @@
...
-
-    def onMessage(self, message):
-        self.inbox.put(message)
-
-    def onError(self, failure):
-        self.errors.put(failure)
-
-
-class TestEventReceiver(TestMessageReceiver):
-    message_class = TransportEvent
```
smn, line 39: These classes were completely not-used.

go/vumitools/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -0,0 +1,106 @@
...
+from twisted.web.http_headers import Headers
+from twisted.web import http
+
+from vumi.transports import Transport
+from vumi.config import ConfigText
+from vumi.message import TransportUserMessage, TransportEvent
+from vumi.utils import http_request_full
+from vumi import log
+
+from go.apps.http_api.client import StreamingClient
```
smn: Because the `StreamingClient` doesn't exist in Vumi this currently all lives in Vumi-Go. This needs to be moved over because we want the transport to live in Vumi. Any suggestions for how we'd like to go about that?

go/vumitools/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -0,0 +1,106 @@
...
+    This transport essentially connects as a client to Vumi Go's streaming
+    HTTP API [1]_.
+
+    It allows one to bridge Vumi and Vumi Go installations.
+
+    .. [1] https://github.com/praekelt/vumi-go/blob/develop/docs/http_api.rst
+
+    """
+
+    transport_type = 'http'
```
smn: Not sure on this yet.

go/vumitools/transports/vumi_bridge/vumi_bridge.py
smn: We could move the whole client to core Vumi and reference that from Vumi Go.

smn:
Stuff I realized that is missing: the message id's between both systems aren't the same, when we do a PUT Vumi Go should return its message_id and the transport should hold on to that to be able to map future events.

go/vumitools/transports/vumi_bridge/vumi_bridge.py
jerith: We can add this to praekelt/vumi#486, I suppose.

smn: We can but I don't want to have to wait for praekelt/vumi#486 to land. Unless you think that's quick?

hodgestar: I think landing the parts from Vumi Go that are needed by the client in the vumi repo can be considered a pre-cursor to praekelt/vumi#486.

go/apps/http_api/client.py
```diff
@@ -64,8 +67,11 @@ class StreamingClient(object):
     def __init__(self):
         self.agent = Agent(reactor)
 
-    def stream(self, message_class, callback, errback, url, headers=None):
-        receiver = VumiMessageReceiver(message_class, callback, errback)
+    def stream(
+        self, message_class, callback, errback, url, headers=None,
```
hodgestar: I think you can just wrap arguments like so:

```
def stream(self, ...,
           on_disconnect=None):
```

`self` underneath `stream(` looks a bit bizarre to me.

hodgestar:
:+1: other than my minor style comment about argument wrapping.

hodgestar:
Would like to see the transport get moved into Vumi fairly quickly though.

smn:
Will move this to Vumi today.

smn:
This now depends on praekelt/vumi#510 landing.

go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -70,8 +51,8 @@ def setUp(self):
         self.conversation.c.delivery_tag_pool = u'pool'
         self.tag = yield self.conversation.acquire_tag()
 
-        self.batch_id = yield self.vumi_api.mdb.batch_start([self.tag],
-                                    user_account=unicode(self.account.key))
+        self.batch_id = yield self.vumi_api.mdb.batch_start(
+            [self.tag], user_account=unicode(self.account.key))
```
smn, line 55: pep8 fix

hodgestar:
:+1:

PR #211, opened 2013-04-17
Contact management for jsbox

justinvdm:


go/vumitools/contact/models.py
```diff
@@ -99,6 +99,16 @@ def new_contact(self, **fields):
         returnValue(contact)
 
     @Manager.calls_manager
+    def update_contact(self, key, **fields):
+        # TODO remove once we have Model.from_data()
```
hodgestar: I think we can just scratch the TODO -- it might be awhile before we get `from_data` and there'll probably be subtle differences between the two approaches away.

go/vumitools/contact/models.py
```diff
@@ -99,6 +99,16 @@ def new_contact(self, **fields):
...
+        contact = yield self.get_contact_by_key(key)
+        if contact is not None:
+            for field_name, field_value in fields.iteritems():
+                setattr(contact, field_name, field_value)
```
justinvdm, line 109: Is there a better way to set a model's field?

go/apps/jsbox/tests/test_contacts.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.check_contact_reply(reply,
+            key=contact.key,
+            name=u'A Random',
+            surname=u'Person',
+            twitter_handle=u'random',
+            msisdn=u'unknown')
+
+    @inlineCallbacks
+    def test_handle_get_or_create(self):
```
smn, line 113: Can we add a test for `name=u'Zoë'`?

justinvdm, line 113: Good idea

smn:
:+1: nice snowman!

PR #215, opened 2013-04-18
Contact JSBox resource needs to be able to update extras and subscriptions

justinvdm:


go/vumitools/contact/models.py
```diff
@@ -130,8 +128,12 @@ def new_smart_group(self, name, query):
         yield group.save()
         returnValue(group)
 
+    @Manager.calls_manager
     def get_contact_by_key(self, key):
-        return self.contacts.load(key)
+        contact = yield self.contacts.load(key)
+        if contact is None:
+            raise RuntimeError("Contact with key '%s' not found." % key)
```
hodgestar: As a general rule, don't raise `RuntimeError` -- it makes catching this specific error hard. Make a specific exception class.

go/apps/jsbox/contacts.py
```diff
@@ -88,9 +86,36 @@ def handle_update(self, api, command):
...
+    def _update_dynamic_field(self, field_name, api, command):
+        try:
+            if 'key' not in command:
+                raise SandboxError("'key' needs to be specified for command")
+
+            store = self._contact_store_for_api(api)
+            contact = yield store.get_contact_by_key(command['key'])
+
+            field = getattr(contact, field_name)
+            field.update(self.pick_fields(command, *field.keys()))
```
hodgestar: I think users need to be allowed to set keys that don't exist yet.

go/apps/jsbox/contacts.py
```diff
@@ -88,9 +86,36 @@ def handle_update(self, api, command):
...
+            yield contact.save()
+        except (SandboxError, RuntimeError) as e:
```
hodgestar: This is why raising `RuntimeError` is bad -- this `except RuntimeError` clause could catch almost anything.

go/apps/jsbox/contacts.py
```diff
@@ -63,12 +63,9 @@ def handle_get_or_create(self, api, command):
...
             contact=contact.get_data()))
 
-    def _filter_contact_fields(self, command):
-        """
-        Returns only the fields of a command that are Contact model fields.
-        """
-        return dict((k, command[k]) for k in Contact.field_descriptors.keys()
-                    if k in command)
+    @staticmethod
+    def pick_fields(collection, *fields):
```
hodgestar, line 67: Plus one on the new name. :)

go/apps/jsbox/contacts.py
```diff
@@ -88,9 +90,42 @@ def handle_update(self, api, command):
...
+                raise SandboxError("'field' needs to be specified for command")
+
+            if 'value' not in command:
+                raise SandboxError("'value' needs to be specified for command")
+
+            store = self._contact_store_for_api(api)
+            contact = yield store.get_contact_by_key(command['key'])
+
+            field = getattr(contact, field_name)
+            field[command['field']] = command['value']
```
justinvdm, line 102: I'm not really happy about having the command look something like `{key: <contact_key>, field: <field_name>, value: <field_value>}`, any ideas on a better way to do this?

hodgestar, line 102: Maybe we should re-name key `contact_key` to differentiate it from `field` better?

go/apps/jsbox/contacts.py
```diff
@@ -88,9 +90,42 @@ def handle_update(self, api, command):
...
+            yield contact.save()
+        except (SandboxError, ContactError) as e:
+            log.warning(str(e))
+            returnValue(self.reply(command, success=False, reason=unicode(e)))
+
+        returnValue(self.reply(
+            command,
+            success=True,
+            contact=contact.get_data()))
```
hodgestar: Should we be returning the contact here? I think just success is enough.

go/apps/jsbox/contacts.py
```diff
@@ -63,12 +63,13 @@ def handle_get_or_create(self, api, command):
...
-        """
-        return dict((k, command[k]) for k in Contact.field_descriptors.keys()
-                    if k in command)
+    @staticmethod
+    def pick_fields(collection, *fields):
+        return dict((k, collection[k]) for k in fields if k in collection)
+
+    @staticmethod
+    def omit_fields(collection, *fields):
+        return dict((k, collection[k]) for k in fields if k not in collection)
```
hodgestar: Is anyone using omit_fields?

justinvdm: Ah, thanks, forgot to remove it

hodgestar:
:+1:

PR #213, opened 2013-04-18
Allow for auto-deploys of JS apps via Git commit hooks.

smn:
I don't want to have to "reload from server" every time my JS changes. I want the option of having it happen automagically.

See: https://help.github.com/articles/post-receive-hooks

### DIFF
--- a/go/api/conversation_api/conversation_api.py
+++ b/go/api/conversation_api/conversation_api.py
@@ -8,6 +8,7 @@ from twisted.internet.defer import inlineCallbacks, Deferred, succeed
 from vumi.config import ConfigText, ConfigDict, ConfigInt
 from vumi.transports.httprpc import httprpc
 from vumi.utils import http_request_full
+from vumi.worker import BaseWorker
 
 from go.apps.http_api.resource import AuthorizedResource, BaseResource
 from go.vumitools.app_worker import GoApplicationWorker
@@ -119,14 +120,14 @@ class AccountResource(AuthorizedResource):
     resource_class = ConversationApiResource
 
 
-class ConversationApiWorker(GoApplicationWorker):
+class ConversationApiWorker(BaseWorker):
 
     worker_name = 'conversation_api_worker'
     CONFIG_CLASS = ConversationApiWorkerConfig
+    SEND_TO_TAGS = frozenset(['default'])
 
     @inlineCallbacks
-    def setup_application(self):
-        yield super(ConversationApiWorker, self).setup_application()
+    def setup_worker(self):
         self.vumi_api = yield VumiApi.from_config_async(self.config)
         config = self.get_static_config()
         self.webserver = self.start_web_resources([
@@ -135,8 +136,7 @@ class ConversationApiWorker(GoApplicationWorker):
         ], config.web_port)
 
     @inlineCallbacks
-    def teardown_application(self):
-        yield super(ConversationApiWorker, self).teardown_application()
+    def teardown_worker(self):
         yield self.webserver.loseConnection()
 
     def setup_connectors(self):
--- a/go/api/conversation_api/conversation_api.py
+++ b/go/api/conversation_api/conversation_api.py
@@ -56,7 +56,7 @@ class ConversationConfigResource(BaseResource):
 
     def fallback_handler(self, request, command, conversation):
         request.setResponseCode(http.BAD_REQUEST)
-        succeed(1)
+        succeed(None)
 
     @inlineCallbacks
     def handle_jsbox(self, request, command, conversation):
@@ -100,7 +100,7 @@ class ConversationApiResource(BaseResource):
         return resource_class(self.worker, self.conversation_key)
 
 
-class ConversationApiWorkerConfig(GoApplicationWorker.CONFIG_CLASS):
+class ConversationApiWorkerConfig(BaseWorker.CONFIG_CLASS):
     worker_name = ConfigText(
         "Name of this tagpool API worker.", required=True, static=True)
     web_path = ConfigText(
--- a/go/api/conversation_api/conversation_api.py
+++ b/go/api/conversation_api/conversation_api.py
@@ -65,26 +65,25 @@ class ConversationConfigResource(BaseResource):
             # update the config blocks
             jsbox_app_config_md = metadata.get('jsbox_app_config', {})
             for key, config_section in jsbox_app_config_md.items():
-                src_url = config_section.get('source_url')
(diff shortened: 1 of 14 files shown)


go/apps/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,115 @@
...
+        request.finish()
+
+    def load_source_from_url(self, url, method='GET'):
+        # primarily here to make testing easier
+        return http_request_full(url, method=method)
+
+    @inlineCallbacks
+    def handle_jsbox(self, request, conversation):
+        metadata = conversation.get_metadata()
+        jsbox_md = metadata.get('jsbox')
```
hodgestar: We probably need to loop over all the bits of jsbox config that can be loaded from a source URL.

go/apps/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,115 @@
...
+                needs to look like, which is updating the source-code for
+                a JSBox application from it's source URL.
+        """
+        user_account = request.getUser()
+        conversation = yield self.get_conversation(user_account)
+        if conversation is not None:
+            handler = getattr(self,
+                              'handle_%s' % (conversation.conversation_type),
+                              lambda r, conversation: succeed(r))
+            yield handler(request, conversation)
```
hodgestar: I think we really need to get a command name in here somewhere otherwise we'll have to change the API later (which would be nice to avoid).

hodgestar: Maybe we can just chop the end off request.path ourselves with a regex? Don't forget to use `*?` for minimal expansion. :)

go/apps/conversation_api/conversation_api.py
smn: You mean `jsbox_app_config` as well? Are there more?

smn: Or are there potentially a number in the `jsbox_app_config` dictionary?

hodgestar: Yep (although jsbox_app_config is slightly different because each sub-key
is an individual thing that potentially has a source url).

go/api/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,140 @@
...
+    def handle_POST(self, request):
+        """
+        NOTE:   This at some point will grow the ability to accept generic
+                commands that a Conversation instance can act on. We're not
+                quite sure yet what that needs to look like so for now
+                we're hardcoding it for the case that we do know what it
+                needs to look like, which is updating the source-code for
+                a JSBox application from it's source URL.
+        """
+        if not (1 <= len(request.postpath) < 2):
```
hodgestar: Is this an odd way of writing `len(request.postpath) != 1` ? :)

go/api/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,140 @@
...
+            request.finish()
+            return
+
+        [command] = request.postpath
+        user_account = request.getUser()
+        conversation = yield self.get_conversation(user_account)
+        if conversation is not None:
+            handler = getattr(self,
+                              'handle_%s' % (conversation.conversation_type),
+                              lambda r, command, conversation: succeed(r))
```
hodgestar: Perhaps this should have a default that sets BAD_REQUEST on the request instead of succeeding?

go/api/conversation_api/conversation_api.py
smn: it's late, yes.

go/api/conversation_api/conversation_api.py
smn: good point, thanks.

go/api/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,140 @@
...
+                        config_section['value'] = response.delivered_body
+
+            # update the application code
+            jsbox_md = metadata.get('jsbox', {})
+            src_url = jsbox_md.get('source_url')
+            if src_url is not None:
+                response = yield self.load_source_from_url(src_url,
+                                                           method='GET')
+                if response.code == http.OK:
+                    jsbox_md['javascript'] = response.delivered_body
```
hodgestar: I think we should be able to abstract out a common bit of code here since `jsbox_md` is pretty similar to `config_section`?

go/api/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,144 @@
...
+    health_path = ConfigText(
+        "The path to server the health resource on.", default='/health/',
+        static=True)
+    redis_manager = ConfigDict(
+        "Redis client configuration.", default={}, static=True)
+    riak_manager = ConfigDict(
+        "Riak client configuration.", default={}, static=True)
+
+
+class AccountResource(AuthorizedResource):
```
hodgestar, line 118: Sub-classing `GoApplicationWorker` here is a bit odd since this isn't actually an `ApplicationWorker` (admittedly it's likely to be a bit fiddly to get the same functionality without sub-class -- hopefully not too fiddly :/)

smn, line 118: This comment ended up in the wrong place, but it's been addressed @hodgestar

go/api/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,143 @@
...
+            yield handler(request, command, conversation)
+        request.finish()
+
+    def load_source_from_url(self, url, method='GET'):
+        # primarily here to make testing easier
+        return http_request_full(url, method=method)
+
+    def fallback_handler(self, request, command, conversation):
+        request.setResponseCode(http.BAD_REQUEST)
+        succeed(1)
```
hodgestar: I would do `succeed(None)` because that's closer to what a function with no return statement would do.

go/api/conversation_api/conversation_api.py
```diff
@@ -0,0 +1,143 @@
...
+            'config': ConversationConfigResource
+        }
+        resource_class = class_map.get(path)
+        if resource_class is None:
+            return NoResource()
+
+        return resource_class(self.worker, self.conversation_key)
+
+
+class ConversationApiWorkerConfig(GoApplicationWorker.CONFIG_CLASS):
```
hodgestar: This should probably also inherit from `BaseWorker.CONFIG_CLASS` maybe?

hodgestar:
:+1: after the two most recent minor comments are addressed.

PR #217, opened 2013-04-22
Contact management fixes

jerith:


jerith:
We should probably check all the places we're calling things that used to return `None` and now raise an exception to make sure we're not broken in places the tests aren't testing.

That's out of scope for this branch, though.

hodgestar:
:+1: after checking the places we're calling things.

justinvdm:
:+1:

PR #219, opened 2013-04-22
ContactsResource update and new api methods need to set contact fields better

justinvdm:
Currently, ContactsResource's `.update()` and `.new()` methods need command objects like this to update a contact's fields: 

``` js
{key: '123abc', some_field: 'some_field_value'}
```

The following is command format is nicer  and allows a contact to have `key` field:

``` js
{key: '123abc', fields: {some_field: 'some_field_value'}}
```


hodgestar:
:+1:

PR #216, opened 2013-04-22
ContactsResource needs docstrings and commenting

justinvdm:


go/apps/jsbox/contacts.py
```diff
@@ -4,13 +4,20 @@
...
-from go.vumitools.contact import Contact, ContactError
+from go.vumitools.contact import Contact, ContactError, ContactNotFoundError
 
 
 class ContactsResource(SandboxResource):
+    """
+    Sandbox resource for accessing, creating and modifying contacts for
+    a Go application.
+
+    See :class:``go.vumitools.contact.Contact`` for a look at the Contact model
```
hodgestar: I think these should only have a single backtic on either side? Could be wrong.

justinvdm: I originally had it like that, but the [docs](http://sphinx-doc.org/rest.html) say differently. I know that in markdown single backticks work, not really sure if it works in reStructuredText.

jerith: You want double backticks for standalone things, single backticks if you're using the colons.

hodgestar: @justinvdm You're reading the wrong docs. :) See http://sphinx-doc.org/markup/inline.html#cross-referencing-objects and follow "Python" from there. These are refs, not code samples.

justinvdm: Ah! Sorry about that :)

go/apps/jsbox/contacts.py
```diff
@@ -30,6 +37,50 @@ def _parse_get(self, command):
...
+                    'surname': 'Person',
+                    'user_account': 'test-0-user',
+                    'bbm_pin': null,
+                    'msisdn': '+27831234567',
+                    'created_at': '2013-04-24 14:01:41.803693',
+                    'gtalk_id': null,
+                    'dob': null,
+                    'groups': ['group-a', 'group-b'],
+                    'facebook_id': null,
+                    '$VERSION': None,
```
smn, line 72: @jerith can this value potentially be set? If so, we need to fix that. (sorry @justinvdm that's got nothing to do with the specifics of this PR)

hodgestar, line 72: `$VERSION` is not settable via the current JS sandbox resource (since it's not a field) but will appear in the data returned and I think returning it will break `handle_save()` so we should pop it in `handle_save()`, I think.

hodgestar:
:+1: although the `$VERSION` issue @smn picked up needs to be handled somewhere (probably #220 is a better branch than this one but a completely new branch is fine too).

PR #222, opened 2013-04-23
ContactsResource needs a save api method

justinvdm:


go/apps/jsbox/contacts.py
```diff
@@ -134,3 +134,27 @@ def handle_new(self, api, command):
...
+
+            if 'key' not in command['contact']:
+                raise SandboxError(
+                    "'key' needs to be specified for as a 'contact' field for "
+                    "command")
+
+            contact_store = self._contact_store_for_api(api)
+
+            # raise an exception if the contact does not exist
+            yield contact_store.get_contact_by_key(command['contact']['key'])
```
hodgestar: I think it's okay to not raise an error if the contact is being created (at least that matches my expectation of what `.save()` would do). Thoughts?

go/vumitools/contact/models.py
```diff
@@ -114,6 +114,17 @@ def update_contact(self, key, **fields):
...
     @Manager.calls_manager
+    def save_contact(self, key, **fields):
+        # ensure no duplicate kwarg when calling `self.contacts()`
+        fields.pop('user_account', None)
+
+        contact = self.contacts(
+            key, user_account=self.user_account_key, **fields)
+
+        yield contact.save()
+        returnValue(contact)
```
hodgestar: I think maybe this method should go on the sandbox resource? Just wondering if everyone needs it.

go/vumitools/contact/models.py
```diff
@@ -114,6 +114,17 @@ def update_contact(self, key, **fields):
         returnValue(contact)
 
     @Manager.calls_manager
+    def save_contact(self, key, **fields):
+        # ensure no duplicate kwarg when calling `self.contacts()`
+        fields.pop('user_account', None)
```
hodgestar: Eep. We need to check that `handle_update` can't override `user_account` either.

go/apps/jsbox/contacts.py
```diff
@@ -134,3 +134,27 @@ def handle_new(self, api, command):
...
         returnValue(self.reply(command, success=True, key=contact.key))
+
+    @inlineCallbacks
+    def handle_save(self, api, command):
+        try:
+            if 'contact' not in command:
+                raise SandboxError(
+                    "'contact' needs to be specified for command")
+
+            if 'key' not in command['contact']:
```
smn: Higher level question, can a key be `None` and also should we allow keys to be set?

go/apps/jsbox/contacts.py
hodgestar: Hmm. I'm having second thoughts here. Probably ensuring the key exists and forcing people to call new is a good idea. Comment retracted.

go/vumitools/contact/models.py
justinvdm: Well caught! Thanks

go/apps/jsbox/contacts.py
justinvdm: Good point, we should probably check if it is `None` or does not exist.

About allowing keys to be set, I think we should maybe ensure that keys are automatically generated uuids, since allowing users to set keys would make the chance of collisions too likely.

go/apps/jsbox/contacts.py
```diff
@@ -134,3 +135,36 @@ def handle_new(self, api, command):
...
+
+    @inlineCallbacks
+    def handle_save(self, api, command):
+        try:
+            if 'contact' not in command:
+                raise SandboxError(
+                    "'contact' needs to be specified for command")
+
+            fields = command['contact']
+            if 'key' not in fields:
```
smn, line 147: If we just do an `if fields.get('key'): ...` here we'll catch both empty strings and `None` which I don't think we're currently catching further below.

PR #224, opened 2013-04-23
tox and CI stuff

jerith:
We need to set up CI for vumi-go and tox is a good way to do that. This means a bunch of (mostly minor) changes and fixes to requirements, `setup.py`, etc.

### DIFF
--- a/requirements.pip
+++ b/requirements.pip
@@ -12,17 +12,19 @@ Fabric==1.2.2
 celery==2.5.3
 django-celery==2.4.2
 django-celery-email
--e git+git://github.com/praekelt/vumi@develop#egg=vumi
--e git+git://github.com/praekelt/vxpolls.git@develop#egg=vxpolls
+https://github.com/praekelt/vumi/tarball/develop#egg=vumi-dev
+https://github.com/praekelt/vxpolls/tarball/develop#egg=vxpolls-dev
 Markdown==2.1.1
 django-registration==0.8
 # django_compressor==1.1.2 on Pypi has a bug, install from GitHub
 # until a new version has been published to pypi
--e git+git://github.com/jezdez/django_compressor@236e0d711#egg=compressor
+# -e git+git://github.com/jezdez/django_compressor@236e0d711#egg=compressor
+django_compressor>=1.2
 lesscpy==0.9h
 xlrd==0.8.0
--e git+git://github.com/earle/django-bootstrap.git@9ec5697#egg=bootstrap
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -1,12 +1,38 @@
...
+                name = "%(name)s==%(version)s" % ver_match.groupdict()
+            install_requires.append(name)
+            dependency_links.append(requirement)
+        else:
+            install_requires.append(requirement)
+    return install_requires, dependency_links
+
+install_requires, dependency_links = parse_requirements("requirements.pip")
+print install_requires
+print dependency_links
```
smn: stray prints

requirements.pip
```diff
@@ -12,17 +12,19 @@ Fabric==1.2.2
...
--e git+git://github.com/praekelt/vxpolls.git@develop#egg=vxpolls
+https://github.com/praekelt/vumi/tarball/develop#egg=vumi-dev
+https://github.com/praekelt/vxpolls/tarball/develop#egg=vxpolls-dev
 Markdown==2.1.1
 django-registration==0.8
 # django_compressor==1.1.2 on Pypi has a bug, install from GitHub
 # until a new version has been published to pypi
--e git+git://github.com/jezdez/django_compressor@236e0d711#egg=compressor
+# -e git+git://github.com/jezdez/django_compressor@236e0d711#egg=compressor
+django_compressor>=1.2
```
smn, line 22: This requirements can go but I suggest doing this in a separate PR as it also touches `go/settings.py`:

```
➜  go git:(feature/VUMIGO-176-endpoint-based-routing) ✗ git grep compress | grep py
go/base/widgets.py:            "%s/lib/codemirror-compressed.js" % (codemirror_path,),
go/settings.py:    'compressor.finders.CompressorFinder',
go/settings.py:    'compressor',
```

smn, line 22: Created #227 to address this.

setup.py
```diff
@@ -1,12 +1,36 @@
...
+
+
+def parse_requirements(filename):
+    install_requires = []
+    dependency_links = []
+    for requirement in listify(filename):
+        if requirement.startswith("#"):
+            continue
+        if requirement.startswith("-e"):
+            continue
```
smn, line 19: You've removed all the ones with `-e` should we log a warning here perhaps?

smn:
:+1: in general, just a question about `-e` instead of skipping them silently.

setup.py
jerith, line 19: I don't know how to do that in a useful way without potentially breaking packaging.

PR #220, opened 2013-04-23
ContactsResource Contact model changes should always be saved

justinvdm:


go/apps/jsbox/contacts.py
```diff
@@ -54,17 +53,26 @@ def handle_get(self, api, command):
     def handle_get_or_create(self, api, command):
         try:
             self._parse_get(command)
-            contact = yield self._contact_store_for_api(api).contact_for_addr(
+            contact_store = yield self._contact_store_for_api(api)
```
hodgestar: This doesn't need to yield, I don't think?

go/apps/jsbox/contacts.py
```diff
@@ -54,17 +53,26 @@ def handle_get(self, api, command):
...
+        except ContactNotFoundError:
+            contact = yield contact_store.new_contact_for_addr(
+                command['delivery_class'],
+                command['addr'])
+            created = True
+        except (SandboxError, ContactError) as e:
             log.warning(str(e))
             returnValue(self.reply(command, success=False, reason=unicode(e)))
 
+        yield contact.save()
```
hodgestar: Maybe we only need to save() if the contact was created?

go/apps/jsbox/tests/test_contacts.py
```diff
@@ -188,7 +191,10 @@ def test_handle_get_or_create_for_unicode_chars(self):
     def test_handle_get_or_create_for_nonexistent_contact(self):
         reply = yield self.dispatch_command('get_or_create',
                                             addr=u'+27831234567')
-        self.check_contact_reply(reply, msisdn=u'+27831234567')
+
+        self.check_reply(reply, created=True)
+        self.check_contact_fields(reply['contact']['key'],
+                                  msisdn=u'+27831234567')
```
hodgestar, line 197: We should add a check that the contact is now stored in Riak, I think.

go/apps/jsbox/contacts.py
hodgestar: Actually, `new_contact_for_addr` saves the address for us via `new_contact` so `.save()` doesn't need to be called here, I think.

go/vumitools/contact/models.py
```diff
@@ -260,39 +265,44 @@ def contact_has_opted_out(self, contact):
...
         returnValue(opt_out)
 
+    def _contact_field_for_addr(self, delivery_class, addr):
+        # TODO: change when we have proper address types in vumi
+        field = {
+            'sms': {'msisdn': '+' + addr.lstrip('+')},
+            'ussd': {'msisdn': '+' + addr.lstrip('+')},
+            'gtalk': {'gtalk_id': addr.partition('/')[0]},
+            'twitter': {'twitter_handle': addr},
+        }.get(delivery_class)
```
hodgestar: I think I still prefer the `if` version since all of the operations that happen to addr for other delivery_classes are non-sensical.

go/apps/jsbox/tests/test_contacts.py
justinvdm, line 197: `.check_contact_fields()` calls `ContactStore.get_contact_by_key()`, which calls `Contact.load()`, so I think it does check that the content is stored in Riak. I think I might be misunderstanding something though.

hodgestar, line 197: Woot.

go/vumitools/contact/models.py
```diff
@@ -83,10 +87,17 @@ def __unicode__(self):
...
 
 class ContactStore(PerAccountStore):
+    NONSETTABLE_CONTACT_FIELDS = ['$VERSION', 'user_account']
+
     def setup_proxies(self):
         self.contacts = self.manager.proxy(Contact)
         self.groups = self.manager.proxy(ContactGroup)
 
+    @classmethod
+    def settable_contact_fields(cls, **fields):
```
justinvdm, line 97: This is my attempt at preventing repitition in `.new_contact()`, `.update_contact()` and `ContactsResource.handle_save()`. Any ideas on a better way of doing this?

hodgestar:
:+1:

PR #225, opened 2013-04-23
ContactsResource update and save methods need to handle group fields

justinvdm:


go/vumitools/contact/models.py
```diff
@@ -88,30 +88,27 @@ def setup_proxies(self):
         self.groups = self.manager.proxy(ContactGroup)
 
     @Manager.calls_manager
-    def new_contact(self, **fields):
+    def new_contact(self, groups=[], **fields):
```
hodgestar: I think do `groups=()` or `groups=None` -- lists are mutable. `fields.pop('groups', [])` was also okay, I think. I think I also vaguely prefer `for x in y: do(x)` over `map(y, do)` (maybe that's just me though).

hodgestar: Do the changes in `new_contact` actually make any sort of difference? If not, can we maybe just leave things as they were?

justinvdm: They don't make a difference, was doing it so that `new_contact()` and `update_contact()` were a bit more consistent. Will be change it back to how it was though (and change `update_contact()` to do the same)

hodgestar:
:+1:

PR #228, opened 2013-04-24
AccountRoutingTableDispatcher needs to handle opt-outs.

smn:
Currently that lives in the application message dispatcher, it needs to be ported over.

### DIFF
--- a/go/vumitools/routing.py
+++ b/go/vumitools/routing.py
@@ -72,7 +72,7 @@ class AccountRoutingTableDispatcher(RoutingTableDispatcher, GoWorkerMixin):
         return mapping.get(conversation_type, conversation_type)
 
     def handle_opt_out(self, msg):
-        return self.publish_message(
+        return self.publish_inbound(
             msg, self.get_static_config().opt_out_connector, 'default')
 
     def process_inbound(self, config, msg, connector_name):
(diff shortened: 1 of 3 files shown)


go/vumitools/routing.py
```diff
@@ -0,0 +1,141 @@
...
+    application_connector_mapping = ConfigDict(
+        "Mapping from conversation_type to connector name.", static=True)
+    message_tag = ConfigList("Tag for the message, if any.")
+    tagpool_metadata = ConfigDict(
+        "Tagpool metadata for the tag attached to the message if any.")
+    conversation_info = ConfigDict(
+        "Conversation info for the tag attached to the message if any.")
+    user_account_key = ConfigText("UserAccount key.")
+    opt_out_connector = ConfigText(
+        "Connector to publish opt-out messages on.", static=True)
```
hodgestar: I would move this to the top along with the other static config item.

hodgestar:
:+1: on merging this into #180. The merge will unfortunately be a little messy because some small changes landed in the routing dispatcher (nothing huge, but enough that I think auto merge will fail).

go/vumitools/routing.py
```diff
@@ -0,0 +1,141 @@
...
+            msg, self.get_static_config().opt_out_connector, 'default')
+
+    def process_inbound(self, config, msg, connector_name):
+        log.debug("Processing inbound: %s" % (msg,))
+        if not config.message_tag:
+            log.warning("No tag found for inbound message: %s" % (msg,))
+            return
+
+        if OptOutMiddleware.is_optout_message(msg):
+            return self.handle_opt_out(msg)
```
jerith, line 89: I think we still need to add the account key to the helper metadata here.

hodgestar:
Looks good to me.

PR #218, opened 2013-04-25
ContactsStore dependents should handle ContactError exceptions properly

justinvdm:


justinvdm:
There were a few places that used `ContactStore.get_contact_by_key()`. I haven't guarded any of them in try-except blocks, because all of them seem to assume we get a contact back instead of None, so I'm not sure which cases need to be guarded, and how to handle these cases.

go/apps/sna/handlers.py
```diff
@@ -12,10 +12,15 @@
 
 class SNAEventHandler(EventHandler):
 
+    @inlineCallbacks
     def find_contact(self, account_key, msisdn):
-        contact_store = ContactStore(self.dispatcher.vumi_api.manager,
-                                        account_key)
-        return contact_store.contact_for_addr('ussd', msisdn)
+        try:
+            contact_store = ContactStore(
```
jerith: No need for this to be in the `try` block.

go/contacts/tests.py
```diff
@@ -388,26 +388,6 @@ def test_contact_key_value_query(self):
...
-            sms_contact.key)
-        self.assertEqual(
-            self.contact_store.contact_for_addr('twitter', u'@someone').key,
-            twitter_contact.key)
-        self.assertEqual(
-            self.contact_store.contact_for_addr('gtalk',
-                u'gtalk@host.com').key, gtalk_contact.key)
-        self.assertRaisesRegexp(ContactError, 'Unsupported transport_type',
-            self.contact_store.contact_for_addr, u'unknown', u'unknown')
-
```
jerith, line 410: Why is this test going away?

justinvdm, line 410: We test `ContactStore.contact_for_addr()` in the `ContactStore` tests over [here](https://github.com/praekelt/vumi-go/blob/develop/go/vumitools/tests/test_contact.py#L189), where I think it should be tested. I have a feeling I'm missing something though. Is there another reason it was being tested here?

jerith, line 410: No idea. "duplicate test" is a good enough reason to remove it.

go/vumitools/app_worker.py
```diff
@@ -203,10 +204,13 @@ def get_contact_for_message(self, message):
...
-            contact = yield user_api.contact_store.contact_for_addr(
-                conv.delivery_class, message.user())
 
-            returnValue(contact)
+            try:
+                contact = yield user_api.contact_store.contact_for_addr(
+                    conv.delivery_class, message.user())
+                returnValue(contact)
+            except ContactError:
+                returnValue(None)
```
jerith: Instead of returning `None` here, we should probably add a `create` param to the method and pass it through to `contact_for_addr()`.

go/apps/surveys/vumi_app.py
```diff
@@ -102,8 +106,12 @@ def end_session(self, participant, poll, message):
...
         # This does that.
-        contact = yield self.get_contact_for_message(message)
-        if contact:
+
+        try:
+            contact = yield self.get_contact_for_message(message, create=False)
+        except ContactError:
+            pass
+        else:
             # Clear previous answers from this poll
```
justinvdm: Is there a better of way letting this block of code run if a contact was found?

jerith: When `contact_for_addr()` always created a contact when there wasn't already one, that check would always have passed. I think it's safe to just pass `create=True` in here and mimic the old behaviour.

jerith:
:+1:

PR #230, opened 2013-04-29
ContactStore.contact_for_addr() needs to use load_all_bunches

smn:
Currently we're doing a `load(keys[0])`. The first key we get back might have been deleted from Riak, `load_all_bunches(keys)` filters out the deleted ones.

When this happens the `load(keys[0])` returns a `None` instead of a contact instance.

### DIFF
--- a/go/vumitools/contact/models.py
+++ b/go/vumitools/contact/models.py
@@ -7,7 +7,7 @@ from twisted.internet.defer import returnValue
 
 from vumi.persist.model import Model, Manager
 from vumi.persist.fields import (Unicode, ManyToMany, ForeignKey, Timestamp,
-                                    Dynamic)
+                                 Dynamic)
 
 from go.vumitools.account import UserAccount, PerAccountStore
 from go.vumitools.opt_out import OptOutStore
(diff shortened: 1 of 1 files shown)


go/vumitools/contact/models.py
```diff
@@ -298,7 +298,11 @@ def contact_for_addr(self, delivery_class, addr, create=True):
         keys = yield self.contacts.search(**field).get_keys()
 
         if keys:
-            returnValue((yield self.contacts.load(keys[0])))
+            contacts = []
+            bunches = yield self.contacts.load_all_bunches(keys)
+            for bunch in bunches:
+                contacts.extend((yield bunch))
+            returnValue(max(contacts, key=lambda c: c.created_at))
```
justinvdm: Can we be sure that `contacts` will be non-zero?

jerith:
:+1:

go/vumitools/contact/models.py
justinvdm: s/non-zero/non-empty

smn: hrm, right, no - thanks!

justinvdm:
:+1:

PR #231, opened 2013-04-29
CSVs need to be zipped when emailed

smn:
For some reason CSV attachments can get folded at 1000 chars, making them hard to parse. The hope is that zipping it will prevent that from happening.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -120,8 +120,7 @@ def export_group_contacts(account_key, group_key, include_extra):
         'group "%s" attached.\n\n' % (len(contact_keys), group.name),
         settings.DEFAULT_FROM_EMAIL, [user_profile.user.email])
 
-    zipio.seek(0)
-    email.attach('contacts-export.zip', zipio.read(), 'application/zip')
+    email.attach('contacts-export.zip', zipio.getvalue(), 'application/zip')
     email.send()
 
 
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -1,6 +1,7 @@
 import sys
 import traceback
 from StringIO import StringIO
+from zipfile import ZipFile
 
 from celery.task import task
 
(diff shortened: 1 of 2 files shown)


go/contacts/tasks.py
```diff
@@ -83,8 +84,8 @@ def export_group_contacts(account_key, group_key, include_extra):
 
     # collect all names that we can export
     fields = ['name', 'surname', 'email_address', 'msisdn', 'dob',
-                    'twitter_handle', 'facebook_id', 'bbm_pin', 'gtalk_id',
-                    'created_at']
+              'twitter_handle', 'facebook_id', 'bbm_pin', 'gtalk_id',
+              'created_at']
```
smn, line 88: pep8 fix

go/contacts/tasks.py
```diff
@@ -95,14 +96,13 @@ def export_group_contacts(account_key, group_key, include_extra):
 
     # write the CSV header
     extra_fields = sorted(extra_fields)
-    writer.writerow(fields + ['extras-%s' % (key,)
-                                    for key in extra_fields])
+    writer.writerow(fields + ['extras-%s' % (key,) for key in extra_fields])
```
smn, line 99: pep8 fix

go/contacts/tasks.py
```diff
@@ -95,14 +96,13 @@ def export_group_contacts(account_key, group_key, include_extra):
...
 
     # loop over the contacts and create the row populated with
     # the values of the selected fields.
     for contact in contacts:
         row = [unicode(getattr(contact, field, None) or '')
-                for field in fields]
+               for field in fields]
```
smn, line 105: pep8 fix

go/contacts/tasks.py
```diff
@@ -110,12 +110,18 @@ def export_group_contacts(account_key, group_key, include_extra):
...
+
+    email = EmailMessage(
+        '%s contacts export' % (group.name,),
+        'Please find the CSV data for %s contact(s) from '
+        'group "%s" attached.\n\n' % (len(contact_keys), group.name),
         settings.DEFAULT_FROM_EMAIL, [user_profile.user.email])
-    email.attach('contacts-export.csv', io.getvalue(), 'text/csv')
+
+    zipio.seek(0)
+    email.attach('contacts-export.zip', zipio.read(), 'application/zip')
```
hodgestar: Is there a reason we do a `.seek(0)` and `.read()` here rather than just doing `.getvalue()`?

hodgestar:
:+1: other than the minor stylistic query to show I read it. :)

go/contacts/tasks.py
smn: `getvalue()` can only be used before `close()` is called: http://docs.python.org/2/library/stringio.html#StringIO.StringIO.getvalue

hodgestar: Hmm. `close()` doc string says _any operations_ after calling `close()` are bad. Maybe we shouldn't use `with` then (since that closes the file)?

hodgestar:
:+1:

PR #227, opened 2013-05-02
drop django-compressor dependency.

smn:
It isn't used anywhere and is just cluttering our settings.py and various packaging things. See #224 for details.


jerith:
:+1:

PR #214, opened 2013-05-02
go_setup_env needs updating for allowing non-localhost connections

smn:
Allows mapped ports from a vagrant host machine etc.
https://gist.github.com/anonymous/23701bd31a944d79f8f0


imsickofmaps:
+1

PR #236, opened 2013-05-02
Better Wikipedia app tests

jerith:
The next set of updates to the vumi-wikipedia app (praekelt/vumi-wikipedia#21) changes a bunch of test machinery and breaks the current Wikipedia app tests. We really need more Go-specific tests for this anyway.


smn:
:+1:

PR #240, opened 2013-05-02
Fix bug in new Outbound resource and add tests.

hodgestar:
Recent updates to the Outbound resource added new methods without tests and these methods contained a subtle bug (helper_metadata wasn't being preserved in replies).

### DIFF
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -48,8 +48,8 @@ class GoOutboundResource(SandboxResource):
         d = reply_func(orig_msg, content, continue_session=continue_session,
                        helper_metadata=helper_metadata)
         d.addCallback(lambda r: self.reply(command, success=True))
-        d.addErrback(lambda f: self.reply(command, success=False,
-                                          reason=unicode(f.getErrorMessage())))
+        d.addErrback(lambda f: self._mkfail(command,
+                                            unicode(f.getErrorMessage())))
         return d
 
     def handle_reply_to(self, api, command):
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -16,25 +16,28 @@ class GoOutboundResource(SandboxResource):
     messages via given tags.
     """
 
+    def _mkfail(self, command, reason):
+        return self.reply(command, success=False, reason=reason)
+
+    def _mkfaild(self, command, reason):
+        return succeed(self._mkfail(command, reason))
+
     def _handle_reply(self, api, command, reply_func):
         if not isinstance(command.get('content'), unicode):
-            return succeed(self.reply(
-                command, success=False,
-                reason=u"'content' must be given in replies."))
+            return self._mkfaild(
+                command, reason=u"'content' must be given in replies.")
         if not isinstance(command.get('in_reply_to'), unicode):
-            return succeed(self.reply(
-                command, success=False,
-                reason=u"'in_reply_to' must be given in replies."))
+            return self._mkfaild(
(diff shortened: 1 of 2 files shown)


go/apps/jsbox/outbound.py
```diff
@@ -3,36 +3,134 @@
...
+
+    Includes support for replying, replying to groups and sending
+    messages via given tags.
+    """
 
     def _handle_reply(self, api, command, reply_func):
+        if not isinstance(command.get('content'), unicode):
+            return succeed(self.reply(
+                command, success=False,
+                reason=u"'content' must be given in replies."))
```
smn: Do we want to allow blank replies?

go/apps/jsbox/outbound.py
```diff
@@ -3,36 +3,134 @@
...
+        if command.get('continue_session', True) not in (True, False):
+            return succeed(self.reply(
+                command, success=False,
+                reason=u"'continue_session' must be either true or false"
+                " if given"))
+        orig_msg = api.get_inbound_message(command['in_reply_to'])
+        if orig_msg is None:
+            return succeed(self.reply(
+                command, success=False, reason=u"Could not find original"
+                " message with id: %r" % command['in_reply_to']))
```
justinvdm: If we refactored this block to raise a `SandboxError` if validation fails, we would only do the `return succeed(self.reply(command, success=False, reason=u"..."))` thing once when we catch the exception, and could log something as well? Definitely not important, your call.

justinvdm:
Other than my minor comment, :+1:

go/apps/jsbox/outbound.py
hodgestar: We do. /me fixes.

go/apps/jsbox/outbound.py
```diff
@@ -3,48 +3,147 @@
...
-        helper_metadata = conv.set_go_helper_metadata()
-        return reply_func(orig_msg, content, continue_session=continue_session,
-                          helper_metadata=helper_metadata)
+        helper_metadata = conv.set_go_helper_metadata(
+            orig_msg['helper_metadata'])
+
+        d = reply_func(orig_msg, content, continue_session=continue_session,
+                       helper_metadata=helper_metadata)
+        d.addCallback(lambda r: self.reply(command, success=True))
+        d.addErrback(lambda f: self.reply(command, success=False,
```
justinvdm: Would it maybe make sense to use `_mkfail()` here?

go/apps/jsbox/outbound.py
hodgestar: Refactored a bit. We really need to add decorators for all this checking.

go/apps/jsbox/outbound.py
hodgestar: Actually, having thought about it a bit more I'm not sure we do. In general, `content` may be `null` in Vumi but the only cases this typically happens are session open and close and I don't think it makes much sense to allow replies that close a session to have no content? Thoughts?

smn: Just copying over from IRC:

```
10:46 <sdehaan> Hodgestar: I guess in an end-session message no content can make sense
10:47 <sdehaan> Hodgestar: I'm -0 because it feels like we're missing something there but I don't have a clear answer on exactly what yet.
10:47 <sdehaan> Hodgestar: so +1 on leaving it
```

hodgestar: I add support for `content: null` but not leaving `content` out entirely (which keeps the feature but requires users ask for null content explicitly).

smn:
:+1:

PR #235, opened 2013-05-02
Client side testing boilerplate

justinvdm:
Since the new Go UI will be needing considerably more complex client side logic than the previous UI, we will need to include client side testing.


smn:
+100

justinvdm:
I'm not sure how, but we will probably need to include these client side tests in the CI we are now doing with Travis. I'll look into doing that in the PR associated to this issue.

go/base/static/js/test/runner.html
```diff
@@ -0,0 +1,35 @@
...
+    <script src="../vendor/jquery.js"></script>
+    <script src="../vendor/jquery.ui.js"></script>
+    <script src="../vendor/jquery.jsPlumb.js"></script>
+    <script src="../vendor/underscore.js"></script>
+    <script src="../vendor/backbone.js"></script>
+
+    <script src="../src/init.js"></script>
+    <script src="../src/campaigns/init.js"></script>
+    <script src="../src/campaigns/setup/init.js"></script>
+    <script src="../src/campaigns/setup/interactive.js"></script>
```
justinvdm, line 27: We can avoid having to specify each and every module, vendor lib and test module if we concat our modules using something like [grunt](http://gruntjs.com/). I know that the current plan is to use django-pipeline, which appears to concatenate the files at runtime. This means we will need to continue specifying everything here. I'm fine with that, though. Any thoughts or suggestions?

justinvdm:
One concern of mine is that the tests are in the static dir (`go/base/static/js/test`), so they are publically viewable, which is a bit wierd. I could move them out of static, but then they are far away from the code they are testing, which isn't good. Any thoughts or suggestions?

smn:
The whole repository is open so it's always publicly viewable, not sure what the problem is with that?

justinvdm:
Not really worried about security at all. I'm more worried that its bad practice, but I'm probably overthinking things.

smn:
I'm happy with them living in `static` for the time being until we find a better way of doing it or find it isn't a problem after all.

peterp:
Pipelines creates a unique filename for the the minified and combined library if changes are detected. (In order to overcome agressive caching.)

I'm guessing that this might be a problem as you'll have to modify the tests if you want to test against the files in production.

peterp:
:+1:

PR #203, opened 2013-05-03
Conversation updates for new routing mechanism

jerith:
Now that we have new routing in place (once #180 lands, at least) we can move on to the conversation model and lifecycle updates that take advantage of it.


go/base/management/commands/go_migrate_conversations.py
```diff
@@ -0,0 +1,68 @@
...
+        self.outln(
+            u'%s %s <%s> [%s]\n    Conversations: %s' % (
+            user.first_name, user.last_name, user.username,
+            user_api.user_account_key, len(conversations)))
+        if migrate:
+            for conv_key in conversations:
+                self.outln(u'  Migrating conversation: %s' % (conv_key,))
+                conv = user_api.get_wrapped_conversation(conv_key)
+                self.outln(u'    - [%s] migrated' % (conv.name,))
+                conv.save()
```
hodgestar, line 62: Should the two lines above maybe be swapped?

go/conversation/templates/generic/confirm.html
```diff
@@ -19,13 +19,13 @@
     <table class="table table-bordered">
         <thead>
             <tr>
-                <th colspan="2">{{conversation.subject}}</th>
+                <th colspan="2">{{conversation.name}}</th>
             </tr>
         </thead>
         <tbody>
             <tr>
                 <th>Message:</th>
```
hodgestar, line 27: Can we update this to "Description:" (I know there are a lot of other places -- no need to do them all in this PR but doing the easily findable ones would save a lot of PRs and such later).

go/conversation/views.py
```diff
@@ -33,13 +33,13 @@ def index(request):
...
+    if query:
         conversations = [c for c in conversations
-                            if c.conversation_type == conversation_type]
+                         if query.lower() in c.name.lower()]
 
-    if query:
+    if conversation_type:
         conversations = [c for c in conversations
-                            if query.lower() in c.subject.lower()]
+                         if c.conversation_type == conversation_type]
```
hodgestar, line 38: Did these swap around for any particular reason?

go/vumitools/conversation/models.py
```diff
@@ -12,40 +12,46 @@
...
     """A conversation with an audience"""
+
+    VERSION = 1
+    MIGRATOR = ConversationMigrator
+
     user_account = ForeignKey(UserAccount)
-    subject = Unicode(max_length=255)
-    message = Unicode()
-    start_timestamp = Timestamp()
+    name = Unicode(max_length=255)
```
hodgestar, line 30: Maybe an index on name would be good too?

go/vumitools/conversation/models.py
```diff
@@ -118,3 +126,43 @@ def new_conversation(self, conversation_type, subject, message,
...
+    def list_active_conversations(self):
+        # We need to list things by both the old-style 'end_timestamp' index
+        # and the new-style 'status' index, at least until we've migrated all
+        # old-style conversations to v1.
+        keys = yield self.conversations.index_lookup(
+            'status', CONVERSATION_RUNNING).get_keys()
+        keys.extend((yield self.conversations.index_lookup(
+            'status', CONVERSATION_DRAFT).get_keys()))
+        keys.extend((yield self.conversations.index_lookup(
+            'end_timestamp', None).get_keys()))
```
hodgestar, line 163: Do we need to use `_list_oldstyle_running_conversations` here too?

hodgestar:
I left a bunch of comments but nothing too serious -- on the whole this looking fine to me.

go/conversation/templates/generic/confirm.html
jerith, line 27: Probably. I did the very minimum UI work possible in here.

go/conversation/views.py
jerith, line 38: I probably rewrote it several times and it just ended up that way. I'll swap them back.

go/base/management/commands/go_migrate_conversations.py
jerith, line 62: Good point, thanks.

go/vumitools/conversation/models.py
jerith, line 30: I don't think it's really necessary, given that we can only do exact match queries, but it can't hurt. I'll add one.

go/vumitools/conversation/models.py
jerith, line 163: The `.index_lookup('end_timestamp', None)` is the only part of that we actually need.

go/conversation/templates/generic/confirm.html
jerith, line 27: On second thoughts, the only conversation which actually uses this field is "bulk_message", for which "Message" is a much better name. I'd rather leave this rename for the more extensive UI work.

hodgestar:
:+1:

go/base/tests/test_go_migrate_conversations.py
```diff
@@ -0,0 +1,79 @@
...
+        super(GoMigrateConversationsCommandTestCase, self).setUp()
+        self.user = self.mk_django_user()
+        self.setup_user_api(self.user)
+
+        self.old_conv_model = self.user_api.conversation_store.manager.proxy(
+            ConversationVNone)
+
+        self.conv1 = self.mkoldconv(subject=u'Old 1')
+        self.conv2 = self.mkoldconv(subject=u'Old 1')
+        self.conv3 = self.mkoldconv(subject=u'Zoë destroyer of Ascii')
```
smn, line 25: :heart:

go/conversation/templates/generic/confirm.html
smn, line 27: :+1: on leaving it as is, less template changes here means less chances of conflicts when I merge this into `develop-frontend`

go/vumitools/conversation/models.py
```diff
@@ -98,7 +107,7 @@ def get_conversation_by_key(self, key):
         return self.conversations.load(key)
 
     @Manager.calls_manager
-    def new_conversation(self, conversation_type, subject, message,
+    def new_conversation(self, conversation_type, name, config,
```
smn: This seems like a fairly scary change in positional arguments, is it absolutely necessary?

smn:
Looks good but a bit worried about the signature change in `new_conversation()`

go/vumitools/conversation/models.py
jerith: I suppose we can make `description` (formerly `message`, but it means something different now) mandatory in there again, but we still need `config` to be mandatory.

smn: +1 on adding `description` and keeping `config`.

PR #242, opened 2013-05-05
Client side error infrastructure

justinvdm:


go/base/static/js/src/errors/errors.js
```diff
@@ -0,0 +1,33 @@
+(function(exports) {
+  var GoError = exports.GoError = function(message) {
+    if (message) { this.message = message; }
+  };
+
+  // Creates a sub-error constructor from 'this' error constructor
+  GoError.suberror = function(name, message) {
```
justinvdm: Any suggestions for a better name than `suberror`?

smn: What is a `suberror` exactly?

go/base/static/js/test/errors/errors.test.js
```diff
@@ -0,0 +1,32 @@
...
+      assert.equal('' + new GoError('Aaah!'), '[GoError: Aaah!]');
+      assert.equal('' + new GoError(), '[GoError]');
+
+      assert.equal('' + new SubError('Aaah!'), '[SubError: Aaah!]');
+      assert.equal('' + new SubError(), '[SubError]');
+    });
+
+    it("should use a default message if no message is provided", function() {
+      var SubError = GoError.suberror('SubError', 'Aaah!');
+      assert.equal(new GoError().message, '');
```
smn, line 28: Where does this `''` empty string come from?

go/base/static/js/src/errors/errors.js
smn: Having read the code, I'm not sure I have a better suggestion. I guess since you're creating a new subclass / type perhaps a name with `make_*` or `create_*` in it could make sense but honestly I am ±0.

justinvdm: `var SomeOtherError = SomeError.suberror('SomeOtherError')` would create a new error type called `SomeOtherError` that has as its parent error `SomError`. In a try-catch block, the exception instance `e` would be an instance of both `SomeOtherError` and `SomeError`, since `SomeOtherError` is a suberror of `SomeError`. Make any sense?

go/base/static/js/test/errors/errors.test.js
justinvdm, line 28: This comes from `Error` (the builtin error), `Error.prototype.message === ''`

smn, line 28: Aah, thanks.

go/base/static/js/src/errors/errors.js
smn: Best suggestion I can come up with now is `subtype` instead of `suberror`

smn:
:+1: curious to see how we can hook this into travis / jenkins setup.

smn:
this as in "frontend js tests"

go/base/static/js/src/errors/errors.js
justinvdm: Ah, sounds better.

PR #251, opened 2013-05-10
Client side needs simple inhertance infrastructure

justinvdm:
When creating with objects that aren't Backbone structures (Model, View, Router, etc), we are left without an easy way to inherit. Instead of verbose workarounds or including another dependency, I think this can be done quite simply by borrowing as much as we can from Backbone. This will also give us some uniformity in our code, since our structures will look similar to Backbone structures.

### DIFF
--- a/go/base/static/js/src/utils/utils.js
+++ b/go/base/static/js/src/utils/utils.js
@@ -1,0 +1,8 @@
+(function(exports) {
+  exports.Extendable = function () {};
+
+  // Backbone has an internal `extend()` function which it assigns to its
+  // structures. We need this function, so we arbitrarily choose
+  // `Backbone.Model`, since it has the `extend()` function we are looking for.
+  exports.Extendable.extend = Backbone.Model.extend;
+})(go.utils = {});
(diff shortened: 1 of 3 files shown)


go/base/static/js/src/utils/utils.js
```diff
@@ -0,0 +1,8 @@
+(function(exports) {
+  exports.Extendable = function () {};
+
+  // Backbone has an internal `extend()` function which it assigns to its
+  // structures. We need this function, so we arbitrarily choose
+  // `Backbone.Model`, since it has the `extend()` function we are looking for.
+  exports.Extendable.extend = Backbone.Model.extend;
+})(go.utils = {});
```
hodgestar, line 8: These 8 lines succinctly encapsulate the insanity of Javascript.

hodgestar:
:+1:

go/base/static/js/src/utils/utils.js
peterp, line 8: Hehehehe.

PR #253, opened 2013-05-11
Contacts Resource API needs fixing

justinvdm:


justinvdm:
#### `contact_key` -> `key`

We now use `key` consistently (instead of `contact_key` for `update_extra()` and `update_subscription()`).
I chose to make this `key` instead of `contact_key` because the docstrings explain what key means, `contact_key` is unnecessarily verbose, and `key` is more consistent. For example, in `save()`, we require one of the fields in `contact` to be `key` (which shouldn't be changed), so if we leave `save()` as is, using `key` everywhere is more consistent.

#### `update_extra()` -> `update_extras()`, `update_subscription()` -> `update_subscriptions()`

`update_extras()` and `update_subscriptions()` can now update multiple fields. Being forced to update one extra or subscription field at a time is an annoying and unneccessary constraint. Also, having to specify `field` and `value` fields is wierd and tedious.

#### All methods return a `contact` object

Returning `contact` in all the api methods allows the user to do chaining and have the correctly updated contact available to them all the time. Secondly, it is common for APIs to return an updated model as a response to an update/save method (github do this with their [HTTP API](http://developer.github.com/v3/)). Lastly, the user doesn't need to hold on to their new field values when they issue an update or save, they have these available to them in the returned contact.

go/apps/jsbox/contacts.py
```diff
@@ -180,85 +190,102 @@ def handle_update(self, api, command):
...
+            - ``success``: set to ``false``
+            - ``reason``: Reason for the failure
 
         Example:
         .. code-block:: javascript
             api.request(
-                'contacts.update_extra',
-                {contact_key: '123abc', field: 'surname', value: 'Jones'},
+                'contacts.update_extras',
+                {fields: {location: 'CPT', beer: 'Whale Tail Ale'}},
```
smn, line 260: Whale Tail Ale ++

go/apps/jsbox/tests/test_contacts.py
```diff
@@ -48,13 +50,18 @@ def tearDown(self):
...
-        for key, expected_value in kw.iteritems():
-            self.assertEqual(reply[key], expected_value)
 
-    def check_contact_reply(self, reply, **expected_fields):
-        for field_name, expected_value in expected_fields.iteritems():
-            self.assertEqual(reply['contact'][field_name], expected_value)
-        self.check_reply(reply)
+        # get a dict of the reply fields that we can pop items off without
+        # worrying about modifying the actual reply
+        reply = json.loads(reply.to_json())
```
smn, line 56: We have [Message.copy()](https://github.com/praekelt/vumi/blob/develop/vumi/message.py#L109) for that.

smn:
:+1: from me on that. It's likely that some JS apps that use contacts will break as a result of this though. As far as I know it's only @imsickofmap's demo apps & https://github.com/praekelt/mama-ussd stuff and my http://github.com/smn/go-break-the-rules things.

go/apps/jsbox/contacts.py
justinvdm, line 260: :)

PR #255, opened 2013-05-12
Client side code needs to be reorganised

justinvdm:
Currently, the client side modules are a bit over-structured, with dirs being created when they don't need to be. This needs to be changed now, while not much of the client side code has been implemented.


peterp:
:+1: 

Just for my own purposes, I'm going to be creating the bulk messaging view now;

would this go into something like `go/base/static/js/src/campaign/bulk_message.js` with a module of `go.campaign.bulkMessage?`

justinvdm:
Good question. I'm thinking `bulkMessage` since, that's the convention we've been using. That ok?

justinvdm:
Erm, sorry that was a vague answer. I meant `go/base/static/js/src/campaign/bulk-message.js`, and `go.campaign.bulkMessage`

PR #256, opened 2013-05-12
Extendable.extend should accept multiple objects as arguments

justinvdm:


go/base/static/js/src/utils.js
```diff
@@ -0,0 +1,19 @@
...
+
+(function(exports) {
+  // Acts as a 'base' for class-like objects which can be extended (with the
+  // prototype chain set up automatically)
+  exports.Extendable = function () {};
+
+  // Backbone has an internal `extend()` function which it assigns to its
+  // structures. We need this function, so we arbitrarily choose
+  // `Backbone.Model`, since it has the `extend()` function we are looking for.
+  var extend = Backbone.Model.extend;
```
smn, line 13: What are the potential down sides of this? What does Backbone.Model internally use to provide this feature?

justinvdm, line 13: I can't think of any serious downside to this approach. Maybe if Backbone decided against using `extend()` in future, we would be forced to grow our own `extend()`, but I don't think that is likely.

Backbone.Model uses the `extend()` method internal to Backbone. Taking a look at [this part of the source](https://github.com/documentcloud/backbone/blob/master/backbone.js#L1537-L1557) might make things clearer.

smn, line 13: My worry is that we're assuming that Model.extend will always be available and will not grow some funky specific features that are built specifically for Model. If we have tests that ensure that a) extend() is available and b) does what we expect it to do then I think we're fine.

peterp, line 13: :+1:

PR #260, opened 2013-05-13
Survey app should fail early if no survey is configured

smn:
Sometimes this can happen if the number is dialed before the setup is completed.


smn:
This is bare minimum maintenance while we're waiting for vxpolls to go away and die.

jerith:
:+1: once Travis confirms.

PR #258, opened 2013-05-14
Campaign Wizard: Bulk Message Composer (Step 2)

peterp:
In step 1 you can select either bulk message or interactive conversation, this is the path that is taken if bulk message is selected.

### DIFF
--- a/go/base/static/js/src/campaign/bulkMessage.js
+++ b/go/base/static/js/src/campaign/bulkMessage.js
@@ -8,23 +8,23 @@
         SMS_TOTAL_CHARS: 160,
 
         events: {
-            'keyup': 'render',
+            'keyup': 'render'
         },
 
         initialize: function() {
             _.bindAll(this, 'render');
         },
         render: function() {
-            $p = this.$el.next();
+            var $p = this.$el.next();
             if (!$p.hasClass('textarea-char-count')) {
                 $p = this.$el.after('<p class="textarea-char-count"/>');
             }
             var totalChars = this.$el.val().length;
-            var totalSMS = Math.ceil(totalChars / this.SMS_TOTAL_CHARS)
+            var totalSMS = Math.ceil(totalChars / this.SMS_TOTAL_CHARS);
             $p.html(totalChars + ' characters used<br>' + totalSMS + ' smses');
 
             return this;
-        },
+        }
     });
     exports.TextareaView = TextareaView;
 
--- a/go/base/static/js/src/campaign/bulkMessage.js
+++ b/go/base/static/js/src/campaign/bulkMessage.js
@@ -5,7 +5,7 @@
 (function(exports) {
 
     var TextareaView = Backbone.View.extend({
-        SMS_TOTAL_CHARS: 160,
+        SMS_MAX_CHARS: 160,
 
         events: {
             'keyup': 'render'
@@ -17,11 +17,12 @@
         render: function() {
             var $p = this.$el.next();
             if (!$p.hasClass('textarea-char-count')) {
-                $p = this.$el.after('<p class="textarea-char-count"/>');
+                $p = $('<p class="textarea-char-count"/>');
+                this.$el.after($p);
             }
-            var totalChars = this.$el.val().length;
-            var totalSMS = Math.ceil(totalChars / this.SMS_TOTAL_CHARS);
-            $p.html(totalChars + ' characters used<br>' + totalSMS + ' smses');
+            this.totalChars = this.$el.text().length;
+            this.totalSMS = Math.ceil(this.totalChars / this.SMS_MAX_CHARS);
+            $p.html(this.totalChars + ' characters used<br>' + this.totalSMS + ' smses');
 
             return this;
         }
--- a/go/base/static/js/src/campaign/bulkMessage.js
+++ b/go/base/static/js/src/campaign/bulkMessage.js
@@ -4,7 +4,25 @@
 
 (function(exports) {
 
-    
+    var TextareaView = Backbone.View.extend({
+        tagName: 'p',
+        SMS_TOTAL_CHARS: 160,
+
+        initialize: function() {
+            _.bindAll(this, 'render');
+            this.$textarea = this.options.$textarea;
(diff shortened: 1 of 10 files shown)


go/campaigns/templates/campaigns/wizard_2_message_bulk.html
```diff
@@ -0,0 +1,38 @@
...
+{{ block.super }}
+<script type="text/javascript">
+    $(function() {
+        // highlight the breadcrumbs.
+        $('.breadcrumbs a').eq(1).addClass('active');
+
+
+        $textarea = $('#frm-campaign textarea')
+        var composer = new go.campaign.bulkMessage.TextareaView({
+            $textarea: $textarea
```
justinvdm: Since the view is a a Textarea view, I think it makes more sense to pass in `'#frm-campaign textarea'` as `el` here, then change to do stuff will `this.$el` instead of `this.$textarea`. Then we can simply replace this block by something like

``` js
var composer = new go.campaign.bulkMessage.TextareaView({el: '#frm-campaign textarea'});
composer.render();
```

go/base/static/js/src/campaign/bulkMessage.js
```diff
@@ -0,0 +1,28 @@
...
+(function(exports) {
+
+    var TextareaView = Backbone.View.extend({
+        tagName: 'p',
+        SMS_TOTAL_CHARS: 160,
+
+        initialize: function() {
+            _.bindAll(this, 'render');
+            this.$textarea = this.options.$textarea;
+            this.$textarea.on('keyup', this.render);
```
justinvdm: If we do the change I mentioned [here](https://github.com/praekelt/vumi-go/pull/258/files#r4215375), we can get rid of line 13 and declare an `events` property for the view that looks something like `{'keyup': 'render'}`.

go/base/static/js/src/campaign/bulkMessage.js
```diff
@@ -0,0 +1,28 @@
...
+        },
+        render: function() {
```
justinvdm, line 17: I think this view's `render()` is simple enough to be tested.

go/base/static/js/src/campaign/bulkMessage.js
```diff
@@ -0,0 +1,31 @@
...
+
+        events: {
+            'keyup': 'render',
+        },
+
+        initialize: function() {
+            _.bindAll(this, 'render');
+        },
+        render: function() {
+            $p = this.$el.next();
```
justinvdm: Global leak

go/base/static/js/src/campaign/bulkMessage.js
```diff
@@ -0,0 +1,31 @@
...
+            if (!$p.hasClass('textarea-char-count')) {
+                $p = this.$el.after('<p class="textarea-char-count"/>');
+            }
+            var totalChars = this.$el.val().length;
+            var totalSMS = Math.ceil(totalChars / this.SMS_TOTAL_CHARS)
```
justinvdm: Missing `;`

go/campaigns/urls.py
```diff
@@ -8,6 +8,6 @@
         name='message_bulk'),
     url(r'^message/(?P<campaign_key>\w+)/conversation/$',
         views.message_conversation, name='message_conversation'),
-    url(r'^recipients/$', views.todo, name='recipients'),
+    url(r'^contacts/(?P<campaign_key>\w+)/', views.todo, name='contacts'),
```
justinvdm, line 11: Sorry for missing this in the review for that PR yesterday

go/campaigns/views.py
```diff
@@ -1,10 +1,11 @@
 from django.http import HttpResponse
 from django.shortcuts import render, redirect
 
-from go.campaigns.forms import CampaignGeneralForm, CampaignConfigurationForm
+from go.campaigns.forms import CampaignGeneralForm, \
+    CampaignConfigurationForm, CampaignBulkMessageForm
```
justinvdm: Sorry for bikeshedding, but I think the preferred approach with vumi stuff is to do one of these three things in situations where need to span multiple lines:

``` python
from go.campaigns.forms import (
    CampaignGeneralForm, CampaignConfigurationForm, CampaignBulkMessageForm)


from go.campaigns.forms import (CampaignGeneralForm,
                                CampaignConfigurationForm,
                                CampaignBulkMessageForm)

from go.campaigns.forms import (
    CampaignGeneralForm,
    CampaignConfigurationForm,
    CampaignBulkMessageForm)
```

go/campaigns/forms.py
```diff
@@ -38,3 +38,8 @@ class CampaignConfigurationForm(forms.Form):
                                          choices=CHANNEL_CHOICES)
 
     keyword = forms.CharField(label="Keyword", max_length=100)
+
+
+class CampaignBulkMessageForm(forms.Form):
+    # pass
```
justinvdm: We can remove this line

justinvdm:
I think we should testing for `TextareaView.render()`. It should be fairly easy. Other than that, and the other really minor comments, looks good.

go/base/static/js/test/tests/campaign.test.js
```diff
@@ -0,0 +1,28 @@
+describe("go.campaign", function() {
+
+  describe(".bulkMessage", function() {
+
+    var $div = $('<div/>');
+    var text = 'Margle. The. World.';
+    var $textarea = $('<textarea>' + text + '</textarea>');
+    $div.append($textarea);
+    view = new go.campaign.bulkMessage.TextareaView({el: $textarea});
```
justinvdm, line 9: I think it might be better to just declare `$div`, `text` and `$textarea` here, then put the assignment and other stuff in this block in `before()` or `beforeEach()`. I don't think it's dangerous to do it this way, but I think its what `before()` is meant to be used for.

justinvdm, line 9: Actually, I'm just being way too fussy, this is fine, sorry.

justinvdm:
:+1: Thanks for the tests!

PR #252, opened 2013-05-15
Conversation definition class

jerith:
We currently have a split of conversation definition and configuration logic across Django views and Vumi application worker code. We need to split this out into its own thing that both sides can reference.


jerith:
Not yet ready to land.

go/base/utils.py
```diff
@@ -109,3 +109,12 @@ def writerows(self, rows):
 def configured_conversation_types():
     return dict((a['namespace'], a['display_name'])
                 for a in settings.VUMI_INSTALLED_APPS.itervalues())
+
+
+def get_conversation_definition(conversation_type):
+    # HACK: Assume 'namespace' is 'conversation_type'.
```
smn, line 122: What's the long term solution for this?

go/conversation/conversation_views.py
```diff
@@ -0,0 +1,583 @@
...
+from go.base import message_store_client as ms_client
+from go.base.utils import (make_read_only_form, page_range_window)
+
+
+class ConversationView(TemplateView):
+    view_name = None
+    path_suffix = None
+    template_base = 'generic'
+
+    # Theis is set in the constructor, but the attribute must exist already.
```
smn, line 33: s/eis/is/

go/conversation/conversation_views.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def get_view_url(self, view_name, **kwargs):
+        kwargs['path_suffix'] = self.view_mapping[view_name].path_suffix
+        return reverse('conversations:conversation', kwargs=kwargs)
+
+    def get_view(self, path_suffix):
+        if path_suffix not in self.path_suffix_mapping:
+            raise Http404
+        return self.path_suffix_mapping[path_suffix].as_view(
+            conversation_views=self)
```
smn: Are we re-inventing stuff that Django's URL resolvers perhaps already give us?

go/base/utils.py
jerith, line 122: We need to consolidate a bunch of things that are unnecessarily different. I'm not sure what the eventual solution will be, but my current plan is a cleanup pass once we can get rid of all the old UI code.

go/conversation/conversation_views.py
jerith: I'm not sure. I didn't really look too deeply into using Django's resolvers outside of urls.py, but it might be worth a shot.

jerith:
So far, I've implemented the core of the conversation definition and conversation action objects (in `go.vumitools.conversation.definition`), generic conversation views that use them (in `go.conversation.conversation_views`) and written conversation definitions for three apps: `bulk_message`, `jsbox` and `surveys`. All of these have varying levels of custom code in them.

The two major differences from the old views are that the initial action hack isn't sent from the definition-based start view and actions are listed in the definition-based show view. You can visit `/conversations/new/<conversation_type>/` to create a new conversation using these views and `/conversations/<conversation_key>/` to view an existing conversation using these views.

go/vumitools/conversation/definition.py
```diff
@@ -0,0 +1,62 @@
...
+    conversation_form = None
+    conversation_group_form = None
+    edit_conversation_forms = None
+    conversation_start_params = None
+    draft_view = None
+
+    actions = ()
+
+    def __init__(self, conv=None):
+        self.conv = conv
```
hodgestar, line 31: I'm still reading but this seems like an odd thing to have. A link from a conversation to its definition makes sense, but having one conversation attached to a definition instance is strange.

go/conversation/views.py
```diff
@@ -72,3 +74,29 @@ def index(request):
...
         'search_form': search_form,
     })
+
+
+@login_required
+def conversation(request, conversation_key, path_suffix):
+    conv = conversation_or_404(request.user_api, conversation_key)
+    conv_def = get_conversation_definition(conv.conversation_type)
+    finder = ConversationViewFinder(conv_def(conv))
+    view = finder.get_view(path_suffix)
```
hodgestar, line 84: Maybe we should wrap these 3 lines into a utility function that lives in `conversation_views` and takes `conv` and `path_suffix` as parameters and returns a view?

go/conversation/templates/generic/action.html
```diff
@@ -0,0 +1,39 @@
...
+    <div class="row">
+        <div class="span12">
+            <div class="pull-right padtop">
+                <input type="submit" name="submit" class="btn btn-success" value="Next">
+            </div>
+            <div class="pull-right padtop" style="margin-right: 10px">
+                <input type="reset" class="btn" value="Cancel">
+            </div>
+            <div id="welcome">
+                <h1>Conversation Action</h1>
```
hodgestar: Should we put the action display name here?

go/conversation/templates/conversation_views/start.html
```diff
@@ -0,0 +1,48 @@
...
+        {% endif %}
+        {% csrf_token %}
+    </div>
+    </form>
+{% endblock %}
+
+
+{% block modals %}
+    {% include "conversation_views/includes/tools.html" %}
+{% endblock %}
```
hodgestar, line 48: Missing newline.

go/vumitools/conversation/definition.py
jerith, line 31: It makes it easier to pass a conversation object into a ConversationAction constructor without having to pass both the definition and the conversation around all the time. I was originally planning to do more with it, but that turned out to not be necessary.

go/conversation/views.py
jerith, line 84: I can take a look, but given the slight differences between all variations we have (which is why they're different views) it might not be a win.

go/conversation/templates/conversation_views/start.html
jerith, line 48: This is a straight copy of our existing views with the minimum necessary modification. I'm not particularly worried about fixing existing style issues that got carried across.

go/conversation/templates/generic/action.html
jerith: I added this template before I gave the new set of views its own copy of the templates and nothing's actually using it. I'll delete it.

jerith: I'll make the change you suggested in the copy of the template we _are_ using, though. Thanks.

go/conversation/conversation_views.py
```diff
@@ -0,0 +1,664 @@
...
+        return self.conversation_views.conversation_form(*args, **kw)
+
+    def get_next_view(self, conversation):
+        if not conversation.is_draft():
+            return 'show'
+        if self.conversation_views.draft_view is not None:
+            return self.conversation_views.draft_view
+        if self.conversation_views.conversation_initiator == 'client':
+            return 'start'
+        return 'people'
```
hodgestar, line 59: Is `get_next_view` still relevant in the new conversation world?

jerith, line 59: Currently, yes. That's something we can fix in the new UI where routing is decoupled from conversation lifecycle.

go/conversation/conversation_views.py
```diff
@@ -0,0 +1,664 @@
...
+    # Provided by conversation definition:
+    tagpool_filter = None
+    conversation_initiator = None
+    conversation_display_name = None
+    # Overridden by conversation definition:
+    conversation_form = ConversationForm
+    conversation_group_form = ConversationGroupForm
+    edit_conversation_forms = None
+    conversation_start_params = None
+    draft_view = None
```
hodgestar, line 602: These attributes seem to duplicate things from the conversation definition? Could we maybe just have a reference to the definition instead?

jerith, line 602: Probably, but I don't think it's worth changing unless we need to change any of these fields.

go/base/utils.py
hodgestar, line 122: I actually don't know that this is a hack -- get method above is called `configured_conversation_types` and returns the `namespace` as the name of conversation type. It might just be that `namespace` is oddly named (at least I'm not seeing `namespace` used except as a conversation type).

jerith, line 122: We need to clean up the places we're using these names and stop having multiple similar things.

hodgestar, line 122: Agreed. Maybe that should be a ticket or a TODO in settings.py where we define this though?

hodgestar:
:+1:

PR #184, opened 2013-05-15
Update setup.py

hodgestar:
Now that the repository has been opened we need to update setup.py to:
- Update the license entries
- Update the trove classifiers
- Make sure it runs


jerith:
:+1:

PR #271, opened 2013-05-15
Zip messages export CSV before emailing.

smn:
CSV balloons to sizes general mail servers aren't willing to accept as attachments.


hodgestar:
:+1: once Travis builds.

PR #268, opened 2013-05-16
Prevent some UI elements from wrapping when the window is resized

peterp:
1. The menu
2. Breadcrumbs and title.
3. Sidebar actions and page actions.

### DIFF
--- a/go/base/static/css/vumigo.less
+++ b/go/base/static/css/vumigo.less
@@ -6,7 +6,9 @@ body {
   color: #c1c1c1;
 }
 
-.container {
+.container-fluid {
+  // prevent wrapping.
+  min-width:650px;
 }
 
 .footer {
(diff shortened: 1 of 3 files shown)


justinvdm:
Other than my comment, I'm :+1: , although my css (and less) reviewing shouldn't be trusted :)

go/base/static/css/vumigo.less
```diff
@@ -6,7 +6,9 @@ body {
   color: #c1c1c1;
 }
 
-.container {
```
justinvdm, line 9: Won't we need to change the things that use the `container` class to use `container-fluid`?

smn:
:+1:

go/base/static/css/vumigo.less
peterp, line 9: The old layout used container; but I changed it to container-fluid. You inadvertently made me consider that I should really change bootstrap classes. Thanks!

PR #273, opened 2013-05-16
Hook up correct conversation values to Dashboard

smn:
Currently all conversations are shown while we should be making a distinction between various states.


smn:
i'll go about filtering on the states next in a separate branch.

smn:
Making the branch filtering part of this branch as there are failing tests on this.

smn:
deployed to staging

PR #279, opened 2013-05-17
JsBox should be able to load URL using Basic Auth.

smn:
I'm not sure if GitHub allows for basic auth access to stuff to begin with but that's a separate issue.

### DIFF
--- a/go/apps/jsbox/tests/test_views.py
+++ b/go/apps/jsbox/tests/test_views.py
@@ -85,7 +85,21 @@ class JsBoxTestCase(DjangoGoApplicationTestCase):
             {'url': 'http://username:password@domain.com'})
         [call] = mocked_get.call_args_list
         args, kwargs = call
-        self.assertEqual(args, ('http://username:password@domain.com',))
+        self.assertEqual(args, ('http://domain.com',))
+        self.assertEqual(kwargs, {'auth': ('username', 'password')})
+        self.assertTrue(mocked_get.called)
+        self.assertEqual(response.content, 'foo')
+        self.assertEqual(response.status_code, 200)
+
+    @patch('requests.get')
+    def test_basic_auth_cross_domain_xhr_with_https_and_port(self, mocked_get):
+        mocked_get.return_value = Mock(text='foo', status_code=200)
+        response = self.client.post(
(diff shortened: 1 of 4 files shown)


go/apps/jsbox/tests/test_views.py
```diff
@@ -60,3 +62,31 @@ def test_edit_conversation(self):
...
+
+    @patch('requests.get')
+    def test_basic_auth_cross_domain_xhr(self, mocked_get):
+        mocked_get.return_value = Mock(text='foo', status_code=200)
+        response = self.client.post(
+            reverse('jsbox:cross_domain_xhr'),
+            {'url': 'http://username:password@domain.com'})
+        [call] = mocked_get.call_args_list
+        args, kwargs = call
+        self.assertEqual(args, ('http://username:password@domain.com',))
```
hodgestar: Seems a bit odd that the username and password are still in the URL? Don't think it'll directly harm anyone but maybe worth a comment and a little extra thought?

hodgestar:
Are people okay with their passwords being stored as plain-text in config? Probably not much we can do to avoid it for this kind of delegated login.

smn:
I don't see any way around this. HTTP Basic Auth requires us to store passwords in plain text, we can add a bit of text that tells people this and make a strong suggestion to use a system that uses revokable-tokens for passwords but that's about the best we can do.

hodgestar:
Any reason to not land this and just require that people use GitHub's Oauth
2 support?

smn:
People should be free to host their JS anywhere they like, not just GitHub.

go/apps/jsbox/tests/test_views.py
hodgestar: Just poking this comment again.

hodgestar:
I think I'm +1 but I'd like GitHub users to use the Oauth2 `access_token` parameter from http://developer.github.com/v3/#authentication.

go/apps/jsbox/tests/test_views.py
smn: Thanks! Fixed.

smn:
Ready for re-review @hodgestar

hodgestar:
:+1:

PR #276, opened 2013-05-20
Make jsPlumb implementation more convincing

peterp:


go/campaigns/templates/campaigns/wizard_2_conversation_tool.html
```diff
@@ -72,9 +72,9 @@
 
 
 <div class="row-fluid">
-    <div class="span12 tool" style="height:900px;">
-        <div class="wrapper" style="height:900px;" class="clearfix">
-            <section class="viewport" style="height:900px;">  
+    <div class="span12 tool" style="height:1800px;">
```
justinvdm: Unless theres a good reason why its here, could we maybe have this styling specified in css?

peterp: It's inline because I believe that the majority of this will be thrown out. Probably not a good enough reason to be lazy.

justinvdm:
:+1:

PR #282, opened 2013-05-20
ContactsResourse save() doesn't change the 'groups' field properly

justinvdm:


justinvdm:
Turns out it does save `groups` fields properly :) Added a comment and changed a test to make sure.

smn:
:+1:

PR #158, opened 2013-05-20
Capture JS sandbox logs and display them in the UI for the conversation.

hodgestar:


go/apps/jsbox/tests/test_log.py
```diff
@@ -0,0 +1,72 @@
...
+from twisted.trial.unittest import TestCase
+from twisted.internet.defer import inlineCallbacks, returnValue
+
+from vumi.application.tests.test_sandbox import (
+    ResourceTestCaseBase, DummyAppWorker)
+
+from go.apps.jsbox.log import LogManager, GoLoggingResource
+
+
+class TestLogManager(TestCase):
```
smn: Where is this being used?

hodgestar:
For now I'm limiting this to the non-UI part. Will create a separate ticket and PR for that.

smn:
:+1: in general though I'm a bit concerned about the `RedisManager` prefix being hardcoded in the view but the config must be made matching for the `LoggingResource` & the redis manager that's being handed to the `LogManager`.

hodgestar:
Updated -- could I get a re-review?

smn:
:+1:

PR #287, opened 2013-05-21
Add backbone-relational.js to vendor libs

justinvdm:
One example use case for backbone-relational is with the plumbing component. Collections of `EndpointModel`s are quite low level structures. It makes more sense for collections of `EndpointModel`s to be nested under their associated `StateModel`s. We then can simply sync the collections of `StateModel`s with the server, without having to also sync the endpoints associated to each separately.


justinvdm:
This is actually linked to ticket #285. I by mistake set the pull request's base to `develop` instead of `develop-frontend`, so I needed to close that PR and create a new one. Sorry about that :/

hodgestar:
:+1: on the changes themselves. I wasn't involved in the discussion about including backbone-relational and I'm not up to speed on what's driving its inclusion so someone else should +1 the concept of including it.

hodgestar:
Now that I've caught up a bit, :+1:.

PR #270, opened 2013-05-16
Conversation lifecycle improvements

jerith:
We currently have routing and certain conversation operations (such as bulk send) tied to conversation lifecycle transitions. We need to decouple these in the backend to make the new routing and UI features possible.

PR #288, opened 2013-05-21
Fix send_message command and add some tests.

smn:
It looks like the send_message command is a bit broken since #270 landed, but we didn't notice because test coverage is poor.

### DIFF
--- a/go/apps/sna/handlers.py
+++ b/go/apps/sna/handlers.py
@@ -106,7 +106,6 @@ class USSDMenuCompletionHandler(SNAEventHandler):
         yield conversation.dispatch_command(
             'send_message', account_key, conversation.key,
             command_data={
-                'conversation_key': conversation.key,
                 'batch_id': batch_id,
                 'to_addr': from_addr,
                 'content': content,
--- a/go/apps/subscription/vumi_app.py
+++ b/go/apps/subscription/vumi_app.py
@@ -64,14 +64,18 @@ class SubscriptionApplication(GoApplicationWorker):
     @inlineCallbacks
     def process_command_send_message(self, account_key, conversation_key,
                                      **kwargs):
+        # FIXME: Fix whatever needs updating here?
         # TODO: Update
+        conv = yield self.get_conversation(user_account_key, conversation_key)
         command_data = kwargs['command_data']
         log.info('Processing send_message: %s' % kwargs)
+        msg_options = command_data['msg_options']
+        self.add_conv_to_msg_options(conv, msg_options)
         yield self.send_message(
                 command_data['batch_id'],
                 command_data['to_addr'],
                 command_data['content'],
-                command_data['msg_options'])
+                msg_options)
 
     @inlineCallbacks
     def collect_metrics(self, user_api, conversation_key):
--- a/go/apps/subscription/vumi_app.py
+++ b/go/apps/subscription/vumi_app.py
@@ -62,7 +62,8 @@ class SubscriptionApplication(GoApplicationWorker):
         return self.vumi_api.mdb.add_event(event)
 
     @inlineCallbacks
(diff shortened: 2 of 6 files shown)


smn:
Looks like the signature for the function changed? Although the way that command dispatching works with `*args, **kwargs` is way too liberal to begin with.

jerith:
There are a couple of other apps that also need this change. Probably best to write a "send `send_message` command" helper function in the base test class and use that for all the app worker tests.

smn:
Potential `send_message` calls:

```
go/apps/bulk_message/vumi_app.py:    def send_message_via_window(self, conv, window_id, batch_id, to_addr,
go/apps/bulk_message/vumi_app.py:            yield self.send_message_via_window(
go/apps/bulk_message/vumi_app.py:    def process_command_send_message(self, user_account_key, conversation_key,
go/apps/bulk_message/vumi_app.py:        log.info('Processing send_message: %s' % kwargs)
go/apps/http_api/vumi_app.py:    def process_command_send_message(self, *args, **kwargs):
go/apps/http_api/vumi_app.py:        log.info('Processing send_message: %s' % kwargs)
go/apps/sequential_send/vumi_app.py:                yield self.send_message(
go/apps/sequential_send/vumi_app.py:    def send_message(self, batch_id, to_addr, content, msg_options):
go/apps/sna/handlers.py:            'send_message', account_key, conversation.key,
go/apps/subscription/vumi_app.py:    def send_message(self, batch_id, to_addr, content, msg_options):
go/apps/subscription/vumi_app.py:    def process_command_send_message(self, *args, **kwargs):
go/apps/subscription/vumi_app.py:        log.info('Processing send_message: %s' % kwargs)
go/apps/subscription/vumi_app.py:        yield self.send_message(
go/apps/surveys/vumi_app.py:    def process_command_send_message(self, *args, **kwargs):
go/apps/surveys/vumi_app.py:        log.info('Processing send_message: %s' % kwargs)
go/conversation/base.py:                    'send_messages': False,
go/conversation/base.py:                    'send_messages': True,
go/conversation/tasks.py:        'send_message', account_key, conversation_key,
go/conversation/templates/generic/start.html:        {% elif send_messages %}
go/vumitools/conversation/utils.py:        I was tempted to make this a generic 'send_message' function but
go/vumitools/conversation/utils.py:            'send_message',
go/vumitools/handler.py:                "send_message",
```

go/apps/sna/handlers.py
```diff
@@ -103,10 +103,13 @@ def handle_event(self, event, handler_config):
...
-            'conversation_key': conversation.key,
-            'batch_id': batch_id,
-            'to_addr': from_addr,
-            'content': content,
-            'msg_options': msg_options,
-        })
+        yield conversation.dispatch_command(
+            'send_message', account_key, conversation.key,
+            command_data={
+                'conversation_key': conversation.key,
```
hodgestar: Should `conversation.key` be here too if it's already passed in as an positional parameter?

go/apps/subscription/vumi_app.py
```diff
@@ -62,15 +62,20 @@ def consume_delivery_report(self, event):
         return self.vumi_api.mdb.add_event(event)
 
     @inlineCallbacks
-    def process_command_send_message(self, *args, **kwargs):
+    def process_command_send_message(self, account_key, conversation_key,
+                                     **kwargs):
+        # FIXME: Fix whatever needs updating here?
```
hodgestar, line 67: Is it possible to write a better #FIXME / #TODO here? I'm not sure what's supposed to be fixed or updated after this PR?

smn, line 67: Me neither. I think the # TODO is from @jerith.

hodgestar, line 67: Let's leave it here for now then?

PR #289, opened 2013-05-24
Client side needs view collections

justinvdm:
We need some extendable base structure that takes in a collection of models and provides:
- Looking up views by their model ids
- Accessing all the views at once
- Removing 'dead' views on render calls (views that no longer have corresponding models in the collection)
- Adding 'new' views (models in the collection that currently do not have views representing them)
- Rendering all the views in one go

Two use cases I've found for this so far:
- Adding/removing/keeping track of endpoint views on a `StateView` in the plumbing component.
- Adding/removing/keeping track of state views in a `PlumbView` in the plumbing component

### DIFF
--- a/go/base/static/js/src/components/structures.js
+++ b/go/base/static/js/src/components/structures.js
@@ -3,7 +3,8 @@
 // Reusable, generic structures for Go
 
 (function(exports) {
-  var merge = go.utils.merge;
+  var merge = go.utils.merge,
+      GoError = go.errors.GoError;
 
   // Acts as a 'base' for class-like objects which can be extended (with the
   // prototype chain set up automatically)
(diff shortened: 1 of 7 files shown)


peterp:
Other than my comment on `undefined` vs. `null` :+1:

go/base/static/js/src/components/structures.js
```diff
@@ -0,0 +1,112 @@
...
+      this.models.on('remove', function(m) { self.remove(m.id); });
+    },
+
+    _modelId: function(m) { return m.id; },
+    _modelIds: function(m) { return this.models.map(this._modelId); },
+
+    // Override to specialise how the view is created
+    create: function(model) { return new Backbone.View({model: model}); },
+
+    add: function(id) {
```
justinvdm: @peterp I can't decide whether it makes more sense for `add()` to accept the model or the id of the model. If we accept a model, should we still add a view for the model, even though it might not be in the collection that the ViewCollection is watching? Any thoughts?

justinvdm: Actually, maybe `add()` and `remove()` should be made internal?

peterp: Hmpf. This is a tough one.

You're trying to associate a model with a view and vice-versa?

What if you added the view as a property on the model?

Also, once you've got a collection of views what would you do with the views?

peterp:
:+1:

PR #290, opened 2013-05-25
Client side needs a lookup group structure

justinvdm:
We need to add a `LookupGroup` structure that keeps track of the `Lookup`s subscribed to it. When items are added to any of the subscribed lookups, the items will automatically be added to the lookup group. In other words, `LookupGroup` can be thought of as a self-maintaining, 'flattened' collection of the all items of its members.

This has a very important use case for the plumbing component in particular. A state view can keep endpoints of different types, and has a `ViewCollection` for each endpoint type. Using a `LookupGroup`, each state view has a self-maintained collection of all the endpoints it contains.

Then, at the top level (the state diagram view), we can use a `LookupGroup` to keep track of all the endpoints in the diagram (all the state views' endpoints), without ever needing to maintain this collection. This is especially necessary to route connection events to the relevant endpoints, since jsPlumb does not trigger events on its endpoint objects when connections occur (something we really need).

### DIFF
--- a/go/base/static/js/src/components/structures.js
+++ b/go/base/static/js/src/components/structures.js
@@ -115,6 +115,7 @@
       lookup.on('remove', this._remove, this);
 
       this.members._add(key, lookup);
+      return this;
     },
 
     unsubscribe: function(key) {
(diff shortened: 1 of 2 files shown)


peterp:
:+1:

PR #291, opened 2013-05-25
Client side needs state machine models

justinvdm:
The state machine models can serve as base models that are extended and specialised for the different parts of the ui where state-like objects exist (for eg, states in the dialogue screen, the model for the routing block, or the campaign details set up screen). 

These models don't necessarily have to serve as the models to be synced with the server side (although they can be), but can be used as intermediate structures between non-state-machine-like objects synced with the server, and the state-machine-like views. This approach ensures we don't store data/attributes in the views (a violation of the MV pattern) in order to cope with the different nature of the objects we sync with the server and the views themselves.

### DIFF
--- a/go/base/static/js/src/components/states.js
+++ b/go/base/static/js/src/components/states.js
@@ -1,0 +1,67 @@
+// go.components.states
+// ====================
+// Models for the state-machine-like parts of the ui.
+//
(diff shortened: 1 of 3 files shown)


peterp:
:+1:

PR #297, opened 2013-05-28
Define go.components.structures exports at the bottom of module

justinvdm:
The `go.components.structures` module contains quite a few internal references, since each structure extends some other structure. It will simplify things if we defined the exports at the bottom of the module.

### DIFF
--- a/go/base/static/js/src/campaign/bulkMessage.js
+++ b/go/base/static/js/src/campaign/bulkMessage.js
@@ -25,6 +25,8 @@
             return this;
         }
     });
-    exports.TextareaView = TextareaView;
 
+    _.extend(exports, {
+      TextareaView: TextareaView
+    });
 })(go.campaign.bulkMessage = {});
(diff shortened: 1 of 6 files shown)


peterp:
:+1: with :100:

go/base/static/js/src/campaign/bulkMessage.js
```diff
@@ -25,6 +25,8 @@
             return this;
         }
     });
-    exports.TextareaView = TextareaView;
 
+    _.extend(exports, {
+      TextareaView: TextareaView
+    });
```
justinvdm, line 31: @peterp I changed all the modules to use the `_.extend(exports, {...});` pattern, even if they only had one export. I know it seems a bit wierd if there is only one export, but I think it makes it easier to add new exports. Otherwise, for example, if I got lazy, I'd start defined new exports as `exports.someThing = someThing`. Eventually, we'd end up with non-uniformity (some modules using `exports.someThing = someThing`, some using `_.extend(exports, {...});`). Are you ok with that? I don't mind changing it back for single exports if you are opposed to doing it this way.

peterp, line 31: Nicely done.

peterp:
:+1:

PR #299, opened 2013-05-29
Remove go.components.structures.ProtectedLookup

justinvdm:
Having things that extend `ProtectedLookup` override its internal methods (`._add()` and `._remove()`) feels a bit wierd. It would probably be better to just use `Lookup`'s `.add()` and `.remove()`, and just expect the things that currently extend `ProtectedLookup` to be used correctly.


peterp:
:+1: also <3

PR #283, opened 2013-05-29
Fix authentication views.

peterp:
Login, logout and password reset forms.

### DIFF
--- a/go/templates/registration/base.html
+++ b/go/templates/registration/base.html
@@ -3,7 +3,7 @@
 {% block content %}
 <div class="row-fluid">
     <div class="span6 offset3 text-center">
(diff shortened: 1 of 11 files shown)


go/templates/registration/base.html
```diff
@@ -3,7 +3,7 @@
 {% block content %}
 <div class="row-fluid">
     <div class="span6 offset3 text-center">
-        <img src="{{ STATIC_URL }}img/logo-lrg.png" alt="Vumi" class="logo-large">
+        <img src="{{ STATIC_URL }}img/logo-lrg.png" alt="VUMI" class="logo-large">
```
justinvdm: For most of the templates in registration, the `alt`s are `Vumi Go`. We should probably make them all the same, unless I am missing something.

justinvdm:
Other than minor comment, :+1:

PR #266, opened 2013-05-29
Data Tables Refactor

peterp:
The tables that are being used everywhere in the app share a common pattern:
1. The have button that are only enabled once a row is checked.
2. The entire row should be clickable to see a detailed item.

The data tables should reflect an interface that's similar to bootstrap as this is the UI component that we're using on the frontend.


go/base/static/js/test/tests/components/tables.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+      go.components.tables.init({
+        tableSelector: $table
+      });
+    });
+
+    it('should propogate events on the td element which holds the checkbox', function() {
+      var td = 'tbody tr:first-child td:first-child';
+      $table.find(td).trigger('click');
+      assert.equal($table.find(td).find('input:checkbox').prop('checked'), true);
```
justinvdm: Sorry for bikeshedding, but can we maybe just use `assert` or `assert.isTrue` for the truth assertions?

go/base/static/js/src/components/tables.js
```diff
@@ -0,0 +1,80 @@
...
+            if (typeof(opts.checkedToggle) !== 'undefined') {
+                $(opts.checkedToggle).prop('disabled', !oneChecked);
+            }
+        });
+
+        // the entire row of the table should act as a link, except for the
+        // header row.
+        $table.find('tbody tr').click(function() {
+            var url = $(this).attr(opts.linkAttribute);
+            if (url.length) window.location = url;
```
justinvdm: Can't we just check `url` instead of `url.length`?

justinvdm:
Can we maybe add a test to see that `checkedToggle` is working properly? Other than that, and the small comments, looks good!

justinvdm:
:+1:

PR #301, opened 2013-05-29
Export of multi-survey messages not working.

smn:
Looks like it was never implemented?


hodgestar:
We need a test for the `_export_conversation_messages` code path. I'll add it.

hodgestar:
Odd. There is a test but how was it passing before?

hodgestar:
Oh, I see -- it was testing the `_export_conversation_messages` from `survey` and not `multi_survey`.

hodgestar:
:+1:

PR #293, opened 2013-05-31
Basic implementation for plumbing components

justinvdm:


hodgestar:
I've reviewed the non-test parts and those are looking good. Still need to review the tests. What's the easiest way to run those? Also, quick question about the state of the PR -- is it up to date with the recent model and view changes in `frontend-develop`?

justinvdm:
@hodgestar Thanks for the review! To answer your question, I still need to merge in the recent `frontend-develop` changes (the recently merged in #305 and #306), and change the plumbing components to work with them. Working on that now. Maybe it would be a good idea to take a look at the tests after I make those changes? Let me know what you think.

hodgestar:
@justinvdm Cool. I'll wait for changes to be merged in.

justinvdm:
@hodgestar Just realised I didn't answer your first question. To run the tests, install [phantomjs](http://phantomjs.org/) (its independant to node and npm), then: 

```
npm install -g mocha mocha-phantomjs
```

and lastly:

```
./run-tests-client.sh
```

go/base/static/js/src/components/plumbing/connections.js
```diff
@@ -0,0 +1,101 @@
...
+
+      // Keep a reference the actual jsPlumb connection
+      this.plumbConnection = options.plumbConnection || null;
+    },
+
+    _plumbOptions: function() {
+      return _.defaults({
+        source: this.source.plumbEndpoint,
+        target: this.target.plumbEndpoint
+      }, _(this).result('plumbOptions'));
```
hodgestar, line 47: Is there a reason for not just using `this.plumbOptions` in places like this?

justinvdm, line 47: It allows us to make `this.plumbOptions` a function or an object, which is helpful if the options aren't static. For example, we could draw the jsPlumb endpoint differently based on some attribute on the model by making `plumbOptions` a function.

justinvdm, line 47: Obviously, it wouldn't be of much help using `_.result()` if `plumbOptions` wasn't overrideable when extending `ConnectionView`.

go/base/static/js/src/components/plumbing/diagrams.js
```diff
@@ -0,0 +1,202 @@
...
+    // belongs to.
+    determineCollection: function(source, target) {
+      var collections = this.members.values(),
+          i = collections.length,
+          c;
+
+      while (i--) {
+        c = collections[i];
+        if (c.accepts(source, target)) { return c; }
+      }
```
hodgestar, line 82: Is there a reason not to do `for (collection in this.members.values()) { ... }`? Also wonder if we need to be a bit more careful about what happens if no collection accepts `(source, target)`. Maybe not for now, might be confusing later though?

justinvdm, line 82: `for (collection in this.members.values()) { ... }` would iterate through the index numbers of the array.

Very good point about the lack of handling about no collection accepting `source` and `target`. I've been thinking of what to do in these cases, but not too sure yet. Will definitely make sure to do something about this at some stage.

go/base/static/js/src/components/plumbing/connections.js
hodgestar, line 47: Cool.

go/base/static/js/src/components/plumbing/diagrams.js
hodgestar, line 82: Oh, yes, I keep forgetting Javascript is nuts. :)

I think we can leave the handling of no collections accepting `source` and `target` and land this PR -- maybe just file a ticket so we don't forget.

go/base/static/js/src/components/plumbing/diagrams.js
```diff
@@ -0,0 +1,207 @@
...
+      // and view since they no longer exists.
+      if (!this.has(connectionId)) { return; }
+
+      // Case 2:
+      // -------
+      // The connection was removed in the UI, so the model and view still
+      // exist. We need to remove them.
+      var source = this.diagram.endpoints.get(sourceId),
+          target = this.diagram.endpoints.get(targetId),
+          collection = this.determineCollection(source, target);
```
hodgestar: I think it would be better to determine the collection from `this.get(connectId)` if we can (since presumably the `connectionId` exists in a collection somewhere)?

hodgestar:
Other than avoiding the potential `null`-collection case in `onPlumbDisconnect` and adding tests for the `null` cases in the two `onPlumb` methods, I think we're good to go. :)

go/base/static/js/src/components/plumbing/diagrams.js
justinvdm, line 82: Good idea, will do :)

go/base/static/js/src/components/plumbing/diagrams.js
justinvdm: Very good point. We shouldn't need to use `determineCollection` here. The connections keep track of their collection, so I'll do that for now. I think in future I'll give `LookupGroup` a way of returning the collection that an existing view belongs to.

hodgestar:
:+1:

PR #305, opened 2013-05-31
Change state machine models to have connections defined explicitly

justinvdm:


hodgestar:
Changes look good. I'm a bit surprised that they didn't require changes in the JS tests you were showing me earlier though?

justinvdm:
@hodgestar Ah, those are part of the plumbing PR. The state machine models were landed in a different PR earlier in the week (I should have probably shown you that PR, sorry).

hodgestar:
In that case :+1: on this -- updated tests can land with the plumbing PR.

PR #306, opened 2013-05-31
Refactor subview structures out of plumbing components

justinvdm:
`SubviewCollection` and `SubviewCollectionGroup` can be created as generic reuseable structures, out of refactored pieces of the plumbing components. As a side effect, this will take away some of the current bloat in the plumbing components.


hodgestar:
:+1:

PR #302, opened 2013-06-01
User Accounts

peterp:
Modals required:
1. Personal details
2. Change password
3. Permissions

### DIFF
--- a/go/account/views.py
+++ b/go/account/views.py
@@ -14,10 +14,6 @@ from go.base.django_token_manager import DjangoTokenManager
 from django.contrib.auth.forms import PasswordChangeForm
 
 
-def index(request):
-    return HttpResponse("I stole this page")
-
-
 @login_required
 def details(request):
     profile = request.user.get_profile()
--- a/go/account/views.py
+++ b/go/account/views.py
@@ -12,6 +12,8 @@ from go.account.forms import EmailForm, AccountForm
 from go.account.tasks import update_account_details
 from go.base.django_token_manager import DjangoTokenManager
 
+from django.contrib.auth.forms import PasswordChangeForm
+
 
 def index(request):
     return HttpResponse("I stole this page")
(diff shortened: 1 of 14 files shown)


go/account/views.py
```diff
@@ -1,19 +1,25 @@
...
 
 from go.account.forms import EmailForm, AccountForm
 from go.account.tasks import update_account_details
 from go.base.django_token_manager import DjangoTokenManager
 
+from django.contrib.auth.forms import PasswordChangeForm
+
 
-@login_required
 def index(request):
```
justinvdm: Do we still use `index` anywhere?

go/urls.py
```diff
@@ -19,11 +16,7 @@ def health(request):
 
     # django-regisration auth, override some views so we can specify
     # our own custom forms to use.
```
justinvdm: Is this comment still relevant?

go/account/views.py
peterp: I don't think so, I'll remove it.

go/urls.py
peterp: Nope. Removed.

go/templates/registration/login.html
```diff
@@ -2,7 +2,7 @@
 {% load url from future %}
 
 {% block auth_form %}
-<form action="{% url 'auth_login' %}" method="post">
+<form action="" method="post">
```
justinvdm, line 5: This is probably me not understanding django or `django.contrib.auth` properly, but why are we removing `{% url 'auth_login' %}`?

peterp, line 5: `action=""` submits the form to the same url, and in this instance the post and get actions of this form are at the same endpoint. So, they're essentially the same thing.

justinvdm, line 5: Ah, ok cool.

justinvdm:
:+1:

PR #311, opened 2013-06-03
Users and permissions

peterp:


go/base/models.py
```diff
@@ -27,13 +27,23 @@ def create_user_profile(sender, instance, created, **kwargs):
...
 class UserProfile(models.Model):
     """A profile for a user"""
     user = models.OneToOneField('auth.User')
     user_account = models.CharField(max_length=100)
+    organisation = models.ForeignKey(UserOrganisation, blank=True, null=True)
+    is_admin = models.BooleanField(default=False)
 
     def __unicode__(self):
-        return u' '.join([self.user.first_name, self.user.last_name])
+        return u"%s's profile" % self.user
```
justinvdm: Can we be sure `django.contrib.auth.models.User` can be stringified? It probably can be, I don't know much about Django and Django auth, but just want to be sure.

go/account/views.py
```diff
@@ -92,3 +91,20 @@ def details(request):
...
+def user_list(request):
+    """Fetch a list of users that belong to the same company in the
+    users profile."""
+
+    # TODO: Each user has to belong to an organisation, make sure the model
+    # reflects that, otherwise we'll introduce a security problem.
+    # grab the list of profiles that also belong to this user's organisation
+    user_list = []
+    user_profile = request.user.get_profile()
+    for profile in UserProfile.objects.filter(
```
justinvdm: Would it be worth just skipping this loop and returning an empty `user_list` if the user doesn't belong to an organisation (to solve the security concern)?

go/base/models.py
peterp: Yup. It can be, but now that you've mention it I've realised that we don't actually use usernames... We only use email addresses.

go/account/views.py
peterp: Nice! :+1:

PR #312, opened 2013-06-03
Basic models for routing screen

justinvdm:


peterp:
:+1:

go/base/static/js/src/campaign/routing/routing-models.js
```diff
@@ -0,0 +1,70 @@
+// go.components.routing (models)
+// ==============================
+// Models for campaign routing screen.
+
+// We use the state machine models as a base to make use of any
+// custom/overriden functionality the models provide over Backbone.Model,
+// seeing as the models for campaign routing are of a state machine nature.
+// NOTE: This is subject to change.
```
hodgestar: I think we can just remove the NOTE and leave the rest of the comment. Inheriting from the state machine models seems to clearly make sense to me. The routing models are a kind of state machine model.

go/base/static/js/src/campaign/routing/routing-models.js
```diff
@@ -0,0 +1,70 @@
...
+var stateMachine = go.components.stateMachine,
+    EndpointModel = stateMachine.EndpointModel,
+    ConnectionModel = stateMachine.ConnectionModel,
+    StateModel = stateMachine.StateModel,
+    StateMachineModel = stateMachine.StateMachineModel;
+
+(function(exports) {
+  var ChannelModel = StateModel.extend({
+    relations: [{
+      type: Backbone.HasOne,
```
hodgestar: I tempt to suggest we leave this as `HasMany` for consistency. No channels currently have multiple endpoints but I also can't really see a reason to rule that out entirely and the routing table does support having different endpoints.

hodgestar:
Other than the suggestion to replace `HasOne` with `HasMany` for the channel model, :+1:

go/base/static/js/src/campaign/routing/routing-models.js
justinvdm: Ah! Good point.

PR #317, opened 2013-06-04
Stop using celery hack for command messages

jerith:


vgeddes:
I see two almost identical copies of the function `_send_one_off_reply()` in `conversation/base.py` and `apps/surveys/views.py` respectively. I think the function is large and complex enough that the duplication should be eliminated. Perhaps the function can be put into a utility module?

jerith:
`surveys` and `multi_surveys` have their own copies of all the view logic for historical reasons, so the new function doesn't add all that much extra. These views are going to be replaced by the new UI as soon as that's ready, so it isn't really worth the effort of factoring out.

vgeddes:
OK, I've read through the PR, and have got a fairly good understanding of the changes. :+1:

go/base/amqp.py
```diff
@@ -28,7 +37,7 @@ def publish(self, message, exchange, routing_key):
             routing_key=routing_key)
 
     def publish_command_message(self, command):
-        return self.publish(command.to_json, exchange=self.default_exchange,
+        return self.publish(command.to_json(), exchange=self.default_exchange,
```
hodgestar, line 40: How did this work before?

go/conversation/base.py
```diff
@@ -308,6 +307,25 @@ def get(self, request, conversation):
                 conversation_key=conversation.key)
         return self.render_to_response(params)
 
+    def send_one_off_reply(self, user_api, conversation, in_reply_to, content):
+        inbound_message = user_api.api.mdb.get_inbound_message(in_reply_to)
+        if inbound_message is None:
+            print 'Replying to an unknown message'
```
hodgestar: Should we log an error here?

go/conversation/base.py
```diff
@@ -308,6 +307,25 @@ def get(self, request, conversation):
                 conversation_key=conversation.key)
         return self.render_to_response(params)
 
+    def send_one_off_reply(self, user_api, conversation, in_reply_to, content):
```
hodgestar: Since this doesn't actually use `self` can we make it a classmethod and avoid the duplicate copies?

go/conversation/base.py
hodgestar: The `print` was there before but no point in leaving bad code around (and I suspect the print was there because it was a celery task).

go/settings.py
```diff
@@ -222,7 +222,7 @@ def abspath(*args):
 
 # If we're running in DEBUG mode then skip RabbitMQ and execute tasks
 # immediate instead of deferring them to the queue / workers.
-CELERY_ALWAYS_EAGER = DEBUG
+CELERY_ALWAYS_EAGER = True
```
hodgestar: Accidental change?

go/vumitools/tests/utils.py
```diff
@@ -47,117 +43,25 @@ def model_eq(m1, m2):
...
-        while True:
-            msg = consumer.fetch()
-            if msg is not None:
-                msgs.append(msg.payload)
-            else:
-                break
-        return [VumiApiCommand(**payload) for payload in msgs]
+    def get_commands(self):
+        commands, self.commands = self.commands, []
+        return commands
```
hodgestar, line 68: This never pushes commands through `json` which I think is a useful feature to keep. It also never checks that commands are routed to the right queues?

go/base/amqp.py
jerith, line 40: I didn't. We weren't using it.

go/settings.py
jerith: Yes, I'll revert it. This was part of trying to figure out why celery wasn't working for me.

go/vumitools/tests/utils.py
jerith, line 68: The purpose of this is to check that the correct commands are sent with the minimum overhead required. In particular, we don't want to have to run a command dispatcher in the unit tests. We test the full path elsewhere.

hodgestar, line 68: Given that this is part of the test utils packages, I'm struggling a bit to see how the "test the full path elsewhere" bit can be guaranteed to be true in general.

go/conversation/base.py
jerith: I'd prefer staticmethod, but other than that I don't have strong opinions either way.

go/vumitools/tests/utils.py
hodgestar, line 68: I'm just concerned we're changing something and its tests but are now checking correctness less carefully.

go/conversation/base.py
hodgestar: Fine with staticmethod too.

go/vumitools/tests/utils.py
jerith, line 68: This still sends the command message, so we can assert that it reaches the destination correctly and we'll get exceptions if `.to_json()` fails. It's just a more direct way to test that our view logic sends appropriate commands without jumping through a bunch of additional hoops that aren't part of the thing we're testing.

hodgestar, line 68: I think what got me started on this was that it seemed a bit odd to have `self._amqp` and publish to it but not use it to get back messages too.

go/conversation/base.py
jerith: I'll turn this into a log message.

go/vumitools/tests/utils.py
jerith, line 68: Messages we publish go into the broker and are delivered to anything listening, which we use in other tests. `get_commands()` is just a convenience method for tests that only care about checking for a command being sent without requiring all the delivery and dispatch infrastructure.

jerith:
:+1: on the last two commits.

PR #315, opened 2013-06-05
Edit user modal

peterp:


peterp:
This turned out to not work very well as a modal, so I created it's own page. I think this requires some additional work, and perhaps a clearer way of implementing everything that makes it future proof.

go/account/views.py
```diff
@@ -102,10 +105,64 @@ def user_list(request):
     user_profile = request.user.get_profile()
     if user_profile.organisation:
         for profile in UserProfile.objects.filter(
-            organisation=user_profile.organisation):
+                organisation=user_profile.organisation):
+            print '-----------'
+            print profile
```
justinvdm: Are these print statements supposed to be here?

peterp: Oh! No.

/me hides in shame.

go/account/views.py
```diff
@@ -102,10 +105,64 @@ def user_list(request):
...
+        user_profile_form = UserProfileForm(request.POST,
+            instance=edit_user_profile)
+
+        if user_form.is_valid() and user_profile_form.is_valid():
+            # TODO: This works fine for editing users, but I think
+            # creating users is slightly more complicated because
+            # it needs to work within riak.
+
+            # TODO: Username isn't required in vumigo, what should we 
+            # use instead?
```
justinvdm, line 157: Currently, we seem to use email addresses as usernames.

go/account/views.py
```diff
@@ -102,10 +105,62 @@ def user_list(request):
...
+            organisation=user_profile.organisation)
+
+    user_form = UserAccountForm(instance=edit_user)
+    user_profile_form = UserProfileForm(instance=edit_user_profile,
+        initial={
+            'organisation': user_profile.organisation
+        })
+
+    if request.method == 'POST':
+        user = user_form  = UserAccountForm(request.POST, instance=edit_user)
```
justinvdm: Do we use `user` here at all? Also, I could be wrong, but I think the PEP8 doesn't like multiple assignments on one line.

justinvdm:
@peterp Other than the minor comments, this looks good. I'd like this to land, but I really think whoever needs to flesh out the logic behind the views needs to take a look, otherwise they won't know where to start. I'm also missing alot of context, since I don't really know how the current Go handles user management under the hood, so I may be missing some really important things. I know we're rushed though, so I understand if this needs to land. @hodgestar would you be able to take a look?

go/account/views.py
peterp, line 157: I think that we'll need to create a definitive registration form that handles all this stuff.

peterp, line 157: For instance in Django you create a user in 1 form/view, and then set the password in another, would the person creating the user set the password, or would we send an email to the person informing them that they now have an account with a link to reset their password?

peterp, line 157: Currently during registration it sets the `username` as the `email`, but I think that the `username` field only is only 35 characters in length in Django. Which could cause problems as it's also unique.

hodgestar, line 157: The 35 character limit has already bitten us so +1 on fixing that if we can (although it's possible someone already updated the schema).

go/contacts/templates/contacts/contact_detail.html
```diff
@@ -20,7 +20,7 @@
 
 
 {% block content_main_list %}
-    <form id="frm-person" method="post" action="" class="form-horizontal">
+    <form class="form-horizontal" method="post" action="">
```
hodgestar, line 23: This change seems unrelated? Just checking it's meant to be here.

go/templates/app.html
```diff
@@ -4,14 +4,17 @@
...
-          <li><a href="{% url 'contacts:people' %}">Contacts</a></li>
-          <li><a href="#">Help</a></li>
+            <li>
+                <a class="brand" href="/">
+                    <img src="{{ STATIC_URL }}img/logo-lrg.png" alt="Vumi Go" class="logo">
+                </a>
+            </li>
+            <li><a href="{% url 'conversations:index' %}">Dashboard</a></li>
+            <li><a href="{% url 'contacts:people' %}">Contacts</a></li>
+            <li><a href="#">Help</a></li>
```
hodgestar, line 17: These also seem unrelated. If they're good they can stay.

hodgestar:
:+1: as soon as you're happy with the two bits I flagged as seemingly unrelated.

go/templates/app.html
peterp, line 17: Sorry about these, I just happened to come across them whilst I was tinkering.

How do you guys usually fix unrelated things that you notice as you're going along?

hodgestar, line 17: Our policy is to do small gist reviews or explain them in the PR like we're doing now. :) Completely happy for them to stay in -- just wanted to be sure they were intended.

hodgestar:
:+1:

PR #320, opened 2013-06-05
Add default anchor behavior for StateViewEndpointCollection

justinvdm:
Currently, endpoints aren't given any default arrangement for `StateView`. While this is obviously customizeable, we need to provide some default behaviour for how endpoints are arranged (using jsPlumb anchors) that can be used as the default for all the endpoint collections on state views in parts of the UI that use the plumbing components.


justinvdm:
Depends on #293, #322, #323 and #324 (the last three are small).

go/base/static/js/src/components/plumbing/endpoints.js
```diff
@@ -62,8 +65,98 @@
...
+
+    // Move the endpoint along its side based on parameter t, where
+    // 0 <= t <= 1.
+    reposition: function(t) {
+      this.plumbAnchor = this.anchor(t);
+      return this;
+    },
+
+    render: function() {
+      EndpointView.prototype.render.call(this);
```
peterp, line 100: What does this do?

go/base/static/js/src/components/plumbing/endpoints.js
```diff
@@ -62,8 +65,98 @@
...
+      {render: false},
+      EndpointViewCollection.prototype.addDefaults),
+
+    defaults: {
+      type: StaticEndpoint,
+      side: 'left',  // the side of the state the collection is drawn on
+      margin: 0.005  // margin spacing on each end of the state side
+    },
+
+    opts: function() {
```
peterp, line 121: Do you return opts to ensure that they're not modified?

go/base/static/js/src/components/plumbing/endpoints.js
justinvdm, line 100: It calls the 'super' render method, which in this case renders the actual jsPlumb endpoint.

peterp:
:+1:

go/base/static/js/src/components/plumbing/endpoints.js
justinvdm, line 121: Not sure if I'm answering your question, but I use a function for opts so that I have access to `this`. For example, that lets me pass on `this.type` to each endpoint view, which I wouldn't be able to do with just an object.

PR #324, opened 2013-06-05
ViewCollectionGroup.prototype.render should render its collections

justinvdm:
ViewCollectionGroup.prototype.render should render its collections, not each view.


hodgestar:
:+1:

PR #327, opened 2013-06-06
LookupGroup doesn't add items to its owner lookup on member add events

justinvdm:


go/base/static/js/src/components/structures.js
```diff
@@ -100,21 +100,35 @@
       // Lookup of item owners by item keys
       this._owners = {};
 
+      // Lookup of the add callbacks bound to member add events. We only need
+      // to bind callbacks for adds and not removes, since we need to know the
+      // owner of an item to be added, while with removes, we know already. We
+      // keep a lookup so we can unbind the add callbacks when the member is
+      // unsubscribed.
+      this._adds = {};
```
hodgestar: Maybe this should be called `_subscribers` instead of `_adds`?

go/base/static/js/test/tests/components/structures.test.js
```diff
@@ -174,6 +184,30 @@ describe("go.components.structures", function() {
...
+      });
+    });
+
+    describe(".remove", function() {
+      it("should remove the key value pair", function() {
+        assert.equal(group.remove('b'), 2);
+        assert.isFalse(group.has('b'));
+      });
+
+      it("should add the item to the owner lookup", function() {
```
hodgestar: I think this should say "should remove the item from"?

go/base/static/js/test/tests/components/structures.test.js
```diff
@@ -187,7 +221,7 @@ describe("go.components.structures", function() {
         group.subscribe('c', lookupC);
 
         lookupC.on('add', function(key, value) {
-          assert.equal(group.get('i'), 9);
+          assertAdded(lookupC, key, value);
```
hodgestar: Does this no longer assert that the key and value are what we expected them to be? Same for the similar change in the removal test below.

hodgestar:
Just some minor comments, +1 once those are resolved though.

go/base/static/js/test/tests/components/structures.test.js
justinvdm: Ah! Well caught, thanks :)

go/base/static/js/src/components/structures.js
justinvdm: We already have a subscribers lookup (`members`, since they are the members of the group). `_subscribers` doesn't really suggest what the callbacks are doing (adding). How about `_memberAdds`?

go/base/static/js/test/tests/components/structures.test.js
justinvdm: Ah, you're right, thanks!

go/base/static/js/src/components/structures.js
hodgestar: I can live with `_memberAdds`. What about `_subscribersToAdd`?

hodgestar:
:+1:

PR #316, opened 2013-06-06
Reusable table modals

peterp:


go/base/static/js/src/components/tables.js
```diff
@@ -4,76 +4,119 @@
...
+                actionPrefix: '_'
+            });
+
+            _.bindAll(this,
+                'toggleAllCheckboxes',
+                'onClick',
+                'onChecked',
+                'followLink',
+                'showConfirmationModal'
+            );
```
justinvdm, line 33: The callbacks defined above in `TableFormView.prototype.events` are automatically bound to the view instance by Backbone, so we probably don't need a few of these bound here. I suppose its not too important to take them out, just don't want us to slow things down unneccesarily. Your call.

go/account/templates/account/user_list.html
```diff
@@ -14,78 +14,86 @@
...
 {% endblock %}
 
 {% block content_main %}
-<div class="row-fluid">
-    <div class="span3 sidebar">
-    </div>
+    <form class="table-form-view" action="" method="post">
+        <div class="row-fluid">
+            <div class="span3 sidebar">
+            </div>
```
justinvdm, line 28: There's probably a good reason for this, but for my understanding, why do we have an empty div here?

go/contacts/templates/contacts/contact_list.html
```diff
@@ -31,21 +31,31 @@
...
+        onCheckedSelector: '.table-form-view-buttons button',
+    });
+
+    $('.table-form-view-buttons button').each(function() {
+        var action = $(this).attr('data-action');
+        if (typeof(action) !== 'undefined') {
+            $(this).click(function() {
+                tableFormView.showConfirmationModal({action: action});
+            });
+        }
```
justinvdm: Is there any way we could factor this code block into `TableFormView`?. It seems to be repeated elsewhere, and also seems like it should belong in the view.

go/base/static/js/src/components/tables.js
peterp, line 33: Oh awesome!

go/account/templates/account/user_list.html
peterp, line 28: It pushes the content up by `span3` which gives it a visual indent that matches the sidebar.

go/contacts/templates/contacts/contact_list.html
peterp: I really wanted to, and I tried, but I kept hitting my head.
1. The buttons aren't children of the form, but I guess I could use the `onCheckedSelector` to get to them like I'm doing on line 45. Perhaps the name `onCheckedSelector` is bad... it describes the object, but not the requirement.

justinvdm: You can do it! :)

go/base/static/js/src/components/tables.js
```diff
@@ -4,76 +4,119 @@
...
             if (typeof(url) !== 'undefined') window.location = url;
-        });
-    };
+        },
+
+        showConfirmationModal: function(options) {
+
+            var $cbs = this.$el.find('tbody input:checked');
+            var numChecked = $cbs.length;
+            var q = "Are you sure you want to <%=action%> this item?";
```
justinvdm: Could we maybe define and compile `q` (and the `numChecked > 1` version) as properties on the prototype? It would be faster and would make editing easier.

go/base/static/js/src/components/tables.js
```diff
@@ -4,76 +4,119 @@
...
+
+            _.bindAll(this,
+                'toggleAllCheckboxes',
+                'onClick',
+                'onChecked',
+                'followLink',
+                'showConfirmationModal'
+            );
+
+            this.on('onChecked', this.onChecked);
```
justinvdm: I was a bit confused by the event name containing an `on`, since the callbacks are usually named `on<eventName>`. Could we maybe change the event name to `checked`? Sorry for bikeshedding, or if I am missing something.

go/base/static/js/test/tests/components/tables.test.js
```diff
@@ -1,54 +1,58 @@
...
-      $checkboxes.each(function() {
-        $(this).trigger('click');
-      });
-      assert.isTrue($table.find('thead input:checkbox').prop('checked'));
+    it('should execute `onCheckedCallback` event when a checkbox is toggled', function() {
 
-    });
+      tableFormView.options.onCheckedCallback = function() {
+        assert.isTrue($cb.prop('checked'));
+        done();
```
justinvdm, line 40: How is this test passing if `it()` does not contain `done` as an arg?

justinvdm, line 40: Ah, sorry, bad question. Its passing since we don't actually wait for `done()` to be called, since `it()`  does not contain `done` as an arg. Lets do that and fix the test if it is failing.

go/base/static/js/test/tests/components/tables.test.js
```diff
@@ -1,54 +1,58 @@
...
+      }
 
-    it('elements in `checkToggle` should toggle `disabled`', function() {
+      var $cb = $form.find('tbody input:checkbox').eq(0);
+      $cb.trigger('click');
+    });
+    
+    it('elements in `onCheckedSelector` should toggle `disabled`', function() {
```
justinvdm: Can we make this sentence read better with `it` in front?

justinvdm:
Really liking the `TableFormView` idea :) Good thing we did it. Other than the comments, looks good.

go/base/static/js/test/tests/components/tables.test.js
```diff
@@ -1,54 +1,64 @@
...
+    });
 
-      var $checkboxes = $table.find('tbody input:checkbox');
-      $checkboxes.each(function() {
-        $(this).trigger('click');
-      });
-      assert.isTrue($table.find('thead input:checkbox').prop('checked'));
+    it('should execute `onCheckedCallback` event when a checkbox is toggled', function() {
+
+      tableFormView.options.onCheckedCallback = function(done) {
```
justinvdm: I think you meant to put `done` in the function defined in `it()`.

justinvdm:
:+1: once [this test](https://github.com/praekelt/vumi-go/pull/316/files#L6R36) is telling the truth.

PR #332, opened 2013-06-07
Add bump-utils script for updating the version number.

hodgestar:
Currently the repository doesn't have one and now that we have docs there are a couple of places that need version bumping.


justinvdm:
:+1:

PR #331, opened 2013-06-07
Directional and bidirectional connection views

justinvdm:


justinvdm:
I've decided to leave bidirectional connections for another PR, since I don't think I need them to do the routing screen basics.

go/base/static/js/src/components/plumbing/connections.js
```diff
@@ -93,9 +97,40 @@
...
+      id: 'head-arrow'
+    }],
+
+    midArrow: [
+      'Arrow', {
+      width: 12,
+      height: 12,
+      location: 0.5,
+      id: 'mid-arrow'
+    }]
```
hodgestar, line 119: Is `midArrow` intentionally unused? Seems like a cool idea to have it.

justinvdm, line 119: Ya, its intentional. I've got a feeling we will be needing it for the dialogue views, so I thought I may as well keep it.

hodgestar:
:+1:

PR #339, opened 2013-06-08
Rename plumb component view types to exclude the 'View' suffix

justinvdm:
I'm suggesting that we exclude the 'View' suffix from the plumbing views. As the views get more and more specialised, we end up with complicated, overly verbose names like `MultiConnectionEndpointView` and `AligningEndpointViewCollection`. Removing the suffix, we can make things a bit simpler. We can ensure we maintain the disctinction between views and models by ensure we keep our `Model` suffix for models (for eg, `EndpointModel`). Since the views are more prone to specialisation, its probably a better decision to remove the suffix for views, as opposed to models.

### DIFF
--- a/go/base/static/js/src/components/plumbing/connections.js
+++ b/go/base/static/js/src/components/plumbing/connections.js
@@ -8,7 +8,7 @@
       SubviewCollection = structures.SubviewCollection;
 
   var plumbing = go.components.plumbing,
-      EndpointView = plumbing.EndpointView;
+      Endpoint = plumbing.Endpoint;
 
   var idOfConnection = function(sourceId, targetId) {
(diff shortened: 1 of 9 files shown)


hodgestar:
My personal instinct is to keep the `View` suffix around -- it's nice to be able to see which things being created are Backbone Views. Maybe the problem is that the class hierarchy is growing too rapidly -- I usually consider that a sign that some base class needs more generic functionality.

That said, I'm not entirely against the change. If after reading this you still want to make it, :+1:.

justinvdm:
Keeping `View` in order to know which things are Backbone Views is a very good point, it is probably worth keeping the suffix for that reason. Closing the PR.

PR #329, opened 2013-06-09
The great cleanup

peterp:
Go over all screens and implement them as they appear in the designs.


go/base/static/css/contacts.less
```diff
@@ -0,0 +1,5 @@
+.contacts {
+
+    
+
+}
```
justinvdm: Do we need this?

go/base/static/css/vumigo.css
```diff
@@ -143,21 +168,26 @@ body {
...
-.content .sidebar .extra-actions a {
+.content .sidebar .extra-actions a,
+.content .sidebar .extra-actions a.active {
   color: #0088cc;
 }
+.content .sidebar .extra-actions a.active {
+  margin: 0;
+  border: 0;
+  background: transparent;
+}
```
justinvdm, line 185: Any reason why we have two `.content .sidebar .extra-actions a.active` rule sets?

go/campaigns/templates/campaigns/wizard_3_contacts.html
```diff
@@ -28,12 +28,12 @@
...
-{% block extrascript %}
-{{ block.super }}
-<script type="text/javascript">
-    $(function() {
-        // highlight the breadcrumbs.
-        $('.breadcrumbs a').eq(2).addClass('active');
+{% block ondomready %}
+    {{ block.super }}
+    // highlight the breadcrumbs.
+    $('.breadcrumbs a').eq(2).addClass('active');
```
justinvdm, line 34: I'm wondering whether we shouldn't rather give the breadcrumb `a` elements ids. Then we can do something like `$('.breadcrumbs #contacts').addClass('active')`. It feels a bit more durable, in case we end up changing things later on, and end up setting the wrong breadcrumb as active without realising.

go/base/static/css/contacts.less
peterp: it's just a stub, I'm sure there are things that'll go in here eventually.

go/base/static/css/vumigo.css
peterp, line 185: The css for the website is generated by .less; so... you're going to see some strange rules, I'll have a look to see why it's producing doubles, but generally I would suggest ignoring the .css files.

justinvdm:
Should we also be removing these `components-table` references, or is there a reason to keep them?

```
account/templates/account/user_list.html
30:                <table class="table components-table">

campaigns/templates/campaigns/incoming_list.html
48:            <table class="table components-table">
```

go/base/static/css/vumigo.css
justinvdm, line 185: Oh right, this is the .css file, not the .less file, sorry :)

go/contacts/templates/contacts/includes/tools.html
```diff
@@ -56,6 +56,21 @@
     </form>
 </div><!--/uplContactFrm-->
 
+<div class="modal hide fade" id="delContact">
+  <form method="post" action="">
```
justinvdm, line 60: Is this `action` intentionally empty? If its something @hodgestar will need to hook up, could we add a `TODO` here?

go/contacts/templates/contacts/includes/tools.html
```diff
@@ -181,28 +196,25 @@
...
+        </fieldset>
+  </div>
+  <div class="modal-footer">
     <a href="#" class="btn" data-dismiss="modal">Cancel</a>
-    <input type="hidden" name="_delete_group_contacts" value="1"/>
-    <button type="submit" class="btn btn-danger" data-loading-text="saving..." >Delete Group's Contacts</button>
-    </div>
-    </form>
-</div><!--/emptyGroup-->
+    <button type="submit" name='_start_campaign' class="btn btn-primary" data-loading-text="saving...">Start Campaign</button>
```
justinvdm: Should `data-loading-text` rather say something like `starting campaign...` ? Also, what is `data-loading-text` used for?

go/contacts/templates/contacts/includes/tools.html
```diff
@@ -56,6 +56,21 @@
...
+<div class="modal hide fade" id="delContact">
+  <form method="post" action="">
+    {% csrf_token %}
+    <input type="hidden" name="_delete">
+    <div class="modal-body">
+      Are you sure you want to delete this item?
+    </div>
+    <div class="modal-footer">
+      <a class="btn" data-dismiss="modal" href="#">Cancel</a>
+      <button type="submit" class="btn btn-primary">OK</button>
```
justinvdm: Should we maybe have a `data-loading-text` attribute here, or wouldn't it make sense?

go/campaigns/templates/campaigns/wizard_3_contacts.html
peterp, line 34: I like that, I actually want to change this to bootstrap breadcrumbs and automate it via that "highlight" links function of mine.

go/contacts/templates/contacts/includes/tools.html
peterp: I actually don't know. It's something that I saw in the other modals from the previous site... I'll investigate.

go/campaigns/templates/campaigns/wizard_3_contacts.html
justinvdm, line 34: Ah, sounds like a good idea

go/contacts/templates/contacts/includes/tools.html
peterp: So this is something that comes with bootstrap, if you click on the button it'll disable and display the text in data-loading-text.

go/contacts/templates/contacts/includes/tools.html
```diff
@@ -56,6 +56,21 @@
...
+<div class="modal hide fade" id="delContact">
+  <form method="post" action="">
+    {% csrf_token %}
+    <input type="hidden" name="_delete">
+    <div class="modal-body">
+      Are you sure you want to delete this item?
+    </div>
+    <div class="modal-footer">
+      <a class="btn" data-dismiss="modal" href="#">Cancel</a>
+      <button type="submit" data-loading-text="saving..." class="btn btn-primary">OK</button>
```
justinvdm, line 68: Shouldn't this say `deleting...` if we are deleting the contact?

justinvdm:
:+1: once [this](https://github.com/praekelt/vumi-go/pull/329#discussion_r4629625) has changed.

PR #350, opened 2013-06-13
A 'following' endpoint is needed for the dialogue screen

justinvdm:
We need an endpoint type that follows the vertical position of a state's child element.


hodgestar:
:+1:

PR #356, opened 2013-06-13
Rpc-ify the routing models

justinvdm:


go/base/static/js/src/components/models.js
```diff
@@ -0,0 +1,30 @@
...
+    // override to specify rpc methods
+    methods: {},
+
+    // override to specify the model relations
+    relations: [],
+
+    sync: function(method, model, options) {
+      // Keep a reference to the model as one of the options so we can pass the
+      // model as one of the rpc method params
+      options.self = model;
```
hodgestar, line 21: Nice solution!

go/base/static/js/test/runner.html
```diff
@@ -21,12 +21,19 @@
     <script src="../vendor/jquery.jsPlumb-1.4.1.leakfix.js"></script>
     <script src="../vendor/lodash.underscore-1.2.1.js"></script>
     <script src="../vendor/backbone-1.0.0.js"></script>
+
+    <!--
+    NOTE: ordering is important here, switching these around results in
+    endless recursion
```
hodgestar: I would explicit name the backbone-rpc and backbone-relational in the comment since this includes file is likely to be shuffled around / edited a fair bit (and the comment / ordering / spacing of lines might get out of sync).

go/base/static/js/test/tests/testHelpers/rpc.js
```diff
@@ -0,0 +1,48 @@
...
+
+(function(exports) {
+  var assertRequest = function(req, url, method, params) {
+    assert.equal(req.url, url);
+    assert.equal(req.data.method, '/' + method);
+    assert.deepEqual(req.data.params, params || []);
+  };
+
+  var response = function(id, data) {
+    return _({id: id, jsonrpc: '2.0', result: data}).extend(data);
```
hodgestar: Seems odd to have `data` in both places. Just checking that this is correct?

go/base/static/js/test/tests/testHelpers/rpc.js
```diff
@@ -0,0 +1,48 @@
...
+
+      restore: function() {
+        stub.restore();
+      },
+
+      respondWith: function(data) {
+        // deep copy the data to ensure it can't be modified (which may cause
+        // obscure test passes/failures)
+        data = JSON.parse(JSON.stringify(data));
+        requests[0].success(response(requests[0].data.id, data));
```
hodgestar: I guess we should add helpers for testing error cases too?

go/settings.py
```diff
@@ -350,7 +350,12 @@ def abspath(*args):
             'js/vendor/jquery-1.9.1.js',
             'js/vendor/lodash.underscore-1.2.1.js',
             'js/vendor/backbone-1.0.0.js',
+
+            # NOTE: ordering is important here, switching these around
+            # results in endless recursion
```
hodgestar: Same comment as for the other note about the ordering.

hodgestar:
Almost +1 -- just a few questions and comments. I would add support for testing error cases here but leave actual tests for simple error cases for another PR unless they're easy to write (since they might reveal a bunch of other work we need to do).

go/base/static/js/test/tests/testHelpers/rpc.js
justinvdm: Thanks for catching this, no idea why I'm extending it with `data` :)

hodgestar:
:+1:

PR #363, opened 2013-06-14
Use Riak index lookups instead of mapreduce where applicable

jerith:


jerith:
This makes the tests slower (About 6 minutes for a full run on my machine to about 7 minutes), but it should make them more robust. There's a bunch more work we can do around removing unnecessary Riak access in tests, but that can wait for another branch.

hodgestar:
:+1: once Travis builds and is successful or fails in unimportant ways.

Issue #362, opened 2013-06-14
SubviewCollectionGroup needs to support creating views from submodel types


justinvdm:

justinvdm:
Turns out the support needed to go in `ViewCollection` (after some necessary refactoring). #365 was created to do that, so I'm closing this ticket.

PR #365, opened 2013-06-14
ViewCollection needs refactoring

justinvdm:
The refactoring includes allowing `ViewCollection` to work with views with subtypes, so the PR solves what #362 was meant to.


hodgestar:
:+1:

PR #367, opened 2013-06-15
ViewCollection should allow view subtypes to be defined in a way that solves circular dependency problems

justinvdm:


go/base/static/js/src/utils.js
```diff
@@ -28,9 +28,27 @@
...
+  // and an optional context to look for the object in
+  var objectByName = function(name, that) {
+    return _(name.split( '.' )).reduce(
+      function(obj, propName) { return obj[propName]; },
+      that || window);
+  };
+
+  // If given a string, gets the object by name, otherwise just returns what
+  // it was given
+  var ensureObject = function(nameOrObj, that) {
```
justinvdm, line 41: I'm not too happy with `ensureObject`. Thoughts on a better name?

hodgestar, line 41: What about `maybeByName`? I don't think `ensureObject` is that bad though.

hodgestar:
:+1:. I suggested a better name for `objectName` but mostly because you asked. I'll leave it up to you to decide which of the two you prefer.

go/base/static/js/src/utils.js
justinvdm, line 41: Sounds good :)

PR #364, opened 2013-06-17
ConversationDefinition config validation and view split

jerith:


jerith:
This ended up being more about splitting the conversation definition than about config validation.

go/vumitools/conversation/definition.py
```diff
@@ -10,7 +10,29 @@ class ConversationDefinitionBase(object):
...
+    actions = ()
+
+    def __init__(self, conv=None):
+        self.conv = conv
+
+    def get_actions(self):
+        return [action(self.conv) for action in self.actions]
+
+    def is_config_valid(self):
+        raise NotImplementedError()
```
hodgestar, line 23: We should probably remove `is_config_valid` from this PR since it's no longer about validation?

go/apps/jsbox/view_definition.py
```diff
@@ -0,0 +1,74 @@
...
+from django.http import HttpResponse
+
+from go.conversation.conversation_views import ConversationView
+from go.vumitools.conversation.definition import ConversationViewDefinitionBase
+from go.apps.jsbox.forms import JsboxForm, JsboxAppConfigFormset
+from go.apps.jsbox.log import LogManager
+
+from urlparse import urlparse, urlunparse
+
+# NOTE: All of this stuff is assumed to be sync, so we don't yield.
```
hodgestar: Some of these notes went away -- I think maybe they all can now?

go/vumitools/conversation/definition.py
jerith, line 23: It's a one-line method that I'm going to be using as soon as this lands. We could probably do it in its own branch with some of the validation logic, but I'd rather not pay the time cost for that.

go/apps/jsbox/view_definition.py
jerith: Good catch. I'll do that.

jerith:
I landed this so I could continue with the rest of the work. Any further concerns can be addressed in a new issue.

PR #370, opened 2013-06-16
ViewCollection shouldn't need to have models backing its views

justinvdm:

PR #372, opened 2013-06-17
ViewCollectionGroup should allow adding and removing items directly

justinvdm:


justinvdm:
Depends on #370

go/base/static/js/src/components/structures.js
```diff
@@ -141,14 +151,26 @@
...
-      lookup.keys().forEach(this.remove, this);
+      lookup.keys().forEach(this.onMemberRemove, this);
       this.members.remove(key);
       return lookup;
     }
   });
 
+  var idOfModel = function(obj) {
+    return obj.id
+      ? obj.id
```
hodgestar, line 165: Should this also do the `_(obj).result('id')` dance?

hodgestar, line 165: After discussion with @justinvdm this is fine -- model.id can never be a function.

go/base/static/js/src/components/structures.js
```diff
@@ -317,6 +330,21 @@
   // A self-maintaining, 'flattened' lookup of the views in a group of view
   // collections.
   var ViewCollectionGroup = LookupGroup.extend({
+    idOfModel: idOfModel,
```
hodgestar: is this reference to `idOfModel` ever used?

hodgestar:
:+1: after resolving the question about `idOfModel` on the ViewCollectionGroup.

go/base/static/js/src/components/structures.js
justinvdm: I don't think it is, thanks, will remove it.

hodgestar:
:+1:

PR #369, opened 2013-06-17
When a state view is removed, relevant endpoints and connection views should be removed

justinvdm:


justinvdm:
Depends on #372 and #370.

hodgestar:
:+1:

PR #349, opened 2013-06-19
Add Riak models for routing blocks.

jerith:
We will need these before implementing the first routing block (or possibly as part of that). See also #277.

### DIFF
--- a/go/vumitools/router/models.py
+++ b/go/vumitools/router/models.py
@@ -1,0 +1,117 @@
+# -*- test-case-name: go.vumitools.router.tests.test_models -*-
+
+from uuid import uuid4
+from datetime import datetime
+
+from vumi.persist.model import Model, Manager
+from vumi.persist.fields import Unicode, ForeignKey, Timestamp, Json, ListOf
+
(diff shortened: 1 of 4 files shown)


go/vumitools/router/models.py
```diff
@@ -0,0 +1,117 @@
...
+        self.status = ROUTER_STOPPED
+
+    def set_status_finished(self):
+        self.archive_status = ROUTER_ARCHIVED
+
+    def __unicode__(self):
+        return self.name
+
+    def get_routing_name(self):
+        return ':'.join((self.conversation_type, self.key))
```
hodgestar: Still says conversation type but I think this whole method can go? At least the equivalent method on conversation doesn't appear to be used anywhere?

hodgestar:
We should have tests for the model methods too, but other than that and the comment about `get_routing_name`, I'm :+1:

go/vumitools/router/models.py
jerith: Good point. I'll delete the one on the conversation model too.

hodgestar:
:+1:

Issue #354, opened 2013-06-13
Basic views for dialogue screen


justinvdm:

justinvdm:
#357 plans out the dialogue screen's views. Since the screen is quite complex, it makes more sense to create a ticket for implementing each basic view from the skeletons made in #357 .

PR #357, opened 2013-06-19
Skeletons for dialogue screen views

justinvdm:
I've decided to create a separate ticket and corresponding PR for first getting the skeletons of the dialogue screen views. I have a feeling the implementation of this screen is going to be non trivial, so I'd rather get the skeleton out first (before #354) so you guys have a better idea of how the views fit together.


hodgestar:
:+1:

PR #383, opened 2013-06-19
Remove GoApplicationRouter and OldGoMessageMetadata classes.

hodgestar:
These are remnants of the old routing infrastructure and are no longer used.


jerith:
:+1:, and it's about time we got around to doing this.

PR #380, opened 2013-06-19
Routing screen backend

jerith:


jerith:
This isn't ready to land yet, but I'd like some feedback on various things.

go/conversation_tmp/views.py
```diff
@@ -210,10 +212,28 @@ def pricing(request):
...
     })
 
+
 @login_required
-def routing(request, campaign_key):
-    # TODO Get initial routing model data so we can bootstrap it to page load
+def routing(request):
+    # TODO: Better Go API client.
+
+    url = 'http://localhost:8001/api/v1/go/api'
```
justinvdm: I don't think we should be hardcoding this. Should we maybe be getting this from `settings.py`?

go/api/views.py
```diff
@@ -0,0 +1,51 @@
...
+    auth = ('session_id', request.COOKIES['sessionid'])
+
+    # FIXME: This hackery shouldn't be necessary:
+    req_data = json.loads(request.body)
+    # XXX: Extract the 'id' from the request to put in the response.
+    #      Is this necessary? If so, the API worker should handle it.
+    req_id = req_data['id']
+    # XXX: Strip leading slashes off the method name.
+    #      The JS and the API worker need to agree on this.
+    req_data['method'] = req_data['method'].lstrip('/')
```
justinvdm: If Backbone.rpc offers an option for omitting the leading slash, we should probably do that instead of adding more hackery to the proxy method. I'm not sure if it needs to be in this PR, your call.

go/conversation_tmp/views.py
jerith: Good catch. I updated it in the api proxy view, but not here.

go/conversation_tmp/views.py
```diff
@@ -210,10 +212,28 @@ def pricing(request):
...
+    auth = ('session_id', request.COOKIES['sessionid'])
+    req_data = {
+        "params": [request.user_api.user_account_key],
+        "jsonrpc": "2.0",
+        "method": "routing_table",
+        "id": None,
+    }
+    data = json.dumps(req_data)
```
justinvdm, line 229: Doesn't txjsonrpc offer some client functionality (like `callRemote()`, which I think we use in the api tests), or are we doing this here since txjsonrpc doesn't support v2?

go/api/views.py
jerith: We have bigger things to worry about at the moment, but we need to clean this up eventually.

go/conversation_tmp/views.py
jerith, line 229: We're doing this the quick and dirty way because I didn't want to waste time running around looking for a non-Twisted JSON-RPC client to use from Django-land. We'll eventually need a proper client library for these APIs, but this is good enough for the single call we need to make here.

justinvdm, line 229: Ok cool, sounds sensible.

justinvdm:
Once [this](https://github.com/praekelt/vumi-go/pull/380#discussion_r4805381) is fixed, I think I'm :+1: on this PR. I do think @hodgestar needs to be made aware of the hackery that needs to be fixed (like needing a better rpc client for making api requests from Django). Not sure if that needs to be done before this PR can land, your call.

jerith:
There's a bunch of txjsonrpc work before we can get rid of all the hackery, and it's all marked with TODOs and such. I'll land this and we can clean up when we're not chasing impossible deadlines.

Issue #234, opened 2013-04-30
Better routing exception handling


jerith:
Certain kinds of "expected" routing failures (such as messages sent to unallocated tags) are currently treated as exceptions and logged as such. We should probably do something a little less drastic with them.

hodgestar:
Closing since this has been superceded by #307.

Issue #233, opened 2013-04-30
Add support for Routing Blocks to the routing dispatcher.


smn:

hodgestar:
Closing since this has been superceded by #307.

PR #307, opened 2013-06-19
Messages from the opt-out worker fail to be routed

hodgestar:
Typical error message:

```
No conversation info found for outbound message: <Message payload="{
    'transport_name': u'safaricom_sna_smpp_transport',
    'in_reply_to': u'f722a7d8-51b6-422a-ad74-45805105abd5',
    'group': None,
    'from_addr': u'XXXXX',
    'timestamp': datetime.datetime(2013, 6, 1, 6, 16, 19, 746126),
    'to_addr': u'+XXXXXXXXXXXX',
    'content': u'You have opted out',
    'routing_metadata': {u'endpoint_name': u'default'},
    ...
```

### DIFF
--- a/go/vumitools/routing.py
+++ b/go/vumitools/routing.py
@@ -222,10 +222,10 @@ class AccountRoutingTableDispatcher(RoutingTableDispatcher, GoWorkerMixin):
 
         if direction == self.INBOUND:
             allowed_types = (
-                GoConnector.CONVERSATION, GoConnector.ROUTING_BLOCK)
+                self.CONVERSATION, self.ROUTING_BLOCK, self.OPT_OUT)
         else:
             allowed_types = (
-                GoConnector.ROUTING_BLOCK, GoConnector.TRANSPORT_TAG)
+                self.ROUTING_BLOCK, self.TRANSPORT_TAG)
 
         if conn.ctype not in allowed_types:
             raise UnroutableMessageError(
@@ -250,6 +250,9 @@ class AccountRoutingTableDispatcher(RoutingTableDispatcher, GoWorkerMixin):
                     "No transport name found for tagpool %r"
                     % conn.tagpool, msg)
             dst_connector_name = transport_name
+
+        elif conn.ctype == conn.OPT_OUT:
+            dst_connector_name = self.opt_out_connector
 
         else:
             raise UnroutableMessageError(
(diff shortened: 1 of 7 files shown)


hodgestar:
This now also covers #233 and #234.

go/vumitools/routing.py
```diff
@@ -3,145 +3,386 @@
...
+
+
 class AccountRoutingTableDispatcherConfig(RoutingTableDispatcher.CONFIG_CLASS,
                                           GoWorkerConfigMixin):
     application_connector_mapping = ConfigDict(
-        "Mapping from conversation_type to connector name.", static=True)
+        "Mapping from conversation_type to connector name.",
+        static=True, required=True)
+    router_connector_mapping = ConfigDict(
+        "Mapping from routing block type to connector name.",
```
jerith: "router_type" rather than "routing block type" -- it's a field name on the model.

go/vumitools/routing.py
```diff
@@ -3,145 +3,386 @@
...
+            config.application_connector_mapping)
+        self.application_connectors = set(
+            config.application_connector_mapping.itervalues())
 
+    @inlineCallbacks
     def teardown_dispatcher(self):
-        # This assumes RoutingTableDispatcher.teardown_dispatcher() is empty.
-        return self._go_teardown_worker()
+        yield super(AccountRoutingTableDispatcher, self).teardown_dispatcher()
+        yield self._go_teardown_worker()
```
jerith, line 238: We should probably tear down in the reverse of the order in which we set up.

jerith:
Looks good to me. :+1:

go/vumitools/tests/test_utils.py
```diff
@@ -126,6 +128,59 @@ def test_set_user_account(self):
...
+        self.assertTrue(md.is_optout_message())
+
+    def test_get_router_key(self):
+        md = self.mk_md()
+        self.assertRaises(KeyError, md.get_router_key)
+        md = self.mk_md(go_metadata={'router_key': 'router-1'})
+        self.assertEqual(md.get_router_key(), 'router-1')
+
+    def test_get_router(self):
+        raise SkipTest("Add test once VumiUserApi has a get_router method")
```
jerith, line 144: As a general rule, I prefer using the `@skip()` decorator unless the skip is conditional.

hodgestar, line 144: I don't think trial has such a decorator (at least I failed to find it). Python `unittest` has one but that's 2.7+ only.

go/vumitools/tests/test_routing.py
```diff
@@ -54,134 +79,248 @@ def get_dispatcher(self, **config_extras):
...
+        self.assert_rkeys_used('router_ri.outbound', 'sphex.outbound')
+        self.with_md(msg, tag=("pool1", "5678"), endpoint='default',
+                     hops=[
+                        ['ROUTING_BLOCK:router:router1:INBOUND', 'other'],
+                        ['TRANSPORT_TAG:pool1:5678', 'default'],
+                     ])
+        self.assertEqual([msg], self.get_dispatched_outbound('sphex'))
 
     @inlineCallbacks
     def test_event_routing(self):
```
jerith: Can you break this method up a bit? I'd like to individually test the various paths an outbound message can take and assert that they're handled correctly instead of having a big monolithic test that's hard to follow.

jerith:
:+1:

I'd like a test for an event with multiple hops, but we can do that when we implement routers.

PR #385, opened 2013-06-20
Plumbing views shouldn't set their element's id attribute to the model's id

justinvdm:
Currently, the plumbing views set their element's id attribute to the model's id so we can identify the views and look them up. This causes issues when the model id fields aren't valid html ids. 

The planned solution is to set a `data-uuid` attribute on the elements, and map this to the model id.


go/base/static/js/test/tests/testHelpers/testHelpers.test.js
```diff
@@ -0,0 +1,38 @@
+describe("go.testHelpers", function() {
```
justinvdm, line 1: I added helper methods for use in the PR, and added tests for them (along with these tests). I later realised I don't need the helpers methods, but thought I may as well leave these tests here, even though its not really relevant to the PR.

jerith:
I've tested this in my branch and it works fine, but I don't know JS well enough to thoroughly review the code changes. I'm happy enough with the workingness to :+1: though.

PR #382, opened 2013-06-21
Sequential Send UI only allows you to select the first conversation.

smn:
The `tagpools_by_delivery_class()` always returns the first result, not the aggregated result.
The bug is in the indentation at https://github.com/praekelt/vumi-go/blob/develop/go/apps/sequential_send/views.py#L102.

Cookies for @imsickofmaps for both hitting the bug and spotting the problem in the code.


jerith:
:+1: once Travis is happy.

PR #388, opened 2013-06-21
ConnectionModel needs custom parsing to set connection uuids

justinvdm:


go/base/static/js/test/tests/campaign/routing/models.test.js
```diff
@@ -36,7 +36,69 @@ describe("go.campaign.routing (models)", function() {
...
+            uuid: 'conversation3',
+            type: 'js-app',
+            name: 'js-app1',
+            description: 'Some JS App',
+            endpoints: [
+              {uuid: 'endpoint10', name: 'default'},
+              {uuid: 'endpoint11', name: 'sms'}]
+          }],
+          routing_entries: [{
+            uuid: 'endpoint1-endpoint4',
```
hodgestar, line 117: Are these `uuid` entries new? Do I need to support them in the Go API?

justinvdm, line 117: This is created automatically from the routing entry's `source` and `target` when I [parse](https://github.com/praekelt/vumi-go/pull/388/files#L0R69) the response. You wouldn't need to support this in the api.

go/base/static/js/src/components/stateMachine.js
```diff
@@ -57,15 +59,19 @@
...
 
-    initialize: function() {
-      // Ensure our connection models ids are generated from their source and
-      // target so we have a consistent way of creating new connection models
-      // in the ui and looking them up later by their source and target
-      //
-      // NOTE: We don't set the id attribute used when syncing the model with
-      // the server, we only set the model's id *property* so we can use it on
-      // the client side
-      this.id = idOfConnection(this.get('source').id, this.get('target').id);
```
hodgestar, line 68: Could you explain to me why it wasn't sufficient to  set `this.id` any more and we had to set an attribute instead?

Does `this.get('source').id == ifOfModel(attrs.source)`?

go/base/static/js/test/tests/campaign/routing/models.test.js
hodgestar, line 117: Woot.

go/base/static/js/src/components/stateMachine.js
hodgestar, line 68: @justinvdm's explanation from IRC:

> fetch() calls parse(), which parses the response from the api, and then 'merges' the response with the current 
> models using set(). I need to know the model's uuid at that stage, instead of setting 'id' in initialize() (which is 
> called after the 'merging'), since it allows backbone to identify connections, and not create duplicates

i.e. setting the `id` attribute happens too late for Backbone to know to merge the identical models.

hodgestar:
:+1:

go/base/static/js/src/components/stateMachine.js
justinvdm, line 68: I think I addressed your first question on IRC, let me know if there is anything else I didn't answer.

Yes, `this.get('source').id == idOfModel(attrs.source)`

justinvdm:
After saying my [explanation from IRC](https://github.com/praekelt/vumi-go/pull/388#discussion_r4818973), I looked through lots of source to figure out what is actually going on. It seems my explanation was quite wrong (sorry!), but that this PR is still necessary.

**TL;DR** Backbone-relational expects the model data parsed from the api response to contain the model ids.

`fetch()` sets a `success` callback for the api request made, which `parse()`es, then `set()`s the attributes given in the api response. If we were using Backbone alone for fetching a collection of `ConnectionModel`s, we would be able to stick with simply setting the model id in `initialize()`. In `Backbone.Collection`'s `set()`, Backbone takes the attributes given in the response and prepares models for them (which involves initializing the model). It then checks which models already exist using their ids. Since `initialize()` would have been called when preparing each model, Backbone would ensure duplicates weren't made.

Now, the reason we still need this PR is because of the way Backbone-relational's `RelationalModel` works. `fetch()` still works the same (calling `parse()`, then `set()` with the api response). `RelationalModel`'s `set()` triggers change events for the relations which have changed. When the relation for the collection of `ConnectionModel`s recognises the change event made, it looks at the connection model objects obtained from the parsed api response and tries to resolve the model ids from this. Since we are still dealing with _objects_ instead of _initialized models_, we still do not have ids for the connections, so duplicates cannot be recognised.

Its likely that I made some mistakes in my explanation, but I do think I'm at least most of the way there.

peterp:
:+1:

PR #394, opened 2013-06-21
Display routing state name instead of description

justinvdm:


jerith:
:+1: based on it working in my browser.

PR #395, opened 2013-06-22
Reset button for routing screen

justinvdm:


justinvdm:
#396 needs to land before this will work

peterp:
:+1:

PR #398, opened 2013-06-22
Change LookupGroup to use listenTo() instead of on() for watching its member's adds()

justinvdm:
This makes things easier, since we no longer need to keep track of each member's `add()` callbacks.


peterp:
:+1:

PR #399, opened 2013-06-23
Plumbing components need refactoring

justinvdm:
Plumbing components should be able to override their default view collection group types. Supporting this properly involves quite a bit of refactoring, and will involve some refactoring of SubviewCollectionGroup (what #400 was meant to solve).

Fixing this now is important, since we will be needing to extend and override the default diagram state group type for the dialogue screen in order to implement the grid layout we need.


peterp:
:+1:

PR #401, opened 2013-06-23
Plumbing components need better namespacing

justinvdm:
All plumbing components are currently namespaced under `go.components.plumbing`. More and more specialised components are being created, so keeping them all under this namespace is getting quite clumsy.


justinvdm:
Depends on #399

peterp:
:+1:

PR #286, opened 2013-06-24
Add rolling restart command to fab scripts

hodgestar:
Currently we restart Vumi Go using a `supervisorctl restart all` command which takes down the entire system for a prolonged period. We need to be able to do rolling restarts to ensure there isn't any down time.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -1,4 +1,4 @@
-from fabric.api import cd, sudo, env
+from fabric.api import cd, sudo, env, puts
 
 env.path = '/var/praekelt/vumi-go'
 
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -48,3 +66,14 @@ def npm_install(package):
 
 def _venv_command(command, user='vumi'):
     return sudo('. ve/bin/activate && %s' % (command,), user=user)
+
+
+def _get_supervisorctl_proxy(serverurl):
+    import supervisor.xmlrpc
+    import xmlrpclib
```
jerith: Why are these imports here?

jerith:
:+1: aside from the question about imports.

fabfile.py
hodgestar: They're here so you don't have to have `supervisor` installed to run commands that don't use this or to import the fab file and list the available commands (although I'm not firmly attached to this decision).

jerith: I thought that was the case. :+1:

hodgestar:
This implementation is all messed up because fab runs all the xmlrpclib commands locally (i.e. it is less magic than I thought -- that may be a good thing).

peterp:
:+1:

PR #409, opened 2013-06-25
Related `select` widget

peterp:


go/channel/templates/channel/new.html
```diff
@@ -2,63 +2,75 @@
...
-    // configuration view for it.
-    $('input[name=channel]:checked').trigger('change');
 
-    $('input[name=channel]').on('change', function() {
-        $('.configuration').show();
-        $('.configuration fieldset').hide();
-        var $el = $('.configuration_' + $(this).attr('value'));
-        $el.show();
+    $('.country select').on('change', function() {
+        var v = $(this).find(':checked').val();
```
justinvdm, line 64: This is probably a case of me not understanding what's going on here, but if `this` is a `<select>`, shouldn't we be looking for `option:selected` instead of `:checked`?

peterp, line 64: It's a standard DOM selector so it's faster than using :selected which is something that's native to jQuery itself: http://api.jquery.com/checked-selector/

justinvdm, line 64: /me tries remember this.

justinvdm:
:+1:

PR #360, opened 2013-06-25
Campaign wizards sections into bootstrap breadcrumbs

peterp:


peterp:
I'm considering the option of us having  to calculate, or store, which `step` the conversation is in in order to make the breadcrumbs work properly.

For instance:
1. Create a new conversation, proceed to step wizard 2.
2. Abort. (And return to dashboard.)
3. Click on the conversation again, and you're on wizard 4. (You should be taken to wizard 2)

Are my assumptions correct?

PR #407, opened 2013-06-25
Move UI things to the right places

jerith:
We have a whole lot of UI stuff (templates, views, forms, etc.) that live in the wrong places, mostly for historical reasons. We need to figure out where these things should live and put them there.

This will potentially break in-flight UI branches, so it would be a good idea to agree on a time to do it.


jerith:
Probably easiest to review this a commit at a time rather than all at once.

jerith:
I accidentally broke the Wikipedia app's views. We need to rewrite them to work with conversation definition stuff anyway, though.

peterp:
This looks good to me. :+1:

hodgestar:
Can we maybe rename all of the `test_new_views.py` to `test_views.py` (or something better if someone can think of something better)?

jerith:
I thought there was something I was forgetting. I'll rename them shortly.

hodgestar:
I think `conversation_screen` does slightly the wrong thing in the case where there is no conversation definition -- it doesn't do anything with `view_name`. I don't know if anything uses it like that.

hodgestar:
I think `go/conversation/templates/conversation/conversation_base.html` might have lost links to messages, contacts and views (see https://github.com/praekelt/vumi-go/commit/24731bb1fd7a55ebb21206e068f61ba715ac5de9#L0L11). Those links probably need to be somewhere in the default view. Are they somewhere else now?

@peterp -- see above too. :)

hodgestar:
Pretty difficult to give a change this big a thorough review but I think it's ready to land after the test renaming and the couple of questions I raised have been discussed.

peterp:
@hodgestar I think we're still trying to figure out how to make the wizard work.

hodgestar:
Understood, but maybe the links should stay there until we've made them
work?

peterp:
Aaah, good point.

jerith:
I took the links out of `conversation_base.html` because they're part of the wizard flow. They're still in `wizard_base.html`.

jerith:
The `conversation_screen` hackery was there as a temporary measure. It's still (kind of) necessary for multi_surveys, but it's doing the wrong thing there. I'm not actually sure what the correct behaviour should be for viewing a broken conversation, but there's a FIXME comment and we can figure it out in a new branch.

hodgestar:
Cool. Can we get tickets for wikipedia and multi-surveys and mention all
the relevant things in their descriptions?

hodgestar:
:+1:

PR #412, opened 2013-06-26
Lookup keys need to be returned in the correct order

justinvdm:


hodgestar:
:+1:

hodgestar:
:+1: (current value: 2 :)

justinvdm:
@hodgestar  :)

PR #410, opened 2013-06-26
ViewCollection needs a way to allow ordering to be manipulated

justinvdm:


justinvdm:
Depends on #412

go/base/static/js/src/components/structures.js
```diff
@@ -107,19 +144,38 @@
...
       return this;
+    },
+
+    rearrange: function() {
+      var keys = Array.prototype.slice.call(arguments);
+
+      if (this.ordered && this.arrangeable) {
+        keys.forEach(function(k, i) { this.arranger(this.get(k), i); }, this);
+        this.sort();
+      }
```
justinvdm, line 183: I'm really not sure if this is the best way to re-order a `Lookup`. I've been stuck thinking of better alternatives for a while, but can't think of anything better. I'm definitely up for suggestions on this!

hodgestar, line 183: I think this seems fine to me. Would it not be easier to just pass a list to `rearrange` to begin with?

hodgestar:
:+1: other than the comment about maybe passing a list into rearrange directly.

go/base/static/js/src/components/structures.js
justinvdm, line 183: @hodgestar Ya, I wasn't too sure if I should be passing in a list or just args. It probably would be easier as a list, will do that.

hodgestar:
:+1:

PR #408, opened 2013-06-26
Basic dynamic grid layout

justinvdm:
This will be used for the dialogue screen.


justinvdm:
Depends on #416, #417 and #410.

go/base/static/js/test/tests/components/views.test.js
```diff
@@ -46,7 +46,7 @@ describe("go.components.views", function() {
 
       it("should position the label at the given position", function() {
         label.render();
-        assertOffset('#dummy .label', {left: 21, top: 73});
+        assertOffset('#dummy .label', {left: 19, top: 80});
```
justinvdm, line 49: This change needed to be made after Boostrap was added to the test runner page, since Bootstrap already uses the `label` css class. I'm assuming we will be using this css class for labels, so I have left the `LabelView` css class name unchanged, and instead changed the tests to cooperate with Bootstrap's `label` css class.

justinvdm:
Dependencies have been landed into `develop-frontend`, ready for review.

go/base/static/js/src/components/grid.js
```diff
@@ -0,0 +1,196 @@
...
+      this.$el.addClass('span' + this.span);
+    },
+
+    spanOfEl: function($el) {
+      var classes = ($el.attr('class') || '').split(' '),
+          i = classes.length;
+
+      while (i--) {
+        var c = classes[i];
+        if (c.substr(0, 4) === 'span') { return parseInt(c.slice(4), 10); }
```
hodgestar: Maybe we want to make this check and bit stronger and test that parseInt doesn't return 'NaN' and maybe that there aren't multiple possible class elements that match?

justinvdm: Good idea. For the case of multiple matching class elements, can we not just return the first match, as we are doing now?

justinvdm: @hodgestar Also, a bit unrelated, but I'd like to ask your opinion. We do `spanOfEl` each time we create a `RowItemView`, which is on every grid render, so I think it slows things down a bit. What are your thoughts on doing something like `this.item.span = span` once we have found the span, so we don't need to call `spanOfEl` every render?

hodgestar: I think on multiple matches we should return `null` or raise an error since returning the first one is just guessing (unless they're all the same I suppose?).

I'd suggest skipping caching unless we actually encounter performance issues -- how slow can splitting and matching a string be? I suspect JS engines that JIT stuff can probably optimize this kind of code pretty well (although who knows for sure?)

go/base/static/js/src/components/grid.js
```diff
@@ -0,0 +1,196 @@
...
+      this._ensureArray(options.items).forEach(this.addItem, this);
+    },
+
+    _ensureArray: function(obj) {
+      return obj instanceof ViewCollection
+        ? obj.values()
+        : obj;
+    },
+
+    viewOptions: function() { return {uuid: 'row' + this.size()}; },
```
hodgestar: Bit unsure about this as a uuid for rows -- seems not that unique. Don't might if it stays like this. Just flagging it so you can consider it briefly.

justinvdm: You're right, this should probably just be `id`.

I realised that `uuid` was the wrong thing to use for `UniqueView` (the view that sets a `data-uuid` attr on its element, so we can use uuids without worrying if they are valid html ids). Once we have more time, I'd like to change this to `uid`.

go/base/static/js/src/components/structures.js
```diff
@@ -150,14 +150,18 @@
 
     indexOfKey: function(k) { return this._indexOf('key', k); },
 
+    last: function() { return this.at(this.size() - 1); },
```
peterp, line 153: can we use `_.last()` here? Is `this` a collection? Maybe it already has a `.last()` method.

go/base/static/js/src/components/grid.js
```diff
@@ -0,0 +1,197 @@
...
+
+    initialize: function(options) {
+      this.items = this._ensureViewCollection(options.items);
+      this.sortableOptions = options.sortableOptions || this.sortableOptions;
+      this.rowType = options.rowType || this.rowType;
+      this.rowItemType = options.rowItemType || this.rowItemType;
+      this.rowCollectionType = options.rowCollectionType
+                            || this.rowCollectionType;
+
+      this.resetRows();
```
peterp: I see this is called in `render()` as well, it is necessary that it's also called here?

go/base/static/js/src/components/structures.js
justinvdm, line 153: `this` is a `Lookup`, so its like a collection, except it can keep anything, not only models. Also, never knew `_.last()` existed, doh :) thanks!

peterp:
Looks good, turns out the scary part was a jquery.simulate. :+1:

go/base/static/js/src/components/grid.js
justinvdm: Ya, we probably don't need this here.

go/base/static/js/src/components/structures.js
justinvdm, line 153: Erm, on second thought, I think we should stick with having it the way it is. We ensure we return an undefined in `at` before looking up `value` on the item, so I'd rather only be doing that handling in one place, than do it here as well. That ok?

peterp, line 153: Cool.

peterp:
:+1:

PR #419, opened 2013-06-26
Fix starting of web service inside Vumi Go API

hodgestar:
Currently:
- The Vumi Go API naively adds a child service inside `setup_worker`. The worker service is already started at this point so that does nothing useful. :/


hodgestar:
This doesn't actually appear to have fixed production. The symptom (losing the connection to Redis after a bit) persists.

hodgestar:
See also praekelt/vumi#577 for the txredis reconnecting issue.

jerith:
:+1:

PR #420, opened 2013-06-27
Provide better error reporting in routing view and add password to the routing API calls.

hodgestar:
Currently:
- The routing view provides very little feedback if its request to the Go API fails.
- Clicking `save` on the routing view requests a username and password.


hodgestar:
Backbone.Rpc doesn't support authentication as far as I can tell (see [backbone.rpc.js line 127](https://github.com/asciidisco/Backbone.Rpc/blob/master/src/backbone.rpc.js#L127) so this might be a bit tricky.

hodgestar:
We need to pass in `username` and `password` options to the `$.ajax` call (see http://api.jquery.com/jQuery.ajax/).

justinvdm:
Backbone's `save()` and `fetch()` allow us to pass in `success` and `error` callbacks, which gets passed to `Backbone.sync`. Backbone.rpc overrides `Backbone.sync` to be rpc-like, but currently, it doesn't do anything with the `error` callback. I suspect this is unintentional, but regardless, we will need to send a PR upstream or something before we will be able to handle erroneous `save()`s and `fetch()`es.

justinvdm:
Ready for review!

hodgestar:
:+1:

PR #422, opened 2013-06-27
Client tests should unregister models from the Backbone.Relational store instead of resetting the store

justinvdm:
We use `Backbone.Relational.store.reset()` to reset the store Backbone.Relational uses to keep track of models. We need to do this since tests will try create models with the same ids and attributes, and Backbone.Relational will complain.

The problem with doing this is that `Backbone.Relational.store.reset()` also resets the submodel types it keeps track of. This means that models with subtypes instantiated after a `Backbone.Relational.store.reset()` won't be instances of their subtype.

To solve this, we should be unregistering the models from the store after each test.


peterp:
:+1:

PR #425, opened 2013-06-27
GridView.p.render() should detach its rows instead of emptying itself

justinvdm:
When `GridView.prototype.render()` does a `this.$el.empty()`, all its' contained elements with jquery objects lose their data. This confuses jsPlumb.

We should instead be detaching each row, before re-appending the new rows later.


hodgestar:
:+1:

PR #426, opened 2013-06-27
Add go_manage_conversation for managing conversations.

hodgestar:
Its first features will be:
- listing conversations for an account
- showing conversation config

### DIFF
--- a/go/base/management/commands/go_manage_conversation.py
+++ b/go/base/management/commands/go_manage_conversation.py
@@ -40,6 +40,8 @@ class Command(BaseCommand):
             self.list_conversations(user_api)
             return
 
+        if 'conversation_key' not in options:
+            raise CommandError('Please specify a conversation key')
         conversation = user_api.get_wrapped_conversation(
             options['conversation_key'])
         if conversation is None:
--- a/go/base/management/commands/go_manage_conversation.py
+++ b/go/base/management/commands/go_manage_conversation.py
@@ -52,9 +52,11 @@ class Command(BaseCommand):
 
     def list_conversations(self, user_api):
         conversations = user_api.active_conversations()
-        for c in conversations:
-            self.stdout.write("%s (type: %s, key: %s)"
-                              % (c.name, c.conversation_type, c.key))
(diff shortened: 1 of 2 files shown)


go/base/management/commands/go_manage_conversation.py
```diff
@@ -0,0 +1,62 @@
...
+        make_option('--list',
+                    dest='list_conversations',
+                    action='store_true', default=False,
+                    help='List the active conversations in this account.'),
+        make_option('--show-config',
+                    dest='show_config',
+                    action='store_true', default=False,
+                    help='Display the config for a conversation'),
+    ]
+    option_list = BaseCommand.option_list + tuple(LOCAL_OPTIONS)
```
jerith, line 29: Why build a list and call `tuple()` instead of just building a tuple?

go/base/management/commands/go_manage_conversation.py
```diff
@@ -0,0 +1,62 @@
...
+        user_api = vumi_api_for_user(user)
+
+        if options.get('list_conversations'):
+            self.list_conversations(user_api)
+            return
+
+        conversation = user_api.get_wrapped_conversation(
+            options['conversation_key'])
+        if conversation is None:
+            raise CommandError('Conversation does not exist')
```
jerith, line 48: We probably need a better error if the conversation isn't specified.

go/base/management/commands/go_manage_conversation.py
hodgestar, line 29: Mostly because it's what we do elsewhere and because square brackets look nicer?

go/base/management/commands/go_manage_conversation.py
hodgestar, line 48: Good point. Will fix and add test.

go/base/management/commands/go_manage_conversation.py
hodgestar, line 29: Fix pushed.

jerith:
:+1:

PR #428, opened 2013-06-28
Hook up conversation lifecycle stuff in the UI

jerith:

PR #429, opened 2013-06-28
Some extra machinery to support the new conversation lifecycle.

jerith:
This is used by #428. I screwed up the issue number in the branch name. :-/

### DIFF
--- a/go/vumitools/conversation/utils.py
+++ b/go/vumitools/conversation/utils.py
@@ -52,6 +52,13 @@ class ConversationWrapper(object):
         yield self._release_batches()
 
     @Manager.calls_manager
+    def archive_conversation(self):
+        self.c.set_status_finished()
+        yield self.c.save()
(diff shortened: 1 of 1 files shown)


go/vumitools/conversation/utils.py
```diff
@@ -52,6 +52,13 @@ def end_conversation(self):
         yield self._release_batches()
 
     @Manager.calls_manager
+    def archive_conversation(self):
+        self.c.set_status_finished()
+        yield self.c.save()
+        yield self._remove_from_routing_table()
+        yield self._release_batches()
```
hodgestar, line 59: I think these last two should actually happen before the status is set to finished?

hodgestar, line 59: Maybe everything has to support slightly inconsistent states anyway, so let's leave this.

hodgestar:
:+1:

PR #430, opened 2013-06-28
Move toy dialogue state view and model out of tests

justinvdm:
We add a toy state model and view subtype to dialogue states for tests. This happens after `DialogueStateModel`'s submodel's are added to `Backbone.Relational.store`, so the store ends up not knowing about the toy dialogue state subtype.

This means we need to move the toy submodel and view out of the test code. Since the toy model and view are used when testing out basic dialogue state functionality, it wouldn't hurt to keep these in the actual code outside of the tests.


hodgestar:
:+1:

PR #433, opened 2013-06-28
Allow HTTP API concurrent connections setting to be configurable and clean-up HTTP authorization hook-up.

hodgestar:
Currently the number of concurrent connections per-account is a constant on the class. It would be useful if it was configurable via the `.yaml` config file.


jerith:
:+1:

PR #438, opened 2013-06-29
End dialogue state

justinvdm:


hodgestar:
Other than the same set of references remaining to `choice` in test docs as in the `freetext` PR, :+1:.

hodgestar:
:+1:

PR #435, opened 2013-06-29
Standalone view for new conversation

jerith:
We used to have one of these, but it was horrible and I made it POST-only. We need a new one.


hodgestar:
Can we get tests for the new GET part of the view? Otherwise :+1:

hodgestar:
Test added. @jerith could you review the test? If you're happy with it, I'm :+1: on everything else.

PR #439, opened 2013-06-29
Choices can't be added to a dialogue choice state after a reset

justinvdm:


go/base/static/js/test/tests/campaign/dialogue/models.test.js
```diff
@@ -1,24 +0,0 @@
-describe("go.campaign.dialogue.models", function() {
```
justinvdm, line 1: These tests went away because we aren't doing the `uuid` initializing in `DialogueStateModel`'s `initialize()` anymore. We now just rely on the `defaults` prototype property Backbone offers.

hodgestar:
:+1:

PR #440, opened 2013-06-29
Reduce reliance on delivery_class

jerith:
There are a bunch of place we rely on a conversation's `delivery_class` field. Conversations are no longer guaranteed to have this.

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -51,8 +51,7 @@ class TestBulkMessageApplication(AppWorkerTestCase):
                 msisdn=from_addr.format(i), groups=[group])
 
         conversation = yield self.create_conversation(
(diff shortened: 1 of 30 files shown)


hodgestar:
:+1:. The `addr_for` hack that assumes the address type is msisdn when we don't know it going to bite us unless we fix it really soon. I've created a Vumi ticket (praekelt/vumi#581) for adding address types but we also need some sort of ticket on the Go side for an interim solution (maybe specifying the message type in the form where we send messages or in the conversation config?).

hodgestar:
And can we get a final commit renaming `new_start()` to `start()`?

jerith:
Agreed on the `addr_for` thing. That needs to live in `develop-frontend` and this needs to live in `develop`, so I'll do it in a new branch.

hodgestar:
:+1:

PR #445, opened 2013-06-29
Cancel button is broken for ChoiceStateEditView

justinvdm:
Clicking cancel confuses the choice endpoints, since they lose their target element temporarily.


peterp:
:+1:

PR #451, opened 2013-07-01
Add means to specify address type when sending messages that aren't replies.

jerith:
In #440 a lot of the delivery_class things got cleaned up but we now need a way for the user to specify the address type to use when sending messages to contacts. Options are:
- Specifying the address type in the message sending form (I think this is my preferred method currently because it allows the user to specify the address type each time).
- Including the address type in the conversation config (this requires the user to know to update the config before sending messages to a new type of address so I'm less keen on this).


jerith:
I'm going to pass the `delivery_class` in with the conversation actions that require it and then have the code that generates the command messages look it up on the conversation model.

jerith:
This needs a second part that will live in develop-frontend.

smn:
:+1:

PR #449, opened 2013-07-01
Dialogue states don't always save fields to model

justinvdm:
Currently, we have a `.save()` for dialogue state views that looks up the values of each element that corresponds to a model attribute, and updates the model. This means that if the view is re-rendered before `.save()` is called, we lose the user's input, and the model can't be updated.

The plan is to get rid of `.save()`, and rather set each model attribute corresponding to an element on that element's change event.


hodgestar:
:+1:

PR #453, opened 2013-07-01
Dialogue state reset confirmation should only be shown for connected states

justinvdm:


justinvdm:
@hodgestar @badlard This PR ensures that the `Changing the message type will break the message's connections` confirmation modal is only shown for connected states. I'm starting to wonder if this is that useful though. The reset action resets the state completely, so maybe we should instead show the modal regardless of whether the state is connected, and change it to say something like `Changing the message type resets the message`. Thoughts?

hodgestar:
@justinvdm I think I agree the confirmation message should always be shown. Perhaps it should read `Changing the message type will break the message's connections and reset all its content`?

justinvdm:
@hodgestar Sounds good, closing this PR.

PR #414, opened 2013-07-02
Tests for wizard and routing views

jerith:


go/routing/views.py
```diff
@@ -23,10 +22,18 @@ def routing(request):
...
     }
     model_data.update(r.json['result'])
 
+    return json.dumps(model_data)
+
+
+@login_required
+def routing(request):
+    model_data = _get_routing_table(
+        request.user_api.user_account_key, request.COOKIES['sessionid'])
```
hodgestar: We should use `request.session.session_key` which works for Django setups that don't use session cookies too.

go/wizard/tests.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_post_create_view_invalid_country(self):
+        response = self.client.post(reverse('wizard:create'), {
+            'conversation_type': 'bulk_message',
+            'name': 'My Conversation',
+            'country': 'Nowhere',
+            'channel': 'longcode:',
+        })
+        # We just assert that we don't redirect here.
+        self.assertNotEqual(response.status_code, 302)
```
hodgestar: I'd prefer if we asserted some specific expected status here (and in the test below).

go/routing/views.py
jerith: Ah, _that's_ how you do it. I couldn't find docs.

go/wizard/tests.py
jerith: I think we're currently doing the wrong thing when we get an invalid form, but I can assert a 200 response instead.

go/routing/tests.py
```diff
@@ -0,0 +1,108 @@
...
+                u'uuid': u'%s:%s' % (pool, tag),
+                u'name': tag,
+                u'tag': [pool, tag],
+                u'description': u"%s: %s" % (
+                    pool.replace('_', ' ').title(), tag),
+                u'endpoints': [{
+                    u'name': u'default',
+                    u'uuid': u'TRANSPORT_TAG:%s:%s:default' % (pool, tag),
+                }],
+            })
```
hodgestar: Can be replaced with a call to `go.api.go_api.api_types.ChannelType.format_channel(tag)`.

go/routing/tests.py
```diff
@@ -0,0 +1,108 @@
...
+                u'uuid': conv.key,
+                u'type': conv.conversation_type,
+                u'name': conv.name,
+                u'description': conv.description,
+                u'endpoints': [{
+                    u'name': u'default',
+                    u'uuid': u'CONVERSATION:%s:%s:default' % (
+                        conv.conversation_type, conv.key),
+                }],
+            })
```
hodgestar: Can be replaced with a call to `go.api.go_api.api_types.ConversationType.format_conversation(tag)`.

go/routing/tests.py
```diff
@@ -0,0 +1,108 @@
...
+        self.client = Client()
+        self.client.login(username=self.user.username, password='password')
+
+    def tearDown(self):
+        views._get_routing_table = self._orig_get_routing_table
+        super(RoutingScreenTestCase, self).setUp()
+
+    def get_routing_table(self, user_account_key, session_id):
+        self.assertEqual(user_account_key, self.user_api.user_account_key)
+        response = self.routing_table_api_response.copy()
```
hodgestar: Maybe we should set `self.routing_table_api_response` to `None` in `setUp` or on the class so that it's always present?

go/routing/tests.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class RoutingScreenTestCase(DjangoGoApplicationTestCase):
+    # Most of the functionality of this view lives in JS, so we just test that
+    # we're correctly injecting initial state into the template.
+
+    def setUp(self):
+        # Monkey patch _get_routing_table
+        self._orig_get_routing_table = views._get_routing_table
+        views._get_routing_table = self.get_routing_table
```
hodgestar: It `setUp` fails part way `tearDown` won't get called which will then make subsequent tests fail in confusing ways possibly. I think the `mock.patch` decorator actually makes sense here -- then we can do:

``` python
@patch('go.routing.views._get_routing_table')
def test_foo(self, mock_routing):
    mock_routing.return_value = self.make_routing_table()
    ...
```

hodgestar:
Lots of minor comments left but :+1: generally.

go/routing/tests.py
```diff
@@ -0,0 +1,81 @@
...
+        for src, dst in routing:
+            routing_table[u'routing_entries'].append(
+                RoutingEntryType.format_entry(mkconn(src), mkconn(dst)))
+
+        return json.dumps(routing_table)
+
+    @patch('go.routing.views._get_routing_table')
+    def test_empty_routing(self, mock_routing):
+        mock_routing.side_effect = self.get_routing_table
+        self.routing_table_api_response = self.make_routing_table()
```
hodgestar, line 62: I think we should be able to do away with `get_routing_table` and just set `mock_routing.return_value` and check that it was called correctly with `mock_routing.assert_called_once_with`?

jerith, line 62: I'd prefer to leave the method in there. It makes it clear what the function signature is and lets us do more complicated stuff later if necessary.

hodgestar, line 62: I don't know what these more complicated things are? My objection is mostly to stashing the routing table on self an then yanking it off again and to asserting things a bit behind the scenes where they're not that visible to someone reading the test.

jerith, line 62: I think the magic in the mocking library makes things less readable and less explicit. Also, I'd rather write one thing that does everything we need in the fake routing table API call than duplicate it all in every test.

hodgestar:
:+1: -- I don't really agree on the mock issue but it's also not really too important. Both ways work. :)

PR #454, opened 2013-07-02
Dialogue connections need to be easier to follow

justinvdm:


go/base/static/css/conversations.less
```diff
@@ -6,8 +6,32 @@
...
+      opacity: 0.5;
+    }
+
+    ._jsPlumb_hover {
+      z-index: 3;
+      opacity: 0.75;
+    }
+
+    *:not(._jsPlumb_connector) {
+      z-index: 2;
```
hodgestar: Will this not override some of the z-indexes above?

go/base/static/js/src/campaign/dialogue/style.js
```diff
@@ -0,0 +1,33 @@
+// go.campaign.dialogue.states.dummy
```
hodgestar: Should be `go.campaign.dialogue.style`?

hodgestar:
:+1: other than the `z-index` question and the fix to the styles file header.

go/base/static/css/conversations.less
justinvdm: Good point, will change this. I don't like using `:not`, but it is the only workaround to get the z indices to work properly.

hodgestar:
:+1:

PR #455, opened 2013-07-02
Prettier titlebars for dialogue states

justinvdm:


go/base/static/css/conversations.less
```diff
@@ -41,11 +41,16 @@
       .user-select(none);
 
       .titlebar {
+        font-size: 16px;
         padding: 8px 6px;
+
+        // Set the margin to the reverse of the state's padding so the titlebar
+        // actually looks like a titlebar
+        margin: -6px;
```
hodgestar: Maybe we can use a variable for this `6px` value?

hodgestar: Although we're not using variables anywhere else really and I don't think here is the right place to start.

hodgestar:
:+1:

go/base/static/css/conversations.less
justinvdm: Hmm, it does sound like a good idea though. I keep forgetting this is less :)

PR #463, opened 2013-07-02
Prettify subscription edit view

peterp:


go/templates/base.html
```diff
@@ -24,6 +20,13 @@
     
     {% compressed_css 'all' %}
 
+    {% compressed_js 'lib' %}
+    {% compressed_js 'go' %}
```
hodgestar: Is it safe to  move these up here? I know we've had issues with Javascript running before the rest of the page has finished loading?

peterp: It's not ideal, because now it'll load and execute the JavaScript before rendering the page.

But... If that was an issue beforehand then it was one that definitely shouldn't be fixed by placing the JavaScript at the bottom of the page because if it needed to work on the DOM then it should wait until the DOM has loaded and attached to the appropriate event.

hodgestar: Can we put it back where it was for now and leave a comment with a TODO? Moving this requires running around and making sure everything that needs to wait for page load does so (I think).

peterp: Ok. :)

go/base/static/js/source_url.js
```diff
@@ -0,0 +1,46 @@
+function SourceUrl(elem_id, dest_id) {
```
hodgestar, line 1: Where was `source_url.js` before? Did it accidentally get deleted at some point?

go/conversation/templates/conversation/edit.html
```diff
@@ -1,57 +1,30 @@
...
-<link type="text/css" href="{{STATIC_URL}}css/smoothness/jquery-ui-1.8.11.custom.css" rel="stylesheet" />
-<link type="text/css" href="{{STATIC_URL}}css/jquery-ui-timepicker.css" rel="stylesheet" />
+{% block extrahead %}
+    {{ edit_forms_media }}
 {% endblock %}
 
-{% block extra_js %}
-<script>
-    $(document).ready(function(){
-        $("#conversation").validate();
```
hodgestar, line 11: Any idea what this was for?

go/base/static/js/source_url.js
peterp, line 1: Yup. I had to go find it on develop.

go/conversation/templates/conversation/edit.html
```diff
@@ -1,57 +1,30 @@
...
+    <form id="form-conversation" class="indent" method="post" action="">
+        {% csrf_token %}
+        {% for edit_form in edit_forms %}
+            <fieldset>
+                {{ edit_form }}
+            </fieldset>
+        {% endfor %}
+
+        <div class="form-actions">
+            <input class="btn btn-primary" type="submit" value="Save">
```
hodgestar: I'm a little worried that if the 'Save' button is all the way at the bottom a large set of edit forms (e.g. in the Javascript application) it'll hard to find. +20 on changing the name to 'Save' -- that's been on our TODO list for forever. :)

go/conversation/templates/conversation/edit.html
peterp, line 11: None. I recall seeing it earlier, with the authentication forms. I couldn't find a library that validates forms. Seems a bit odd to blanket validate a form when the form doesn't provide any metadata in order to do that.

go/base/static/js/source_url.js
hodgestar, line 1: Doh. Thanks for noticing and yoinking it back in.

go/conversation/templates/conversation/edit.html
hodgestar, line 11: I assume the idea was to have a library that highlighted any missing information in a form? Can bootstrap forms do that?

hodgestar, line 11: I suspect it's an attempt to use http://jqueryvalidation.org/documentation/.

hodgestar, line 11: The develop branch has:

```
$ find go -name '*validate*'
go/base/static/js/jquery.validate.min.js
go/base/static/js/libs/jquery.validate.js
```

Maybe we should put this back too?

hodgestar, line 11: So, I looked at the HTML source generated and since it doesn't look like jquery.validate is getting given anything interesting lets leave it out for now and put it back later if we decide to do something interesting with it.

hodgestar:
:+1:

PR #464, opened 2013-07-03
Channel dashboard view

jerith:
Mostly so I have a way to view and release channels.

### DIFF
--- a/go/channel/templatetags/channel_tags.py
+++ b/go/channel/templatetags/channel_tags.py
@@ -1,3 +1,5 @@
+import urllib
+
 from django import template
 
 from go.channel.views import get_channel_view_definition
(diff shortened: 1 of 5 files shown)


go/channel/templatetags/channel_tags.py
```diff
@@ -10,4 +12,5 @@
 def channel_screen(channel, view_name='show'):
     # TODO: Update this when we have proper channels.
     view_def = get_channel_view_definition(channel)
-    return view_def.get_view_url(view_name, channel_key=channel.key)
+    return view_def.get_view_url(
+        view_name, channel_key=urllib.quote(channel.key))
```
hodgestar: Do we actually need the urllib.quote here? If so, I think we messed up somewhere in writing `get_view_url`.

jerith: We need it because the "channel key" is actually a "pool:tag" string. But you're right, it should go in `get_view_url()` which is the current hack around not actually having a proper channel model yet.

jerith: Actually, it needs to be `CheapPlasticChannel.key`, since that's where the hackery lives.

hodgestar:
:+1:

PR #467, opened 2013-07-03
Check action is enabled before showing action form or processing GETs or POSTs.

hodgestar:
Currently actions are checked for being enabled just before being performed (which is  good) but users could do with feedback a little earlier (so they don't fill out forms or get asked for confirmations).


jerith:
:+1: on the Python. I'll leave the template and JS to someone else.

peterp:
:+1:

PR #460, opened 2013-07-04
Prettify JavaScript edit view

peterp:


peterp:
Ignore that comment above, I undid the fix.

go/base/static/js/src/widgets/initCodemirror.js
```diff
@@ -0,0 +1,12 @@
+$(function () {
+
+    var $textareas = $('textarea[data-widget=codemirror]');
+
+    $textareas.each(function(i, el) {
+        var cm = CodeMirror.fromTextArea(el, codemirrorConfig[$(el).attr('id')]);
+        // TODO: figure out where this comes from?
```
hodgestar: Maybe update this comment to explain that 'on_source_update' is called by the source URL widget?

hodgestar:
Approach looks cool.

go/base/static/js/src/go.js
```diff
@@ -10,4 +10,4 @@
   // Models created automatically when added to Backbone.Collection won't be
   // recognised as instances of Backbone.Model, so we need to change this.
   Backbone.Collection.prototype.model = Backbone.Model;
-})(window.go = {});
+})(window.go = window.go || {});
```
justinvdm: Could we rather define `configs` inside here (`exports.configs = {}`). That would git rid of the `|| {}` part here and the defining we are doing [here](https://github.com/praekelt/vumi-go/pull/460/files#L3R117), so we can ensure we are only ever defining things in one place.

justinvdm: Or does the codemirror setup code get reached before this? :/

justinvdm:
:+1: `SourceUrl` probably needs some tlc, but that's probably for another day.

go/templates/base.html
```diff
@@ -23,6 +20,12 @@
     
     {% compressed_css 'all' %}
 
+    <script type="text/javascript">
+        // namespacing
```
hodgestar: Can we get more of a comment here? Maybe "creating namespaces ahead of time of time so that Django widgets have a place to stash their configuration even though other Javascript hasn't loaded yet"?

hodgestar:
:+1: once there's more of a comment.

PR #466, opened 2013-07-04
Refactor app view test infrastructure

jerith:
There's a bunch of variance and duplication is both view and app worker tests. We need to clean up and standardise this to make tests more readable and more maintainable. In the interests of not updating the whole world, this issue only touches the view tests.

### DIFF
--- a/go/apps/multi_surveys/tests/test_views.py
+++ b/go/apps/multi_surveys/tests/test_views.py
@@ -1,14 +1,12 @@
 from datetime import date
 from zipfile import ZipFile
 from StringIO import StringIO
-import uuid
 
 from django.core.urlresolvers import reverse
 from django.core import mail
 
 from go.vumitools.tests.utils import VumiApiCommand
 from go.apps.tests.base import DjangoGoApplicationTestCase
-from go.base.tests.utils import declare_longcode_tags
 
 
 class MultiSurveyTestCase(DjangoGoApplicationTestCase):
@@ -20,7 +18,7 @@ class MultiSurveyTestCase(DjangoGoApplicationTestCase):
             VXPOLLS_REDIS_CONFIG=self._persist_config['redis_manager'])
 
     def add_tagpool_to_conv(self):
-        declare_longcode_tags(self.api)
+        self.declare_tags(u'longcode', 4)
         self.add_tagpool_permission(u'longcode')
         conv = self.get_wrapped_conv()
         conv.c.delivery_class = u'sms'
@@ -232,7 +230,7 @@ class MultiSurveyTestCase(DjangoGoApplicationTestCase):
         self.client.post(reverse('multi_survey:start', kwargs={
             'conversation_key': self.conv_key}))
 
-        self.put_sample_messages_in_conversation(1)
+        print self.put_sample_messages_in_conversation(1)
         conversation = self.get_wrapped_conv()
         [msg] = conversation.received_messages()
         response = self.client.post(reverse('multi_survey:show', kwargs={
--- a/go/apps/multi_surveys/tests/test_views.py
+++ b/go/apps/multi_surveys/tests/test_views.py
@@ -185,15 +185,13 @@ class MultiSurveyTestCase(DjangoGoApplicationTestCase):
 
     def test_aggregates(self):
         self.setup_conversation(started=True)
-        self.put_sample_messages_in_conversation(
-            10, start_date=date(2012, 1, 1), time_multiplier=12)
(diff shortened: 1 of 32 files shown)


go/apps/multi_surveys/tests/test_views.py
```diff
@@ -112,8 +120,12 @@ def test_start(self):
         """
         Test the start conversation view
         """
+        self.setup_conversation(with_group=True, with_contact=True)
+        self.add_tagpool_to_conv()
+
         response = self.client.post(reverse('multi_survey:start', kwargs={
             'conversation_key': self.conv_key}))
+        print repr(response.content)
```
hodgestar: Stray print.

go/apps/tests/base.py
```diff
@@ -16,67 +19,50 @@ class DjangoGoApplicationTestCase(VumiGoDjangoTestCase):
...
             'description': u"Test message",
-            'delivery_class': u"sms",
-            'delivery_tag_pool': u"longcode",
-            'groups': [self.group_key],
             'config': {},
-            }
+        }
+        if with_group:
+            if self.group is None:
+                self._create_group(with_contact)
```
hodgestar: Since `_create_group` sets `self.group` maybe this check for `if self.group is None` should just go inside `_create_group`? And same for `_create_contact`? Maybe they should also be called `_setup_X` instead of `_create_X` then?

jerith: I had the checks in there, but then moved them out to make it possible to override self.<whatever> if necessary.

I don't really care much about the naming, but I think "create" is slightly more accurate than "setup". The only thing that's actually using them right now is `setup_conversation()`.

go/conversation/view_definition.py
```diff
@@ -191,7 +188,7 @@ def send_one_off_reply(user_api, conversation, in_reply_to, content):
                 "conversation_key": conversation.key,
                 "to_addr": inbound_message['from_addr'],
                 "content": content,
-                "msg_options": msg_options,
+                "msg_options": {'in_reply_to': in_reply_to},
```
hodgestar, line 191: This seems a bit unrelated to the test refactor -- it's a fairly drastic change to what options we include with messages. Given the presence of new routing I'm +1 on the change but I'd like us to remove `make_message_options` entirely then so there aren't bits left around to trip up people who read it. It would probably be best for this change to get into develop sooner rather than later so we can try it out.

go/apps/tests/base.py
jerith: Maybe I should get rid of the separate methods and just roll all of that code into `setup_conversation()`?

go/apps/tests/base.py
```diff
@@ -16,67 +19,50 @@ class DjangoGoApplicationTestCase(VumiGoDjangoTestCase):
...
 
-        # Also a contact
+    def _create_contact(self):
         self.contact = self.contact_store.new_contact(
             name=self.TEST_CONTACT_NAME, surname=self.TEST_CONTACT_SURNAME,
-            msisdn=u"+27761234567")
-        self.contact.add_to_group(self.group)
-        self.contact.save()
+            msisdn=u"+27761234567", groups=[self.group])
         self.contact_key = self.contact.key
```
hodgestar: We should kill `self.contact_key` and `self.group_key` at some point since they're a bit pointless, I think?

hodgestar: (maybe not in this PR since they were here already)

go/apps/tests/base.py
hodgestar: I happy with rolling the code into setup_conversation if that's the way you
want to go. I just found the current split of functionality odd.

go/conversation/view_definition.py
jerith, line 191: This module only exists in develop-frontend and it was only working because the tests were setting up old-style conversations with tags on them. Fixing the tests uncovered a bug in the code, albeit code that we're never calling in production. The old UI code in develop still has tags on conversations.

go/apps/tests/base.py
jerith: This PR is a good place for it. I'll do it at the same time as collecting all the `setup_conversation()` stuff.

go/conversation/view_definition.py
hodgestar, line 191: Cool. I created #471 to remind us.

hodgestar:
Looks good. Ping me for a review of the `_create_group` / `_create_contact` update when it's ready.

hodgestar:
:+1:

PR #471, opened 2013-07-04
Remove make_message_options since new routing no longer needs it.

jerith:
The introduction of new routing means that outbound messages from conversations no longer need tags and other metadata in order to be routed correctly (in fact these are ignored except by the message storing middleware).

### DIFF
--- a/go/apps/multi_surveys/vumi_app.py
+++ b/go/apps/multi_surveys/vumi_app.py
@@ -124,7 +124,9 @@ class MultiSurveyApplication(MamaPollApplication, GoApplicationMixin):
 
         # We reverse the to_addr & from_addr since we're faking input
         # from the client to start the survey.
-        from_addr = msg_options.pop('from_addr')
+        from_addr = msg_options.pop('from_addr', None)
+        msg_options.setdefault('transport_type', None)
+        msg_options.setdefault('transport_name', None)
         conversation.set_go_helper_metadata(
             msg_options.setdefault('helper_metadata', {}))
         msg = TransportUserMessage(from_addr=to_addr, to_addr=from_addr,
--- a/go/vumitools/conversation/utils.py
+++ b/go/vumitools/conversation/utils.py
@@ -162,13 +162,6 @@ class ConversationWrapper(object):
             groups.extend((yield bunch))
         returnValue(groups)
 
-    @Manager.calls_manager
-    def make_message_options(self, tag):
-        yield self.get_tagpool_metadata('msg_options')  # force cache update
-        msg_options = yield self.user_api.msg_options(
-            tag, self._tagpool_metadata)
-        returnValue(msg_options)
-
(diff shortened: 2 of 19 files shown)


hodgestar:
The `develop-frontend` branch is already having some work done in this direction as part of #466.

go/apps/multi_surveys/vumi_app.py
```diff
@@ -124,7 +124,9 @@ def start_survey(self, to_addr, conversation, **msg_options):
 
         # We reverse the to_addr & from_addr since we're faking input
         # from the client to start the survey.
-        from_addr = msg_options.pop('from_addr')
+        from_addr = msg_options.pop('from_addr', None)
+        msg_options.setdefault('transport_type', None)
+        msg_options.setdefault('transport_name', None)
```
hodgestar, line 129: Just want to think a bit about the consequences of setting these to None.

go/vumitools/conversation/utils.py
```diff
@@ -162,13 +162,6 @@ def get_groups(self):
             groups.extend((yield bunch))
         returnValue(groups)
 
-    @Manager.calls_manager
-    def make_message_options(self, tag):
-        yield self.get_tagpool_metadata('msg_options')  # force cache update
-        msg_options = yield self.user_api.msg_options(
```
hodgestar, line 168: Should we also be removing `user_api.msg_options` in this PR?

go/apps/multi_surveys/vumi_app.py
hodgestar, line 129: Okay, I think I'm happy with this.

go/vumitools/conversation/utils.py
jerith, line 168: Good point. Only http_api and jsbox seem to be using it.

hodgestar, line 168: Actually, `user_api.msg_options()` can stay, just most of it can be deleted (it just needs to add the tag and maybe not much else).

jerith, line 168: Adding the tag is the problem, since we're not going to have one to add once the old UI goes away.

hodgestar, line 168: You're right. I think we can just have it set the endpoint instead then?

jerith, line 168: There's nothing left that actually needs this stuff, so I can just remove it all.

go/apps/multi_surveys/vumi_app.py
jerith, line 129: Yes, they're all allowed to be `None` as long as they're passed explicitly. The transports need them, but they get set by the dispatcher before they get there.

smn:
Wow. This is really happening! \o/

hodgestar:
:+1:

PR #481, opened 2013-07-05
Allow conversations in transitional states to be stopped

jerith:
If a conversation somehow gets stuck in a transitional state, there's currently no way to get it out. We should be able to stop such a conversation.


hodgestar:
:+1:

PR #386, opened 2013-07-05
Add routing table helper function for computing the set of channels reachable from a given conversation.

hodgestar:
This is needed by the conversation show view (and will be pretty helpful for users :)

### DIFF
--- a/go/testsettings.py
+++ b/go/testsettings.py
@@ -18,4 +18,6 @@ if os.environ.get('VUMIGO_FAST_TESTS'):
         }
     }
 
+EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'
+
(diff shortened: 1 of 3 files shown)


hodgestar:
This is going to need a separate part that lives in `develop-frontend`.

justinvdm:
Could we maybe add a bit more complex example and correspond tests for `transitive_targets()` and `transitive_sources()`? I've got a feeling the current example might be too basic to find out if we have those working properly.

justinvdm:
:+1: once we have multi-step tests for `transitive_targets()` and `transitive_sources()`.

hodgestar:
Ready for a re-review.

justinvdm:
:+1:

PR #477, opened 2013-07-05
Separate groups and smart groups

peterp:


go/contacts/templates/contacts/base.html
```diff
@@ -10,7 +10,7 @@
 {% endblock %}
 
 {% block content_breadcrumbs %}
-    <form id="search-filter" name="search-filter" action="" method="get" class="navbar-search pull-right">
+    <form id="search-filter"name="search-filter" action="" method="get" class="navbar-search pull-right">
```
justinvdm: Intended change?

peterp: yup, there was a space missing behind id and name.

justinvdm: I'm confused, didn't we remove the space in this PR?

peterp: No, no. I'm actually the one who's confused.

justinvdm:
:+1: Once the unintended space change is gone (unless the change was intended and I'm just not understanding).

go/contacts/views.py
```diff
@@ -70,7 +70,10 @@ def groups(request):
         for group_bunch in contact_store.groups.load_all_bunches(keys):
             groups.extend(group_bunch)
     else:
-        groups = contact_store.list_groups()
+        if type == 'smart':
+            groups = contact_store.list_smart_groups()
+        else:
+            groups = contact_store.list_groups()
```
hodgestar, line 94: I would have thought that if `type` equals `static` only static groups would be displayed?

Perhaps `type` should default to `None` which displays all groups and there should be an additional `if type == 'static':` branch in the `if / else`?

PR #478, opened 2013-07-05
Accounts changes

peterp:
1. Hide the admin settings section.

### DIFF
--- a/go/account/templates/account/details.html
+++ b/go/account/templates/account/details.html
@@ -31,8 +31,7 @@
         </form>
     </div>
 
-    <!---
-    TODO: Return when appropriate.
+    {# TODO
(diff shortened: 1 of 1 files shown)


hodgestar:
I don't think we should remove the "How often do you want to receive an account summary via email" since account summary emails are a feature we currently use.

hodgestar:
I've confirmed with Ben that we need to keep the "How often do you ..." link.

go/account/templates/account/details.html
```diff
@@ -49,6 +51,7 @@
             </li>   
         </ul>
     </div>
+    -->
```
hodgestar: Shall we also use Django block comments here?

PR #413, opened 2013-07-05
Conversation definition for wikipedia app

jerith:
We need a conversation definition and related views for the wikipedia app.

The new conversation definition isn't configurable, because the old one wasn't. We need to fix that, but it's a new piece of work that can go in a new branch.


hodgestar:
:+1:

PR #474, opened 2013-07-06
Delete contacts group and contact

peterp:
Causes a CSRF validation error. Not even sure if we have hooks to delete things.


go/contacts/views.py
```diff
@@ -57,6 +58,18 @@ def groups(request):
...
+            for group_key in groups:
+                group = contact_store.get_group(group_key)
+                tasks.delete_group.delay(request.user_api.user_account_key,
+                                         group.key)
+                messages.info(request,
+                        '%d groups will be deleted shortly.' % len(groups))
+                # TODO: Should the contacts be removed as well?
+        elif '_export':
+            pass
+            # TODO: Export contacts.
```
justinvdm: Could `tasks.export_group_contacts` maybe be used here?

go/contacts/views.py
```diff
@@ -57,6 +58,18 @@ def groups(request):
...
                 return redirect(_group_url(smart_group.key))
+        elif '_delete':
+            groups = request.POST.getlist('group')
+            for group_key in groups:
+                group = contact_store.get_group(group_key)
+                tasks.delete_group.delay(request.user_api.user_account_key,
+                                         group.key)
+                messages.info(request,
+                        '%d groups will be deleted shortly.' % len(groups))
+                # TODO: Should the contacts be removed as well?
```
justinvdm: @hodgestar thoughts on this?

go/contacts/views.py
```diff
@@ -309,40 +322,50 @@ def people(request):
...
-                return redirect(_group_url(group.key))
+        if '_delete' in request.POST:
+            contacts = request.POST.getlist('contact')
+            for person_key in contacts:
+                contact = contact_store.get_contact_by_key(person_key)
+                contact.delete()
+            messages.info(request, '%d Contacts deleted' % len(contacts))
+
+        elif '_export' in request.POST:
+            # TODO: Do we have an export method?
```
justinvdm, line 339: Just to make sure I'm understanding things, this view (`_people()`) is for operating on a bunch of selected contacts, so we can't use `tasks.export_group_contacts` here, right?

go/contacts/views.py
```diff
@@ -309,40 +322,50 @@ def people(request):
 def _people(request):
     contact_store = request.user_api.contact_store
     group = None
+    upload_contacts_form = UploadContactsForm()
 
     if request.method == 'POST':
         # first parse the CSV file and create Contact instances
         # from them for attaching to a group later
```
justinvdm: I think it might make more sense for this comment to be with the code it was referring to (the uploading contact stuff that was moved down).

justinvdm:
@peterp @hodgestar I'm wondering if we should be adding tests for the contact deleting and exporting. @peterp would you be able to do this? Happy to help out if you feel like you're dealing with a backend rabbit hole.

go/contacts/views.py
hodgestar: I don't think the contacts should be deleted here so let's just remove the TODO for now.

go/contacts/views.py
hodgestar: It can -- don't mind if that's in this PR or another one.

go/contacts/views.py
```diff
@@ -309,40 +322,50 @@ def people(request):
 def _people(request):
     contact_store = request.user_api.contact_store
     group = None
+    upload_contacts_form = UploadContactsForm()
```
justinvdm: Is this line needed?

go/contacts/views.py
peterp, line 339: I don't think so either.

peterp:
Ready for a re-review.

go/contacts/views.py
```diff
@@ -57,6 +58,21 @@ def groups(request):
...
+                                         group.key)
+            messages.info(request, '%d groups will be deleted '
+                                   'shortly.' % len(groups))
+        elif '_export_group_contacts' in request.POST:
+            groups = request.POST.getlist('group')
+            for group_key in groups:
+                tasks.export_group_contacts.delay(
+                    request.user_api.user_account_key, group_key, True)
+            messages.info(request, 'The export is scheduled and should '
+                                   'complete within a few minutes.')
```
hodgestar, line 75: We should have a test for contact exporting too, I think?

PR #468, opened 2013-07-08
Inline add dialogue state button

justinvdm:


go/base/static/js/src/campaign/dialogue/states/states.js
```diff
@@ -277,9 +278,38 @@
...
 
+      var $add = $('<button>')
+        .addClass('add btn btn-primary')
+        .text('+');
+
+      var $addContainer = $('<div>')
+        .addClass('item add-container')
+        .append($add);
+
+      this.add('add-btn', $addContainer, {index: Infinity});
```
hodgestar, line 289: Woot. :)

go/base/static/js/src/campaign/dialogue/states/states.js
```diff
@@ -277,9 +278,38 @@
...
+
+    onAddClick: function(e) {
+      e.preventDefault();
+      this.states.add();
+    },
+
+    onReorder: function(keys) {
+      this.items
+        .get('add-btn')
+        .data('grid:index', Infinity);
```
hodgestar, line 307: Wouldn't the add button have its index set to `Infinity` already?

go/base/static/js/src/campaign/dialogue/states/states.js
```diff
@@ -277,9 +278,38 @@
...
+
+      keys.pop();
```
hodgestar, line 310: This seems to rely on the keys having the add button index at the end already? Maybe it would be better for the caller to do this pop?

hodgestar:
Other than my questions about `reOrder` (which are probably just me not understanding quite what's going on), :+1:.

go/base/static/js/src/campaign/dialogue/states/states.js
justinvdm, line 310: This is an event callback for `reorder` events triggered by `GridView`. Since `GridView` doesn't know about the add button (since it is generic), doing it before the event is triggered doesn't make much sense.

go/base/static/js/src/campaign/dialogue/states/states.js
justinvdm, line 307: Ya, I'd like the add button to be set to `Infinity` already. Problem is, when the user reorders the items, we do a `Lookup.prototype.rearrange()`, which sets the index of each item based on its current position. By the time `onReorder()` is reached, the `grid:index` value of the add button would be, for example, 5 when there are 4 dialogue states. Any new dialogue state would be appended after the add button, so we need to reset the add button's index to `Infinity` each reorder.

Maybe I should add a comment on this? I'm up for suggestions on a better way of doing this.

hodgestar, line 307: I think just a comment is fine.

hodgestar:
:+1:

PR #484, opened 2013-07-09
Restrict new conversations to allowed types

hodgestar:
We're not checking app permissions anywhere in the conversation creation flow. We really need to do this.

### DIFF
--- a/go/conversation/forms.py
+++ b/go/conversation/forms.py
@@ -2,19 +2,20 @@ from django import forms
 
 from bootstrap.forms import BootstrapForm
 
-from go.base.utils import configured_conversation_types
-
 
 class NewConversationForm(forms.Form):
-
-    TYPE_CHOICES = configured_conversation_types().items()
 
     name = forms.CharField(label="Conversation name", max_length=100)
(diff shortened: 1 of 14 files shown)


go/conversation/forms.py
```diff
@@ -2,19 +2,20 @@
 
 from bootstrap.forms import BootstrapForm
 
-from go.base.utils import configured_conversation_types
-
 
 class NewConversationForm(forms.Form):
 
-    TYPE_CHOICES = configured_conversation_types().items()
```
justinvdm, line 10: Since we are now getting the conversation types from the user api maybe `go.base.utils.configured_conversation_types()` can go? It doesn't seem to be used anywhere else.

hodgestar, line 10: I was actually wondering the same thing. The question is whether it's ever useful for the UI to be able to list all conversation types regardless of whether the user has access. I suspect it might be useful later when we start adding admin pages.

justinvdm, line 10: Ah, that makes sense.

justinvdm:
:+1:

PR #494, opened 2013-07-09
User should be directed to the conversation edit or show pages after clicking “save” at the end of the wizard.

hodgestar:
Currently they are always directed to the show page.

If the conversation type is editable, they should be directed to the edit page instead.


peterp:
:+1:

PR #497, opened 2013-07-10
Better logic for enabling the send message and survey actions.

hodgestar:
Currently the send message and send survey actions are enabled when a bulk message or survey conversation is activated. Ideally this logic should be extended so that the action is only enabled if the conversation is:
- activated
- linked to a channel that supports sending messages that aren't replies
- has one or more contact groups assigned

### DIFF
--- a/go/apps/bulk_message/definition.py
+++ b/go/apps/bulk_message/definition.py
@@ -12,10 +12,8 @@ class BulkSendAction(ConversationAction):
     needs_running = True
 
     def check_disabled(self):
-        channels = self._conv.get_channels()
-        for channel in channels:
-            if channel.supports_generic_sends():
-                return None
+        if self._conv.has_channel_supporting(generic_sends=True):
+            return None
         return ("This action needs channels capable of sending"
                 " messages attached to this conversation.")
 
--- a/go/apps/bulk_message/definition.py
+++ b/go/apps/bulk_message/definition.py
@@ -11,6 +11,14 @@ class BulkSendAction(ConversationAction):
     needs_group = True
     needs_running = True
 
+    def check_disabled(self):
+        channels = self._conv.get_channels()
+        for channel in channels:
+            if channel.supports_generic_sends():
+                return None
+        return ("This action needs channels capable of sending"
+                " messages attached to this conversation.")
+
     def perform_action(self, action_data):
         return self.send_command(
             'bulk_send', batch_id=self._conv.get_latest_batch_key(),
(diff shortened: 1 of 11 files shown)


hodgestar:
The group logic is already in place so we just need to add the channel logic.

go/apps/bulk_message/definition.py
```diff
@@ -11,6 +11,14 @@ class BulkSendAction(ConversationAction):
     needs_group = True
     needs_running = True
 
+    def check_disabled(self):
+        channels = self._conv.get_channels()
+        for channel in channels:
+            if channel.supports_generic_sends():
+                return None
+        return ("This action needs channels capable of sending"
+                " messages attached to this conversation.")
```
justinvdm, line 18: This `check_disabled()` and the one in `SendSurveyAction` look identical. Is there any way we could refactor this check into `ConversationAction`?

hodgestar, line 18: Not every action needs to be disabled if there is no channel to send on, so
we can't just put check_disabled on the base class. We could perhaps create
a helper method but it really doesn't have anything to do with the base
action functionality so I'm -1 on that. Open to more suggestions on how to
structure this though?

justinvdm, line 18: Maybe we need an abstract `ConversationSendAction` and we can put this `check_disabled()` there? I know we should be careful not to misuse inheritance though, but I think it might make sense in this case? Let me know what you think.

justinvdm:
Other than the `check_disabled()` refactoring question, looks good. :+1: once we've resolved that.

go/base/tests/utils.py
```diff
@@ -195,6 +196,12 @@ def add_messages_to_conv(self, message_count, conversation, reply=False,
             messages.append((msg_in, msg_out, ack))
         return messages
 
+    def add_channel_to_conversation(self, conv, tag):
```
justinvdm, line 199: This PR looks like it adds an identical `add_channel_to_conversation()` method in `GoAppWorkerTestMixin`. It looks like we need it in both places though, but I'm just a bit worried about the repitition. If you think refactoring this to avoid repitition is overkill, I'm happy for this to stay as is.

justinvdm:
:+1: Left one more comment, but it doesn't really need to be addressed in this PR (or at all). Your call.

go/base/tests/utils.py
hodgestar, line 199: It does (this is what I was mentioning on IRC). There doesn't appear to be a suitable base class to put it on though). I'm going to add a comment and then land.

PR #504, opened 2013-07-10
Fix positioning of arrow on dialogue state connections

justinvdm:
Currently, the dialogue state connection arrows are positioned near the middle of the connection. This screws up the arrow a bit when connections are made from the last state on a row to the first state on the next row.


hodgestar:
:+1:

PR #490, opened 2013-07-10
Allow users to disable the dialogue state reset modal

justinvdm:


go/base/static/js/src/campaign/dialogue/states/states.js
```diff
@@ -113,12 +119,14 @@
...
-          if (submit) { this.state.reset($option.val()); }
-          else { this.$('.type').val(this.state.typeName); }
-        }.bind(this));
+      this.resetModal
+        .once(
+          'ok',
+          function() { this.state.reset($option.val()); }, this)
+        .once(
+          'cancel',
+          function() { this.$('.type').val(this.state.typeName); }, this)
```
hodgestar, line 131: Won't this mean that if 'ok' fires, the cancel handler will still be listening and if 'cancel' fires 'ok' will still be listening? Maybe we should just bind both of these with 'on' when the modal is created?

hodgestar:
Just the question about binding the 'ok' and 'cancel' event handlers, otherwise :+1:.

go/base/static/js/src/campaign/dialogue/states/states.js
justinvdm, line 131: This can't be bound when the modal is created, since the modal needs to work across all states in the session, but the events are only relevant to this particular `.type` change event. It is a good point though. Maybe I should unbind the events when the modal is hidden?

hodgestar, line 131: Unbinding on hide sounds like a good plan. And maybe a note on why 'once'
won't work?

hodgestar:
:+1:

PR #509, opened 2013-07-12
Support for partial view templates

justinvdm:
This is needed for #459


go/base/static/js/src/components/views.js
```diff
@@ -158,10 +158,48 @@
     }
   });
 
+  var TemplateView = Backbone.View.extend({
+    jst: _.template(''),
```
peterp: Could this potentially give you an undesired result in that you might be expecting output and you get none, but it wouldn't be clear why?

go/base/static/js/src/components/views.js
```diff
@@ -158,10 +158,48 @@
...
+
+    partials: {},
+
+    data: {},
+
+    initialize: function(options) {
+      if (options.data) { this.data = options.data; }
+
+      if (options.jst) { this.jst = options.jst; }
+      this.partials = _({}).extend(this.partials, options.partials || {});
```
peterp, line 172: `({})` ascii art as code. :P

peterp:
:+1:

go/base/static/js/src/components/views.js
justinvdm, line 172: Sigh, was wondering why I never saw anyone else use the `_()` thing with an empty object before.

go/base/static/js/src/components/views.js
justinvdm: Good point. Hmm, maybe I should rather have `jst: null`, so we are forced to override it? Thoughts?

peterp: Sounds good to me.

PR #499, opened 2013-07-12
Gray action buttons on conversation show page are too large

peterp:
One the conversation show page the gray action buttons span half the page. They should span the width of the text only and be aligned with the right hand side of the page.

screen shot 2013-07-10 at 11 30 38 am

### DIFF
--- a/go/conversation/templates/conversation/show.html
+++ b/go/conversation/templates/conversation/show.html
@@ -108,7 +108,7 @@
 
     <div class="span5">
         {% if actions %}
-        <div class="sub-actions">
+        <div class="sub-actions pull-right">
             <ul class="nav nav-pills nav-stacked">
                 {% for action in actions %}
                     {% with is_disabled=action.is_disabled %}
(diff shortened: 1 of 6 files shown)


go/conversation/templates/conversation/show.html
```diff
@@ -108,7 +108,7 @@ <h4 class="control-label">
 
     <div class="span5">
         {% if actions %}
-        <div class="sub-actions">
+        <div class="sub-actions pull-right">
```
justinvdm, line 114: The ticket description says these actions should be aligned to the right and that the width should span the width of the text. This probably does the "aligning to the right" part, but I think we still need to fix the width part of the ticket description.

peterp, line 114: The width is arbitrary as it depends on the width of the text in the buttons. There's no way to effectively match them together given the current structure of the HTML.

justinvdm, line 114: Ah ok cool

justinvdm:
:+1:

PR #511, opened 2013-07-15
Reset the routing diagram's model without needing the api

justinvdm:
When the user clicks the reset button, it makes more sense to just restore the model to a backup than to call `diagram.model.fetch()` and issue an unnecessary api request.


hodgestar:
:+1:

PR #516, opened 2013-07-16
Change bulk message action button to "Write and send bulk message"

peterp:
The bulk message conversation's show page currently contains an action button named "Send bulk message". This needs to be changed to "Write and send bulk message".


justinvdm:
:+1:

PR #515, opened 2013-07-16
Change the conversation show page's 'Messages' edit button to redirect to the right place

peterp:
The 'Messages' edit button should redirect to the appropriate app edit page instead of a wizard edit screen.


justinvdm:
:+1:

PR #517, opened 2013-07-16
Allow conversations to be considered editable without needing edit forms

justinvdm:


justinvdm:
Ready for review, although tests should probably be added still.

hodgestar:
:+1: -- would be nice to have direct tests for this but since it's existing code it's not essential that that happens in this branch.

PR #512, opened 2013-07-17
Move dialogue screen into survey edit page

justinvdm:

PR #521, opened 2013-07-19
Bring back vxpolls surveys and move newer surveys into their own app.

hodgestar:
The new survey builder is wonderful but doesn't map at all well to the old vxpolls surveys so we need to temporarily support both while people migrate their old surveys over. This requires:
- restoring some of the pieces of the vxpolls survey support that was recently removed.
- adding a new survey runner that supports the new survey builder.


hodgestar:
Ready for review if Travis builds succeed.

go/apps/surveys/tests/test_views.py
```diff
@@ -263,41 +335,6 @@ def test_message_results(self, Client, MatchResult):
...
-        [(file_name, contents, mime_type)] = email.attachments
-        self.assertEqual(file_name, 'messages-export.zip')
-
-        zipfile = ZipFile(StringIO(contents), 'r')
-        csv_contents = zipfile.open('messages-export.csv', 'r').read()
-
-        # 1 header, 5 sent, 5 received, 1 trailing newline == 12
-        self.assertEqual(12, len(csv_contents.split('\n')))
-        self.assertEqual(mime_type, 'application/zip')
-
```
justinvdm, line 300: I'm not sure if these tests should be removed. I think I may have removed them in #512, so maybe something happened when those changes were reverted?

justinvdm:
Other than the comment, looks good. :+1: when that is resolved.

go/apps/surveys/tests/test_views.py
hodgestar, line 300: These are now handled by and tested on the base class, I think, so there's not need to test them here any more. There's no implementations for them in `surveys/view_definition.py` that I can see.

justinvdm, line 300: Ah ok, makes sense then.

justinvdm:
:+1:

PR #522, opened 2013-07-19
Replace Backbone.Rpc

justinvdm:


go/base/static/js/src/components/rpc.js
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      contentType: 'application/json; charset=utf-8',
+      type: 'POST',
+      dataType: 'json',
+      data: rpcData(method, model)
+    };
+  };
+
+  var sync = function(method, model, options) {
+    options = _(options || {}).extend(ajaxOptions(method, model, options));
```
hodgestar: Do we maybe want this `extend` the other way around so that options can override what ajaxOptions returns?

go/base/static/js/src/components/rpc.js
```diff
@@ -0,0 +1,42 @@
...
+      method: spec.method,
+      params: spec.params.map(function(p) {
+        return p === 'self'
+          ? model
+          : model.get(p);
+      })
+    });
+  };
+
+  var ajaxOptions = function(method, model, options) {
```
hodgestar: This doesn't appear to do anything with options?

go/base/static/js/src/components/rpc.js
```diff
@@ -0,0 +1,42 @@
...
+    };
+  };
+
+  var sync = function(method, model, options) {
+    options = _(options || {}).extend(ajaxOptions(method, model, options));
+
+    var success = options.success;
+    if (success) { options.success = function(resp) { success(resp.result); }; }
+
+    return Backbone.sync(method, model, options);
```
hodgestar, line 47: Maybe this should be left to the success callback to call? Let me keep reading a bit more though.

hodgestar, line 47: Ah. Have read up now. This part looks good.

go/base/static/js/src/components/rpc.js
justinvdm: Good idea :)

hodgestar:
A few minor comments, otherwise looks good.

go/base/static/js/src/components/rpc.js
justinvdm: True, will fix.

go/base/static/js/src/components/rpc.js
```diff
@@ -0,0 +1,42 @@
...
+      dataType: 'json',
+      data: rpcData(method, model)
+    };
+  };
+
+  var sync = function(method, model, options) {
+    options = _(options || {}).extend(ajaxOptions(method, model, options));
+
+    var success = options.success;
+    if (success) { options.success = function(resp) { success(resp.result); }; }
```
hodgestar: A JSON-RPC request can return a 200 HTTP response but still have failed and be returning its error as part of the JSON response. In this case `resp.result` will be `null` and `resp.error` will contain the error code.

hodgestar:
:+1:. 53 -- at least it's prime. :)

justinvdm:
:) hehehe, ya, prime will have to do.

PR #518, opened 2013-07-20
Add All channels link to dashboard template

peterp:
Please add a channels link in the left nav. This should be styled the same as the "All campaigns" link See screenshot:
screen shot 2013-07-17 at 4 36 57 pm

Please add the 'body' of the current channels page: http://qa-vumi.za.prk-host.net/channels/  to the 'body' of the dashboard page. This view is displayed when a user clicks "All channels" in the left nav. In other words, instead of having a separate channels page, we display the channels in the dashboard template when a user clicks "All channels" See screenshot: 
screen shot 2013-07-17 at 5 21 12 pm

### DIFF
--- a/go/conversation/templates/conversation/dashboard_base.html
+++ b/go/conversation/templates/conversation/dashboard_base.html
@@ -1,0 +1,73 @@
+{% extends "app.html" %}
+{% load url from future %}
+
+
+{% block content_extraclass %}dashboard campaigns{% endblock %}
+
+{% block content_title %}Dashboard{% endblock %}
+
+{% block content_actions_left %}
+    <a href="{% url 'wizard:create' %}" class="btn btn-primary">New conversation</a>
+{% endblock %}
+
+{% block content_actions_right %}
+    <div class="table-form-view-buttons pull-left">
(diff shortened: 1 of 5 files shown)


go/conversation/templates/conversation/dashboard_base.html
```diff
@@ -0,0 +1,73 @@
...
+                        -->
+                        <li><a href="{% url 'conversations:index' %}?conversation_status=draft">Deactivated</a></li>
+                        <li><a href="{% url 'conversations:index' %}?conversation_status=finished">Archived</a></li>
+                    </ul>
+                </li>
+
+
+                <li>
+                    <a href="{% url 'channels:index' %}">All channels</a>
+                </li>
```
hodgestar, line 55: If this nav is shared perhaps it needs to be in a shared template?

peterp, line 55: That's why I placed it in `dashboard_base.html`

hodgestar, line 55: Ah. My bad. Doh. :)

hodgestar:
:+1:

PR #485, opened 2013-07-22
Message export Zip doesn't apply any compression.

smn:
A 16MB CSV file still stays 16MB when zipped because we're not applying any compression value. It defaults to [`ZIP_STORED`](http://docs.python.org/2/library/zipfile#zipfile.ZipFile). 

Thanks to @highvoltage for pointing this out.


jerith:
There's a broken test (I think we fixed it in develop-frontend but not develop) that needs to be fixed. After that, :+1:.

PR #530, opened 2013-07-22
Clean up buttons on channel page

peterp:
Click on an individual channel name on the dashboard page (currently still under this URL on staging: http://qa-vumi.za.prk-host.net/channels/)

The user is directed to an individual channel page: http://qa-vumi.za.prk-host.net/channels/xmpp_vumigo_solo%3Axmpp-solo-0001/

We need to clean the buttons up on this page. See screen shot.

screen shot 2013-07-22 at 10 26 27 am


go/channel/templates/channel/show.html
```diff
@@ -2,17 +2,12 @@
...
 
 {% block content_actions_right %}
-    {# TODO: Better URL than conversation index. #}
-    <a href="{% url 'conversations:index' %}">Cancel</a>
-    <form action="{% channel_screen channel 'release' %}" class="form-horizontal" method="post">
-        {% csrf_token %}
-        <input type="submit" class="btn btn-primary" name="start" value="Release">
-    </form>
+<div class="pull-right">
+    <a href="{% channel_screen channel 'release' %}" class="btn btn-primary action">Release</a>
```
hodgestar: I don't think this will work if it's a GET request? At least it _shouldn't_ work. :)

go/channel/templates/channel/channel_base.html
```diff
@@ -4,25 +4,6 @@
...
-    {# TODO: Put something sensible in here. #}
-{% endblock %}
-
-{% block content_main_header %}
-    {% if messages %}
-    <div class="row-fluid">
-        <div class="span12">
-            {% include "base/includes/messages.html" %}
-        </div>
-    </div>
```
hodgestar, line 22: Are we going to put this messages stuff in a base template somewhere? Doesn't have to be in this PR -- just checking that it'll happen somewhere.

hodgestar:
Two small questions but otherwise looks good.

go/channel/templates/channel/show.html
peterp: Hmpf. Can we make it work for get requests? We're not posting anything?

go/channel/templates/channel/channel_base.html
peterp, line 22: I'll do it in here, saves me from having to do it in another PR.

go/channel/templates/channel/show.html
hodgestar: It's an action that changes state on the server so good practice would say "no, it can't be a GET request". You can argue with @jerith about it if you like. :)

hodgestar:
:+1:

PR #531, opened 2013-07-22
Remove 'channel' column from dashboard

peterp:
We'll add it back in the next release. But for 0.5b1 please remove the channel column from the dashboard.

screen shot 2013-07-22 at 1 43 36 pm


hodgestar:
:+1:

PR #523, opened 2013-07-22
Hook up old survey edit pages into new views.

hodgestar:
The old survey edit views were replaced by new dialogue ones but now that we have the old survey views back we need to hook them up properly.

This follows on from #521.


go/apps/surveys/templates/surveys/edit.html
```diff
@@ -1,35 +1,28 @@
-{% extends "app.html" %}
+{% extends "conversation/edit.html" %}
```
peterp, line 1: Does this generally serve as a base for all conversations types?

go/apps/surveys/templates/surveys/edit.html
```diff
@@ -1,35 +1,28 @@
...
-    {{ block.super }}
-    <link type="text/css" href="{{STATIC_URL}}css/smoothness/jquery-ui-1.8.11.custom.css" rel="stylesheet" />
-    <link type="text/css" href="{{STATIC_URL}}css/jquery-ui-timepicker.css" rel="stylesheet" />
+{% block content_extraclass %}campaigns surveys{% endblock %}
+
+{% block content_actions_left %}
+{% endblock %}
+
+{% block content_actions_right %}
+    <a href="{% url 'conversations:index' %}">Cancel</a>
```
peterp, line 10: This should probably return you back to the view which brought you here.

peterp:
:+1:

go/apps/surveys/templates/surveys/edit.html
hodgestar, line 1: I think so? :)

go/apps/surveys/templates/surveys/edit.html
hodgestar, line 10: Hmm. After the wizard that wouldn't make sense since the going back to the wizard screen wouldn't be useful. Maybe I need a different name for the button? Are you happy for me to merge this for now and we can deal with this issue a separate PR if needed? If you have a good suggestion right now though I'm happy to implement it though.

PR #277, opened 2013-07-22
Write key-word routing block.

jerith:
This is the first routing block so it'll likely require some infrastructure to be built.

### DIFF
--- a/go/vumitools/app_worker.py
+++ b/go/vumitools/app_worker.py
@@ -372,9 +372,9 @@ class GoApplicationMixin(GoWorkerMixin):
 
     @inlineCallbacks
     def get_message_config(self, msg):
-        if isinstance(msg, TransportEvent):
-            msg = yield self.find_message_for_event(msg)
-
+        # By the time we get an event here, the metadata has already been
+        # populated for us from the original message by the routing table
+        # dispatcher.
         msg_mdh = self.get_metadata_helper(msg)
         conversation = yield msg_mdh.get_conversation()
 
@@ -389,7 +389,8 @@ class GoRouterMixin(GoWorkerMixin):
     @inlineCallbacks
     def get_message_config(self, msg):
         # By the time we get an event here, the metadata has already been
-        # populated for us from the original message.
+        # populated for us from the original message by the routing table
+        # dispatcher.
         msg_mdh = self.get_metadata_helper(msg)
         router = yield msg_mdh.get_router()
 
--- a/go/vumitools/app_worker.py
+++ b/go/vumitools/app_worker.py
@@ -441,14 +441,7 @@ class GoRouterConfig(BaseWorker.CONFIG_CLASS, GoWorkerConfigMixin):
 
 class GoRouterWorker(GoRouterMixin, BaseWorker):
     """
-    A base class for Vumi Go router worker.
-
-    Configuration parameters:
-
-    :type worker_name: str
-    :param worker_name:
-        The name of this worker, used for receiving control messages.
-
+    A base class for Vumi Go router workers.
     """
     CONFIG_CLASS = GoRouterConfig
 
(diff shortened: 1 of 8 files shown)


go/vumitools/app_worker.py
```diff
@@ -414,3 +429,112 @@ def _publish_message(self, message, endpoint_name=None):
...
+
+class GoRouterWorker(GoRouterMixin, BaseWorker):
+    """
+    A base class for Vumi Go router worker.
+
+    Configuration parameters:
+
+    :type worker_name: str
+    :param worker_name:
+        The name of this worker, used for receiving control messages.
```
hodgestar: This bit of the docstring can go? Since it's generated for us automatically?

go/vumitools/app_worker.py
```diff
@@ -414,3 +429,112 @@ def _publish_message(self, message, endpoint_name=None):
...
+    worker_name = None
+
+    def setup_router(self):
+        return self._go_setup_worker()
+
+    def teardown_router(self):
+        return self._go_teardown_worker()
+
+    def _publish_message(self, message, endpoint_name=None):
+        if not self.get_metadata_helper(message).get_conversation_info:
```
hodgestar: Maybe .get_router_info? And we need to fix the error message below.

go/vumitools/app_worker.py
```diff
@@ -414,3 +429,112 @@ def _publish_message(self, message, endpoint_name=None):
...
+            log.error(
+                "Conversation metadata missing for message for %s: %s" % (
+                    type(self).__name__, message))
+        return super(GoApplicationWorker, self)._publish_message(
+            message, endpoint_name)
```
hodgestar: Actually, maybe this whole method can just be deleted?

go/vumitools/app_worker.py
```diff
@@ -414,3 +429,112 @@ def _publish_message(self, message, endpoint_name=None):
...
+    def handle_outbound(self, config, msg):
+        raise NotImplementedError()
+
+    def handle_event(self, config, event):
+        raise NotImplementedError()
+
+    def _mkhandler(self, handler_func, connector_name):
+        def errback(f):
+            log.error("Error routing message for %s" % (connector_name,))
+            log.error(f)
```
hodgestar: This swallows error messages which will be an issue if higher level stuff wants to publish `nacks` ever. Maybe not an issue for now though?

go/vumitools/app_worker.py
```diff
@@ -414,3 +429,112 @@ def _publish_message(self, message, endpoint_name=None):
...
+            connector.set_default_inbound_handler(
+                self._mkhandler(self.handle_inbound, connector_name))
+            connector.set_default_event_handler(
+                self._mkhandler(self.handle_event, connector_name))
+            return connector
+
+        def add_ro_handlers(connector, connector_name):
+            connector.set_default_outbound_handler(
+                self._mkhandler(self.handle_outbound, connector_name))
+            return connector
```
hodgestar, line 504: Any reason we're returning `connector` here?

hodgestar:
Looks good -- left a few comments.

go/vumitools/app_worker.py
jerith, line 504: The dispatcher code I copied this from returns it. Nothing's using it, but I generally prefer callbacks to return whatever they were given unless they specifically need to return something different.

go/vumitools/app_worker.py
```diff
@@ -378,8 +366,34 @@ def add_conv_to_msg_options(self, conv, msg_options):
...
+
+class GoRouterMixin(GoWorkerMixin):
+    def get_config_for_router(self, router):
+        config_data = GoApplicationConfigData(self.config, router)
+        return self.CONFIG_CLASS(config_data)
+
+    @inlineCallbacks
+    def get_message_config(self, msg):
+        # By the time we get an event here, the metadata has already been
+        # populated for us from the original message.
```
hodgestar: Might be good to say what has populated the metadata? I assume it's meant to be the routing table router? Perhaps we should raise a suitable error if its not populated? Can the same change be made for GoApplicationMixin?

go/vumitools/app_worker.py
hodgestar, line 504: Cool.

hodgestar:
We need a unit test for the new routing helper function and I left a question about where metadata comes and whether we can optimize things the same way in conversation workers. Looks good otherwise.

hodgestar:
:+1:

PR #525, opened 2013-07-22
Add choice to define whether users can interact multiple times w/ survey

peterp:
On the dialogue (survey) creator screen - when a user picks "Recipients are shown text and the session ends" from the 'Type of message" drop down, we need to add the ability for the user to define whether the end user can interact multiple times with the survey, or just complete the survey once.

I was thinking radio buttons would be a good approach. See screen shot
screen shot 2013-07-19 at 4 26 12 pm

And there is a (small) error in the screen shot. Text for the second radio button should read: "Enable users to complete this dialogue one time only"


badlard:
Here's the latest solution.

screen shot 2013-07-22 at 11 03 14 am

hodgestar:
:+1:

PR #469, opened 2013-07-23
Update management commands for new conversation lifecycle

jerith:


jerith:
This seems to be making some other tests fail (both on my machine and on Travis, but different ones). I'll have to figure out what's going on before it can land.

jerith:
Replaced by #611.

PR #532, opened 2013-07-23
Views for keyword router

jerith:
This is the part of #277 that needs to live in `develop-frontend`.


jerith:
This isn't finished yet, but I'd like a review on the bits I've done so far.

go/base/utils.py
```diff
@@ -131,3 +143,18 @@ def get_conversation_view_definition(conversation_type, conv=None):
                 return ConversationViewDefinitionBase(conv_def)
             return app_pkg.view_definition.ConversationViewDefinition(conv_def)
     raise Exception("Can't find conversation_type: %s" % (conversation_type,))
+
+
+def get_router_view_definition(router_type, router=None):
+    # Scoped import to avoid circular deps.
+    from go.router.view_definition import RouterViewDefinitionBase
+    # HACK: Assume 'namespace' is 'router_type'.
```
hodgestar, line 151: Stll not sure this is a hack. Maybe let's just document them as the same thing? There's probably room to clear this up somehow (i.e. put the name as a constant in the module or something).

hodgestar, line 151: Not for this PR -- just reminding us.

go/base/utils.py
```diff
@@ -131,3 +143,18 @@ def get_conversation_view_definition(conversation_type, conv=None):
...
+    for module, data in settings.VUMI_INSTALLED_ROUTERS.iteritems():
+        if data['namespace'] == router_type:
+            app_pkg = __import__(module,
```
hodgestar: Maybe `router_pkg`?

go/router/templates/router/dashboard.html
```diff
@@ -0,0 +1,51 @@
...
+
+
+{% block content_extraclass %}dashboard routers{% endblock %}
+
+{% block content_title %}All routers{% endblock %}
+
+
+{% block content_actions_right %}
+    <div class="table-form-view-buttons pull-left">
+        <button class="btn" data-action="archive" disabled="disabled">Release</button>
```
hodgestar: s/Release/Archive/ ?

go/router/templates/router/new.html
```diff
@@ -0,0 +1,24 @@
+{% extends "app.html" %}
+
+{% block content_title %}New Router{% endblock %}
+
+{% block content_main %}
+<div class="row">
+    <div class="span12">
+        {% include "base/includes/messages.html" %}
+    </div>
+</div>
```
hodgestar, line 10: @peterp is planning to put messages in a different content block so we don't have to duplicate this everywhere.

hodgestar:
Looking good so far!

hodgestar:
Other than the suggestion to more neatly update the set of endpoints and to perhaps make `default` less special by just putting it in the default static endpoint lists, :+1:.

I have a question about whether we need an option to turn off regexes (and probably also a case-sensitive / not option) but that's for another PR, I think.

jerith:
I'll change the name of the edit fields. We can change the behaviour in develop.

hodgestar:
:+1:

PR #537, opened 2013-07-23
Replace fakeServer with sinon.fakeServer

justinvdm:
Our client side tests currently use our own `fakeServer` instead of `sinon.fakeServer`. This decision was made after testing out `sinon.fakeServer` and finding out that it didn't work well with phantomjs.

I've just tried it out again, and seem to have it working. I'm guessing this change is because I am using newer phantomjs and mocha-phantomjs versions on my new dev machine, but it could well be a result of a mistake I made when I first tried out `sinon.fakeServer`.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #540, opened 2013-07-24
Allow disabling of HTTP api concurrency limit.

smn:
Setting it to less than zero should stop concurrency cap.


jerith:
:+1: when Travis is happy.

PR #543, opened 2013-07-25
Refactor choice state choices into their own view

justinvdm:
Most of the dialogue state view is being taken over by functionality and behaviour for the individual choices. Things will be simplified if these are put into their own view.


peterp:
:+1: once travis is happy!

justinvdm:
Depends on #546

PR #550, opened 2013-07-26
Add transport_type to outbound messages in the routing table dispatcher

jerith:


go/vumitools/routing.py
```diff
@@ -378,6 +378,11 @@ def set_destination(self, msg, target, direction):
                     "Transport name %r found in tagpool metadata for pool"
                     " %r is invalid." % (transport_name, conn.tagpool), msg)
             dst_connector_name = transport_name
+            msg['transport_name'] = transport_name
+
+            transport_type = tagpool_metadata.get('transport_type')
+            if transport_type is not None:
+                msg['transport_type'] = transport_type
```
hodgestar, line 385: Should we not just set `transport_type` to `None` if the tagpool_metadata doesn't contain it? Any transport_type that was set was certainly bogus. Perhaps we should log an error message telling people their tagpool config is borked if transport_type ends up as None?

jerith, line 385: Reply messages will already have the `transport_type` set. I'd rather not break that.

hodgestar, line 385: Hmm. I'm trying to simplify the set of possible cases. The current code will override the reply's transport_type. We already override the transport_name in replies so maybe overriding the transport_type too isn't too bad?

jerith, line 385: We need `transport_name` set in order to route outbound messages. We don't need `transport_type` for anything except the occasional transport that can handle multiple types. I'd rather not require everything to always be configured everywhere if we can avoid it.

hodgestar, line 385: If we're not going to always set it maybe we should just have TaggingMiddleware set it for us? I'd like the routing table dispatcher to guarantee a consistent set of things to transports.

jerith, line 385: We don't really have that consistency to begin with, so we can't really guarantee it without potentially throwing away information. We should probably sit down and figure this stuff out properly at some point, but not in this PR.

hodgestar, line 385: Yes we can -- we can consistently not set transport_name or transport_type and fix the current issue with a one line change to TaggingMiddleware config?

jerith, line 385: We may need different `transport_type` depending on tagpool, which would make the `TaggingMiddleware` config (and code) much messier.

hodgestar, line 385: Not really understanding how that would be the case? If the few cases where we have different tagpools on the same transport we already use multiple TaggingMiddlewares so it's just one transport_type in the msg_template entries for each transport that needs then?

hodgestar:
:+1:

PR #549, opened 2013-07-26
Backport some changes from develop-frontend to develop

jerith:
There are some bits of code (mostly around routers, but elsewhere as well) that have been changed in develop-frontend but not develop. We should backport these changes to develop to avoid merge conflicts.


jerith:
This is probably best done after #532 lands.

hodgestar:
:+1:

PR #524, opened 2013-07-29
Add 'stored value name' and 'searchable' option to Message screens

peterp:
On the dialogue (survey) creator screen we need to add the option for users to select whether the question should appear in search results (i.e. saved as extras in contacts) and we need to add a field where the user defines what the actual answer to the question will be stored as in the DB. 

See screenshot and let me know what questions you may have.
screen shot 2013-07-19 at 4 41 44 pm

### DIFF
--- a/go/base/static/css/conversations.less
+++ b/go/base/static/css/conversations.less
@@ -1,3 +1,14 @@
+.popover {
+  z-index: 20 !important;
+  font-size: 11px;
+  width: 220px;
+
+  input {
+    font-size:11px;
+    width:90%;
+  }
+}
+
 
 .dialogue {
   .actions {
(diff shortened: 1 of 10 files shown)


hodgestar:
Code looks good but I'm getting some odd positioning issues on the new pop overs and hovering information text bubbles. Sometimes hovering doesn't work for me (or perhaps the text bubble appears somewhere I can't see it?) and after creating a second state pop over for the first state appear in the top-left corner of the browser rather than next to the (i) they're supposed to appear next to.

hodgestar:
I'm going to :+1: so we can work on the issues as smaller PRs. Please open tickets before merging though.

PR #475, opened 2013-07-29
Group contacts exporting is not working

peterp:
The original code was written to export a single group at a time.

The new implementation allows you to export multiple groups at a time.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -1,7 +1,7 @@
 import sys
 import traceback
 from StringIO import StringIO
-from zipfile import ZipFile
+from zipfile import ZipFile, ZIP_DEFLATED
 
 from celery.task import task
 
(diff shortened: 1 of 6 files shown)


justinvdm:
Ready for review

go/contacts/tasks.py
```diff
@@ -51,42 +51,33 @@ def delete_group_contacts(account_key, group_key):
...
-    :param str account_key:
-        The account holder's account key
-    :param str group_key:
-        The group to export contacts for (can be either static or smart groups)
-    :param bool include_extra:
-        Whether or not to include the extra data stored in the dynamic field.
-    """
+def zipped_file(filename, data):
+    zipio = StringIO()
+    zf = ZipFile(zipio, "a")
```
hodgestar: We should probably add `compression=ZIP_DEFLATED` here so that the file is actually compressed (not directly related to this PR but it's a small change we've missed before.

hodgestar:
Other than the `ZIP_DEFLATED` suggestion, :+1:.

go/contacts/tasks.py
justinvdm: Ah, good idea

hodgestar:
:+1:

PR #552, opened 2013-07-29
Hook up keyword router to wizard

jerith:


go/wizard/views.py
```diff
@@ -1,89 +1,149 @@
...
+        # Create channel
+        tag = self._create_channel(request, chan_form.cleaned_data)
+
+        # Create conversation
+        view_def, conversation = self._create_conversation(
+            request, conv_form.cleaned_data)
+        messages.info(request, 'Conversation created successfully.')
+
+        # Set up routing
+        logger.info("POST: %s" % (request.POST,))
```
hodgestar: Stray debug log?

go/wizard/views.py
```diff
@@ -1,89 +1,149 @@
...
+        keyword = wiz_form.cleaned_data['keyword']
+        logger.info("keyword: %s" % (keyword,))
```
hodgestar: And here?

go/wizard/views.py
```diff
@@ -1,89 +1,149 @@
...
+        if keyword:
+            # Create router
+            router = self._create_keyword_router(
+                request, keyword, conversation.name)
+            self._setup_keyword_routing(
+                request, conversation, tag, router, keyword)
+        else:
+            self._setup_basic_routing(request, conversation, tag)
```
hodgestar, line 74: Don't we also need to handle the case of adding a keyword to an existing keyword router?

jerith, line 74: Not for the "new channel" case. Using existing keyword routers is out of scope for this branch.

hodgestar, line 74: Do we have a relevant ticket for using existing keyword routers?

hodgestar:
:+1: as soon as we're sure we have a ticket for using existing keyword routers.

jerith:
I think #493 is sufficient for now.

hodgestar:
I added a comment to #493 describing the current status.

:+1:

PR #558, opened 2013-07-30
Fix router directionality

jerith:
We have "INBOUND" meaning one side of a router in some places and the other side in other places. We need to standardise on "INBOUND" mapping to a `ReceiveInboundConnector` on the router.


jerith:
Depends on #559.

hodgestar:
:+1:

PR #560, opened 2013-07-30
Duplicate id when recreating deleted routing connection

justinvdm:
On the routing screen, it's possible to remove a connection between two endpoints and then create a new connection with the same source and destination. This causes a "Duplicate id" warning to be displayed on the JS console, but the new connection is displayed on the page. When the routing table is saved, however, the new connection is not included.


justinvdm:
Ready for review

hodgestar:
:+1:

PR #563, opened 2013-07-31
Router API functionality

jerith:
This is the precursor (in `develop`) to #562 (in `develop-frontend`).

### DIFF
--- a/go/vumitools/account/models.py
+++ b/go/vumitools/account/models.py
@@ -168,6 +168,17 @@ class RoutingTableHelper(object):
                                                      conv.key))
         self.remove_connector(conv_conn)
 
+    def remove_router(self, router):
+        """Remove all entries linking to or from a given router.
+
+        Useful when archiving a router to ensure it is no longer present in the
+        routing table.
+        """
+        self.remove_connector(str(GoConnector.for_router(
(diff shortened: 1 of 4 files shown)


go/vumitools/account/models.py
```diff
@@ -168,6 +168,17 @@ def remove_conversation(self, conv):
...
+    def remove_router(self, router):
+        """Remove all entries linking to or from a given router.
+
+        Useful when archiving a router to ensure it is no longer present in the
+        routing table.
+        """
+        self.remove_connector(str(GoConnector.for_router(
+            router.router_type, router.key, GoConnector.INBOUND)))
+        self.remove_connector(str(GoConnector.for_router(
+            router.router_type, router.key, GoConnector.OUTBOUND)))
```
hodgestar, line 180: Could we get a unit test for this? Doesn't have to be too extensive since it just calls remove_connector.

hodgestar:
:+1: after a test for the remove_router helper method is added.

go/vumitools/account/models.py
jerith, line 180: Whoops, I thought I was missing a test somewhere.

hodgestar:
:+1:

PR #562, opened 2013-07-31
Remove routing entries when archiving router

jerith:


jerith:
I didn't add a test for actually cleaning up the routing table, but I did test it manually. I'd really like to get #564 done before mucking about with routing tables any more.

hodgestar:
Would really like this to have a test. Is writing a test for it particularly difficult?

jerith:
It's about ten lines of ugly boilerplate that makes the test less readable. We already have a test around `archive_router()`, so we're pretty safe.

hodgestar:
Not ideal but I'm :+1: on this. The method itself is tested -- it's just the results that aren't completely asserted on.

PR #533, opened 2013-07-31
Implement new Javascript sandbox based polls.

hodgestar:
This will be a separate Vumi application worker that builds on top of the generic sandbox and supplies a static Javascript program capable of executing a JSON poll description as outlined in #335.

### DIFF
--- a/go/apps/dialogue/tests/dummy_polls.py
+++ b/go/apps/dialogue/tests/dummy_polls.py
@@ -1,0 +1,63 @@
+simple_poll = {
+    "conversation": "conversation-key",
+    "start_state": {"uuid": "choice-1"},
+    "poll_metadata": {
+        "repeatable": True,
+    },
+    "states": [
+        {
+            # these are common to all state types
+            "uuid": "choice-1",  # name is unique
+            "name": "Message 1",  # a friendly name for the user to see
+            "store_as": "message-1",
+            "type": "choice",  # menu of options
+            "entry_endpoint": None,  # null for the start state
+            "store_on_contact": True,
+            # choice specific
+            "text": "What is your favourite colour?",
+            "choice_endpoints": [  # these are actually also the endpoints
+                {"value": "value-1", "label": "Red", "uuid": "endpoint-1"},
+                {"value": "value-2", "label": "Blue", "uuid": "endpoint-2"}
+            ],
+        },
+        {
+            "uuid": "freetext-1",
+            "name": "Message 2",
+            "store_as": "message-2",
+            "type": "freetext",
+            "store_on_contact": True,
+            "entry_endpoint": "endpoint-3",
+            # freetext specific
+            "exit_endpoint": "endpoint-4",
+            "text": "What is your name?",
+        },
+        {
+            "uuid": "end-1",
+            "name": "Ending 1",
+            "store_as": "ending-1",
+            "type": "end",
+            "store_on_contact": True,
+            "entry_endpoint": "endpoint-5",
+            # end specific
+            "text": "Thank you for taking our survey",
+        },
+        {
+            "uuid": "unknown-1",
+            "type": "unknown-or-garbled",
+        },
+    ],
+    "connections": [
+        {
+            "source": {"uuid": "endpoint-1"},
+            "target": {"uuid": "endpoint-3"},
+        },
+        {
+            "source": {"uuid": "endpoint-2"},
+            "target": {"uuid": "endpoint-5"}
+        },
+        {
+            "source": {"uuid": "endpoint-4"},
+            "target": {"uuid": "endpoint-5"}
+        },
+    ],
+}
--- a/go/apps/dialogue/tests/test_vumi_app.py
+++ b/go/apps/dialogue/tests/test_vumi_app.py
@@ -39,7 +39,7 @@ class DialogueApplicationTestCase(AppWorkerTestCase):
             ),
             'sandbox': {
                 'config': {
-                    'cls': 'go.apps.jsbox.vumi_app.ConversationConfigResource'
+                    'cls': 'go.apps.dialogue.vumi_app.PollConfigResource',
                 },
                 'kv': {
                     'cls': 'vumi.application.sandbox.RedisResource',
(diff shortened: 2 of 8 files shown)


go/apps/dialogue/tests/dummy_polls.py
```diff
@@ -0,0 +1,63 @@
+simple_poll = {
```
justinvdm, line 1: This isn't too important, but I think having the dummy poll data in two places will mean having to update two files, and there is a chance that the files will diverge by mistake. Would it maybe be better to include the sample data as a JSON file, then `json.loads()` it for python and `require()` it for js?

hodgestar, line 1: I think my intuition here was that I wanted the JS and Python tests to be separate from each other. Could probably be convinced that combining them is a good idea though.

justinvdm, line 1: My only worry is that it will make maintaining things a bit difficult (say, example, if we added/removed/changed one of the fields), since the sample data is in two places. If you don't feel its too important, I'm okay with us leaving it as is. We can always combine them at a later stage.

hodgestar, line 1: Cool. Then my vote is we leave them separate and see how things go. My guess is that the JS one will grow and get lots of new types of states and the Python one will not really change (recording guess here so I can see how I score on predicting the future :).

go/apps/dialogue/vumi_app.js
```diff
@@ -0,0 +1,139 @@
+var vumigo = require("vumigo_v01");
+var jed = require("jed");
+
+if (typeof api === "undefined") {
+    // testing hook (supplies api when it is not passed in by the real sandbox)
+    var api = this.api = new vumigo.dummy_api.DummyApi();
+}
```
justinvdm, line 7: This isn't for this PR, but just something we should maybe make note of. Couldn't we just give mocha a test helper module using `mocha --require`, and simply put `var api = this.api = new vumigo.dummy_api.DummyApi()` in there?

hodgestar, line 7: Interesting suggestion. I was thinking of making the exist test hooks duplicate what the sandbox does a little more (i.e. actually load the app inside a separate JS context to which we add a dummy API).

hodgestar, line 7: I created praekelt/vumi-jssandbox-toolkit#23 for this.

go/apps/dialogue/vumi_app.py
```diff
@@ -1,120 +1,97 @@
...
-        })
-        yield super(DialogueApplication, self).end_session(participant, poll,
-            message)
+        # TODO: This generates a fake message id that is then used in
+        #       the reply to field of the outbound message. We need to
+        #       write special version of the GoOutboundResource that
+        #       will set in_reply_to to None on these messages so the
+        #       invalid ids don't escape into the rest of the system.
+        msg = TransportUserMessage(from_addr=to_addr, to_addr=None,
+                                   content=None, **msg_options)
```
justinvdm, line 94: I'm happy with this staying as is, but just wondering if there is some way we could identify these 'fake' messages when looking at logs or sentry. Maybe setting the content to '[start of dialogue]' or something?

hodgestar, line 94: I was having similar ideas. We can't set `content` to anything other than `None` because that will cause the sandbox to try treat it as user input. We could set some `helper_metadata` though.

hodgestar, line 94: Hmm. Actually, helper_metadata won't work because that appears in replies. I could add another random field though.

justinvdm, line 94: Hmm, I see. I'm +1 on adding additional field in order to identify these message in the logs, your call though.

go/apps/dialogue/vumi_app.js
justinvdm, line 7: Thanks! Cool, having the test infrastructure mimic what the actual sandbox does as closely as possible is probably a good thing, so +1 on that approach.

go/apps/dialogue/tests/test_vumi_app.py
```diff
@@ -1,393 +1,133 @@
...
-        yield self.dispatch_to_conv(msg, self.conversation)
-        [reply] = yield self.wait_for_dispatched_messages(1)
-        self.assertTrue('Service Unavailable' in reply['content'])
-        self.assertEqual(reply['session_event'],
-                         TransportUserMessage.SESSION_CLOSE)
+        conversation.set_go_helper_metadata(msg['helper_metadata'])
+        yield self.store_outbound_msg(msg, conversation)
+        event = self.mkmsg_ack(user_message_id=msg['message_id'])
+        conversation.set_go_helper_metadata(event['helper_metadata'])
+        yield self.dispatch_event(event)
```
justinvdm: Could you explain where the assertion is being done in this test?

go/apps/dialogue/tests/test_vumi_app.py
```diff
@@ -1,393 +1,133 @@
...
+
+    @inlineCallbacks
+    def test_user_message(self):
+        conversation = yield self.setup_conversation()
+        yield self.start_conversation(conversation)
+        msg = self.mkmsg_in()
+        yield self.dispatch_to_conv(msg, conversation)
+        [reply] = yield self.get_dispatched_outbound()
+        self.assertEqual(reply["content"],
+                         "What is your favourite colour?\n1. Red\n2. Blue")
```
justinvdm, line 122: This test asserts that we reply to the first user message correctly. I'm wondering whether it would help to add another test to check that we respond to user input correctly with the appropriate next state for subsequent messages. It does feel like it is testing what the js app tests are already testing though, so I'm not sure on this one.

hodgestar, line 122: This is really just testing that the message makes it in to the Javascript app and that a response comes out. I think tests of state logic can all live in `test_vumi_app.js`.

go/apps/dialogue/tests/test_vumi_app.py
hodgestar: It's not. I should fix that. The question is what to assert? Hmm.

go/apps/dialogue/tests/test_vumi_app.py
justinvdm, line 122: Cool, sounds reasonable.

justinvdm:
:+1:

PR #493, opened 2013-07-31
Connect up "select an existing channel" and keyword routing block widgets in wizard

jerith:
Currently these are just dummy items. This probably needs #277 to be finished first.

### DIFF
--- a/go/wizard/templates/wizard_views/wizard_1_create.html
+++ b/go/wizard/templates/wizard_views/wizard_1_create.html
@@ -87,7 +87,7 @@
             <div class="span4">
                 <div class="box">
                     <label for="id_{{ wizard_form.new_keyword.name }}">
-                      {{ wizard_form.bew_keyword.label }}
+                      {{ wizard_form.new_keyword.label }}
                     </label>
                     {{ wizard_form.new_keyword }}
                 </div>
--- a/go/wizard/templates/wizard_views/wizard_1_create.html
+++ b/go/wizard/templates/wizard_views/wizard_1_create.html
@@ -72,24 +72,23 @@
             </div>
         </fieldset>
 
-        {# TODO: This form needs to be built in Django land. #}
         <fieldset class="configuration_existing row-fluid boxes">
             <div class="span4">
                 <div class="box">
-                    <label for="id_{{ wizard_form.existing_router.name }}">
-                        {{ wizard_form.existing_router.label }}
+                    <label for="id_{{ router_form.existing_router.name }}">
+                        {{ router_form.existing_router.label }}
                     </label>
 
-                    {{ wizard_form.existing_router }}
+                    {{ router_form.existing_router }}
                 </div>
             </div>
 
             <div class="span4">
                 <div class="box">
-                    <label for="id_{{ wizard_form.new_keyword.name }}">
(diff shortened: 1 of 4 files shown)


hodgestar:
The remaining work is handling the case of adding keywords to existing keyword routers. To do this we need to:
- Retrieve a list of keyword routers and their current keywords to display.
- Display the list and allow the user to select a keyword router and enter a new keyword (current keywords should be displayed nearby).
- Create the new conversation, add the new keyword to the existing router and update the routing table entries.

jerith:
The "existing keywords" still needs to be hooked up (maybe @peterp or @justinvdm could help with this?) and validation isn't as good as I'd like. Ready for initial review, though.

go/wizard/templates/wizard_views/wizard_1_create.html
```diff
@@ -77,23 +76,20 @@
...
+                    {{ wizard_form.existing_router }}
                 </div>
             </div>
 
             <div class="span4">
                 <div class="box">
-                    <label for="id_keyword2">
-                      Define a new keyword
+                    <label for="id_{{ wizard_form.new_keyword.name }}">
+                      {{ wizard_form.bew_keyword.label }}
```
hodgestar: s/bew/new/ ?

go/wizard/views.py
```diff
@@ -37,56 +37,81 @@ class WizardCreateView(BaseWizardView):
...
-                'show', conversation_key=conversation.key))
+    def _handle_existing_channel(self, request, router_key, keyword, conv):
+        # TODO: Check that keyword is unused.
+        router = request.user_api.get_router(router_key)
+        endpoint = 'keyword_%s' % (keyword.lower(),)
+        # TODO: Better way to manage this kind of thing.
+        kem = router.config.setdefault('keyword_endpoint_mapping', {})
+        kem[keyword] = endpoint
+        if endpoint not in router.extra_outbound_endpoints:
+            router.extra_outbound_endpoints.append(endpoint)
```
hodgestar, line 112: Would this allow `default` to be added to the extra_outbound_endpoints?

go/wizard/views.py
```diff
@@ -37,56 +37,81 @@ class WizardCreateView(BaseWizardView):
...
-        pool, tag = channel_data['channel'].split(':')
-        if tag:
-            return request.user_api.acquire_specific_tag((pool, tag))
-        else:
-            return request.user_api.acquire_tag(pool)
+        self._setup_keyword_routing(request, conv, None, router, endpoint)
+
+    def _get_existing_keywords(self, request):
+        routers = request.user_api.active_routers()
+        return routers
```
hodgestar: I think this function is unused now?

hodgestar:
:+1: on landing this as soon as the the two remaining minor comments have been addressed. We can create new tickets for improving the validation and displaying the existing keywords.

go/wizard/views.py
jerith, line 112: Good catch.

jerith, line 112: Actually, it won't. We add `keyword_default` as the endpoint name.

hodgestar:
:+1:

PR #335, opened 2013-08-01
New polls saving and loading functions to the Go API.

hodgestar:
The new polls will directly support the JSON format below. Initially three kinds of questions will be supported -- choice states, free text states and end states.

Example JSON:

``` javascript
var poll = {
  conversation: "conversation-key",
  start_state: {uuid: "state-1"},
  poll_metadata: {
    repeatable: true,
  },
  states: [
    {
        // these are common to all state types
        uuid: "state-1", // name is unique
        name: "Message 1", // a friendly name for the user to see
        store_as: "message-1",
        type: "choice", // menu of options
        entry_endpoint: null, // null for the start state
        store_on_contact: true,
        // choice specific
        text: "What is your favourite colour?",
        choice_endpoints: [ // these are actually also the endpoints
          {value: "value-1", label: "Red", uuid: "endpoint-1"},
          {value: "value-2", label: "Blue", uuid: "endpoint-2"}
        ]
    },
    {
        uuid: "state-2",
        name: "Message 2",
        store_as: "message-2",
        type: "freetext",
        store_on_contact: true,
        entry_endpoint: "endpoint-3",
        // freetext specific
        exit_endpoint: "endpoint-4",
        text: "What is your name?"
    },
    {
       uuid: "state-3",
       name: "Ending 1",
       store_as: "ending-1",
       type: "end",
       store_on_contact: true,
       entry_endpoint: "endpoint-5",
       // end specific
       text: "Thank you for taking our survey"
    }
  ],
  connections: [
     {
       source: {uuid: "endpoint-1"},
       target: {uuid: "endpoint-2"}
     }
  ]
};

```

### DIFF
--- a/go/api/go_api/action_dispatcher.py
+++ b/go/api/go_api/action_dispatcher.py
@@ -1,10 +1,15 @@
 """Conversation, router and channel action dispatch handlers for Go API."""
 
+from txjsonrpc.jsonrpclib import Fault
+
 from vumi import log
 
 
-class ActionError(Exception):
+class ActionError(Fault):
     """Raise this to report an error from within an action handler."""
+
+    def __init__(self, msg, fault_code=400):
+        super(ActionError, self).__init__(fault_code, msg)
 
 
 class ActionDispatcher(object):
--- a/go/api/go_api/action_dispatcher.py
+++ b/go/api/go_api/action_dispatcher.py
@@ -2,6 +2,8 @@
 
 from txjsonrpc.jsonrpclib import Fault
 
+from twisted.internet.defer import maybeDeferred
+
 from vumi import log
 
 
@@ -22,22 +24,29 @@ class ActionDispatcher(object):
     def __init__(self, user_api):
         self.user_api = user_api
 
+    def _log_error(self, err, obj, action, params):
+        log.err(err,
+            "Action %(action)r on %(type_name)r %(obj)r (key: %(key)r)"
+            " with params %(params)r failed." % {
+                "action": action,
+                "type_name": self.dispatcher_type_name,
+                "obj": obj,
+                "key": obj.key,
+                "params": params,
+                })
+        return err
+
+    def _log_success(self, result, obj, action, params):
+        log.info("Performed action %r on %s %r." % (
+            action, self.dispatcher_type_name, obj.key))
+        return result
+
     def dispatch_action(self, obj, action, params):
         handler = getattr(self, "handle_%s" % action, self.unknown_action)
-        try:
-            return handler(obj, **params)
-        except:
-            log.err(None,
-                "Action %(action)r on"
-                " %(type_name)r %(obj)r (key: %(key)r)"
-                " with params %(params)r failed." % {
-                    "action": action,
-                    "type_name": self.dispatcher_type_name,
-                    "obj": obj,
-                    "key": obj.key,
-                    "params": params,
-                    })
-            raise
+        d = maybeDeferred(handler, obj, **params)
+        d.addErrback(self._log_error, obj, action, params)
+        d.addCallback(self._log_success, obj, action, params)
+        return d
 
     def unknown_action(self, obj, **kw):
         raise ActionError("Unknown action.")
--- a/go/api/go_api/action_dispatcher.py
+++ b/go/api/go_api/action_dispatcher.py
@@ -26,7 +26,7 @@ class ActionDispatcher(object):
 
     def _log_error(self, err, obj, action, params):
         log.err(err,
-            "Action %(action)r on %(type_name)r %(obj)r (key: %(key)r)"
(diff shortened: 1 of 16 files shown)


go/api/go_api/action_dispatcher.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def dispatch_action(self, obj, action, params):
+        handler = getattr(self, "handle_%s" % action, self.unknown_action)
+        d = maybeDeferred(handler, obj, **params)
+        d.addErrback(self._log_error, obj, action, params)
+        d.addCallback(self._log_success, obj, action, params)
+        return d
+
+    def unknown_action(self, obj, **kw):
+        raise ActionError("Unknown action.")
```
justinvdm: I'm wondering if it would be helpful to pass in the action into `unknown_action` and add that to the error message.

justinvdm: Oh right, the errback would display the action, sorry.

go/api/go_api/tests/test_action_dispatcher.py
```diff
@@ -0,0 +1,74 @@
...
+    @inlineCallbacks
+    def test_dispatch_action(self):
+        user_api = Mock(do_thing=Mock())
+        dispatcher = SimpleActionDispatcher(user_api)
+        obj = Mock(key="abc")
+        with LogCatcher() as lc:
+            result = yield dispatcher.dispatch_action(
+                obj, "do_thing", {"foo": "bar"})
+            [msg] = lc.messages()
+            self.assertEqual(msg, "Performed action 'do_thing' on None 'abc'.")
```
justinvdm: This really isn't too important, but would it maybe help to give `SimpleActionDispatcher` a `dispatcher_type_name` so we can ensure that `dispatcher_type_name` is being used correctly in the logged message, and that `None` is not appearing in the message for another reason?

go/apps/dialogue/templates/dialogue/edit.html
```diff
@@ -46,8 +46,12 @@
...
       // button.
 
-      // TODO uncomment once the screen is hooked up to the api
-      //diagram.model.save();
+      diagram.model.save({}, {
+        sessionId: "{{ session_id|safe }}",
+        error: function() {
+          bootbox.alert("Something bad happened, changes couldn't be saved.");
+        }
+      });
```
justinvdm, line 54: I think this will work once we have the api methods hooked up to the client side dialogue model.

go/apps/dialogue/view_definition.py
```diff
@@ -14,10 +15,23 @@ class DialogueEditView(ConversationTemplateView):
     template_base = 'dialogue'
 
     def get(self, request, conversation):
-        # TODO Use api to get model data and bootstrap it to page load
+        r = client.rpc(
+            request.session.session_key, 'conversation_action',
+            [request.user_api.user_account_key,
+             conversation.key, 'get_poll'])
```
justinvdm: I don't know how I feel about having part of the rpc method name as one of the rpc params. I think this is more the job of namespacing the methods (somethink like `conversation/get` or `conversation/get_poll`). If you are strongly for this approach that's okay, but at the very least we will need to change the way we support rpc calls on the client side for this to work.

go/apps/dialogue/view_definition.py
```diff
@@ -14,10 +15,23 @@ class DialogueEditView(ConversationTemplateView):
...
+
+        if r.status_code != 200:
+            raise GoApiError(
+                "Failed to load routing table from Go API:"
+                " (%r) %r." % (r.status_code, r.text))
+
+        model_data = {'campaign_id': request.user_api.user_account_key}
```
justinvdm: I think we will need to add `conversation.key` to the model data since it is needed for the dialogue api's rpc methods.

justinvdm:
Looks good. I'm a bit worried about having part of the rpc method as once of the rpc params, as I commented on [here](https://github.com/praekelt/vumi-go/pull/335#discussion_r5536604). Will re-review once we've come to a consensus on that.

go/apps/dialogue/view_definition.py
hodgestar: I thought about doing it the way you suggest but I steered away from it initially because there are some complications -- once needs to introspect all possible conversation and router definitions up front and construct txjsonrpc server objects for them. That said, now that I've built all the infrastructure I have a better idea of how to achieve this so I'll give it a try.

justinvdm: Ah, I see. Thanks, let me know how it goes. If its starting to feel contrived, we can try get the client side to work with the current solution.

hodgestar:
Ready for review again although the model saving still needs hooking up (that can be a separate ticket if needed).

justinvdm:
:+1: I now you see what you meant by 'scariness' :) In my opinion, I think it is well worth it, we now have a really nice looking API!

PR #574, opened 2013-08-02
Change client side module structure to match the current structure of server side modules

justinvdm:
The client side modules were initially structured with the design prototypes' structuring in mind. We have moved far away from the prototypes, yet our structuring of the client side remains the same. This makes placing new modules in the current structure a headache, and needs to be changed.


go/base/static/js/test/tests/apps/dialogue/testHelpers.js
```diff
@@ -1,8 +1,8 @@
-// go.campaign.routing.testHelpers
+// go.routing.testHelpers
 // ===============================
```
hodgestar: Row of equals is too long now.

go/base/static/js/test/tests/routing/testHelpers.js
```diff
@@ -1,8 +1,8 @@
-// go.campaign.routing.testHelpers
+// go.routing.testHelpers
 // ===============================
```
hodgestar: And here.

hodgestar:
Other than the tiny nitpicks, :+1:. :)

PR #570, opened 2013-08-02
Fix highlighting of selected dashboard screen.

peterp:
Currently on QA `All campaigns`, `Activated`, `Deactivated`, `Archived`, `All channels` and `All routers` are not highlighted by a red background when they're selected but I think they should be.


hodgestar:
:+1:

PR #568, opened 2013-08-02
Check that user has permission to access campaign in Go API

hodgestar:
Currently any user who can login to the Go API can access any campaign id. This is a giant security hole.


hodgestar:
Ready for review but will be easier to look at once #335 has landed.

justinvdm:
:+1:

PR #575, opened 2013-08-02
ConfirmView doens't unbind its events correctly

justinvdm:
`Backbone.View.prototype.remove()` invokes `.stopListening()` and removes its element, but doesn't invoke it's `.off()` to unbind events bound using `.on()`. This is sensible behaviour, since listeners should be using `listenTo()`, and `.on()` should be used internally by the view instance itself.

Currently, we use `.on()` for `ConfirmView` events, since this allows us to unbind both `cancel` and `ok` events after one of them is triggered. We either need to find a way to use `listenTo()`, and still only have `cancel` and `ok` events behave as they currently do (the preferable approach), or have `ConfirmView.prototype.remove()` invoke its `.off()`.

### DIFF
--- a/go/base/static/js/src/components/views.js
+++ b/go/base/static/js/src/components/views.js
@@ -116,12 +116,18 @@
       return this;
     },
 
+    remove: function() {
+      this.off();
+      return ConfirmView.__super__.remove.call(this);
+    },
+
     render: function() {
       this.$el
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review

hodgestar:
:+1:

PR #526, opened 2013-08-02
Header buttons on dashboard view don't do anything

justinvdm:
On the conversations page: http://qa-vumi.za.prk-host.net/conversations/
and the channels page: http://qa-vumi.za.prk-host.net/channels/
the buttons in the header don't complete the action when clicked.

See screenshots. 

Note that there is already a ticket open for enabling the "copy" button.

screen shot 2013-07-19 at 5 26 11 pm
screen shot 2013-07-19 at 5 30 54 pm


justinvdm:
Ready for review

justinvdm:
Depends on #575

go/base/static/js/src/components/tables.js
```diff
@@ -3,129 +3,116 @@
...
+        // Follow the link associated with the row
+        if (typeof url !== 'undefined') { window.location = url; }
+      },
+
+      'click tbody tr td *': function(e) {
+        // Prevent the column's click events from propagating to its elements
+        e.stopPropagation();
+      }
+    }
+  });
```
justinvdm, line 113: I decided to refactor `TableFormView`, since I thought I would be needing it for the inline actions in the dashboard screen, After more looking around and quite a bit of discussion on IRC, I realised that the inline buttons wouldn't be that simple to support inside `TableFormView`, so I went on to add `go.components.actions` and its friends. I decided to keep the refactoring and cleaning up of `TableFormView`, even though it isn't anymore necessary for this PR.

go/conversation/templates/conversation/includes/actions.html
```diff
@@ -0,0 +1,33 @@
...
+    </button>
+{% else %}
+    {# TODO: Figure out what to put here. #}
+    <button
+     class="btn btn-primary action" 
+     data-action='stop'
+     data-url="{% conversation_screen conversation 'stop' %}">
+     In transition: attempt to stop
+    </button>
+{% endif %}
```
hodgestar, line 40: The logic here probably need to be a broken up a little more. I think each action needs its own `if` block that checks specifically whether that action should be shown. You can view conversation states as a sub-set of the cross-product of `starting, running, stopping, stopped` and `active, archived`. `stopped` is the only running state that may be paired with `archive`. So
- You can activate a conversation if it is stopped and not archived.
- You can archive a conversation if it is stopped and not archived.
- You can stop a conversation if it is not stopped.

I think that covers all the cases.

hodgestar:
Other than the comment about which states to display actions in, :+1:

justinvdm:
Ready for re-review

go/base/static/js/src/components/tables.js
```diff
@@ -3,129 +3,116 @@
...
+        if (submit) { $form.submit(); }
+        $input.remove();
+      });
+    },
+
+    onAction: function(e) {
+      this.showConfirmationModal({action: $(e.target).attr('data-action')});
+    },
+
+    events: {
```
peterp, line 81: Events are usually a pretty good _table of contents_ for view based code so I usually like to place them at the top of the file.

peterp, line 81: Agh. I see why you placed it here.

go/base/static/js/src/components/tables.js
```diff
@@ -3,129 +3,116 @@
...
+        this.$headActionMarker().prop('checked', $(e.target).prop('checked'));
+        this.refreshButtons();
+      },
+
+      'change td:first-child input': function(e) {
+        this.$headActionMarker().prop('checked', this.allChecked());
+        this.refreshButtons();
+      },
+
+      'click td:first-child': function(e) {
```
peterp, line 94: There was a pretty decent comment that clarified why this was here originally. It's not immediately obvious when looking at this code.

go/base/static/js/src/components/tables.js
peterp, line 113: Goodbye dear code :/

go/base/static/js/src/components/tables.js
justinvdm, line 81: I used to thinks so too, but realised that things start looking like this:

``` javascript
Backbone.View.extend({
  events: {
    'click .thing-a': 'onThingAClick',
    'click .thing-a': 'onThingAChange',
    'click .thing-b': 'onThingBClick',
    'click .thing-b': 'onThingBChange',
    'click .thing-c': 'onThingCClick',
    'click .thing-d': 'onThingDClick',
  },

  ...

  onThingAClick: function() {
    ...
  },

  onThingAChange: function() {
    ...
  },

  onThingBClick: function() {
    ...
  },

  onThingBChange: function() {
    ...
  },

  onThingCClick: function() {
    ...
  },

  onThingDClick: function() {
    ...
  }
});
```

Compare that to this:

``` javascript
Backbone.View.extend({
  events: {
    'click .thing-a': function() {
      ...
    },

    'click .thing-a': function() {
      ...
    },

    'click .thing-b': function() {
      ...
    },

    'click .thing-b': function() {
      ...
    },

    'click .thing-c': function() {
      ...
    },

    'click .thing-d': function() {
      ...
    },
  }
});
```

In other words, the 'table of contents' starts to become a redundant piece of code and adds bloat to the view. In my opinion, using `click .thing-a` is simpler and more expressive than `onThingAClick`. I don't think we have the time for coming to a consensus on code style at this point though, so I'll change it to the 'table of contents' style.

justinvdm, line 81: Based on IRC discussion, we've decided to leave it as is.

hodgestar:
:+1:

PR #576, opened 2013-08-03
Hook up dialogue saving in Javascript models.

justinvdm:
Everything else should have been covered by #335 so this is hopefully the last little bit of hooking up to do for dialogue saving and loading.


hodgestar:
:+1:

PR #571, opened 2013-08-04
Clicking on edit next to details on the conversation show screen just returns to the show screen.

smn:
We need it to go to a view for editing details.


smn:
this one?

vumi_go

hodgestar:
Yep.

smn:
I'm guessing that having a template-level option of seeing whether a given conversation is editable would be better long term.

smn:
My thinking is that that can possibly be addressed better when all views have view_definitions, otherwise we're layering too many hacks.

go/conversation/templates/conversation/edit_detail.html
```diff
@@ -0,0 +1,34 @@
...
+{% endblock %}
+
+{% block extramedia %}
+    {{ edit_form_media }}
+{% endblock %}
+
+{% block ondomready %}
+    $('.content .actions .right button').click(function() {
+        $('#form-conversation').submit();
+    });
```
hodgestar, line 33: Maybe it would be better for the button to have an id rather than being identified by it's position within the page hierarchy?

hodgestar:
:+1: other than the suggestion about the id. Some tests would also be good but I'm not going to let that stop the branch landing at this point.

go/conversation/templates/conversation/edit_detail.html
smn, line 33: True but this comes straight from the existing [edit.html template](https://github.com/praekelt/vumi-go/blob/537c0b3e4af4903e189f079a16179f41c28ae18d/go/conversation/templates/conversation/edit_detail.html#L31-L33) and possibly it'd be better to address that in a separate PR? Not sure.

hodgestar, line 33: Happy for it to be addressed separately or left for now. The person who adds a cancel button can fix it.

PR #579, opened 2013-08-04
Fix routing and dialogue edit screens

justinvdm:


justinvdm:
Ready for review

go/base/management/commands/go_setup_env.py
```diff
@@ -406,6 +406,10 @@ def write_supervisor_config_file(self, program_name, worker_class,
             fp.write(self.auto_gen_warning)
             cp = ConfigParser()
             cp.add_section(section)
+            cp.set(
+                section,
+                "environment",
+                "DJANGO_SETTINGS_MODULE=go.settings")
```
justinvdm, line 412: The api worker needs this. Happy to do this an alternative way.

go/api/go_api/action_dispatcher.py
```diff
@@ -51,7 +51,7 @@ def mk_jsonrpc_handler(name, handler, type_name):
 
         # we use compile and exec to get exactly the signature we need
         code = compile((
-            "def %s(self, campaign_key, %s_key, params):"
+            "def %s(self, campaign_key, %s_key, params={}):"
```
hodgestar: Having mutable default arguments is a bad idea (since there's only ever one copy but `params['foo'] = 'bar'` will accidentally modify it). We should set this to `None` here and then check for `None` were we dispatch to the action handler with `**params` in `dispatch_action`.

go/base/management/commands/go_setup_env.py
hodgestar, line 412: This looks good to me.

hodgestar:
Other than the mutable argument default thing, :+1:.

hodgestar:
:+1:

PR #557, opened 2013-08-04
Better handling of invalid forms

smn:
There are a bunch of places in the UI where we do horrible things in the face of invalid form input. We need to track these down (they mostly have relevant TODO comments) and handle the validation errors better.


hodgestar:
:+1:

PR #585, opened 2013-08-05
Remove migration from pre-routing-table conversations

jerith:
We no longer have old pre-routing-table conversations anywhere and we no longer support old-style routing anyway. Keeping that cruft around is making #581 unnecessarily hard.


hodgestar:
:+1:

smn:
\o/ for removing old hacks

PR #586, opened 2013-08-05
Keep track of dialogue start state on client side

justinvdm:


hodgestar:
After discussion with @justinvdm we still need to:
- Reset the `start_state` when states are added or removed (added is not needed right now but probably doesn't hurt).
- Fix `vumi_app.js` to support start state being `null` (for when there are no states) and to look up the start state name from the `uuid` attribute inside `start_state`.

hodgestar:
We're going to handle the case where the start state is `null` or invalid in a separate PR.

hodgestar:
:+1:

hodgestar:
Second :+1:

Issue #278, opened 2013-05-17
Add per-channel, per-routing-block and per-conversation message batches.


hodgestar:
This will allow us to track which messages reach / flow through each channel, routing-block and conversation in Vumi Go. It requires extending the message store to allow messages to belong to multiple batches and de-coupling batch-id creation from tag acquisition.

hodgestar:
We can leave the per-channel part of this out of this issue and open a new one for it in 0.5b2 if that makes life easier. Channels are a bit trickier because they are currently maintained in tagpools only and need to be migrate to also have Riak models.

hodgestar:
Left to do:
- [x] New middleware for storing messages received or sent by  router and conversation workers on the correct batch (see #599).
- [x] Migration script to give active conversations that are using their tags batch their own batch instead (see #601).

hodgestar:
All done.

PR #581, opened 2013-08-06
Set conversation and router batches at creation time

jerith:
This is necessary for #278.

### DIFF
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -260,8 +260,15 @@ class VumiUserApi(object):
             conversation_type, name, description, config, batch_id, **fields)
         returnValue(conv)
 
-    def new_router(self, *args, **kw):
-        return self.router_store.new_router(*args, **kw)
+    @Manager.calls_manager
+    def new_router(self, router_type, name, description, config,
+                   batch_id=None, **fields):
+        if not batch_id:
+            batch_id = yield self.api.mdb.batch_start(
+                tags=[], user_account=self.user_account_key)
+        router = yield self.router_store.new_router(
+            router_type, name, description, config, batch_id, **fields)
+        returnValue(router)
 
     @Manager.calls_manager
     def list_conversation_endpoints(self):
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -373,6 +373,7 @@ class VumiUserApi(object):
 
     @Manager.calls_manager
     def _update_tag_data_for_acquire(self, user_account, tag):
+        yield self.api.mdb.batch_start([tag], user_account=user_account.key)
         user_account.tags.append(tag)
         tag_info = yield self.api.mdb.get_tag_info(tag)
         tag_info.metadata['user_account'] = user_account.key.decode('utf-8')
@@ -448,6 +449,11 @@ class VumiUserApi(object):
             tag_info = yield self.api.mdb.get_tag_info(tag)
             del tag_info.metadata['user_account']
             yield tag_info.save()
+            # NOTE: This loads and saves the CurrentTag object a second time.
+            #       We should probably refactor the message store to make this
+            #       less clumsy.
+            if tag_info.current_batch.key:
+                yield self.api.mdb.batch_done(tag_info.current_batch.key)
 
             # Clean up routing table entries.
             routing_table = yield self.get_routing_table(user_account)
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -605,35 +605,6 @@ class VumiApi(object):
     def get_user_api(self, user_account_key):
         return VumiUserApi(self, user_account_key)
 
-    def batch_start(self, tags):
-        """Start a message batch.
-
-        :type tags: list of str
-        :param tags:
-            A list of identifiers for linking replies to this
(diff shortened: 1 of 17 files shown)


go/vumitools/api.py
```diff
@@ -441,6 +456,11 @@ def release_tag(self, tag):
             tag_info = yield self.api.mdb.get_tag_info(tag)
             del tag_info.metadata['user_account']
             yield tag_info.save()
+            # NOTE: This loads and saves the CurrentTag object a second time.
+            #       We should probably refactor the message store to make this
+            #       less clumsy.
+            if tag_info.current_batch.key:
+                yield self.api.mdb.batch_done(tag_info.current_batch.key)
```
hodgestar, line 466: Is this likely to cause eventual consistency problems?

go/vumitools/api.py
```diff
@@ -366,6 +380,7 @@ def validate_routing_table(self, user_account=None):
 
     @Manager.calls_manager
     def _update_tag_data_for_acquire(self, user_account, tag):
+        yield self.api.mdb.batch_start([tag], user_account=user_account.key)
```
hodgestar, line 386: Maybe a comment that we're ignoring the result of `batch_start` here and fishing the batch_id out of the current tag info later?

go/vumitools/conversation/utils.py
```diff
@@ -260,8 +262,22 @@ def old_start(self, no_batch_tag=False, batch_id=None, acquire_tag=True,
...
+                # to confirm that. All of this will go away soon anyway.
+                batch = batches[0]
+                if batch_tags and batch_tags[0] not in batch.tags:
+                    batch.tags.append(batch_tags[0])
+                    yield batch.save()
+                batch_id = batch.key
+            else:
+                # We may not have a batch here if the conversation was created
+                # before new conversations got batches.
+                batch_id = yield self.start_batch(*batch_tags)
```
hodgestar, line 279: Maybe we need a method for this common logic? Maybe `get_or_create_batch`?

go/vumitools/tests/test_api.py
```diff
@@ -248,15 +183,46 @@ def test_declare_acquire_and_release_tags(self):
...
+        tag_info = yield self.vumi_api.mdb.get_tag_info(tag)
+        batch_id = tag_info.current_batch.key
+        to_addr = "+12310001"
+        msgs = [self.mkmsg_in(content=msg, to_addr=to_addr, message_id=str(i),
+                              transport_type="sms")
+                for i, msg in enumerate(("msg1", "msg2"))]
+        for msg in msgs:
+            yield self.vumi_api.mdb.add_inbound_message(msg, batch_id=batch_id)
+        api_msgs = yield self.vumi_api.mdb.batch_inbound_keys(batch_id)
+        self.assertEqual(sorted(api_msgs), ['0', '1'])
```
hodgestar: It looks like these two tests are just testing functionality from the message store?

hodgestar:
Initial review done. Left a few questions.

go/vumitools/api.py
jerith, line 466: No, because we load the CurrentTag object again and save the new one. The consistency problems arise when there's no clear ordering of updates.

go/vumitools/conversation/utils.py
jerith, line 279: Maybe if we need to change it. This stuff all belongs to old-UI code that's going away imminently.

go/vumitools/tests/test_api.py
jerith: Originally they were testing proxy methods that are no longer there. Now they're testing that acquiring and releasing tags sets up the batches properly. We could simplify this to just assert on the existence of the batches, but given that finding the messages is pretty important I'd rather make sure the message store isn't changing underneath us here.

hodgestar: Oh, I see -- they're testing `user_api.acquire_specific_tag`. Can we fix the test name then?

go/vumitools/conversation/utils.py
hodgestar, line 279: I can live with this being here for now.

hodgestar:
Other than the test name fix request, :+1:.

go/vumitools/tests/test_api.py
hodgestar: I think the storing of messages here is all irrelevant to the test? We could just assert that tag_info.current_batch.key is correct?

go/vumitools/conversation/tests/test_utils.py
```diff
@@ -126,7 +127,7 @@ def test_get_latest_batch_key(self):
         conv = yield self.user_api.get_wrapped_conversation(self.conv.key)
         batch_key = yield conv.get_latest_batch_key()
         self.assertEqual(batch_key, batch2)
-        self.assertEqual(len(conv.batches.keys()), 2)
+        self.assertEqual(len(conv.batches.keys()), 3)
```
smn, line 130: @jerith what's causing this? The batch creation when the conversation is created?

jerith, line 130: Yes.

smn:
:+1:

hodgestar:
:+1:

PR #591, opened 2013-08-06
Update setup_env to match the new dialogue application

justinvdm:


setup_env/applications.yaml
```diff
@@ -42,10 +66,6 @@ dialogue:
     send_to:
       default:
         transport_name: "invalid broken transport"
-    vxpolls:
-      prefix: vxpolls
-    middleware:
-      - logging_mw: vumi.middleware.logging.LoggingMiddleware
```
hodgestar, line 48: Maybe we should keep the logging middleware?

hodgestar:
Other than the suggestion to keep the logging middleware, :+1:.

justinvdm:
Ready for re-review

hodgestar:
:+1:

setup_env/applications.yaml
```diff
@@ -32,8 +32,32 @@ bulk_message:
 dialogue:
   class: go.apps.dialogue.vumi_app.DialogueApplication
   config:
-    batch_size: 2
     metrics_prefix: dialogue_application
+    env:
+      NODE_PATH: "/usr/lib/node_modules"
```
jerith, line 37: I'm pretty sure this isn't cross-platform.

justinvdm, line 37: Agreed. We shouldn't be globally installing and using node modules anyways. Hopefully we can solve this along with the rest of the js environment set up stuff in one fell swoop when we get around to it.

PR #592, opened 2013-08-06
Clean up code that pokes the message store directly

jerith:
This is necessary to make Go work with the changes in praekelt/vumi#586, shouldn't break with older vumi versions.


hodgestar:
:+1:

PR #598, opened 2013-08-07
Add batch_id to fake channel model

jerith:


jerith:
This will probably need some work in develop-frontend as well.

hodgestar:
:+1:

PR #588, opened 2013-08-07
Can't move past wizard screen

jerith:
On staging - on the wizard screen: http://qa-vumi.za.prk-host.net/wizard/create/

I provide a conversation name and description.
I pick 'dialogue' as the conversation type
I select "set up a new channel" and pick a channel
I click the 'save' button
Instead of being directed to the survey builder, I remain on the wizard screen: http://qa-vumi.za.prk-host.net/wizard/create/ and get a blank page

screen shot 2013-08-06 at 2 00 56 pm

screen shot 2013-08-06 at 2 01 50 pm

PR #596, opened 2013-08-07
Useful error pages

justinvdm:
We currently have empty error pages, which makes it really hard to figure out what's going wrong. At the very least we should have "404" and "500" pages that tell the client that there was a problem.

We haven't noticed this much before because we run locally with `DEBUG=True`, but it made #588 harder to debug.

### DIFF
--- a/go/templates/base.html
+++ b/go/templates/base.html
@@ -30,7 +30,7 @@
 
 </head>
 <body>
-    <div class="container-fluid main-container">
+    <div class="container-fluid main-container {% block container_extraclass %}{% endblock %}">
 
         <!--[if lt IE 7]>
         <p class=chromeframe>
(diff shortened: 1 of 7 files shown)


go/templates/base.html
```diff
@@ -30,7 +30,7 @@
 
 </head>
 <body>
-    <div class="container-fluid main-container">
+    <div class="container-fluid main-container {% block container_extraclass %}{% endblock %}">
```
justinvdm, line 33: I needed to add this so I could control the styling of the container for error pages. For example, the footer needs to be center aligned on the error page, otherwise the error page looks wierd.

justinvdm:
I did some syling, but it probably needs to belooked at/changed by someone with more design experience.

peterp:
:+1:

PR #605, opened 2013-08-08
Routers should have a start/stop lifecycle

jerith:


go/routers/keyword/vumi_app.py
```diff
@@ -29,9 +29,17 @@ def lookup_target(self, config, msg):
...
+            log.debug("Ignoring inbound for stopped router: %s" % (msg,))
+            return
         log.debug("Handling inbound: %s" % (msg,))
         return self.publish_inbound(msg, self.lookup_target(config, msg))
 
     def handle_outbound(self, config, msg, conn_name):
+        # TODO: Push this up a level or two.
+        if not config.get_conversation().running():
+            log.debug("Ignoring outbound for stopped router: %s" % (msg,))
+            return
```
hodgestar, line 43: Not sure what these changes are doing in this PR?

jerith, line 43: Stopped routers shouldn't process messages. #604 is for adding this to conversations.

go/vumitools/api.py
```diff
@@ -526,6 +526,45 @@ def _remove_from_routing_table(self, router):
         rt_helper.remove_router(router)
         yield user_account.save()
 
+    @Manager.calls_manager
+    def start_router(self, router=None):
```
hodgestar, line 530: This isn't directly relevant to this PR, but the way `VumiRouterApi` deals with routers is rather wat. I feel the constructor should either take `router_type, router_key` or `router_object` and figure out the other. Passing in an optional router object to a method on an instance that already has router_key and router_type attributes is bizarre.

hodgestar:
:+1:

go/vumitools/api.py
jerith, line 530: We don't want to keep a router model object on the API instance because that makes it much easier to accidentally save it twice and cause consistency issues. On the other hand, we don't want to have to load the thing from Riak again if we've already loaded it to look something up and haven't saved it.

I'd like to be able to build an API instance by passing in a model object and having it grab the relevant fields instead of making the caller do that. (Out of scope for this PR -- we need to take a look at all the classes in vumitools.api once develop-frontend has landed and we can get rid of all the cruft.)

PR #604, opened 2013-08-12
Conversations shouldn't process messages unless they're running.

jerith:


jerith:
This depends on praekelt/vumi#589 and also cleans up the similar behaviour in routers from #605.

hodgestar:
:+1:

PR #589, opened 2013-08-12
Implement "Display this question in search results" in dialogue vumi_app.js

justinvdm:
This requires checking the value of the flag in the appropriate state types in `vumi_app.js` and saving the result to the question to contact store extras if the flag is set to `True`.

### DIFF
--- a/go/apps/dialogue/vumi_app.js
+++ b/go/apps/dialogue/vumi_app.js
@@ -1,5 +1,6 @@
 var vumigo = require("vumigo_v01");
 var jed = require("jed");
+var Q = require("q");
 
 if (typeof api === "undefined") {
     // testing hook (supplies api when it is not passed in by the real sandbox)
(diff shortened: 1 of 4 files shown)


justinvdm:
Ready for review. The changes in [this gist](https://gist.github.com/justinvdm/6210979) need to land in [vumi-jssandbox-toolkit](https://github.com/praekelt/vumi-jssandbox-toolkit) and get published before this can land.

hodgestar:
Changes in gist +1ed.

hodgestar:
New vumi-jssandbox-toolkit with gist included released.

go/apps/dialogue/vumi_app.js
```diff
@@ -94,18 +94,37 @@ function DialogueStateCreator() {
         return creator(state_name, im, state_description);
     };
 
+    self.store_answer = function(answer, im, state_description) {
```
hodgestar: Since only `state_description.store_as` is used from `state_description` perhaps we can just pass in `store_as` directly?

go/apps/dialogue/vumi_app.js
```diff
@@ -94,18 +94,37 @@ function DialogueStateCreator() {
...
+
+        api.request('contacts.get_or_create', {
+            addr: msg.from_addr,
+            delivery_class: msg.helper_metadata.delivery_class
+        }, function(reply) {
+            if (reply.success) {
+                api.request('contacts.update_extras', {
+                    key: reply.contact.key,
+                    fields: fields
+                }, function() {});
```
hodgestar: Probably best if `store_answer` returned a promise?

go/apps/dialogue/vumi_app.js
```diff
@@ -94,18 +94,37 @@ function DialogueStateCreator() {
...
+            }
+        });
+    };
+
     self.choice_state_creator = function(state_name, im, state_description) {
         var choices = state_description.choice_endpoints.map(
             function (c) { return new Choice(c.value, c.label); });
         return new ChoiceState(
             state_name,
             function (choice) {
```
hodgestar: You can add an extra argument `done` to this function and call that with the result `get_next_state()` once the answer has been stored.

go/apps/jsbox/vumi_app.py
```diff
@@ -83,6 +83,22 @@ def user_api_for_api(self, api):
     def get_config(self, msg):
         return self.get_message_config(msg)
 
+    def infer_delivery_class(self, msg):
+        return {
+            'smpp': 'ussd',
```
hodgestar: I think the inferred type for 'smpp' should be 'sms'?

hodgestar:
Left a few comments but otherwise looks good.

go/apps/dialogue/vumi_app.js
```diff
@@ -94,18 +95,52 @@ function DialogueStateCreator() {
...
+        
+        p.then(function(reply) {
+            if (!reply.success) { return; }
+
+            return self.api_request(
+              'contacts.update_extras',
+              {key: reply.contact.key, fields: fields});
+        });
+
+        return p;
```
hodgestar: I think you should maybe return the result of `p.then` here? Q promises don't work quite like Twisted deferreds (p.then returns a new promise, not p).

hodgestar:
:+1:

PR #601, opened 2013-08-12
Migration script to give active conversations that are using their tags batch their own batch instead.

hodgestar:
Part of #278.

### DIFF
--- a/go/base/management/commands/go_migrate_conversations.py
+++ b/go/base/management/commands/go_migrate_conversations.py
@@ -26,11 +26,11 @@ class Migration(object):
                 return migrator_cls
         return None
 
-    def run(self, conv):
+    def run(self, user_api, conv):
         if not self._dry_run:
-            self.migrate(conv)
+            self.migrate(user_api, conv)
 
-    def applies_to(self, conv):
+    def applies_to(self, user_api, conv):
         """Whether this migration applies to the given conversation.
 
         Should return True if this conversation requires migrating, False
@@ -38,7 +38,7 @@ class Migration(object):
         """
         return True
 
-    def migrate(self, conv):
+    def migrate(self, user_api, conv):
         """Perform the migration."""
         raise NotImplementedError("Migration %s not implemented."
                                   % (self.name,))
@@ -50,7 +50,7 @@ class UpdateModels(Migration):
         "Load and re-save all conversations, triggering any pending model"
         " migrators in the process.")
 
-    def migrate(self, conv):
+    def migrate(self, user_api, conv):
         conv.save()
 
 
@@ -62,10 +62,15 @@ class SeparateTagBatches(Migration):
         " this batch to a new batch and assign the new batch to the"
         " conversation.")
 
-    def applies_to(self, conv):
-        raise NotImplementedError("TODO: implement")
+    def applies_to(self, user_api, conv):
+        tag = (conv.delivery_tag_pool, conv.delivery_tag)
+        if not conv.active() or None in tag:
+            return False
+        conv_batches = conv.batches.keys()
+        tag_info = user_api.api.mdb.get_tag_info(tag)
+        return tag_info.current_batch.key in conv_batches
 
-    def migrate(self, conv):
+    def migrate(self, user_api, conv):
         raise NotImplementedError("TODO: implement")
 
 
@@ -137,7 +142,8 @@ class Command(BaseCommand):
         all_keys = user_api.conversation_store.list_conversations()
         conversations = [user_api.get_wrapped_conversation(k)
                          for k in all_keys]
-        conversations = [c for c in conversations if migrator.applies_to(c)]
+        conversations = [c for c in conversations
+                         if migrator.applies_to(user_api, c)]
         self.outln(
             u'%s %s <%s> [%s]\n  Migrating %d of %d conversations ...' % (
             user.first_name, user.last_name, user.username,
@@ -145,7 +151,7 @@ class Command(BaseCommand):
         for conv in conversations:
             self.outln(u'    Migrating conversation: %s [%s] ...'
                        % (conv.key, conv.name), ending='')
-            migrator.run(conv)
+            migrator.run(user_api, conv)
             self.outln(u' done.')
 
     def handle(self, *usernames, **options):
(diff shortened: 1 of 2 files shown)


go/base/management/commands/go_migrate_conversations.py
```diff
@@ -7,6 +8,86 @@
...
+    @classmethod
+    def migrator_classes(cls):
+        return cls.__subclasses__()
+
+    @classmethod
+    def migrator_class(cls, name):
+        for migrator_cls in cls.migrator_classes():
+            if migrator_cls.name == name:
+                return migrator_cls
+        return None
```
jerith, line 27: Maybe raise an exception instead of returning `None` here? Saves us checking return values everywhere like we're in C.

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -7,6 +8,86 @@
...
+        if not self._dry_run:
+            self.migrate(user_api, conv)
+
+    def applies_to(self, user_api, conv):
+        """Whether this migration applies to the given conversation.
+
+        Should return True if this conversation requires migrating, False
+        otherwise.
+        """
+        return True
```
jerith: I'd rather raise `NotImplementedError` here as well so we don't forget to do this and accidentally try to rerun non-idempotent migrations (like tag splitting).

go/base/management/commands/go_migrate_conversations.py
hodgestar, line 27: There's only one call and I'd have to catch the exception to report a nice error anyway but happy to change if you feel strongly.

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -7,6 +8,86 @@
...
+        conv.save()
+
+
+class SeparateTagBatches(Migration):
+    name = "separate-tag-batches"
+    help_text = (
+        "Look for active conversations which have a tag and where that tag's"
+        " current batch is the same as the conversation's batch. Copy this"
+        " this batch to a new batch and assign the new batch to the"
+        " conversation.")
```
jerith: "this this" in the text. Yay linebreaks! ;-)

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -7,6 +8,86 @@
...
+
+    def applies_to(self, user_api, conv):
+        tag = (conv.delivery_tag_pool, conv.delivery_tag)
+        if not conv.active() or None in tag:
+            return False
```
jerith: I'm pretty sure we have active conversations that predate setting `delivery_tag` on the conversation.

jerith: Probably best to just split any batch on the conversation that has tags.

go/base/management/commands/go_migrate_conversations.py
jerith, line 27: I'm just a little concerned that we're using return values instead of exceptions in a bunch of places and occasionally being bitten by it.
I'd be a little happier with a generic error reporter (maybe using a custom exception), but it's not really that big a deal here.

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -7,6 +8,90 @@
...
+    @classmethod
+    def migrator_class(cls, name):
+        for migrator_cls in cls.migrator_classes():
+            if migrator_cls.name == name:
+                return migrator_cls
+        return None
+
+    def run(self, user_api, conv):
+        if not self._dry_run:
+            self.migrate(user_api, conv)
```
jerith, line 31: Do we maybe want a (by default do-nothing) dry run method that we call here? That would give us a place to inspect the things that are going to be migrated and report on what we might have to do, as opposed to just reporting which conversations need migrating.

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -48,21 +140,50 @@ def find_accounts(self, *usernames):
...
+        return migrator_cls(dry_run)
+
+    def handle_user(self, user, migrator):
         user_api = vumi_api_for_user(user)
-        conversations = user_api.conversation_store.list_conversations()
+        all_keys = user_api.conversation_store.list_conversations()
+        conversations = [user_api.get_wrapped_conversation(k)
+                         for k in all_keys]
+        conversations = [c for c in conversations
+                         if migrator.applies_to(user_api, c)]
```
jerith, line 168: I'm not enthusiastic about loading all conversations into memory and then filtering. Maybe we could loop over the keys loading one at a time and stick the applicable ones in a list?

go/base/tests/test_go_migrate_conversations.py
```diff
@@ -16,19 +16,25 @@ def setUp(self):
...
-        self.conv3 = self.mkoldconv(subject=u'Zoë destroyer of Ascii')
-        self.convs = [self.conv1, self.conv2, self.conv3]
-
         self.command = go_migrate_conversations.Command()
         self.command.stdout = StringIO()
         self.command.stderr = StringIO()
 
+    def handle_command(self, migration_name=None, list=False, dry_run=False):
+        self.command.handle(migration_name=migration_name, list=list,
+                            dry_run=dry_run)
```
jerith, line 30: I'd rather not use `list` as an identifier here.

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -7,6 +8,93 @@
...
+
+    def applies_to(self, user_api, conv):
+        mdb = user_api.api.mdb
+        conv_batches = conv.batches.keys()
+        if len(conv_batches) != 1:
+            return True
+        for batch_id in conv_batches:
+            tag_keys = mdb.current_tags.index_keys('current_batch', batch_id)
+            if tag_keys:
+                return True
```
jerith, line 77: You can simplify this logic a bit here, since you're guaranteed to always have exactly one batch.

hodgestar, line 77: I think this structure captures what is described in the help text better -- i.e. migrate if there isn't exactly one batch or if any of the batches has a tag associated.

jerith:
:+1:

PR #610, opened 2013-08-14
Make routing errors reported a bit more useful.

smn:
Currently RoutingTableDispatcher logs when it encounters a message it is failing to route but it doesn't provide information on the address of the message trying to be routed which is often useful in order to determine whether this is an urgent issue or not.


smn:
hrm, looking through `vumi/sentry.py` now I doubt this works :p

smn:
Sad PR has no code. Closing because useless.

PR #611, opened 2013-08-14
Update setup_env for new UI

jerith:
This replaces #469 which is a PR onto the now-obsolete develop-frontend.


go/base/management/commands/go_setup_env.py
```diff
@@ -302,36 +256,55 @@ def assign_application(self, account, application_module):
...
-                    'foo': 'bar',
-                }
-            },
-            {
-                ...
-            }]
+    def setup_channels(self, user, channels):
+        user_api = vumi_api_for_user(user)
+        for channel in channels:
+            tag = tuple(channel.split(':'))
```
smn, line 262: do we want to expand this to make sure we only get two?

setup_env/workers.yaml
```diff
@@ -0,0 +1,203 @@
...
+      riak_manager: *RIAK-MANAGER
+      send_to: *SEND-TO
+
+  jsbox:
+    class: go.apps.jsbox.vumi_app.JsBoxApplication
+    config:
+      metrics_prefix: "go.accounts."
+      env:
+        NODE_PATH: "/usr/lib/node_modules"
+      app_context: "{require: function(m) { if (m == 'jed' || m == 'vumigo_v01') return require(m); return null; }, Buffer: Buffer}"
```
smn, line 161: `q` & `libxmljs` are missing from this list.

smn:
:+1: other than the two minor comments

go/base/management/commands/go_setup_env.py
jerith, line 262: We could, but the error will be caught later anyway so it isn't that big a deal.

setup_env/workers.yaml
jerith, line 161: This is copied directly from what was there at the time. I'll check what's there now and update accordingly.

jerith, line 161: Yes, this is what we have in develop. I'm inclined to leave it as-is and fix it later.

PR #612, opened 2013-08-14
Better conversation and router management commands

jerith:
We currently have two management commands for conversations (`go_manage_conversation` and `go_start_conversation`) and none for routers. We need to consolidate the former and build the latter.

### DIFF
--- a/go/base/command_utils.py
+++ b/go/base/command_utils.py
@@ -1,0 +1,53 @@
+from optparse import make_option
+
+from django.core.management.base import BaseCommand, CommandError
+from django.contrib.auth.models import User
+
+from go.base.utils import vumi_api_for_user
+
+
+def make_command_opt_str(command_name):
+    return '--%s' % (command_name.replace('_', '-'),)
+
+
+def make_command_option(command_name, **kw):
+    opt_str = make_command_opt_str(command_name)
+    return make_option(
+        opt_str, dest='command', action='append_const', const=command_name,
+        **kw)
+
+
+class BaseGoAccountCommand(BaseCommand):
+    ACCOUNT_OPTIONS = [
+        make_option('--email-address',
+                    dest='email_address',
+                    help='Email address for the Vumi Go user'),
+    ]
+    LOCAL_OPTIONS = []
+    option_list = (
+        BaseCommand.option_list
+        + tuple(ACCOUNT_OPTIONS)
+        + tuple(LOCAL_OPTIONS)
+    )
+
+    def handle(self, *args, **options):
+        if 'email_address' not in options:
+            raise CommandError("--email-address must be specified")
+        try:
+            self.user = User.objects.get(email=options['email_address'])
+        except User.DoesNotExist, e:
+            raise CommandError(e)
+
+        self.user_api = vumi_api_for_user(self.user)
+
+        commands = options.get('command', [])
+        if not commands:
+            return self.handle_no_command(*args, **options)
+        if len(commands) > 1:
+            raise CommandError(
+                "Multiple command options provided, only one allowed: %s" % (
+                    ' '.join(make_command_opt_str(cmd) for cmd in commands)))
+        return getattr(self, 'handle_command_%s' % commands)(*args, **options)
+
+    def handle_no_command(*args, **options):
+        raise CommandError('Please specify an action')
--- a/go/base/command_utils.py
+++ b/go/base/command_utils.py
@@ -18,17 +18,15 @@ def make_command_option(command_name, **kw):
 
 
 class BaseGoAccountCommand(BaseCommand):
-    ACCOUNT_OPTIONS = [
+    option_list = BaseCommand.option_list + (
         make_option('--email-address',
                     dest='email_address',
                     help='Email address for the Vumi Go user'),
-    ]
-    LOCAL_OPTIONS = []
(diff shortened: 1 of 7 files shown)


go/base/command_utils.py
```diff
@@ -0,0 +1,54 @@
...
+        return [opt.const for opt in self.option_list
+                if opt.action == 'append_const' and opt.dest == 'command']
+
+    def handle(self, *args, **options):
+        if 'email_address' not in options:
+            raise CommandError("--email-address must be specified")
+        try:
+            self.user = User.objects.get(email=options['email_address'])
+        except User.DoesNotExist, e:
+            raise CommandError(e)
```
hodgestar, line 37: It would be nice to be able to see a list of commands without having to specify an email address.

go/base/management/commands/go_manage_conversation.py
```diff
@@ -1,64 +1,78 @@
...
+        conversation = self.get_conversation(options)
         self.stdout.write(pformat(conversation.config))
         self.stdout.write("\n")
+
+    def handle_command_start(self, *apps, **options):
+        conversation = self.get_conversation(options)
+        if conversation.archived():
+            raise CommandError('Archived conversations cannot be started')
+        if conversation.running() or conversation.stopping():
+            raise CommandError('Conversation already running')
```
hodgestar, line 58: Maybe `starting` should be in this list too?

go/base/management/commands/go_manage_conversation.py
```diff
@@ -1,64 +1,78 @@
...
-                    dest='email_address',
-                    help='Email address for the Vumi Go user'),
+    option_list = BaseGoAccountCommand.option_list + (
+        make_command_option(
+            'list', help='List the active conversations in this account.'),
+        make_command_option(
+            'show_config', help='Display the config for a conversation'),
+        make_command_option('start', help='Start a conversation'),
+        make_command_option('stop', help='Stop a conversation'),
+        make_command_option('archive', help='Archive a conversation'),
```
hodgestar, line 20: I think it would be good to have a --show that shows all of a conversations other fields too. I think we could use `pformat(conversation.get_data())` as a first approximation.

go/base/management/commands/go_manage_router.py
```diff
@@ -0,0 +1,80 @@
...
+    def handle_command_show_config(self, *args, **options):
+        router = self.get_router(options)
+        self.stdout.write(pformat(router.config))
+        self.stdout.write("\n")
+
+    def handle_command_start(self, *apps, **options):
+        router = self.get_router(options)
+        if router.archived():
+            raise CommandError('Archived routers cannot be started')
+        if router.running() or router.stopping():
```
hodgestar, line 56: We need to include `starting()` here too.

go/base/management/commands/go_manage_conversation.py
hodgestar, line 20: Maybe `--show` should just replace `--show-config` although it might be useful to have both.

hodgestar:
Suggested some improvements but generally really like.

go/base/command_utils.py
jerith, line 37: Good point. I think `--help` does that, though.

go/base/management/commands/go_manage_conversation.py
jerith, line 58: No. This needs to be able to resend a `start` command to a conversation that's stuck because the previous command somehow got lost.

go/base/management/commands/go_manage_conversation.py
jerith, line 20: Agreed, but I was going for the smallest necessary amount of work to get this landed first. I also want to allow modification of other conversation fields (especially `status`), but I think that can have its own branch.

go/base/management/commands/go_manage_router.py
jerith, line 56: See my comment above.

go/base/command_utils.py
hodgestar, line 37: Good point. :)

go/base/management/commands/go_manage_conversation.py
hodgestar, line 58: I think that should be handled by a stop-then-start cycle to avoid conversations being started twice in a row (if, for example, a start message is queued and becomes unstuck or something).

go/base/management/commands/go_manage_conversation.py
hodgestar, line 20: `--show` seems kinda fundamental and only a few lines of code plus a test?

go/base/management/commands/go_manage_conversation.py
jerith, line 58: Conversations should (and do, unless they override the handler) ignore start messages if they aren't in the "starting" state.

go/base/management/commands/go_manage_conversation.py
jerith, line 20: Perhaps. If you want to add it I'm happy to review. :-)

go/base/management/commands/go_manage_conversation.py
hodgestar, line 58: Woot. Awesome. In that case I'm fine with this bit.

PR #603, opened 2013-08-14
Static reply app

jerith:
It would be useful to have a Vumi Go app that replies to all incoming message with a preconfigured reply. Also very straightforward to implement.

### DIFF
--- a/go/apps/static_reply/vumi_app.py
+++ b/go/apps/static_reply/vumi_app.py
@@ -24,4 +24,5 @@ class StaticReplyApplication(GoApplicationWorker):
     def consume_user_message(self, message):
         config = yield self.get_message_config(message)
         if config.reply_text:
-            yield self.reply_to(message, config.reply_text)
+            yield self.reply_to(
+                message, config.reply_text, continue_session=False)
--- a/go/apps/static_reply/vumi_app.py
+++ b/go/apps/static_reply/vumi_app.py
@@ -1,0 +1,27 @@
+# -*- test-case-name: go.apps.static_reply.tests.test_vumi_app -*-
+# -*- coding: utf-8 -*-
+
+from twisted.internet.defer import inlineCallbacks
+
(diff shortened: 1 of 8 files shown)


go/apps/static_reply/vumi_app.py
```diff
@@ -0,0 +1,27 @@
...
+    Application that replies to incoming messages with a configured response.
+    """
+    worker_name = 'static_reply_application'
+    CONFIG_CLASS = StaticReplyConfig
+
+    @inlineCallbacks
+    def consume_user_message(self, message):
+        config = yield self.get_message_config(message)
+        if config.reply_text:
+            yield self.reply_to(message, config.reply_text)
```
hodgestar: Setting this up on USSD with a suitable message could be amusing. :)

go/conversation/view_definition.py
```diff
@@ -328,7 +328,10 @@ def process_forms(self, request, conversation):
             # Is this a good idea?
             if not edit_form.is_valid():
                 return self._render_forms(request, conversation, edit_forms)
-            config[key] = self.process_form(edit_form)
+            if key is None:
+                config = self.process_form(edit_form)
+            else:
+                config[key] = self.process_form(edit_form)
```
hodgestar, line 334: Useful.

hodgestar:
:+1:

go/conversation/view_definition.py
jerith, line 334: That was always supposed to be like that (and it works that way when building the forms from config data), but nothing actually used it and this side was broken.

go/apps/static_reply/vumi_app.py
jerith: Maybe we should explicitly end the session here?

hodgestar: Explicitly ending the session sounds sensible.

PR #622, opened 2013-08-15
Tagpool permissions look in the wrong place for used tag counts

jerith:
We're still looking at `delivery_tagpool` on conversations instead of the tags owned by the account.


hodgestar:
:+1:

PR #626, opened 2013-08-15
Fix dialogue vumi_app.js bugs

hodgestar:
These include:
- Use of Q promises doesn't work in production.
- Test data and means of looking up entry_endpoints on states is incorrect.


justinvdm:
:+1: Lots of stuff I should've caught in reviews, sorry about that :/

PR #628, opened 2013-08-16
Message confirmation screen formatting display is broken

justinvdm:
If a user has enabled SMS confirmation for their account, they have to confirm that a message can be delivered, before it is actually delivered. The user confirms the message in a web view, the URL of which is included in the SMS confirmation message that they receive.

The confirmation page web view was not updated with the new UI - the formatting is very broken (see screenshot) For example the text is black on a black background. Please can you apply the latest styling to this page so it's legible and looks decent.

screen shot 2013-08-16 at 11 39 07 am

### DIFF
--- a/go/conversation/templates/conversation/confirm.html
+++ b/go/conversation/templates/conversation/confirm.html
@@ -2,19 +2,18 @@
 
 {% block content_extraclass %}confirm{% endblock %}
 
-{% block content_title %}{{ action_display_name }} confirmation{% endblock %}
+{% block content_title %}{{ action_name }} confirmation{% endblock %}
 
 {% block content_main %}
     {% include "base/includes/messages.html" %}
 
     <div class='row-fluid details'>
-        <div class='span7'>
+        {% if not success %}
+          <p>Please review the details below and click confirm if everything is in order.</p>
+        {% endif %}
+
+        <div class='span6'>
             <h4>Conversation details</h4>
-            <p>
-            {% if not success %}
-                Please review the details below and click confirm if everything is in order.
-            {% endif %}
-            </p>
             <table class="conversation table table-bordered">
                 <thead>
                     <tr>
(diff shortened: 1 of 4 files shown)


go/conversation/templates/conversation/confirm.html
```diff
@@ -1,50 +1,63 @@
...
+                        </td>
+                    </tr>
+                    <th>Groups</th>
+                        <td>
+                            {% for group in conversation.get_groups %}
+                                <span class="label label-inverse">{{ group.name }}</span>
+                            {% endfor %}
+                        </td>
+                    </tr>
+                </tbody>
```
hodgestar, line 47: Since we now confirm actions rather than conversation starts, I think we need to display something about the action being confirmed (i.e. the action type and the action data). I think a separate table might be best.

hodgestar:
Looks good. Just think we're missing information about the action itself in the confirmation screen.

go/conversation/templates/conversation/confirm.html
justinvdm, line 47: I'm wondering whether the action type and data are too lower level. The action's display nameis shown as the title, not sure what else to show.

jerith, line 47: If we're confirming a bulk send, it's good to know what message is being sent.

justinvdm, line 47: Don't we have the form below for that? We would just need to improve the bulk send form. Or wouldn't it contain the message being sent out?

hodgestar, line 47: I think the form below is just the confirmation form? I.e. the token and a "confirm" or "deny" button.

justinvdm:
Ready for re-review. We now display action details in the confirm screen. I've noticed that we don't have tests for the conversation confirm screen. We should probably do that, not sure if it should happen in this PR?

hodgestar:
Re tests -- it could go either way but given that we're already mostly done with this one I would create a separate issue and PR for it.

hodgestar:
:+1:

PR #632, opened 2013-08-18
Router dashboard should use per-router buttons (the same way the conversation dashboard uses per-conversation buttons).

hodgestar:


justinvdm:
:+1:

PR #636, opened 2013-08-19
Using Source URL for loading JS box app 404s

justinvdm:
skitch


jerith:
I'd really rather not push a generic HTTP proxy up to the top level where everything can use it. Can you revert this change and somehow make the JS call "/conversations/<key>/cross-domain-xhr/" instead?

justinvdm:
Turns out that solving this properly is more work than it appears: #640

smn:
:+1: and :cake: for CSRF

PR #641, opened 2013-08-19
Some modules are missing from setup_env jsbox application config

smn:
`q` & and `libxmljs` are missing.


jerith:
We might need the same change for the dialogue app worker as well.

smn:
The dialogue app doesn't need it because that doesn't run 3rd party code.

jerith:
:+1:

PR #633, opened 2013-08-19
Add incoming_list to conversation view definition.

hodgestar:
Currently conversations incoming_list posts message export and once-off reply requests back to itself but these are implemented on ShowConversationView. We need an IncomingListView that uses all of what is in ShowConversationView.post.


jerith:
This breaks a bunch of tests in its current state.

jerith:
I'm going to skip the router side of things here, because it means either a bunch of duplicated code or a refactoring of the view logic that I don't yet have a good handle on. Since we don't need the router view as urgently as the conversation one I think it'll be okay to finish this in a new ticket later.

go/apps/bulk_message/tests/test_views.py
```diff
@@ -382,20 +383,19 @@ def test_action_bulk_send_confirm(self):
             delivery_class=conversation.delivery_class,
             content='I am ham, not spam.', dedupe=True))
 
-    @skip("The new views don't have this.")
```
hodgestar, line 385: Yay for removing a bunch of skips.

hodgestar:
:+1:

PR #639, opened 2013-08-19
Remove a bunch of old conversation lifecycle cruft (part 1)

jerith:
This is the first part of the cleanup work to remove all the obsolete code that supported the old conversation lifecycle. I plan to remove the worst of the obsolete hackery, but I want to leave more invasive work (like #624) for different tickets to make reviewing a little easier.


go/vumitools/app_worker.py
```diff
@@ -13,8 +13,6 @@
 from go.vumitools.utils import MessageMetadataHelper
 from go.vumitools.conversation.models import (
     CONVERSATION_STARTING, CONVERSATION_STOPPING)
```
hodgestar: We should probably remove these too and use the `.starting()` and `.stopping()` helper methods on conversation instead.

go/vumitools/conversation/utils.py
```diff
@@ -64,13 +52,9 @@ def archive_conversation(self):
         yield self._release_batches()
 
     @Manager.calls_manager
-    def _release_batches(self, release_tags=False):
+    def _release_batches(self):
         for batch in (yield self.get_batches()):
             yield self.mdb.batch_done(batch.key)
```
hodgestar, line 57: Hmm. I wonder if there is even really a point in releasing these batches anymore? They should never have tags assigned to them?

hodgestar:
:+1: -- two comments that I'm happy to have addressed in a later clean-up (maybe just note them somewhere if you don't address them here).

go/vumitools/app_worker.py
jerith: Good point. I'll do that.

go/vumitools/conversation/utils.py
jerith, line 57: I'd rather call `batch_done()` in case we decide to have that do more than just update tag data.

smn:
:+1:

PR #642, opened 2013-08-19
Change client side action views to use the csrf token cookie instead of using the token in the template

justinvdm:


hodgestar:
:+1:

PR #643, opened 2013-08-19
Fix setup_env some more

jerith:
There are a few broken things in setup_env that need fixing. This is a catchall for those.

### DIFF
--- a/setup_env/workers.yaml
+++ b/setup_env/workers.yaml
@@ -87,6 +87,10 @@ applications:
       send_to: *SEND-TO
       middleware:
         - logging_mw: vumi.middleware.logging.LoggingMiddleware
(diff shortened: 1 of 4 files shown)


setup_env/workers.yaml
```diff
@@ -87,6 +87,10 @@ applications:
       send_to: *SEND-TO
       middleware:
         - logging_mw: vumi.middleware.logging.LoggingMiddleware
```
smn, line 89: we may as well do the `__hidden__` trick for these default middlewares as well.

smn:
:+1:

setup_env/workers.yaml
jerith, line 89: I'd rather have them explicit, since we need configuration for them anyway. I don't really have a particularly strong opinion on the matter, though.

PR #644, opened 2013-08-19
We need a way to access Groups from the Sandbox

smn:
Currently we have a ContactsResource but not a GroupsResource.

It should allow:
1. creation of groups (static + smart).
2. updating of group name.
3. getting member counts.

Deleting & clearing of group members I'm leaving as an action in the UI for now.

### DIFF
--- a/go/apps/jsbox/contacts.py
+++ b/go/apps/jsbox/contacts.py
@@ -446,7 +446,7 @@ class GroupsResource(SandboxResource):
             groups=[group.get_data() for group in groups]))
 
     @inlineCallbacks
-    def handle_get_by_key(self, api, command):
+    def handle_get(self, api, command):
         try:
             contact_store = self._contact_store_for_api(api)
             group = yield contact_store.get_group(command['key'])
@@ -464,7 +464,7 @@ class GroupsResource(SandboxResource):
             group=group.get_data()))
 
     @inlineCallbacks
-    def handle_get(self, api, command):
+    def handle_get_by_name(self, api, command):
         try:
             contact_store = self._contact_store_for_api(api)
             keys = yield contact_store.groups.search(
@@ -494,7 +494,7 @@ class GroupsResource(SandboxResource):
             command, success=True, group=group.get_data()))
 
     @inlineCallbacks
-    def handle_get_or_create(self, api, command):
+    def handle_get_or_create_by_name(self, api, command):
         try:
             contact_store = self._contact_store_for_api(api)
             keys = yield contact_store.groups.search(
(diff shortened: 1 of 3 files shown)


smn:
Still needs some docstrings on the individual `handle_*` functions for RTFD.org

go/apps/jsbox/contacts.py
```diff
@@ -405,3 +405,157 @@ def handle_save(self, api, command):
...
+        try:
+            contact_store = self._contact_store_for_api(api)
+            keys = yield contact_store.groups.search(
+                name=command['name']).get_keys()
+        except (SandboxError,) as e:
+            log.warning(str(e))
+            returnValue(self.reply(command, success=False, reason=unicode(e)))
+        except (Exception,) as e:
+            # NOTE: Hello Riakasaurus, you raise horribly plain exceptions on
+            #       a MapReduce error.
```
jerith, line 601: Not just Riakasaurus. It inherited its "just raise `Exception`" behaviour from the other riak client.

jerith:
Can we factor out all the exception handling stuff into a wrapper function or something? It's a lot of boilerplate and I've lost my place in the review twice already because of it.

go/apps/jsbox/contacts.py
```diff
@@ -405,3 +405,157 @@ def handle_save(self, api, command):
...
+            command,
+            success=True,
+            group=group.get_data()))
+
+    @inlineCallbacks
+    def handle_get(self, api, command):
+        try:
+            contact_store = self._contact_store_for_api(api)
+            keys = yield contact_store.groups.search(
+                name=command['name']).get_keys()
```
jerith, line 542: I really wouldn't expect a method called `handle_get` to be doing a name search. Can we rename this operation `find_by_name` or something?

jerith:
Looks good once my comments are addressed. If we're going to change `handle_get`'s name (and I'll leave it up to you to decide if we do) I'd prefer to do that in this PR. The rest can probably happen later.

jerith:
:+1: -- we can figure out how to get rid of all the exception handler boilerplate later.

go/apps/jsbox/contacts.py
```diff
@@ -1,7 +1,7 @@
 # -*- test-case-name: go.apps.jsbox.tests.test_contacts -*-
 # -*- coding: utf-8 -*-
 
-from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.defer import inlineCallbacks, returnValue, DeferredList
```
jerith, line 4: Why add an unused import here?

PR #649, opened 2013-08-20
Allow for GroupsResource to list all known groups.

smn:
GroupsResource currently can only list all groups by searching for something that happens to match everything.


jerith:
:+1:

PR #637, opened 2013-08-19
Can't search across groups

justinvdm:
I took the following steps:
- on the groups page: https://go.vumi.org/contacts/groups/ , I have a group called "Ben"
- in the search box at the top of the groups page I search for "Ben"
- after I click the search box, nothing happens (i.e. I am shown the same page)

Additional notes:
- the same solution should also apply to the smart groups page: https://go.vumi.org/contacts/groups/smart/
- Group search is a requirement for Rachel at Sisi Ni Amani.

screen shot 2013-08-19 at 1 21 35 pm

PR #638, opened 2013-08-21
Adding a search box to the groups screen

justinvdm:
I took the following steps:
- on the overview/show screen for my bulk message conversation
- I click on the 'edit' button next to the "contacts' header
- I am directed to the 'contacts-edit' screen

We need to have a search box on the 'contacts-edit' screen that a user can use to search across contact (smart and regular) groups.

This is a requirement for Rachel from Sisi Ni Amani who has hundreds of groups in her account.

This ticket is dependent on #637 

screen shot 2013-08-19 at 1 36 20 pm

screen shot 2013-08-19 at 1 27 35 pm


justinvdm:
Ready for review

go/base/static/js/src/contacts/models.js
```diff
@@ -0,0 +1,23 @@
...
+// ==================
+
+(function(exports) {
+  var Model = go.components.models.Model;
+
+  var GroupModel = Model.extend({
+    idAttribute: 'key',
+
+    urls: {
+      show: function() { return '/contacts/group/' + this.id + '/'; }
```
hodgestar: Would be better if this was passed in from Django and not hardcoded.

go/base/static/js/src/conversation/models.js
```diff
@@ -0,0 +1,54 @@
...
+// ==========================
+
+(function(exports) {
+  var Model = go.components.models.Model;
+
+  var ConversationGroupsModel = Model.extend({
+    idAttribute: 'key',
+
+    url: function() {
+      return '/conversations/' + this.id + '/edit_groups/';
```
hodgestar, line 11: Would also be better if this was sent in from Django.

go/conversation/view_definition.py
```diff
@@ -456,48 +457,38 @@ def _render_groups(self, request, conversation):
...
-
-        pagination_params = urlencode({
-            'query': query,
-        })
+        model_data = {
+            'key': conversation.key,
+            'groups': [{
+                'key': group.key,
+                'name': group.name,
+                'inConversation': group.key in selected_groups
```
hodgestar: We could stick the "view_url" and "edit_url" in here?

hodgestar:
Looks good -- just the comments about generating group URLs in Django. Otherwise :+1:.

go/base/static/js/src/contacts/models.js
justinvdm: Cool, that's probably a better idea.

go/base/static/js/src/conversation/models.js
justinvdm, line 11: @hodgestar With Backbone, you usually set the url as one of the model prototype's properties, and use that for syncing the model with the server side (as opposed to getting an attribute from the server side). Maybe we should keep the url the models use to sync with the server side in the Backbone Models, and have the ['secondary' urls](https://github.com/praekelt/vumi-go/pull/638#discussion_r5903604) sent over as the data that becomes the model's attributes? I think that might make sense, since once we move the contact management to the http api, the url property on the model's prototype will go, and the 'secondary' urls will continue to be used as the model's attributes. What do you think?

hodgestar, line 11: Sounds good.

justinvdm:
Ready for re-review

hodgestar:
:+1:

PR #651, opened 2013-08-22
Handle text fields that weren't set for dialogue states

justinvdm:


justinvdm:
@hodgestar I'm wondering whether we should log a warning in `InteractionMachine` for states with empty questions instead of doing that in the dialogue app's `vumi_app.js`. Thoughts?

hodgestar:
I think ensuring that text is not null is fine. A warning in the UI might be cool too. I'm -1 on putting anything directly in InteractionMachine -- other people's code should be able to do what it wants (even if it's wrong).

hodgestar:
I think I'm :+1: on this as is -- it means we'll respond with empty text rather than `null` which seems like a good change (hopefully I'm understanding what this does correctly).

justinvdm:
Thanks :) Also, I agree, other people's code should be able to do what it wants (good point! :) I think once we've reached a point where there are less fires breaking out, we should definitely add some validation to the UI for the dialogue screen.

PR #624, opened 2013-08-22
Clean up conversation model

jerith:
The conversation model still contains obsolete fields that we shouldn't be using for anything anymore, particularly `delivery_tag_pool` and `delivery_tag`. We also need to swap out `batches` for a single `batch` foreign key.


hodgestar:
:+1: on this but lets land it on Monday when everyone is back.

jerith:
There's still some cleanup work I'd like to do (mostly removing `get_batches()` and `get_latest_batch_key()` which are used all over the place) but otherwise this branch is pretty much finished.

hodgestar:
The fix-batches migration will currently display a lot of errors for new conversation models but it should probably just skip them (since they can't have anything other than 1 batch).

hodgestar:
Cool. Reviewed everything so far.

hodgestar:
:+1:

PR #662, opened 2013-08-26
Clicking 'save' button on 'edits_group' page does not direct back to the 'show' page

justinvdm:
These are the steps I took:
- on the conversation 'show' page: 
- clicked on the 'edit' button next to the Contact Groups header
- on the 'edit_groups' page I selected multiple groups and clicked 'save'
- I clicked the 'OK' button on the 'groups saved successfully' pop up
- I am then returned to the 'edit_groups' page

I should be directed back to the conversation show page after saving my groups.

Also, the groups on the 'edit_groups' page should be sorted by group name

screen shot 2013-08-26 at 9 38 19 am

### DIFF
--- a/go/base/static/js/src/contacts/models.js
+++ b/go/base/static/js/src/contacts/models.js
@@ -6,7 +6,7 @@
 
   var GroupModel = Model.extend({
     relations: [{
-      type: Backbone.HasMany,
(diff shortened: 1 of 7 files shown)


hodgestar:
:+1:

PR #665, opened 2013-08-26
Can't delete a contact from a group

justinvdm:
I took the following steps:
- I am on a group page: e.g. https://go.vumi.org/contacts/group/5844c59dd26045b2955ee284413e3f3e/
- I check the box next to a contact row
- I click the delete btn in the header
- I get an error screen (attached)

Note: I don't get this error when deleting a contact from the contacts screen (i.e. not from a group)

screen shot 2013-08-26 at 1 21 40 pm

screen shot 2013-08-26 at 1 19 36 pm

### DIFF
--- a/go/contacts/templates/contacts/group_detail.html
+++ b/go/contacts/templates/contacts/group_detail.html
@@ -7,7 +7,7 @@
 
 {% block content_actions_right %}
     <div class="table-form-view-buttons pull-left">
-        <button class="btn" data-action="delete" disabled="disabled">Delete</button>
+        <button class="btn" data-action="remove" disabled="disabled">Remove Contact(s)</button>
     </div>
     <div class="pull-left ">
         With group:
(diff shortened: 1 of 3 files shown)


justinvdm:
@badlard `Delete` isn't named properly, since we are removing contacts from the group. Changing this to `Remove contacts`. That ok?

On a related note, the action buttons for our 'tables'/'lists' in the UI need some improvement. A bunch of buttons with text will fill up the actions bar pretty quickly, and is a bit unsightly. I think the fact that we need to specify `With group:` in the action bar is one symptom of the bad design here. Maybe we should talk about this with the UX guys that will be onboard soon?

justinvdm:
We still have a button for removing contacts from smart groups, although it doesn't really make sense. It is out of the scope of this ticket, but I've created #666 for it.

justinvdm:
Ready for review

hodgestar:
:+1:

PR #670, opened 2013-08-28
Sequential Send UI disallows specifying message content with line breaks

smn:
Currently a text input field is provided for saving text in.
Sometimes one needs to specify a message with more structure using new lines, to do that we need to use textareas for storing input.


jerith:
:+1:

PR #671, opened 2013-08-28
Remove a bunch of old conversation cruft (part 2)

jerith:
Now that #624 has landed we can clean up a bunch of the now-obsolete helpers and abstractions we have in the conversation code, starting with `get_batches()` and `get_latest_batch()`.

PR #672, opened 2013-08-28
Remove a bunch of old UI cruft

jerith:
There are a bunch of things that were used by the old UI that are no longer called anywhere and can be removed. This work follows on from #671 and friends.


go/vumitools/api.py
```diff
@@ -52,19 +52,6 @@ class TagpoolSet(object):
...
-
-        :param function filter_func:
-            A function f(pool, metadata) that should return True if the
-            pool should be kept and False if it should be discarded.
-        """
-        new_pools = dict((pool, metadata)
-                         for pool, metadata in self._pools.iteritems()
-                         if filter_func(pool, metadata))
-        return self.__class__(new_pools)
-
```
hodgestar, line 67: Not really sure why this is being deleted? Seems useful?

hodgestar:
Other than the question about `select`, :+1:.

go/vumitools/api.py
jerith, line 67: The only thing that called it was the old conversation form that was removed. We can always bring it back (or write a similar thing) if we need it again.

hodgestar:
:+1:

PR #393, opened 2013-08-29
Test cases for new conversation views

jerith:
We currently test all this stuff through the various apps, which means there's a lot of duplication and some things get left out.


jerith:
We still need a few more tests, but I think we can land this and do the rest in a new branch.

smn:
:+1: in general but not convinced the message store / listing tests need to be skipped.

smn:
:+1:

PR #669, opened 2013-08-30
Create billing module (Django models + Twisted Web REST API)

andreit:


smn:
@andreit is this ready for review and @hodgestar will you be reviewing this initially?

andreit:
Hi Simon, Yes. Simon and I discussed it briefly on Tuesday.

hodgestar:
@andreit @smn Yep, I'll be reviewing. Moved the issue into milestone 0.5b2 so it doesn't fall off my radar.

go/billing/admin.py
```diff
@@ -0,0 +1,118 @@
...
+                    'credit_balance')
+
+    search_fields = ('account_number', 'user__email', 'description')
+    readonly_fields = ('credit_balance',)
+    inlines = (ItemInline,)
+    actions = ['load_credits']
+
+    def load_credits(self, request, queryset):
+        selected = request.POST.getlist(admin.ACTION_CHECKBOX_NAME)
+        redirect_url = "credits/load/?ids=%s" % (",".join(selected))
```
hodgestar: Is there a better way to construct this redirect URL?

andreit: This is what the Django admin example looks like in the docs. Will look for a better solution

go/billing/api.py
```diff
@@ -0,0 +1,817 @@
...
+from twisted.internet import defer
+from twisted.web.resource import Resource
+from twisted.web.server import NOT_DONE_YET
+
+from django.contrib.auth.hashers import make_password
+
+from go.billing import settings as app_settings
+from go.billing.utils import RealDictConnectionPool, JSONEncoder
+
+log.startLogging(sys.stdout)
```
hodgestar: Not sure it's a good idea to do this at the module level.

go/billing/utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def closed(self):
+        """Return ``True`` if the underlying connection is closed
+        ``False`` otherwise
+
+        """
+        if self._connection:
+            return self._connection.closed
+        return True
+    closed = property(closed)
```
hodgestar: Maybe use `@property` here since it's a bit cleaner and used elsewhere?

hodgestar:
Btw, I'll be making silly bikeshed comments until I've understood  the structure properly and then I'll be able to make real comments. :)

andreit:
Cool thanks, you can give me a call if something doesn't make sense :)

go/billing/admin.py
andreit: I can just store the selected account ids on the session instead of the URL; Then the URL will just be: "credits/load/". Let me know if you are happy with that.

go/billing/api.py
andreit: Cool, will move it to the runbillingserver management command; just before it starts the reactor

go/billing/forms.py
```diff
@@ -0,0 +1,44 @@
+from decimal import *
```
hodgestar: Eep. Never use "import *" in actual code (for a start it makes various static checks impossible and probably worse it makes things much harder for human readers too).

hodgestar: It looks like this can be replaced with `from decimal import Decimal`?

andreit: Oops, yes.. copy/paste from docs.

go/billing/models.py
```diff
@@ -0,0 +1,129 @@
...
+        return self.account_number
+
+
+class CostOverride(models.Model):
+    """Allows for a per-account cost override"""
+
+    account = models.ForeignKey(Account)
+    tag_pool = models.ForeignKey(TagPool)
+    message_direction = models.CharField(max_length=20,
+                                         choices=BaseCost.DIRECTION_CHOICES)
```
hodgestar: `CostOverride` and `BaseCost` look sort of identical except for the presence of the `account` field and I imagine they'll also have to stay in sync as they evolve. Do you think it would make sense to combine them and just allow `account` to be `null` for the base cost? We should maybe also specify that the `account, tag_pool, message_direction` combination should be unique?

andreit: We can do that. I think it might make the query we use to calculate the message cost simpler too. The advantage of having them separate right now is that it is more intuitive to add cost overrides for an account via the Django admin inline forms, but if that UI will eventually move to the main Vumi Go interface its irrelevant. I have used South migrations for the tables. I don't think Django supports composite keys, perhaps there is a way to add them to the migrations.

hodgestar: Ah! What I was thinking off is only available in 1.5 -- https://docs.djangoproject.com/en/1.5/ref/models/options/#django.db.models.Options.index_together. We can add it once we migrate to 1.5 (i.e. not right now :).

go/billing/api.py
```diff
@@ -0,0 +1,814 @@
...
+    """
+    content_type = request.getHeader('Content-Type')
+    if request.method == 'POST' and content_type == 'application/json':
+        return json.loads(request.content.read(), parse_float=decimal.Decimal)
+    return None
+
+
+class Root(Resource):
+    """The root resource"""
+
```
hodgestar, line 55: I would put all the calls to `.putchild()` in `Root.__init__` and avoid having the global instance (more anti-global comments to come :).

go/billing/api.py
```diff
@@ -0,0 +1,814 @@
...
+from twisted.internet import defer
+from twisted.web.resource import Resource
+from twisted.web.server import NOT_DONE_YET
+
+from django.contrib.auth.hashers import make_password
+
+from go.billing import settings as app_settings
+from go.billing.utils import RealDictConnectionPool, JSONEncoder
+
+_connection_pool = None  # The txpostgres connection pool instance
```
hodgestar: I think `_connection_pool` should be passed into the Root resource constructor and then to the sub-resources (then we can avoid the whole global connection_pool entirely).

go/billing/api.py
```diff
@@ -0,0 +1,814 @@
...
+            SELECT id, email, first_name, last_name
+            FROM auth_user
+        """
+
+        result = yield _connection_pool.runQuery(query)
+        defer.returnValue(result)
+
+    def render_POST(self, request):
+        """Handle an HTTP POST request"""
+        data = _parse_json(request)
```
hodgestar: Probably we need some error handling here so that the server returns a `4xx` error instead of a `500` when sent bad JSON? And in similar places elsewhere.

go/billing/api.py
```diff
@@ -0,0 +1,814 @@
...
+
+            if tag_pool_name:
+                query += " AND t.name = %(tag_pool_name)s"
+                params['tag_pool_name'] = tag_pool_name
+            if message_direction:
+                query += " AND c.message_direction = %(message_direction)s"
+                params['message_direction'] = message_direction
+
+            result = yield _connection_pool.runQuery(query, params)
+        else:
```
hodgestar: I think you can just do `defer.returnValue(result)` here and avoid some `if result` and `result = None` running around.

go/billing/api.py
```diff
@@ -0,0 +1,814 @@
...
+            LIMIT %(limit)s
+        """
+
+        try:
+            offset = int(page_number) * int(items_per_page)
+        except ValueError:
+            offset = 0
+        try:
+            limit = int(items_per_page)
+        except ValueError:
```
hodgestar, line 658: You probably also want to catch `TypeError` here.

go/billing/admin.py
```diff
@@ -0,0 +1,119 @@
...
+            fields=('account_number',), extra=0)
+
+        if request.method == 'POST':
+            formset = CreditLoadFormSet(request.POST, queryset=queryset)
+            if formset.is_valid():
+                for form in formset:
+                    form.load_credits()
+                del request.session['account_ids']
+                messages.info(request, _("Credits loaded successfully."))
+                return HttpResponseRedirect('../../')
```
hodgestar: Is this redirecting to the base admin URL? We should use https://docs.djangoproject.com/en/dev/ref/contrib/admin/#reversing-admin-urls more.

go/billing/management/commands/runbillingserver.py
```diff
@@ -0,0 +1,39 @@
...
+        if len(args) > 0:
+            try:
+                port = int(args[0])
+            except ValueError:
+                raise CommandError("%s is not a valid port number.\n" %
+                                  (args[0],))
+
+        def connection_established(connection_pool):
+            from twisted.web.server import Site
+            reactor.listenTCP(port, Site(api.root))
```
hodgestar: We should use Twisted endpoints here since that allows us to automatically support SSL and IPv6. See http://twistedmatrix.com/documents/current/api/twisted.internet.endpoints.html#serverFromString. The idea would be for the endpoint string to live in config instead of port.

go/billing/templates/admin/billing/account/credit_load_form.html
```diff
@@ -0,0 +1,91 @@
+{% extends "admin/base_site.html" %}
+{% load i18n admin_static admin_modify %}
+{% load url from future %}
+{% load admin_urls %}
+
+{% block extrahead %}{{ block.super }}
+    {% url 'admin:jsi18n' as jsi18nurl %}
+    <script type="text/javascript" src="{{ jsi18nurl|default:"../../../jsi18n/" }}"></script>
```
hodgestar: Do we need this default? If so, we should set its URL properly.

hodgestar: Oh. It looks like this is an entire copy of a base form? Do we need all of this or could we get away with overriding a smaller subset?

go/billing/tests.py
```diff
@@ -0,0 +1,357 @@
+import json
```
hodgestar: Could you move this file to `go/billing/tests/test_api.py`? Ideally we should also have some tests for the other modules (i.e. forms, models, utils, admin) too, but it's not super urgent.

hodgestar:
Done reviewing for now. Looking good generally! :cake:

hodgestar:
We also need to make sure builds pass on travis (hopefully that's just a matter of updating dependencies).

andreit:
Thanks. Will fix these and let you know.

andreit:
@hodgestar looks like the Travis build is failing because the test cases try to create a test postgres database. pytest uses the Django DB settings to figure out what the test db name should be. Is it even possible for Travis to spin up a test postgres db?

jerith:
You can configure it in `.travis.yml`. The db-specific docs for that are at http://about.travis-ci.org/docs/user/database-setup/#PostgreSQL and there's a section at the top of the same page that describes how to tell Travis to install and start postgres.

go/testsettings.py
```diff
@@ -8,6 +8,16 @@
...
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'NAME': 'go',
+        'USER': 'postgres',
+        'PASSWORD': '',
+        'HOST': '',
+        'PORT': '',
+    }
+}
```
hodgestar, line 22: I wonder if there's a way to avoid running against postgres on developers machines (since random developers won't have a 'go' database) but still run against postgres on Travis? How difficult would that be?

go/billing/utils.py
```diff
@@ -0,0 +1,115 @@
...
+        else:
+            return json.JSONEncoder.default(self, obj)
+
+
+def real_dict_connect(*args, **kwargs):
+    kwargs['connection_factory'] = psycopg2.extras.RealDictConnection
+    return psycopg2.connect(*args, **kwargs)
+
+
+class RealDictConnection(txpostgres.Connection):
```
hodgestar: I'm not entirely sure what the `RealDict` in the name refers to?

go/billing/utils.py
```diff
@@ -0,0 +1,115 @@
...
+    @property
+    def closed(self):
+        """Return ``True`` all the connections are closed
+        ``False`` otherwise
+
+        """
+        for c in self.connections:
+            if not c.closed:
+                return False
+        return True
```
hodgestar: You could also use `return all(c.closed for c in self.connections)`.

go/testsettings.py
andreit, line 22: Since we are not using an ORM, we might have some issues with certain SQL queries that use some postgres specific syntax. I will give it some thought.

go/billing/utils.py
andreit: Thanks, will update.

go/billing/utils.py
andreit: It refers to the psycopg2 RealDictConnection which returns each result row as a python dict, which makes it very easy to convert to JSON object for the response.

go/testsettings.py
andreit, line 22: If we want the API to support SQLite we can look into using this ORM (https://github.com/fiorix/txdbapi), but I am afraid it might make the queries less efficient and/or fire multiple queries in places where postgres can achieve the same result in 1 query.

hodgestar, line 22: You could also look at https://pypi.python.org/pypi/alchimia/0.3-dev.

andreit, line 22: This might work. If you are happy with using it in production I will rewrite the API queries. Just keep in mind that in some places I might have to fire off multiple queries where I currently have a single query.

hodgestar, line 22: I think multiple queries in a few places are okay.

go/billing/utils.py
hodgestar: Maybe we should name it `DictRowConnection`? Bike-shedding a bit here. :)

hodgestar:
I've asked @smn to look at a few specific things but other than that and the few issues I've raised in the second round of reviewing, I think this branch is looking good.

andreit:
@hodgestar do you want me to port the queries to alchemia, or shall we do that in a separate issue?

hodgestar:
It's not imperative it happen in this issue, but I think that the test
suite should pass with SQLite before we land this so that developers can
run tests locally.

andreit:
@hodgestar The best way I see this happening is to port to alchemia :) Unfortunately the code is too txpostgres dependent.

hodgestar:
I think on further reflection lets not let the SQLite / alchimia porting hold up landing this branch. I've created a separate ticket for that work -- let's just make sure it happens very shortly after this lands.

andreit:
I agree and will be more than happy to port the code to alchemia once we have the billing module working

hodgestar:
@andreit I think mostly was is left is to fix the Travis tests. One of the failures looks billing related. The others look more obscure (maybe urls.py for billing message up some conversation routes somehow? maybe a merge went wrong). Shout if you'd like to me look at any of them.

andreit:
@hodgestar there are still some failing tests, they don't seem to be billing related though.

hodgestar:
These tests pass on my local machine using `SQLite` but fail using `postgres` so it looks like something we're relying on in these tests doesn't play nicely with postgres.

hodgestar:
:+1:

andreit:
@hodgestar Awesome!!! Thanks.

go/billing/models.py
```diff
@@ -0,0 +1,114 @@
...
+    credit_balance = models.IntegerField(default=0)
+    alert_threshold = models.DecimalField(
+        max_digits=10, decimal_places=2, default=0.0,
+        help_text=_("Low-credits notification will be sent when the "
+                    "credit balance reaches the alert threshold percentage"))
+
+    alert_credit_balance = models.IntegerField(default=0)
+
+    def __unicode__(self):
+        return self.account_number
```
smn: Bike shed: I'd return a combination of `self.user` and `self.account_number` here, primarily because no-one's going to remember the account_number `uuid4().hex` but chances are the user's username is memorable because it should match their email address.

Feel free to ignore if you disagree or have burnt enough hours on this PR as is.

go/billing/models.py
```diff
@@ -0,0 +1,114 @@
...
+    account = models.ForeignKey(Account, blank=True, null=True)
+    tag_pool = models.ForeignKey(TagPool)
+    message_direction = models.CharField(max_length=20,
+                                         choices=DIRECTION_CHOICES)
+
+    message_cost = models.IntegerField(
+        default=0, help_text=_("The base message cost in cents."))
+
+    markup_percent = models.DecimalField(
+        max_digits=10, decimal_places=2, default=0.0,
```
smn, line 61: I think @badlard has a different default markup. It may be a good idea to keep that as a default rather?

go/billing/models.py
```diff
@@ -0,0 +1,114 @@
...
+        help_text=_("The markup percentage."))
```
smn: More bikeshedding: I'd add an example value here. I was running the `resulting_price()` function below with sample `markup_percent` values like `0.2` and the resulting value just didn't make sense until I realized it was expecting `markup_percent` to be something like `20.0`

go/billing/models.py
andreit: Cool, will update.

go/billing/models.py
hodgestar, line 61: I'd like this default to stay zero -- there is already a way of setting global defaults in the database. Details of Praekelt's finances don't really belong in the Vumi Go repository.

smn, line 61: You're completely right. Burn it!

burn the shed

go/testsettings.py
```diff
@@ -8,6 +8,16 @@
     'FAKE_REDIS': 'sure',
 }
 
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'NAME': 'go',
+        'USER': 'postgres',
+        'PASSWORD': '',
```
hodgestar, line 18: Can we perhaps set the username and password here to `go` and not have the database accessed as the `postgres` user? It will make running the tests as a developer easier while we wait for the `alchimia` branch to land. This will require a change to `.travis.yml` too.

hodgestar:
I think @smn's comments have been addressed. I added one last request to have the database user changed to `go` rather than relying on logging in as a user named postgres (which will likely to annoying for local development).

hodgestar:
I've just merged in develop to check that tests still pass. Assuming they do, let's land this. :)

andreit:
The 2.7 build failed for some reason. How do I tell Travis to rebuild?

hodgestar:
There's a button on the Travis CI page -- I've clicked it.

PR #246, opened 2013-06-27
Run tests against real redis on Travis

smn:

PR #684, opened 2013-09-04
Clean up test classes

jerith:
We're constructing test classes from mixins in a bunch of places when we should just be using some preconstructed base classes. (This work is factored out of some stuff I started to do in #246.)

### DIFF
--- a/go/api/go_api/tests/test_session.py
+++ b/go/api/go_api/tests/test_session.py
@@ -1,14 +1,15 @@
 """Tests for go.api.go_api.session."""
 
-from unittest import TestCase
-
 from go.api.go_api.session import SessionStore, CreateError
-from go.vumitools.tests.utils import GoPersistenceMixin
+from go.vumitools.tests.utils import GoTestCase
 
 
-class SessionStoreTestCase(TestCase, GoPersistenceMixin):
+class SessionStoreTestCase(GoTestCase):
     def mk_session_store(self, session_key=None):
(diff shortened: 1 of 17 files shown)


go/api/go_api/tests/test_session.py
```diff
@@ -1,14 +1,15 @@
...
+from go.vumitools.tests.utils import GoTestCase
 
 
-class SessionStoreTestCase(TestCase, GoPersistenceMixin):
+class SessionStoreTestCase(GoTestCase):
     def mk_session_store(self, session_key=None):
         ss = SessionStore(session_key)
+        # We don't grab the Riak manager here because we don't use Riak in
+        # these tests.
+        self._persist_redis_managers.append(ss.vumi_api.redis)
```
hodgestar, line 12: Why do we have to manually grab the `redis` store here? And anyway to avoid it?

hodgestar:
Other than the one question, looks good.

go/api/go_api/tests/test_session.py
jerith, line 12: We have to do that everywhere we create a VumiApi object, it's just usually hidden inside test helpers.

hodgestar, line 12: Cool. Would be nice to hide all of this somehow but let's worry about that some other time.

hodgestar:
:+1:

PR #602, opened 2013-09-04
Routing page connectors sometimes get messed up.

justinvdm:
On the routing page: http://qa-vumi.za.prk-host.net/routing/
I move a conversation block from the button of the right hand column to the top
I change connectors from one end point to another
I hit the 'save' button
Nothing seems to happen
I go back to the dashboard and then back to the routing page, and my changes are lost.

screen shot 2013-08-08 at 3 09 29 pm


justinvdm:
We need to wait for our jsPlumb fork's [#3](https://github.com/praekelt/jsPlumb/pull/3) and [#4](https://github.com/praekelt/jsPlumb/pull/4) to land and for the build to be included here, but otherwise, ready for review.

go/base/static/js/test/runner.html
```diff
@@ -14,7 +14,7 @@
     <script src="../vendor/jquery-1.9.1.js"></script>
     <script src="../vendor/jquery.cookie-1.3.1.js"></script>
     <script src="../vendor/jquery.ui-1.10.3.js"></script>
-    <script src="../vendor/jquery.jsPlumb-1.4.1.leakfix.js"></script>
+    <script src="../vendor/jquery.jsPlumb-1.5.2.js"></script>
```
hodgestar: We should maybe mark this as our own special release somehow.

hodgestar: Perhaps .praekelt.js?

hodgestar:
Other than the small comment about the new filename, :+1:.

go/base/static/js/test/runner.html
justinvdm: Good idea :)

hodgestar:
:+1:

PR #685, opened 2013-09-04
Improve usability of routing screen - POA200975

justinvdm:
It is quite difficult to get a hold of connections to drag them around. I aim to improve this a bit in the PR to be attached to this ticket.

There are more serious usability problems, the most apparent being the spaghetti monster currently living on the routing screen. This problem is way more involved, though, so I'm leaving that for some time in the near future.

### DIFF
--- a/go/base/static/css/routing.less
+++ b/go/base/static/css/routing.less
@@ -1,6 +1,7 @@
 .routing {
   #diagram {
     .column-container {
+      .user-select(none);
       text-align:center;
 
       .title {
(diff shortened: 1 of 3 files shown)


go/base/static/css/routing.less
```diff
@@ -1,6 +1,7 @@
 .routing {
   #diagram {
     .column-container {
+      .user-select(none);
```
hodgestar, line 4: Nice!

hodgestar:
:+1:

PR #687, opened 2013-09-04
Clicking 'save' btn on survey screen does not direct user to 'show' page

justinvdm:
- I set up my survey screens on the survey creator page
- I clicked the 'save' btn on the survey creator page
- I remained on the survey creator page once I received the "save successful" notification

screen shot 2013-09-04 at 4 39 16 pm
I should be directed to the 'show' screen automatically once the save has completed successfully

### DIFF
--- a/go/base/static/js/src/apps/dialogue/views.js
+++ b/go/base/static/js/src/apps/dialogue/views.js
@@ -23,8 +23,7 @@
       });
 
       this.listenTo(this.save, 'success', function() {
-        // TODO better way of specifying urls than 'hardcoding' them
-        go.utils.redirect('/conversations/' + this.model.id + '/');
+        go.utils.redirect(this.model.get('urls').get('show'));
       });
 
       go.apps.dialogue.style.initialize();
--- a/go/base/static/js/src/apps/dialogue/views.js
+++ b/go/base/static/js/src/apps/dialogue/views.js
@@ -22,6 +22,11 @@
         useNotifier: true
       });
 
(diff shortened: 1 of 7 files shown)


hodgestar:
While we're here we should also check that `Cancel` does something (i.e. goes back to the show screen). I know that `Cancel` doesn't for the routing editor.

go/base/static/js/src/apps/dialogue/views.js
```diff
@@ -22,6 +22,11 @@
         useNotifier: true
       });
 
+      this.listenTo(this.save, 'success', function() {
+        // TODO better way of specifying urls than 'hardcoding' them
+        go.utils.redirect('/conversations/' + this.model.id + '/');
```
justinvdm: I'm not too happy with having this 'hardcoded' here. The alternatives I can think of are:
- Giving the url as part of the model data in the api response
- Set up some infrastructure on the client side for keeping track of urls (although this would mean updating these urls whenever they change on the server side).

Thoughts/ideas?

hodgestar: I think adding it to the model data is the cleanest for now.

justinvdm: Cool, agreed. Should we do that before this PR lands?

hodgestar: I think so.

justinvdm:
Ready for re-review

hodgestar:
:+1:

PR #676, opened 2013-09-04
Remove check boxes on channels page

hodgestar:
- On the channels page: https://go.vumi.org/channels/
- selecting the checkboxes and then clicking the 'release' button in the header doesn't do anything.
- this is freaking out our beta users
- let's kill the checkboxes and release button and remove them from the UI

screen shot 2013-08-29 at 4 55 22 pm


hodgestar:
I'll change this to match the multi-button thing we use for campaigns.

hodgestar:
I want to factor out the common code among all the action classes but I'm leaving that for #688.

go/channel/templates/channel/show.html
```diff
@@ -28,3 +25,13 @@
...
 </div>
 {% endblock %}
+
+{% block ondomready %}
+    var actions = new go.channel.show.ChannelActionsView({
+        el: '.content .actions .right'
+    });
+
+    $('.content .sub-actions .disabled').popover({placement: 'left'});
+    $('.content .sub-actions .disabled').on('click', function() { return false; });
```
justinvdm: I'm not too sure if these lines are needed. I also see them in the router template, which was probably what was being used as a reference, but I'm wondering if both can go?

justinvdm: Ah, I see these lines are in the conversation template too. That template has a `.sub-actions` element, so maybe its only relevant to that template?

hodgestar: They're not needed but I was intentionally keeping them for consistency (and to save someone the trouble of adding them if they added an action that needed them).

justinvdm: I see. I think we should be avoiding returning `false` in ui event callbacks (it can have unintended side-effects), and should use `e.preventDefault` instead. Since this is scattered all over the place, maybe we should leave that for a different PR.

justinvdm:
:+1:

justinvdm:
Another :+1: :)

PR #690, opened 2013-09-05
Better Redis and Riak cleanup in tests

jerith:
More yak shaving for #246.

### DIFF
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -515,6 +515,11 @@ class VumiApi(object):
         self.session_manager = SessionManager(
             self.redis.sub_manager('session_manager'))
         self.mapi = sender
+        self._persist_collect_hook(self)
+
+    def _persist_collect_hook(self, vumi_api):
+        # This allows us to collect riak/redis managers for cleanup in tests.
+        pass
 
     @staticmethod
     def _parse_config(config):
--- a/go/vumitools/tests/utils.py
+++ b/go/vumitools/tests/utils.py
@@ -13,7 +13,7 @@ from vumi.message import TransportEvent
 from vumi.application.tests.test_base import ApplicationTestCase
 from vumi.tests.utils import VumiWorkerTestCase, PersistenceMixin
 
-from go.vumitools.api import VumiApiCommand
+from go.vumitools.api import VumiApiCommand, VumiApi
 from go.vumitools.account import UserAccount, RoutingTableHelper
 from go.vumitools.contact import Contact, ContactGroup
 from go.vumitools.utils import MessageMetadataHelper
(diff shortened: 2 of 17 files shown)


go/vumitools/api.py
```diff
@@ -515,6 +515,11 @@ def __init__(self, manager, redis, sender=None):
         self.session_manager = SessionManager(
             self.redis.sub_manager('session_manager'))
         self.mapi = sender
+        self._persist_collect_hook(self)
+
+    def _persist_collect_hook(self, vumi_api):
+        # This allows us to collect riak/redis managers for cleanup in tests.
+        pass
```
hodgestar, line 522: I think we can avoid this test hook by just patching the class with a function that remembers the API creation and then returns an api object?

go/vumitools/tests/utils.py
```diff
@@ -100,14 +119,23 @@ def _list_accounts(self, manager):
...
+            # clear properties from their associated buckets.
+            accounts = yield self._list_accounts(manager)
+            yield manager.purge_all()
+            # This must happen after the objects are deleted, otherwise the
+            # indexes don't go away.
+            yield self._clear_bucket_properties(accounts, manager)
+        # Hackety hack. Attempt to properly close any lingering connections so
+        # Riak doesn't barf on OSX where it's limited to 1024 file descriptors
+        # in its select() loop.
+        manager.client = None
```
hodgestar, line 138: Maybe riak managers should also get a close method?

hodgestar, line 138: Happy for us just to start a Vumi ticket for this rather than delaying this PR.

hodgestar:
Other than the suggestion about the testing hook, looks good.

go/vumitools/api.py
jerith, line 522: I don't understand what you're suggesting.

go/vumitools/tests/utils.py
jerith, line 138: Riak clients don't have a close method. I plan to push that line of code up into vumi at some point, but it's only the vumi-go tests that get hurt by this at the moment.

hodgestar, line 138: Yes, I was suggesting we add one to riak _managers_.

go/vumitools/api.py
hodgestar, line 522: I'm suggestion we patch go.vumitools.api.VumiApi with something like:

``` python

def _collect_api(self, *args, **kws):
     api = self._orig_api_cls(*args, **kws)
     self._apis.append(api)
     return api
```

jerith, line 522: That doesn't work well when modules do `from go.vumitools.api import VumiApi`.

hodgestar, line 522: Hmm. What about a general utility like

``` python
class RecordCreate(object):
    def __init__(self):
        self._clses = {}

    def attach(self, cls, func):
        orig_init = cls.__init__

        def wrap(self, *args, **kws):
            orig_init(self, *args, **kws)
            func(self)
            return

        self._clses[id(cls)] = orig_init
        cls.__init__ = wrap

    def detach(self, cls):
        cls.__init__ = self._clses[id(cls)]
```

Used as:

``` python

class A(object):
    def __init__(self):
        print "A.__init__"

def prnt(obj):
    print "REC:", obj

r = RecordCreate()
r.attach(A, prnt)
A()
r.detach(A)
A()
```

jerith, line 522: Maybe. I'd rather not go down that rabbit hole today, though.

hodgestar, line 522: I can land a commit with this if you like and then you can review it?

hodgestar, line 522: Hmm. Maybe I'd be happier if VumiApi just had a class variable called "_on_create" that was set to "None" and was overridden by tests. That seems to be a fairly common sort of idiom in Twisted.

hodgestar, line 522: Maybe in future it would be cool to have a sort of persistence manager object that classes could have a default version of and that tests could override with their own? Then it'd work a bit like `clock()` does in quite a few places in Twisted.

hodgestar:
:+1:

PR #577, opened 2013-09-05
Script for setting up JavaScript environment

justinvdm:
We're doing a lot of things in JavaScript at the moment, but we don't have an easy way to set up the required environment to run or test our JS code.

### DIFF
--- a/globals.npm
+++ b/globals.npm
@@ -1,0 +1,3 @@
+mocha@~1.12.1
+grunt-cli@~0.1.9
+yuglify@0.1.4
--- a/paths.yml
+++ b/paths.yml
@@ -1,0 +1,3 @@
+jsbox_apps:
+  tests:
+    - 'go/apps/**/test_vumi_app.js'
--- a/paths.yml
+++ b/paths.yml
@@ -1,3 +1,3 @@
-jsbox_apps:
-  tests:
+tests:
+  jsbox_apps:
     - 'go/apps/**/test_vumi_app.js'
--- a/paths.yml
+++ b/paths.yml
@@ -1,3 +1,0 @@
-tests:
-  jsbox_apps:
-    - 'go/apps/**/test_vumi_app.js'
--- a/utils/js-env-setup.sh
+++ b/utils/js-env-setup.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
 
-npm install --prefix=$VIRTUAL_ENV -g `cat globals.npm | tr '\n' ' '`
+PROJECT_ROOT="`dirname $0`/../"
+npm install --prefix=$VIRTUAL_ENV -g `cat $PROJECT_ROOT/globals.npm | tr '\n' ' '`
 npm install
--- a/utils/js-env-setup.sh
+++ b/utils/js-env-setup.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+npm install --prefix=$VIRTUAL_ENV -g `cat globals.npm | tr '\n' ' '`
+npm install
--- a/utils/js-env-setup.sh
+++ b/utils/js-env-setup.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
+
 npm install --prefix=$VIRTUAL_ENV -g `cat globals.npm | tr '\n' ' '`
 npm install
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -12,6 +12,10 @@ module.exports = function (grunt) {
     }
   });
 
+  grunt.registerTask('test:jsbox_apps', [
+    'mochaTest:jsbox_apps'
+  ]);
+
   grunt.registerTask('test', [
     'mochaTest'
   ]);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,7 +4,7 @@ module.exports = function (grunt) {
   grunt.loadNpmTasks('grunt-mocha-test');
 
   grunt.initConfig({
-    paths: require('paths.yml'),
+    paths: require('js_paths.yml'),
     mochaTest: {
       jsbox_apps: {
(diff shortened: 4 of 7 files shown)


justinvdm:
Ok so I'm trying to get the npm packages that need to be installed globally (eg, mocha) to work with virtualenv. We have two options:
1. Have a seperate list of npm packages that need to be installed globally, then do an `npm install -g --prefix=$VIRTUAL_ENV`. The problem with this is that updating the packages requires manual intervention (we can't just do an `npm update`).
2. Have the 'global' packages as `devDependencies` in the `package.json` and install them locally with `npm install`. Then, we will have the binaries we need available in `node_modules/.bin`. We can have the js env setup script symlink all the files in `node_modules/.bin` to `$VIRTUAL_ENV/bin`. The problem with this is that we need to maintain those symlinks, and that running the script multiple times will give errors about failed symlinks (since the files already exist).

Both options have downsides, so I'd like some input on this before going further. Thoughts?

hodgestar:
I think I kinda like option 1 more -- everything in the same virtualenv seems kinda cool and neat. What if we add a script that calls npm and adds --prefix to the virtualenv bin folder?

justinvdm:
Cool, option 1 it is :) Erm, I'm not sure I understand your suggestion, could you explain?

justinvdm:
Ok, so I haven't added the env setup stuff I will be doing for the client side scripts to this PR. It will involve pulling in karma, and a few other things, so I'd rather the basics were down before I get into that stuff.

justinvdm:
So, I think it might make sense to add a note here on how to do stuff with the js parts of Go once this lands (we should probably add it to the readme some time soon).

We are using [grunt](http://gruntjs.com/) to manage our tasks for our js parts. Currently, the only tasks will be testing tasks, but hopefully we can eventually include linting, concatenating, minifying, client-side template compiling, etc. Django Pipeline is currently doing the concatenating, minifying and client-side template compiling. However, we've come across a few problems (for eg, we haven't gotten minification working yet). It might be worthwhile doing this stuff with grunt eventually, so all our js tasks are done in one place.

Ok, so `setup_env.sh` will first set up the js environment before doing what it used to do. If you just want to set up the js environment, use `utils/js-env-setup.sh`.

To run all the tests (which are currently just the jsbox tests, although this will change soon):

```
$ grunt test
```

To run only specific tests (for eg, just the jsbox app tests, or just the client side tests), specify the task name:

```
$ grunt test:jsbox_apps
```

Just running `grunt` will run the default task, which is currently just the tests, although it will soon include other tasks (like linting).

justinvdm:
Ready for review.

globals.npm
```diff
@@ -0,0 +1,3 @@
+mocha@~1.12.1
+grunt-cli@~0.1.9
+yuglify@0.1.4
```
hodgestar, line 3: I think perhaps this file should go in the `setup_env` folder?

paths.yml
```diff
@@ -0,0 +1,3 @@
+tests:
+  jsbox_apps:
+    - 'go/apps/**/test_vumi_app.js'
```
hodgestar: Could we give this file a more descriptive name? Or maybe just put the path directly into the grunt file?

utils/js-env-setup.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+npm install --prefix=$VIRTUAL_ENV -g `cat globals.npm | tr '\n' ' '`
```
hodgestar: This should probably search for globals.npm relative to the script file rather than the current directory. E.g.

``` bash
SETUP_ENV_DIR="`dirname $0`/setup_env"

npm install --prefix=$VIRTUAL_ENV -g `cat "$SETUP_ENV_DIR"/globals.npm | tr '\n' ' '`
```

hodgestar:
Looks good -- just a few niggly comments.

paths.yml
justinvdm: The reason this is not directly in the Gruntfile is because we need the paths in a bunch of places (`settings.py`, `karma.conf.js`).

globals.npm
justinvdm, line 3: All our other dependencies are in the project root, I'm not sure how I feel about putting them in `setup_env`.

utils/js-env-setup.sh
justinvdm: Good point, /me fixes

justinvdm: Oh, actually I'm not sure if we should do this if we are keeping `globals.npm` in the project root. If are, I'm not sure if the change still makes sense (correct me if I'm wrong)?

Also, `js-env-setup.sh` is in `utils/` currently. Should I move it to the project root?

paths.yml
hodgestar: Ah, I see -- I need to look at the later PRs too. In that case, can we just have a more descriptive name?

utils/js-env-setup.sh
hodgestar: I think it still makes sense -- one might run js-env-setup.sh from somewhere odd. Happy for it to stay in utils (that just means the SETUP_ENV_DIR will be set differently).

paths.yml
justinvdm: Ok cool, good idea. Any suggestions? `js_paths`?

utils/js-env-setup.sh
justinvdm: Cool, makes sense, \me fixes.

paths.yml
hodgestar: Sounds good? (assuming you meant js_paths.yml). Is this used for things other than tests? If not, in favour of `js_test_paths.yml`.

hodgestar:
:+1: -- we can figure out how to make globals.npm nice another day.

PR #692, opened 2013-09-06
Use karma for our client side testing

justinvdm:
Whenever we add a new client side script, we need to manually add it in both `settings.py`, and in our test `runner.html` file. Keeping the paths in these different places in sync is problematic, so the current plan is to specify all our js paths we need to know about in a single yaml file. In order to do this, we need to replace what we currently using to run our tests (mocha-phantomjs) with [karma](http://karma-runner.github.io/0.10/index.html). We will still be using mocha and phantomjs, so the only thing that is changing is the way we run the tests.

Karma also has a few other things we can make use of later, like running our tests in chrome and firefox, and will make adding coverage for our client side tests easier.


justinvdm:
Depends on #577, but otherwise, ready for review.

Gruntfile.js
```diff
@@ -1,13 +1,41 @@
...
+        src: ['<%= paths.tests.jsbox_apps.spec %>'],
+      }
+    },
+    jst: {
+      options: {
+        processName: function(filename) {
+          // process the template names the arb Django Pipelines way
+          return filename
+            .replace('go/base/static/templates/', '')
+            .replace(/\..+$/, '')
```
hodgestar: What are we replacing here? (I know what the regex does -- I'm just wondering what it's needed for).

justinvdm: We are removing the file extensions.

go/settings.py
```diff
@@ -354,14 +361,17 @@ def abspath(*args):
 
 
 # PIPELINES CONFIGURATION
+paths = yaml.safe_load(open(os.path.join(PROJECT_ROOT, '..', 'go_paths.yml')))
```
hodgestar, line 364: Little bit worried about how this and `static_paths` interact with production and/or `collectstatic`. Probably `static_paths` needs to use `STATIC_ROOT` or something? Although when one hasn't run `collectstatic` (e.g. in dev mode) it might get more complicated.

Gruntfile.js
justinvdm: This stuff is a direct copy of what I was using in the runner.html. Now that we have node stuff available, I'll use `path.basename`, seems to show what I'm doing here a bit clearer.

hodgestar: Woot.

go/settings.py
hodgestar, line 364: Ah. This is used by PIPELINE_JS which always refers to paths within the Python package. Now just a bit worried that `PROJECT_ROOT` is not the correct thing to look for paths relative too.

hodgestar, line 364: New plan from IRC: Replace this settings.py stuff with grunt tasks.

hodgestar:
:+1:

hodgestar:
I wouldn't mind keeping some sort of comment about why template names are the way they are (even if we don't mention Django pipelines). Otherwise why we're doing this is a bit obscure. Otherwise +1.

PR #691, opened 2013-09-06
Use bower for our client side dependencies

justinvdm:
Currently, we are manually downloading each client side dependency script each time we need to add or update one. We are in need of a package manager for these dependencies, which is exactly what [bower](http://bower.io) does.


justinvdm:
The bower README advises that the bower_components (its like node_modules) dir gets added to project's repo. Our bower_components dir is 11M. I'm going to go against their advice, since our js env setup script does a `bower install`, which will install the bower components. This will save people from painfully slow clones of our repo. Thoughts?

justinvdm:
Depends on #692 and (transitively) on #577, but otherwise, ready for review.

bower.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "vumi-go",
+  "version": "0.0.0",
```
hodgestar: Maybe this should match Vumi Go's current version number (i.e. 0.5.0a)?

js_paths.yml
```diff
@@ -20,17 +20,17 @@ client:
...
+      - 'go/base/static/bower_components/jquery/jquery.js'
+      - 'go/base/static/bower_components/jquery.cookie/jquery.cookie.js'
+      - 'go/base/static/bower_components/jsPlumb/dist/js/jquery.jsPlumb-1.5.2.js'
+      - 'go/base/static/bower_components/jquery-ui/ui/jquery-ui.js'
+      - 'go/base/static/bower_components/js-base64/base64.js'
+      - 'go/base/static/bower_components/node-uuid/uuid.js'
+      - 'go/base/static/bower_components/backbone/backbone.js'
+      - 'go/base/static/bower_components/backbone-relational/backbone-relational.js'
+      - 'go/base/static/bower_components/bootstrap/js/bootstrap.js'
+      - 'go/base/static/bower_components/bootbox/bootbox.js'
```
hodgestar, line 33: So these now refer to paths not in the repo?

hodgestar:
Do we need something to call `bower install` when installing in production?

justinvdm:
@hodgestar Ya, in production, we will need to run `utils/js-setup-env.sh`, which will do the bower installing.

bower.json
justinvdm: Maybe. It is also listed as `0.0.0` in `package.json`, so I'll change both if we decide to do that. My only worry is that bower and npm both require the versioning to be semantic versioning. I guess I'm okay with using whichever versioning we are using anyways, but it does break the way bower and npm upgrade packages. We probably won't be publishing for either, but I'm still a bit concerned. What do you think?

js_paths.yml
justinvdm, line 33: Ya, we need to run `bower install` (or `utils/js-setup-env.sh`) before these will be available. I mentioned something about this [here](https://github.com/praekelt/vumi-go/pull/691#issuecomment-23949286). What are your thoughts?

bower.json
hodgestar: `package.json` is also wrong, I think. :)

hodgestar: The version in setup.py should be canonical. We should maybe also update `utils/bump-version.sh`.

justinvdm: Erm, I've got a feeling you missed [my reply comment](https://github.com/praekelt/vumi-go/pull/691/files#r6286757). Just to be sure, did you see this?

hodgestar:
:+1:

PR #681, opened 2013-09-09
SMS confirmation messages should not be displayed in the Inbox

jerith:
From Rachel:

Hey Ben,

Quick UI note - previously, Simon had done the code so you couldn't see the message sent for confirmation on the platform (since it's a security measure, showing it on the platform defeats the purpose) - on the new UI, it shows up as a sent message so you can see the URL on the platform. 

Cheers,
Rachel

screen shot 2013-09-02 at 11 14 58 am

### DIFF
--- a/go/conversation/view_definition.py
+++ b/go/conversation/view_definition.py
@@ -221,11 +221,11 @@ class MessageListView(ConversationTemplateView):
         inbound_message_paginator = Paginator(
             PagedMessageCache(conversation.count_replies(),
                 lambda start, stop: conversation.received_messages(
-                    start, stop, batch_id)), 20)
+                    start, stop)), 20)
         outbound_message_paginator = Paginator(
             PagedMessageCache(conversation.count_sent_messages(),
-                lambda start, stop: conversation.sent_messages(start, stop,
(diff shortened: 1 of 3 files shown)


jerith:
I think I missed this when I cleaned up all the conversation batch code. We need to get better about test coverage. :-/

go/conversation/view_definition.py
```diff
@@ -221,11 +221,11 @@ def get(self, request, conversation):
         inbound_message_paginator = Paginator(
             PagedMessageCache(conversation.count_replies(),
                 lambda start, stop: conversation.received_messages(
-                    start, stop, batch_id)), 20)
+                    start, stop)), 20)
```
justinvdm, line 224: I'm sure there is a valid reason for this change, but just for my understanding, could you explain why we need it (and the change below)?

justinvdm, line 224: Ah ok, this one was explained on IRC:

> 21:25 < jerith> The bug was that we passed the batch key into the message list methods which used them as the "show sensitive messages" arg.

justinvdm:
:+1:

PR #661, opened 2013-09-11
There is no way to add a new smart group.

smn:
There is currently no way to add a new smart group.


smn:
Added the group name. Frontend ppl will have to decide on styling but at least now we don't get lost trying to figure out which group we're looking at.

vumi_go

hodgestar:
:+1:

PR #647, opened 2013-09-11
Display timestamp in "Last edited" column

smn:
On the dashboard, under the "Last edited" column we currently show: weeks, days, since the last edit.

Instead, let's show the timestamp of the last edit in the format (written month, day, year) e.g. October 4, 2009

screen shot 2013-08-20 at 2 54 03 pm


smn:
No either shows "today", "yesterday" or the date. If one hovers over the value you get the exact timestamp in the tooltip.

fullscreen_2013_09_11_10_22_am

hodgestar:
:+1:

PR #704, opened 2013-09-12
Add js setup to fab

justinvdm:


fabfile.py
```diff
@@ -9,6 +9,7 @@ def deploy_go():
         sudo("find go -name '*.pyc' -delete")
         _venv_command('./ve/bin/django-admin.py collectstatic --pythonpath=. '
                       '--settings=go.settings --noinput')
+        sudo('./utils/js-setup-env.sh')
```
hodgestar: This probably needs a `user="vumi"` at the end (the way _venv_command does). It might also need to be  run with _venv_command because it uses the `VIRTUAL_ENV` environment variable.

justinvdm:
Ready for re-review

hodgestar:
:+1:

PR #698, opened 2013-09-15
Unable to export vxpolls data for survey with lots of users

smn:
It takes too long to collect all the data and haproxy kills the web request. We should turn this into a celery task that sends an email (or redirects somewhere once the download is ready).

### DIFF
--- a/go/apps/surveys/definition.py
+++ b/go/apps/surveys/definition.py
@@ -26,7 +26,11 @@ class SendSurveyAction(ConversationAction):
 class DownloadUserDataAction(ConversationAction):
     action_name = 'download_user_data'
     action_display_name = 'Download User Data'
-    redirect_to = 'user_data'
+
+    def perform_action(self, action_data):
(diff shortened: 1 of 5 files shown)


smn:
Not quite ready for review yet, tests are missing. @badlard and @hodgestar this addresses the issue raised for the GHR project.

hodgestar:
Yay! Tx.

smn:
Making sure `SurveyTestCase.test_export_user_data` fails, leaving this for monday.

smn:
Ok tests passing ready for review.
I was a bit surprised I had to use HTTP POST for the action, but I'm assuming the UI does that by default(?). Leaving that detail for the reviewer.

hodgestar:
:+1:. Having to POST actions is expected (GET's shouldn't trigger actions).

smn:
Right, this triggers an email. For some reason I was still reasoning about it being just a download, in which case a GET could make sense.

PR #564, opened 2013-09-16
Clean up routing table helpers and API

jerith:
The current routing table helpers are a bit clumsy and we have to manually wrap them around routing tables. We can do better.


go/channel/views.py
```diff
@@ -96,12 +96,10 @@ def new_channel(request):
...
     # TODO: Replace this with a real thing when we have channel models.
-    pool, _, tag = channel_key.partition(':')
-    tags = user_api.list_endpoints()
-    if (pool, tag) not in tags:
-        raise Http404
-
-    return user_api.get_channel((pool, tag))
+    for channel in user_api.active_channels():
+        if channel.key == channel_key:
+            return channel
```
hodgestar, line 101: Shouldn't this rather do a `.get_channel()` and check for `None`?

go/vumitools/routing_table.py
```diff
@@ -0,0 +1,403 @@
...
+        source_connector ->
+            source_endpoint_1 -> [destination_connector, destination_endpoint]
+            source_endpoint_2 -> [..., ...]
+
+    in order to make storing the mapping as JSON easier (JSON keys cannot be
+    lists).
+    """
+
+    def __init__(self, routing_table=None):
+        # XXX: Kill this check later.
```
hodgestar: Anything preventing us from killing this now?

go/vumitools/routing_table.py
```diff
@@ -0,0 +1,403 @@
...
+        self.remove_connector(tag_conn)
+
+    def add_oldstyle_conversation(self, conv, tag, outbound_only=False):
+        """XXX: This can be removed when old-style conversations are gone."""
+        conv_conn = str(GoConnector.for_conversation(conv.conversation_type,
+                                                     conv.key))
+        tag_conn = str(GoConnector.for_transport_tag(tag[0], tag[1]))
+        self.add_entry(conv_conn, "default", tag_conn, "default")
+        if not outbound_only:
+            self.add_entry(tag_conn, "default", conv_conn, "default")
```
hodgestar: I think we might be able to get rid of `add_oldstyle_conversation`?

go/channel/views.py
hodgestar, line 101: Maybe this is fine for now.

hodgestar:
Two extra things I think we can remove now but otherwise, :+1:.

go/vumitools/routing_table.py
jerith: Nope, I just forgot about it.

go/apps/tests/base.py
```diff
@@ -52,8 +52,18 @@ def setup_conversation(self, started=False, with_group=False,
...
+            channel = self.user_api.get_channel(tag)
+            user_account = self.user_api.get_user_account()
+            rt = user_account.routing_table
+            rt.add_entry(
+                self.conversation.get_connector(), 'default',
+                channel.get_connector(), 'default')
+            rt.add_entry(
+                channel.get_connector(), 'default',
+                self.conversation.get_connector(), 'default')
+            user_account.save()
```
hodgestar: Could we make the channel addition a separate method (even if it's just an _method on the same class) so it'll be easier to see the intention?

hodgestar:
Re-reviewed.

hodgestar:
:+1:

PR #713, opened 2013-09-17
HTTP API worker's PUT features should be splittable into a separate worker process.

vgeddes:
Because of design limitations of the current HTTP API (largely due to our AMQP Consumer base class not making it easy to create auto-deleting queues) it is not safe to run multiple instances of the HTTP API vumi_app worker.

As a side-effect all handling of `PUT` replies by 3rd party apps also are currently funnelled through this single worker.

The concurrency limitations only apply for the streaming GETs, not for any of the PUT replies. We should be able to run multiple HTTP API workers that only accept PUTs on multiple servers to be able to load balance the replies or outbound messages coming from 3rd party applications.

:cake: and :cookie: for @vgeddes for pointing this out.


go/apps/http_api/resource.py
```diff
@@ -46,7 +46,7 @@ def get_conversation(self, user_account, conversation_key=None):
         return user_api.get_wrapped_conversation(conversation_key)
 
 
-class StreamResource(BaseResource):
+class OutgoingMessageResource(BaseResource):
```
smn: I'm not against renaming but I do think keeping the word `Stream` in there is important as that's all this resource does.

smn:
Can you attend to the build failure in Travis and write tests to cover the new `ImcomingHTTPWorker`?

vgeddes:
yup will do, that's next on my todo for this issue.

smn:
Cool, apologies if I reviewed this too early!

smn:
@hodgestar would you mind have a look at this one as well? It splits a PUT capable resource into a separate worker. 

(cc @sheetalgordhan & @highvoltage -- this is for GFM)

hodgestar:
@vgeddes @smn Quick high-level question: what is this setup supposed to look like in production? I.e. is there a dispatcher for inbound / outbound messages? Where do conversation command messages and metrics processing messages go? Will we need some new sort of dispatcher?

smn:
Before I provide my 2c, @vgeddes what are your thoughts?

vgeddes:
From a very high level, I suggest we run two `InboundHttpWorker` workers on each Go host, and have HAProxy load balance between them. This would provide a measure of redundancy. We would need to make sure that HAProxy correctly routes `PUT` requests to `InboundHttpWorker` workers instead of the singleton OutboundStreamingHttpWorker. Here's a Howto for doing that: http://stackoverflow.com/questions/5673335/how-to-configure-haproxy-to-send-get-and-post-http-requests-to-two-different-app

We would definitely need a dispatcher between Go and the inbound/outbound workers. It would make sure that outbound messages from Go are routed to the outbound http worker, and that inbound messages from the inbound http workers are routed to Go. This dispatcher would take the place of the 'http_api_transport' values in `routing_table_dispatcher.yaml`.

The above solution should work for event and metric messages too, as far as I can see from looking at the http_api worker code.

Not too sure about conversation command messages. I don't think the http_api worker handles conversation messages, so there shouldn't be a problem there either.

hodgestar:
How does the `http_api_nostream` affect this issue?

vgeddes:
@hodgestar I think this issue is pretty much obsoleted by `http_api_nostream`.  I am happy with the new non-streaming architecture and think it is a superior approach. At least I gained some experience working on the http_api in the this PR, which helped me review  the `http_api_nostream` branch.  Anyway, marking this issue as closed.

hodgestar:
I doubt the non-streaming architecture is superior -- there are advantages
to long lived connections (e.g. no TCP setup overhead, more immediate
notification of connection failures). Our current implementation had issues
though.

PR #716, opened 2013-09-20
Change text on 'save' buttons

hodgestar:
- From the show page, I click on "write and send bulk message" btn
- I land on the /action/bulk_send page
- I define my message and click 'save'
- Change the wording of the button from 'save' to 'send'

This is requested by Indrani and makes sense to me.

screen shot 2013-09-19 at 11 40 04 am

Also on the /action/download_user_data page which I access by clicking on the 'download user data' button the conversation show page
Change the button text from 'save' to 'download'

screen shot 2013-09-19 at 11 46 34 am

### DIFF
--- a/go/conversation/templates/conversation/action.html
+++ b/go/conversation/templates/conversation/action.html
@@ -4,7 +4,7 @@
 
 {% block content_actions_right %}
 <div class="pull-right">
-    <button type="submit" name="submit" class="btn btn-primary" value="Save">{{ action.action_display_verb|default:"Save" }}</button>
+    <button type="submit" name="submit" class="btn btn-primary" value="Apply">{{ action.action_display_verb|default:"Save" }}</button>
 </div>
 {% endblock %}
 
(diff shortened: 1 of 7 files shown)


go/conversation/templates/conversation/action.html
```diff
@@ -1,10 +1,10 @@
 {% extends "app.html" %}
 
-{% block content_title %}{{ action_display_name }}{% endblock %}
+{% block content_title %}{{ action.action_display_name }}{% endblock %}
 
 {% block content_actions_right %}
 <div class="pull-right">
-    <button type="submit" name="submit" class="btn btn-primary" value="Save">Save</button>
+    <button type="submit" name="submit" class="btn btn-primary" value="Save">{{ action.action_display_verb|default:"Save" }}</button>
```
justinvdm: Does having a `value` attribute of `"Save"` still make sense if this isn't always a save?

justinvdm:
Looks good. Just a small comment, otherwise +1.

go/conversation/templates/conversation/action.html
hodgestar: Maybe "Apply" would be better?

justinvdm: +1 on "Apply"

justinvdm:
:+1:

PR #720, opened 2013-09-25
Account change confirmation email not received

justinvdm:
On my account page: https://go.vumi.org/account/details/
I uncheck the box "Receive SMS to confirm start of campaign"
I then click 'save'
I get a notification saying the link has been sent to my email.
However I never receive the email with the confirmation

### DIFF
--- a/go/account/tests.py
+++ b/go/account/tests.py
@@ -32,9 +32,14 @@ class AccountTestCase(VumiGoDjangoTestCase):
             'existing_password': 'password',
             '_account': True,
             })
+
         token_url = response.context['token_url']
-        self.confirm(token_url)
+
+        [email] = mail.outbox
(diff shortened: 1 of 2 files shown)


justinvdm:
After a crazy amount of trial and error, I figured out that the confirmation email was actually being sent, but google was spam filtering it because of the spam-like title: "Confirm account detail changes". Changing the title to "Confirm Vumi Go account detail changes".

justinvdm:
Hrm, so that too is being spam filtered out. Will try something else. Any ideas?

justinvdm:
"Vumi Go account detail change confirmation" works if everyone is okay with that

hodgestar:
:+1: and tx for improving the test while you were there.

PR #714, opened 2013-09-30
For dialogues accept both the number and text response by default

justinvdm:
This issue came from Eve, who is using Vumi Go in Tanzania to run Afropinions.

Eve:
- set up a dialogue over SMS
- when she dialed in she was unable to move past the first question

The reason was that she was responding with the actual 'text' answer, and not the number. for e.g. responding 'red', not '1' to the question 'what is your favorite color'.

In the old UI, the user could define the answer choices. In this example they would be '1', 'red', '2', 'blue' etc

In the new dialogue UI, we should accept either the number or the text by default, instead of the current solution, which is to only accept the number.

screen shot 2013-09-17 at 12 59 57 pm

### DIFF
--- a/go/apps/dialogue/tests/test_vumi_app.js
+++ b/go/apps/dialogue/tests/test_vumi_app.js
@@ -12,6 +12,12 @@ function poll_tester(poll) {
     });
 }
 
+function clone(obj) {
+  var dup = {};
+  for (var k in obj) { dup[k] = obj[k]; }
+  return dup;
+}
+
 describe("choice states", function() {
     var tester;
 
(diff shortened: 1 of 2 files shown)


justinvdm:
Needs [vumi-jssandbox-toolkit#31](https://github.com/praekelt/vumi-jssandbox-toolkit/pull/31) to land and get published before this can land, but ready for review.

hodgestar:
Maybe we need a Javascript test that this option is honoured somehow?

hodgestar:
And we still need UI for this right?

justinvdm:
+1 on a test. We probably need some kind of UI, should we add that in this PR? Not sure on a place for it though. We could put it next to the 'Enable users to submit multiple surveys' checkbox, but I think that already looks out of place. Also, we are soon going to run out of space in the 'action bar' for ui things like checkboxes, so we definitely need a better long term solution. @hodgestar @badlard thoughts?

hodgestar:
Can we maybe put the options under a gear menu that pops up the actual
options form?

Schiavo
Simon

justinvdm:
Sounds like a plan. Don't think it should be part of this PR, that okay?

hodgestar:
Separate PR is fine by me.

badlard:
@justinvdm @Hodgestar Hrm - I thought we agreed not to have UI for this and instead default to accepting both the number and text response? That certainly is my vote - there is a lot going on already on the page

hodgestar:
:+1:

PR #733, opened 2013-10-08
Organize vendor libraries by component, then resource type

justinvdm:
Currently, the way we organise vendor libraries often looks something like `go/base/static/vendor/<component_name>/<component_name>/<resource_filename>`. This wierdness came about when we started using the [grunt-bower](https://github.com/yatskevich/grunt-bower-task) task for vendor libraries that needed some shimming to get them to export to `go/base/static/vendor` properly.

We can clean this up a bit by configuring grunt-bower to organize vendor libraries like this: `go/base/static/vendor/<component_name>/<resource_type>/<resource_filename>`. We should be able to get away with fewer shims if we do things this way.

While the way things currently are works, we will need this change anyways in order to get grunt-bower and bootstrap working together properly once we upgrade to bootstrap 3.


justinvdm:
Ready for review. `collectstatic` seems to be working with the changes. Happy to test it out on vumigo_solo if anyone would prefer that.

hodgestar:
This looks sane to me. I'd appreciate it be tried out on vumigo_solo first since it's hard to review properly, but otherwise :+1:.

PR #725, opened 2013-10-09
Can't upload contacts

hodgestar:
I took the following steps:
logged into Lebo's account (lebohang@praekelt.com)
navigated to: https://go.vumi.org/contacts/people/
clicked on "import contacts"
selected a CSV from my drive
named a new group and clicked upload (I got the same when I tried to upload to an existing group)
Got the attached error message.

I also, weirdly, never received an email with the error message

screen shot 2013-10-02 at 12 04 01 pm


hodgestar:
This example fails because there is only a single column (and thus the format sniffer fails to determine the column delimiter).

go/contacts/parsers/csv_parser.py
```diff
@@ -7,11 +8,28 @@
...
 class CSVFileParser(ContactFileParser):
 
+    @classmethod
+    def _sniff(cls, csvfile):
+        sample = csvfile.read(1024)
+        csvfile.seek(0)
+
+        try:
+            dialect = csv.Sniffer().sniff(sample)
+        except (csv.Error,), e:
```
jerith: Maybe `except csv.Error as e:` instead?

go/contacts/parsers/csv_parser.py
```diff
@@ -7,11 +8,28 @@
...
+            if re.search(",|\t", sample) is None:
```
justinvdm, line 19: Is tab a valid delimeter for csv files? If there is no standard for valid csv delimeters, should we maybe be including other delimeters?

jerith:
:+1: modulo minor style comment.

go/contacts/parsers/csv_parser.py
hodgestar, line 19: Absolutely anything is potentially a valid delimiter (at least as far as Python's `csv` module is concernded). I think comma and tab are the most common though. Happy to add others to the list if there are suggestions.

justinvdm, line 19: Ok cool. I'm happy with us keeping those two as the only delimeters, we can always add more later.

justinvdm:
:+1:

PR #732, opened 2013-10-10
Create billing router which talks to the billing API

andreit:


go/vumitools/routing.py
```diff
@@ -152,6 +153,12 @@ class AccountRoutingTableDispatcherConfig(RoutingTableDispatcher.CONFIG_CLASS,
     opt_out_connector = ConfigText(
         "Connector to publish opt-out messages on.",
         static=True, required=True)
+    billing_inbound_connector = ConfigText(
+        "Connector to publish inbound messages on.",
+        static=True, required=True)
+    billing_outbound_connector = ConfigText(
+        "Connector to publish outbound messages on.",
+        static=True, required=True)
```
hodgestar: We should perhaps make these optional, especially at first so that we can turn billing on and off easily.

go/vumitools/routing.py
```diff
@@ -519,7 +528,12 @@ def process_inbound(self, config, msg, connector_name):
         dst_connector_name, dst_endpoint = yield self.set_destination(
             msg, target, self.INBOUND)
 
-        yield self.publish_inbound(msg, dst_connector_name, dst_endpoint)
+        if msg_mdh.is_paid():
+            yield self.publish_inbound(msg, dst_connector_name, dst_endpoint)
```
hodgestar: This logic needs to be rewritten a bit because it's important to always call `set_destination` with the correct destination (otherwise things like the list of hops the message has taken get messed up). To do that I think we need to:
- Set `dst_connector_name` to `self.billing_inbound_connector` and the `dst_endpoint` to `default` higher up in this function if the source of the message is a transport and billing is activated.
- Add a case to `set_destination` to handle messages being sent to the billing dispatcher (i.e. having one of the two billing connectors as `dst_connector_name`).

go/vumitools/routing.py
```diff
@@ -556,7 +570,12 @@ def process_outbound(self, config, msg, connector_name):
         dst_connector_name, dst_endpoint = yield self.set_destination(
             msg, target, self.OUTBOUND)
 
-        yield self.publish_outbound(msg, dst_connector_name, dst_endpoint)
+        if msg_mdh.is_paid():
+            yield self.publish_outbound(msg, dst_connector_name, dst_endpoint)
+        else:
```
hodgestar: The same comment applies here as in `process_inbound` except that this time we need to send to billing if it's about to go to a transport.

go/vumitools/routing.py
```diff
@@ -641,3 +660,97 @@ def process_event(self, config, event, connector_name):
...
+    receive_inbound_connector = ConfigText(
+        "Connector that will receive inbound messages and events.",
+        required=True, static=True)
+
+    receive_outbound_connector = ConfigText(
+        "Connector that will receive outbound messages.",
+        required=True, static=True)
+
+
+class BillingWorker(BaseWorker, GoWorkerMixin):
```
hodgestar: I think you should just inherit from `Dispatcher` and assert in `setup_dispatcher` that there is only one inbound and one outbound connector in the connector lists (and probably save those to useful attribute names). Arguably the base class could support this usecase better but I think it's still nicer than copying most of the class.

go/vumitools/routing.py
```diff
@@ -641,3 +660,97 @@ def process_event(self, config, event, connector_name):
...
+        self.publish_inbound(msg, connector_name, endpoint_name)
+
+    def process_outbound(self, config, msg, connector_name):
+        log.debug("Processing outbound: %r" % (msg,))
+        msg_mdh = self.get_metadata_helper(msg)
+        # TODO: Create transaction by calling the billing API
+        msg_mdh.set_paid()
+        connector_name = self.get_configured_ri_connector()
+        endpoint_name = msg.get_routing_endpoint()
+        self.publish_outbound(msg, connector_name, endpoint_name)
```
hodgestar: `process_inbound` and `process_outbound` look good.

go/vumitools/utils.py
```diff
@@ -90,6 +97,9 @@ def set_user_account(self, user_account):
             'user_account': user_account,
         })
 
+    def set_paid(self):
+        self._go_metadata.update({'is_paid': True})
+
```
hodgestar, line 102: :cake: for the `is_paid` metadata idea.

hodgestar:
Left some comments and it needs some tests but structure is looking good.

andreit:
@hodgestar Hi Simon, could you please review my last changes. Thanks

go/vumitools/routing.py
```diff
@@ -641,3 +675,50 @@ def process_event(self, config, event, connector_name):
...
+    worker_name = 'billing_dispatcher'
+
+    @inlineCallbacks
+    def setup_dispatcher(self):
+        yield self._go_setup_worker()
+        self.unpause_connectors()
+
+    @inlineCallbacks
+    def teardown_dispatcher(self):
+        self.pause_connectors()
```
hodgestar: `pause_connectors` was recently updated to return a deferred that fires once the messages currently being processed are done, so this should now become `yield self.pause_connectors()`.

hodgestar:
This looks mostly good. I think the next steps are:
- split the billing dispatcher out into its own module
- write tests for the billing dispatcher and the new logic in the application dispatcher

The intended links for inbound messages are:
- Message from transport --> billing dispatcher inbound
- Billing dispatcher outbound --> routing table desitnation for 

The intended links for outbound messages are:
- Message about to be sent to transport --> billing dispatcher outbound
- Billing dispatcher inbound --> transport for the message tag

I don't think that's quite want's implemented now but adding test for all four cases will help sort everything out.

I would also suggest handling the billing inbound and outbound message a little like the opt_out messages are handled (i.e. write a separate method for them) rather than piggybacking on the logic for the messages directly from the routing table.

andreit:
@hodgestar Do you have any suggestions on the new home for the billing dispatcher module?

hodgestar:
@andreit Maybe `go.vumitools.billing_worker`?

andreit:
@hodgestar Cool, sounds good.

go/vumitools/utils.py
andreit, line 102: @hodgestar Quick question. Does an outbound reply message contain a copy of the inbound message's metadata?

hodgestar, line 102: @andreit Nope -- just the inbound message's UUID. One can use that to retrieve the original message if one absolutely has to though (since we store messages in a message store).

andreit, line 102: @hodgestar I am experiencing something weird, the first time an outbound message goes into the ARTD's process_outbound it has is_paid set to True in the go metadata.

andreit, line 102: @hodgestar Maybe this has something to do with it. In `TransportUserMessage.reply` there is a comment: 

`FIXME: helper_metadata should *not* be copied to the reply message. 
We only do it here because a bunch of legacy code relies on it.`

Perhaps the `is_paid` flag should go somewhere else?

andreit, line 102: @hodgestar Or I could just have 2 flags; `is_inbound_paid` and `is_outbound_paid`?

hodgestar, line 102: The `helper_metadata` issue is a bit complicated. For some items (e.g.
`is_paid`) copying makes no sense. For other items (e.g. `irc_channel`)
copying does make sense. Two different flags seems a bit ugly but I think
let's go with that for now. Another option would be to clear `is_paid` if
it is seen on an outbound message directly from a conversation or routing
block? Not entirely sure which these options is better.

andreit, line 102: I think we should go with option 2 for now; It will not bloat the `helper_metadata` with an extra field and will keep the `BillingDispatcher` implementation cleaner.

andreit:
@hodgestar Cleaned up the routing logic, but still not 100% sure if my method of figuring out the original message destination after billing is correct. That's in the `publish_inbound_from_billing` and `publish_outbound_from_billing` methods

hodgestar:
Looking better. There are probably a few to many places where `find_target` is called (I originally found it easiest if that was only called from one place). The next step is definitely to start adding tests for this though (there are a lot for the routing dispatcher already).

andreit:
@hodgestar I need some advice. What is the best way to figure out the destination connector name for an outbound message which come back from billing. I used the `metadata_helper.tag`, but it doesn't seem too reliable. 

Should I maybe look at the hops to figure out where it came from and try figure out the destination?

hodgestar:
@andreit `metadata_helper.tag` should be the right thing to look at for outbound messages. Hops are definitely the wrong thing (a hop for the next step of outbound messages doesn't even exist).

Maybe we just need to fix whatever is making `.tag` unreliable?

andreit:
@hodgestar I am having trouble with the "outbound message from conversation" test with billing enabled. Probably not calling `set_destination` at the right time so it never sets the `tag` then when it comes back from the billing worker there is no `tag`. Might have to rethink the flow.

hodgestar:
@andreit The outbound message tests should work if billing is not
configured but fail if billing is configured? I recommend adding a new set
of tests for when billing is enabled and leaving the old set of tests in
place?

andreit:
@hodgestar Yes, I have done that. I just took most of the original ones and trying to make them work with billing enabled. They are part of a new TestCase class.

hodgestar:
@andreit if you push what you have I can look at the failing test and maybe
see what's happening?

andreit:
@hodgestar Seems like I had pushed them earlier, the last 5 tests are still failing. Thanks for the help.

andreit:
@hodgestar The problem I am having is related to outbound messages which are not replies, so they don't have a `tag` set when they come back from billing; as in the `test_outbound_message_from_conversation_in_app1` test case. Does it make sense to set the tag before sending to billing?

hodgestar:
@andreit Yep, I think it makes sense to set the tag before sending it to billing. Before it was set before we sent it to the transport but now at that point we no longer have the information (i.e. the previous routing table step).

go/vumitools/api.py
```diff
@@ -646,3 +651,32 @@ def event(cls, account_key, conversation_key, event_type, content):
...
+        data = json.dumps(data)
+        headers = {'Content-Type': 'application/json'}
+        log.debug("Sending billing request to %r: %r" % (url, data))
+        response = yield http_request_full(url, data, headers=headers)
+        # TODO: Check for a non 200 response.code and do something
+        log.debug("Got billing response: %r" % (response.delivered_body,))
+        result = json.loads(response.delivered_body,
+                            parse_float=decimal.Decimal)
+
+        returnValue(result)
```
hodgestar: This probably belongs in the same file as the billing worker -- `go.vumitools.api` is for the internal Go API and I don't think billing is needed in that yet.

andreit:
@hodgestar Travis says all is well :) Could you please review and let me know if anything is still outstanding. Thanks.

go/vumitools/api.py
```diff
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 
 """Convenience API, mostly for working with various datastores."""
-
```
hodgestar, line 5: Can we revert this line deletion? It's unrelated and I like the extra line.

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,145 @@
...
+        url = urljoin(self.base_url, "/transactions")
+        data = json.dumps(data)
+        headers = {'Content-Type': 'application/json'}
+        log.debug("Sending billing request to %r: %r" % (url, data))
+        response = yield http_request_full(url, data, headers=headers)
+        log.debug("Got billing response: %r" % (response.delivered_body,))
+        if response.code != 200:
+            raise BillingError(response.delivered_body)
+        result = json.loads(response.delivered_body,
+                            parse_float=decimal.Decimal)
```
hodgestar: Could you factor out the bit that does the API call, URL joining, etc? That'll make it easier to add more methods later.

hodgestar:
The file `test_billing.py` should be called `test_billing_worker.py` and should contain unit tests for all the code in `billing_worker.py`.

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,145 @@
...
+
+class BillingDispatcherConfig(Dispatcher.CONFIG_CLASS, GoWorkerConfigMixin):
+
+    api_url = ConfigText(
+        "Base URL of the billing REST API",
+        static=True, required=True)
+
+    billing_api = ConfigText(
+        "Python path to the billing api proxy class",
+        static=True, required=True)
```
hodgestar: I think this part can be dropped and the class can just be referenced directly in the Python code? We do `load_class_by_name` in places were we need to be able to plugin a new class at run time but I don't think this is one of those cases?

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,145 @@
...
+        log.debug("Processing inbound: %r" % (msg,))
+        msg_mdh = self.get_metadata_helper(msg)
+        try:
+            yield self.create_transaction_for_inbound(msg)
+            msg_mdh.set_paid()
+        except Exception as error:
+            log.err(error.message)
+        connector_name = self.get_configured_ro_connectors()[0]
+        endpoint_name = msg.get_routing_endpoint()
+        yield self.publish_inbound(msg, connector_name, endpoint_name)
```
hodgestar: I think you can always leave the endpoint out here (or set it to `None`) since the billing router only uses the default endpoint.

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,145 @@
...
+        log.debug("Processing outbound: %r" % (msg,))
+        msg_mdh = self.get_metadata_helper(msg)
+        try:
+            yield self.create_transaction_for_outbound(msg)
+            msg_mdh.set_paid()
+        except Exception as error:
+            log.err(error.message)
+        connector_name = self.get_configured_ri_connectors()[0]
+        endpoint_name = msg.get_routing_endpoint()
+        yield self.publish_outbound(msg, connector_name, endpoint_name)
```
hodgestar: The same endpoint comment applies here.

go/vumitools/routing.py
```diff
@@ -484,6 +507,57 @@ def publish_outbound_optout(self, config, msg):
...
+        dst_connector_name, dst_endpoint = yield self.set_destination(
+            msg, target, self.INBOUND)
+
+        yield self.publish_inbound(msg, dst_connector_name, dst_endpoint)
+
+    @inlineCallbacks
+    def publish_inbound_from_billing(self, config, msg):
+        """Publish an inbound message to its intended destination
+        after billing.
+
```
hodgestar: Stray blank line.

go/vumitools/routing.py
```diff
@@ -484,6 +507,57 @@ def publish_outbound_optout(self, config, msg):
...
+        dst_connector_name, dst_endpoint = yield self.set_destination(
+            msg, target, self.OUTBOUND)
+
+        yield self.publish_outbound(msg, dst_connector_name, dst_endpoint)
+
+    @inlineCallbacks
+    def publish_outbound_from_billing(self, config, msg):
+        """Publish an outbound message to its intended destination
+        after billing.
+
```
hodgestar: Another stray line.

go/vumitools/routing.py
```diff
@@ -484,6 +507,57 @@ def publish_outbound_optout(self, config, msg):
...
+    @inlineCallbacks
+    def publish_inbound_from_billing(self, config, msg):
+        """Publish an inbound message to its intended destination
+        after billing.
+
+        """
+        msg_mdh = self.get_metadata_helper(msg)
+        src_conn = str(GoConnector.for_transport_tag(*msg_mdh.tag))
+        target = self.find_target(config, msg, src_conn)
```
hodgestar: I'd really like `.find_target` to only be called from one place in the code for each of inbound / outbound messages (it's just a sort of intuition about what's most maintainable). Could the `find_target` bit be combined with the call in `process_inbound`?

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,145 @@
...
+            msg_mdh.tag[1], self.MESSAGE_DIRECTION_INBOUND)
+
+    @inlineCallbacks
+    def create_transaction_for_outbound(self, msg):
+        """Create a transaction for the given outbound message"""
+        msg_mdh = self.get_metadata_helper(msg)
+        yield self.billing_api.create_transaction(
+            msg_mdh.get_account_key(), msg_mdh.tag[0],
+            msg_mdh.tag[1], self.MESSAGE_DIRECTION_OUTBOUND)
+
```
hodgestar, line 132: We need to add a `process_event` implementation that just publishes the event.

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        log.debug("Processing inbound: %r" % (msg,))
+        msg_mdh = self.get_metadata_helper(msg)
+        try:
+            yield self.create_transaction_for_inbound(msg)
+            msg_mdh.set_paid()
+        except Exception as error:
+            log.err(error.message)
+        connector_name = self.get_configured_ro_connectors()[0]
```
hodgestar: I would just do this in dispatcher setup.

go/vumitools/tests/test_billing.py
```diff
@@ -0,0 +1,137 @@
...
+            "tag_name": tag_name,
+            "message_direction": message_direction,
+            "message_cost": 80,
+            "markup_percent": decimal.Decimal('10.0'),
+            "credit_amount": -35,
+            "credit_factor": decimal.Decimal('0.4'),
+            "created": "2013-10-30T10:42:51.144745+02:00",
+            "last_modified": "2013-10-30T10:42:51.144745+02:00",
+            "status": "Completed"
+        }
```
hodgestar: I think we also need a test for the real BillingApi class.

go/vumitools/tests/test_billing.py
```diff
@@ -0,0 +1,137 @@
...
+            md.set_paid()
+        if hops is not None:
+            rmeta = RoutingMetadata(msg)
+            for src, dst in zip(hops[:-1], hops[1:]):
+                rmeta.push_hop(src, dst)
+        if outbound_hops_from is not None:
+            rmeta = RoutingMetadata(msg)
+            outbound_rmeta = RoutingMetadata(outbound_hops_from)
+            rmeta.set_outbound_hops(outbound_rmeta.get_hops())
+        return msg
```
hodgestar: I don't think you need the hops stuff or endpoint setting here.

go/vumitools/tests/test_billing.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @inlineCallbacks
+    def test_inbound_message(self):
+        yield self.get_dispatcher()
+        msg = self.with_md(self.mkmsg_in(), user_account="12345",
+                           tag=("pool1", "1234"))
+
+        yield self.dispatch_inbound(msg, 'billing_dispatcher_ri')
+        self.assert_rkeys_used('billing_dispatcher_ri.inbound',
+                               'billing_dispatcher_ro.inbound')
```
hodgestar: I think this assert doesn't do anything useful (we should just assume `dispatch_inbound` and `get_dispatched_inbound` do the right thing).

go/vumitools/tests/test_billing.py
```diff
@@ -0,0 +1,137 @@
...
+        reply = msg.reply(content="Reply")
+        returnValue((msg, reply))
+
+    @inlineCallbacks
+    def mk_msg_ack(self, **kw):
+        "Create and store an outbound message, then create an ack for it."
+        msg = self.with_md(self.mkmsg_out(), **kw)
+        yield self.vumi_api.mdb.add_outbound_message(msg)
+        ack = self.mkmsg_ack(user_message_id=msg['message_id'])
+        returnValue((msg, ack))
```
hodgestar: These make message methods can probably just be removed and you can use `self.make_inbound`, `self.make_outbound` and `self.make_ack` directly.

go/vumitools/tests/test_billing.py
```diff
@@ -0,0 +1,137 @@
...
+        msg = self.with_md(self.mkmsg_out(), user_account="12345",
+                           tag=("pool1", "1234"))
+
+        yield self.dispatch_outbound(msg, 'billing_dispatcher_ro')
+        self.assert_rkeys_used('billing_dispatcher_ro.outbound',
+                               'billing_dispatcher_ri.outbound')
+
+        self.with_md(msg, is_paid=True)
+        self.assertEqual(
+            [msg], self.get_dispatched_outbound('billing_dispatcher_ri'))
```
hodgestar: We need a test for an event message.

go/vumitools/api.py
andreit, line 5: Cool, must have deleted it by accident.

go/vumitools/billing_worker.py
andreit: Did that so that I could plug in a dummy `BillingApi` for testing purposes. Is there a better way?

hodgestar: Just override the `.billing_api` attribute on the worker after you create
it?

go/vumitools/routing.py
andreit: I followed the PEP257 recommendation for multi-line docstrings (http://www.python.org/dev/peps/pep-0008/#documentation-strings). I can remove it.

go/vumitools/tests/test_routing.py
```diff
@@ -279,15 +281,34 @@ def get_dispatcher(self, **config_extras):
...
+                "billing_dispatcher_ro"
+            ],
+            "api_url": "http://127.0.0.1:9090/",
+            "billing_api": "go.vumitools.tests.test_billing.BillingApiMock",
+            "metrics_prefix": "bar"
+        }
+        config.update(config_extras)
+        billing_dispatcher = yield self.get_worker(
+            self.mk_config(config), BillingDispatcher)
+        returnValue(billing_dispatcher)
```
hodgestar: These tests shouldn't actually create a `BillingDispatcher` -- they should just test that messages are dispatched to the right connectors by the routing table worker.

go/vumitools/routing.py
hodgestar: We don't really follow PEP257 so I'd rather be consistent with the other
code in our code base.

go/vumitools/tests/test_routing.py
hodgestar: Doing this will also break up a bunch of the new tests into smaller ones, which is good.

hodgestar:
I left a bunch of specific comments, but other than those it's looking great and pretty close to being ready to land.

go/vumitools/routing.py
andreit: Ok, will fix.

go/vumitools/tests/test_billing.py
andreit: Any suggestions on how to approach this? Should I somehow fake the actual HTTP request?

hodgestar: I would do something like:

``` python
self.patch(<billing-api-module>, 'http_request_full',
self.dummy_http_request_full)
```

go/vumitools/tests/test_routing.py
andreit: Ok, So I don't actually need to check if is_paid get set in the metadata? I just have to test whether the message gets sent to the billing worker?

hodgestar: The `is_paid` check should be in the tests for the billing worker, not the
tests for the billing routing.

andreit:
@hodgestar Where is that elusive Rebuild button on Travis? I don't see it.

hodgestar:
Top right on either the build or job page (I think it has a refresh icon).

andreit:
Ah, needed to sign in first.

andreit:
@hodgestar I think all the comments have been addressed, could you please review when you get a chance. Thanks.

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,164 @@
...
+            'account_number': account_number,
+            'tag_pool_name': tag_pool_name,
+            'tag_name': tag_name,
+            'message_direction': message_direction
+        }
+
+        result = yield self._call_api("/transactions", data=data,
+                                      method='POST')
+
+        returnValue(result)
```
hodgestar: If you like you can probably just remove the `inlineCallbacks` and do `return self._call_api(...)` (since there's only one deferred step).

go/vumitools/tests/test_billing_worker.py
```diff
@@ -0,0 +1,218 @@
...
+            "metrics_prefix": "bar"
+        }
+        config.update(config_extras)
+        billing_dispatcher = yield self.get_worker(
+            self.mk_config(config), BillingDispatcher)
+        billing_dispatcher.billing_api = BillingApiMock(config["api_url"])
+        returnValue(billing_dispatcher)
+
+    def with_md(self, msg, user_account=None, conv=None, router=None,
+                tag=None, is_paid=False):
```
hodgestar: You can probably also remove the `conv` and `router` options since those aren't being used.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -0,0 +1,218 @@
...
+    @inlineCallbacks
+    def test_outbound_message(self):
+        yield self.get_dispatcher()
+        msg = self.with_md(self.mkmsg_out(), user_account="12345",
+                           tag=("pool1", "1234"))
+
+        yield self.dispatch_outbound(msg, 'billing_dispatcher_ro')
+        self.with_md(msg, is_paid=True)
+        self.assertEqual(
+            [msg], self.get_dispatched_outbound('billing_dispatcher_ri'))
```
hodgestar, line 231: Could we get tests for the failure cases that happen when inbound and outbound messages without `user_account` or `tag` are sent are sent to billing? I know this isn't supposed to happen but I'd rather be sure that that case is handled properly too (probably by logging the appropriate error?).

go/billing/api.py
```diff
@@ -56,8 +59,7 @@ def _parse_json(self, request):
         """
         content_type = request.getHeader('Content-Type')
         if request.method == 'POST' and content_type == 'application/json':
-            return json.loads(request.content.read(),
-                              parse_float=decimal.Decimal)
+            return json.loads(request.content.read(), parse_float=parse_float)
```
hodgestar: Maybe if would be cleaner to add a custom JSONDecoder sub-class that passes in parse_float? That makes it symmetric with the encoding case.

go/vumitools/tests/test_routing.py
```diff
@@ -279,7 +280,7 @@ def get_dispatcher(self, **config_extras):
             "router_outbound_connector_mapping": {
                 "router": "router_ri",
             },
-            "opt_out_connector": "optout",
+            "opt_out_connector": "optout"
```
hodgestar: Trailing commas are preferred since that means one doesn't have to modify the line above in order to add an item at the end.

go/vumitools/tests/test_routing.py
```diff
@@ -581,3 +594,198 @@ def test_outbound_message_gets_transport_fields(self):
...
+        yield self.dispatch_outbound(reply, 'optout')
+        self.assert_rkeys_used('optout.outbound', 'sphex.outbound')
+        hops = [
+            ['OPT_OUT', 'default'],
+            ['TRANSPORT_TAG:pool1:1234', 'default'],
+        ]
+        self.with_md(reply, tag=tag, user_account=self.user_account_key,
+                     hops=hops)
+
+        self.assertEqual([reply], self.get_dispatched_outbound('sphex'))
```
hodgestar: This test made me realize that optout messages need to go via billing too (in both directions).

go/vumitools/routing.py
```diff
@@ -547,12 +625,29 @@ def process_outbound(self, config, msg, connector_name):
             yield self.publish_outbound_optout(config, msg)
             return
 
+        if connector_type == self.BILLING:
+            yield self.publish_outbound_from_billing(config, msg)
+            return
+
+        if connector_type == self.CONVERSATION or \
+                connector_type == self.ROUTER:
```
hodgestar: You can use brackets here to avoid having to use a end-of-line slash, e.g. `if (connector_type == ... or ...):`.

go/vumitools/routing.py
```diff
@@ -505,10 +570,17 @@ def process_inbound(self, config, msg, connector_name):
...
-        if (connector_type == self.TRANSPORT_TAG
-                and msg_mdh.is_optout_message()):
-            yield self.publish_inbound_optout(config, msg)
-            return
+        if connector_type == self.TRANSPORT_TAG:
+            if msg_mdh.is_optout_message():
+                yield self.publish_inbound_optout(config, msg)
+                return
+
+            if self.billing_inbound_connector and not msg_mdh.is_paid():
```
hodgestar: I'm not actually sure we want the `is_paid()` check here -- every message from a transport should be routed via billing if billing is active regardless of what the message contains?

hodgestar:
I think I've done with this round of reviewing now. Nearly there!

go/vumitools/billing_worker.py
andreit: Will do, I think I was doing something with `result` before.

go/vumitools/tests/test_billing_worker.py
andreit, line 231: @hodgestar I assume these messages should just get published back to the ARTD, after logging an error?

hodgestar, line 231: Yep. Maybe we can do something more advanced in future. For now let's just
make sure we get an error logged and the message keeps going to its
destination (with is_paid still False).

go/vumitools/routing.py
andreit: Yeah, its probably not necessary. I will take it out.

andreit:
@hodgestar Tried to route the opt-out messages via billing, test cases pass, but the `process_inbound` and `process_outbound` methods are getting nasty :) Could you please glance over them. Thanks

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,177 @@
...
+        Any errors are logged and the message is allowed to continue on its
+        path and fulfill its destiny.
+        """
+        log.debug("Processing inbound: %r" % (msg,))
+        msg_mdh = self.get_metadata_helper(msg)
+        try:
+            yield self.create_transaction_for_inbound(msg)
+            msg_mdh.set_paid()
+        except Exception as error:
+            log.err(error.message)
```
hodgestar: Twisted's `log.err` is a bit strange and doesn't take a message -- it takes a Failure containing an exception (I forget this all the time). But you can also just do `log.err()` and it'll log the current exception for you.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -0,0 +1,241 @@
...
+        self.assertEqual(
+            [msg], self.get_dispatched_inbound('billing_dispatcher_ro'))
+
+    @inlineCallbacks
+    def test_inbound_message_without_tag(self):
+        yield self.get_dispatcher()
+        msg = self.with_md(self.mkmsg_in(), user_account="12345")
+        yield self.dispatch_inbound(msg, 'billing_dispatcher_ri')
+        self.assertEqual(
+            [msg], self.get_dispatched_inbound('billing_dispatcher_ro'))
```
hodgestar, line 220: These should also test that the correct error is logged. We have the `LogCatcher` for that.

go/vumitools/billing_worker.py
```diff
@@ -0,0 +1,177 @@
...
+        """Create a transaction for the given inbound message"""
+        msg_mdh = self.get_metadata_helper(msg)
+        account_number = msg_mdh.get_account_key()
+        if not account_number:
+            raise BillingError(
+                "No account number found for message %r" % (msg,))
+
+        if not msg_mdh.tag:
+            raise BillingError(
+                "No tag found for message %r" % (msg,))
```
hodgestar: Since we use this logic twice maybe we should create a helper method that looks for the account and tag and raises the necessary exceptions?

hodgestar:
I've now gone through all the changes except the possible clean-up of the `process_inbound` and `process_outbound` methods which I'll look at tomorrow.

go/vumitools/billing_worker.py
andreit: Ok, Didn't know that. :)

go/vumitools/tests/test_billing_worker.py
andreit, line 220: Used the Twisted `flushLoggedErrors` to test that `BillingError` was thrown and changed the exception message to only contain the `message_id`

hodgestar:
I think I've cleaned up the billing routing. Please review my commits though.

hodgestar:
Once you've done reviewing I think we can land the branch. Maybe best to leave landing for Monday morning though.

andreit:
@hodgestar Thanks for that, just a quick question... do opt-out messages always go directly to the transport? and never via a router?

hodgestar:
@andreit Yep. Messages from the opt-out worker always go straight to a transport.

andreit:
Thanks for your help with this. Looks much neater now. Travis is also happy. :+1:

hodgestar:
:D Have a good weekend!

andreit:
@hodgestar Are you happy for me to merge this pull request?

hodgestar:
@andreit I'm trying it out on an internal test instance now (since it's a
fairly big change).

hodgestar:
Test instance looks good. Landing.

andreit:
Yay. :)

PR #519, opened 2013-10-11
Run Javascript tests in Travis

hodgestar:
Currently the Vumi Go Javascript tests aren't run by Travis. See https://github.com/praekelt/go-nike-ghr/blob/develop/.travis.yml for how we might start hooking this up as an additional set of tests for Travis to run.


justinvdm:
:+1:

PR #736, opened 2013-10-11
Add coverage for Javascript tests.

hodgestar:


hodgestar:
@justinvdm Ready for review, I think.

package.json
```diff
@@ -23,7 +26,9 @@
     "grunt-karma": "~0.6.2",
     "karma-phantomjs-launcher": "~0.1.0",
     "grunt-bower-task": "~0.3.2",
-    "grunt-exec": "~0.4.2"
+    "grunt-exec": "~0.4.2",
+    "karma-coverage": "~0.1.0",
+    "grunt-mocha-cov": "0.0.7"
```
justinvdm, line 31: Do we need to add `blanket` here, or does it not matter since `grunt-mocha-cov` has it as a dependency?

requirements.pip
```diff
@@ -5,7 +5,7 @@ gunicorn==0.15.0
 pep8==0.6.1
 pyflakes==0.4.0
 supervisor==3.0a10
-coverage==3.4
+coverage==3.7
 South==0.7.3
 psycopg2==2.4
 Fabric==1.2.2
```
justinvdm, line 11: Should we add `overalls` here?

justinvdm, line 11: Ah, I see we have it separate in `.travis.yml`. I'm guessing we only really need `overalls` for travis.

justinvdm:
Just a few small comments for places where things are probably the way they should be, but would just like to make sure. Otherwise, +1.

package.json
hodgestar, line 31: `grunt-mocha-cov` has it as a dependency so we're fine.

requirements.pip
hodgestar, line 11: I was following the pattern of how the original `coveralls` library is used.

hodgestar:
@justinvdm Happy with my responses?

justinvdm:
Happy :) :+1:

PR #729, opened 2013-10-13
Upgrade to Bootstrap 3

justinvdm:
Diamondash currently uses Bootstrap 3. Although it would probably take less work to downgrade diamondash to Bootstrap 2, we should probably be upgrading to Bootstrap 3 before starting on the new UI changes we have planned.


justinvdm:
Ready for review. Although there were quite a few changes, the addition and deletion statistics are probably skewed by the vendor file changes in `go/base/static/vendor` after upgrading Bootstrap and Bootbox. Wondering if there is something we could do (cherry picking or something?) to exclude the vendor file changes and make reviewing easier?

hodgestar:
I think it's easy enough to just scroll past the bits in `vendor` for now.

hodgestar:
We should fix the failing Travis tests (some seem related).

go/apps/jsbox/view_definition.py
```diff
@@ -23,11 +23,23 @@ def get(self, request, conversation):
...
+    def get(self, request, conversation):
+        edit_forms = dict(self.make_forms_dict(conversation))
+
+        return self.render_to_response({
+            'conversation': conversation,
+            'jsbox_form': edit_forms['jsbox'],
+            'jsbox_app_config_forms': edit_forms['jsbox_app_config'],
+            'edit_forms_media': self.sum_media(edit_forms.values())
+        })
+
```
hodgestar, line 42: Would it be simpler to not override `get` and just access the forms in `edit_forms` directly from within the template? E.g. `edit_forms.jsbox`?

go/conversation/templates/conversation/dashboard.html
```diff
@@ -19,6 +19,7 @@
                 <!-- TODO: We're only storing the created_at value :\ -->
                 <th>Last Edited</th>
                 <th></th>
+                <th></th>
```
hodgestar, line 22: Did the dashboard grow an extra column intentionally?

go/templates/registration/activation_complete.html
```diff
@@ -1,17 +1,9 @@
-{% extends "base.html" %}
+{% extends "registration/base.html" %}
```
hodgestar, line 1: Did this do the wrong thing before?

hodgestar:
I've gone through this. Left some minor questions. Let's fix the tests tomorrow morning and then have a look at it on staging with @badlard before merging to develop.

go/apps/jsbox/view_definition.py
justinvdm, line 42: I don't think accessing it as `edit_forms.jsbox` is possible. To make sure, I tried that before overriding `get` to find out that it doesn't work. I'd much prefer doing it that way, though, so if you have an idea to get it working that way, I'd be happy to make the change.

go/conversation/templates/conversation/dashboard.html
justinvdm, line 22: I found out it had one `<th>` too few, so the table header border that Bootstrap 3 adds wasn't drawing properly.

hodgestar, line 22: Woot.

go/templates/registration/activation_complete.html
justinvdm, line 1: Ya, previously it was redoing everything that was in base (including the logo and footer links). I assumed this was just a change that wasn't made when the UI changed.

justinvdm:
Cool, sounds good. Happy to deploy to vumigo_solo after the comments have been resolved and Travis is happy.

go/templates/registration/activation_complete.html
hodgestar, line 1: Tx for fixing.

hodgestar:
:+1: once @badlard is happy with it.

hodgestar:
:+1:. Let's land first thing tomorrow morning.

PR #165, opened 2013-10-16
Add UI for setting HTTP API tokens

smn:
Currently the UI is missing API token support.

### DIFF
--- a/go/apps/http_api/view_definition.py
+++ b/go/apps/http_api/view_definition.py
@@ -1,5 +1,44 @@
-from go.conversation.view_definition import ConversationViewDefinitionBase
+from django import forms
 
+from go.conversation.view_definition import (
+    ConversationViewDefinitionBase, EditConversationView)
+
+
+class TokenForm(forms.Form):
+    api_tokens = forms.CharField(
+            help_text='The access token for this HTTP Conversation.',
+            required=True)
+    push_message_url = forms.CharField(
+        help_text='The URL to forward messages to via HTTP POST.',
+        required=False)
+    push_event_url = forms.CharField(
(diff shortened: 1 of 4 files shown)


smn:
@hodgestar @jerith @justinvdm ready for review

go/apps/http_api/view_definition.py
```diff
@@ -1,5 +1,43 @@
...
+from django import forms
 
+from go.conversation.view_definition import (
+    ConversationViewDefinitionBase, EditConversationView)
+
+
+class TokenForm(forms.Form):
+    api_tokens = forms.CharField(
+            help_text='The access token for this HTTP Conversation.',
+            required=True)
```
justinvdm, line 10: I think we should either do validation or cleaning up of this field so we can ensure url-safe token values. Not sure if it needs to be done in this PR though.

go/apps/http_api/view_definition.py
```diff
@@ -1,5 +1,43 @@
...
+    push_message_url = forms.CharField(
+        help_text='The URL to forward messages to via HTTP POST.',
+        required=False)
+    push_event_url = forms.CharField(
+        help_text='The URL to forward events to via HTPT POST.',
```
justinvdm: Typo :)

go/apps/http_api/view_definition.py
justinvdm, line 10: On second thought, maybe cleaing up or validating this field isn't too important.

justinvdm:
Other than the typo, :+1:. We can always consider generating the tokens ourselves instead of requiring the user to do that in a later PR.

hodgestar:
:+1:

justinvdm:
:+1:

PR #741, opened 2013-10-16
Jsbox application conversation can not be updated

justinvdm:
Browse to conversation -> Edit -> Save -> 500 error.

Affects multiple conversations on multiple accounts.


hodgestar:
:+1:

PR #745, opened 2013-10-25
CSV contact parser fails on some single-column files

jerith:
This is due to it guessing the wrong field delimiter and incorrectly splitting the line into multiple fields.


hodgestar:
:+1:

PR #743, opened 2013-10-25
Publish metrics for message stats

justinvdm:
Currently, we do counts to calculate our message stats. It would be helpful if we published metrics for these message stats. This will be a good starting point for displaying per-conversation dashboards (with diamondash), since message metrics will be metrics displayable for all conversation types.


justinvdm:
A few applications are already overriding `collect_metrics()`, some doing message metric collection already. I think we should keep it this way for, since `GoWorkerMixin.collect_metrics()` is going to be changing soon, so the overridden `collect_metrics()` methods living on the application worker classes (`SubscriptionApplication`, `BulkMessageApplication`, etc) will go away during this change. Thoughts?

jerith:
:+1:, but only after we've done the Graphite work we discussed offline.

hodgestar:
:+1:

PR #243, opened 2013-10-25
Bump Django version to 1.5.1

hodgestar:
This includes removing the `{% load url from future %}` lines in templates.


jerith:
Looks good, but the issue description mentions removing `{% load url from future %}` from templates which we haven't done here.

hodgestar:
Good point. I'll do that.

jerith:
:+1: once Travis is happy.

PR #747, opened 2013-10-25
Replace all the horrible mkmsg_*() methods with GoMessageHelper

jerith:
I didn't split up the commits this time because too much stuff uses base classes from other places.


go/vumitools/tests/utils.py
```diff
@@ -197,12 +197,14 @@ def poll_metrics(self, assert_prefix=None, app=None):
...
         if batch_id is None and conv is not None:
             batch_id = conv.batch.key
         return self.user_api.api.mdb.add_outbound_message(
             msg, batch_id=batch_id)
 
     def store_inbound_msg(self, msg, conv=None, batch_id=None):
+        assert False
         if batch_id is None and conv is not None:
             batch_id = conv.batch.key
         return self.user_api.api.mdb.add_inbound_message(
```
hodgestar: Maybe these two methods can just be deleted now?

hodgestar:
Other than deleting the two methods that start with "assert False", I think this is good.

go/vumitools/tests/utils.py
jerith: Oops. I was supposed to do that.

hodgestar:
:+1: once Travis is happy.

PR #292, opened 2013-10-27
HTTP API can stall when defined URL for HTTP Posting to times-out.

smn:
If an API partner is not using the streaming capabilities but instead is using the forwarding via HTTP POST and their URL times out the HTTP API transport can stall and there will be a back log of messages.

### DIFF
--- a/go/apps/http_api/tests/test_vumi_app.py
+++ b/go/apps/http_api/tests/test_vumi_app.py
@@ -446,7 +446,6 @@ class StreamingHTTPWorkerTestCase(AppWorkerTestCase):
         from go.apps.http_api import vumi_app
 
         def timeout_raiser(*args, **kw):
-            print "o/"
             raise HttpTimeoutError()
         self.patch(vumi_app, 'http_request_full', timeout_raiser)
 
@@ -463,7 +462,7 @@ class StreamingHTTPWorkerTestCase(AppWorkerTestCase):
         with LogCatcher(message='Timeout') as lc:
             yield self.dispatch_to_conv(msg, self.conversation)
             [timeout_log] = lc.messages()
-        self.assertTrue(timeout_log.endswith(self.mock_push_server.url))
+        self.assertTrue(self.mock_push_server.url in timeout_log)
 
     @inlineCallbacks
     def test_post_inbound_event(self):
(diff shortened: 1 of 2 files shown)


jerith:
Due to the way we handle timeouts in `http_request_full()`, builds will fail until vumi#634 lands.

jerith:
praekelt/vumi#634 landed and I reran the latest build, which now passes.

go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -442,6 +442,29 @@ def test_post_inbound_message(self):
         posted_msg = TransportUserMessage.from_json(posted_json_data)
         self.assertEqual(posted_msg['message_id'], msg['message_id'])
 
+    def _patch_http_request_full(self):
+        from go.apps.http_api import vumi_app
+
+        def timeout_raiser(*args, **kw):
+            print "o/"
```
hodgestar: Stray print?

go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -442,6 +442,29 @@ def test_post_inbound_message(self):
...
+            'push_message_url': self.mock_push_server.url,
+        })
+        yield self.conversation.save()
+
+        self._patch_http_request_full()
+        msg = self.mkmsg_in(content='in 1', message_id='1')
+        with LogCatcher(message='Timeout') as lc:
+            yield self.dispatch_to_conv(msg, self.conversation)
+            [timeout_log] = lc.messages()
+        self.assertTrue(timeout_log.endswith(self.mock_push_server.url))
```
hodgestar: I think it would be better to do `url in timeout_log` here since that matches what's required more closes (and means that a change to the log message text won't cause this test to fail unnecessarily).

go/apps/http_api/tests/test_vumi_app.py
jerith: Yes, thanks.

go/apps/http_api/tests/test_vumi_app.py
jerith: Good point.

hodgestar:
:+1:

PR #749, opened 2013-10-28
Update metric name construction for app workers

justinvdm:
Currently, converstation metrics are constructed in the form `<prefix>.<account_key>.<conversation_key>.<metric_name>.<agg_method>`, and account metrics in the form  `<prefix>.<account_key>.<store_name>.<metric_name>.<agg_method>`. The plan is to change these to  `<prefix>.<account_key>.conversations.<conversation_key>.<metric_name>.<agg_method>` and `<prefix>.<account_key>.stores.<store_name>.<metric_name>.<agg_method>` respectively.

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -231,8 +231,8 @@ class TestBulkMessageApplication(AppWorkerTestCase):
         yield self.dispatch_command(
             'collect_metrics', conversation_key=conv.key,
             user_account_key=self.user_account.key)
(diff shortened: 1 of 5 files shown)


hodgestar:
I think we should factor these out into functions so that we can easily see what metrics look like. Maybe a small `go.metrics` module that includes `conversation_metric` and `account_store` functions that return appropriate names given an account and conversation or store name?

hodgestar:
Hmm. Although they are all currently in `app_worker`. I think :+1: for now.

We should wait for the migrations to be ready before landing.

PR #748, opened 2013-10-28
Change default worker metric aggregation method from MAX to LAST

justinvdm:


hodgestar:
:+1: (needs migrations to be ready to land)

Issue #721, opened 2013-09-23
Email address not updating in the UI


badlard:
Kgosi in Control Room needs to download Guinness VIP stats each week 
GVIP is run in Wambui's account 
Kgosi needed to update Wambui's email address from wambui@praekelt.com to controlroom@praekelt.com 
I changed the email address in go.vumi.org/admin to controlroom@praekelt.com but Kgosi still didn't receive the download in his email. 
I then logged into Wambui's account and changed the email address in the UI 
I received the automated 'account change' email (because I'm an admin) and verified the change by clicking on the link 
Kgosi then received the data download, however the email listed in /account is still wambui@praekelt.com 

In short, the email address didn't update in the UI

hodgestar:
#746 fixed this too.

PR #746, opened 2013-10-29
User account names can't be more than 32 characters long.

hodgestar:
Which considering they're email addresses is ludicrously short.


hodgestar:
http://www.caktusgroup.com/blog/2013/08/07/migrating-custom-user-model-django/ has a bunch of useful information on how to migrate to the custom user models.

hodgestar:
This branch should also aim to address https://github.com/praekelt/vumi-go/issues/721

hodgestar:
@andreit Just a heads up that I have a migration for the billing app in this branch so there is a change we'll clash (shouldn't be too difficult to sort out though).

hodgestar:
The migration is rather complex because the `auth` app doesn't use South and Django's `syncdb` doesn't even create the `auth_user` table if a custom user model is used. Bit stuck on how to make forward progress. @smn -- any suggestions?

hodgestar:
By commenting out the following at the end of `django.contrib.auth.management.__init__` I can now make migrations work:

```
signals.post_syncdb.connect(create_permissions,
    dispatch_uid="django.contrib.auth.management.create_permissions")
signals.post_syncdb.connect(create_superuser,
    sender=auth_app, dispatch_uid="django.contrib.auth.management.create_superuser")
```

Not quite sure if I've gone down the right road here. I can probably disconnect these from inside go.base if needed?

hodgestar:
Yay! Some tests are running and passing on Travis.

hodgestar:
Ready for review! :zap:

go/base/models.py
```diff
@@ -13,18 +16,58 @@ def get_account_store():
...
+            email, password,
+            is_superuser=True, is_staff=True, is_active=True)
+
+
+class GoUser(AbstractBaseUser, PermissionsMixin):
+    objects = GoUserManager()
+
+    email = models.EmailField(max_length=254, unique=True)
+    first_name = models.CharField(max_length=254)
+    last_name = models.CharField(max_length=254)
```
jerith, line 45: Do we really need two name fields here? If not, can we merge them while we're doing this stuff anyway?

hodgestar, line 45: I agree that one name field is probably better but I'd like @badlard to okay a change like that since it's more of a UI thing currently and I'd definitely rather do it in a separate PR since this one is complex enough already (once the user table is in South, migrations will be fairly easy).

hodgestar, line 45: Combining the name fields isn't high on my priority list right now.

go/account/views.py
```diff
@@ -32,7 +32,7 @@ def details(request):
     account_form = AccountForm(request.user, initial={
         'name': request.user.first_name,
         'surname': request.user.last_name,
-        'email_address': request.user.username,
+        'email_address': request.user.email,
```
smn, line 35: This is the only place where we expose the email as something different than the username. Since we're not communicating this people could think that changing this also changes their username and the two fields would get out of sync.

go/account/tests.py
```diff
@@ -46,18 +45,14 @@ def test_update_details(self):
...
-            'existing_password': 'password',
-            'new_password': 'new_password',
-            '_account': True,
+            'old_password': 'password',
+            'new_password1': 'new_password',
+            'new_password2': 'new_password',
+            '_password': True,
             })
-        token_url = response.context['token_url']
-        self.confirm(token_url)
```
smn, line 60: This was at one point required as a security measure. What's our plan to replace it?

go/account/views.py
smn, line 35: Ah username's completely gone?

/me pulls foot out of mouth.

go/base/models.py
```diff
@@ -13,18 +16,58 @@ def get_account_store():
...
-    # Enable search for the contact & group stores
-    user_api.contact_store.contacts.enable_search()
-    user_api.contact_store.groups.enable_search()
+class GoUserManager(BaseUserManager):
+    def _create_user(self, email, password, **kw):
+        if not email:
+            raise ValueError('Users must have an email address')
+        email = self.normalize_email(email)
+        user = self.model(email=email, password=password, **kw)
+        user.set_password(password)
```
smn, line 25: Why the double password setting with `password=password` and `set_password(password)` ?

go/base/models.py
```diff
@@ -47,3 +90,51 @@ def __unicode__(self):
...
+                Permission.objects.get_or_create(
+                    content_type=ct, codename=codename, name=name)
+
+
+def fix_auth_post_syncdb():
+    """Disconnects post_syncdb signals so that syncdb doesn't attempt to call
+       create_permissions or create_superuser before South migrations have had
+       chance to run. Connects a new post_syncdb hook that creates permissions
+       during tests.
+       """
```
smn, line 129: Is this a bug in South?

go/account/views.py
hodgestar, line 35: Yep, username is gone completely. :)

go/billing/settings.py
```diff
@@ -9,6 +9,9 @@
     settings, 'BILLING_ENDPOINT_DESCRIPTION_STRING',
     "tcp:9090:interface=127.0.0.1")
 
+AUTH_USER_TABLE = getattr(
+    settings, 'AUTH_USER_MODEL', 'auth_user').replace('.', '_')
+
```
smn, line 18: the user table is configurable per model.

``` python
>>> from django.conf import settings
>>> app_name, model_name = settings.AUTH_USER_MODEL.split('.')
>>> from django.db import models
>>> model = models.get_model(app_name, model_name)
>>> model._meta.db_table
u'base_gouser'
>>>
```

go/base/models.py
smn, line 129: Discussed in IRC.

smn:
:+1: in general, just two minor things: 
1. Why we're supplying the password param twice.
2. A small issue around the db table that may not be important right now.

go/base/models.py
hodgestar, line 25: Fixed.

go/base/models.py
hodgestar, line 129: Explanation from IRC:

```
[16:20:41] <Hodgestar> So, the madness of auth / South / syncdb:
[16:20:57] <Hodgestar> * auth has two post_syncdb hooks.
[16:21:06] <Hodgestar> * tests use syncb even for things that use south.
[16:21:35] <Hodgestar> * our Travis tests test migrations by setting up the postgres database at the start
[16:22:09] <Hodgestar> * so for tests we need to have the syncdb hooks but for the test migration on Travis we can't.
```

hodgestar:
I've addressed the two comments from @smn. Stuff I'd still like to do:
- [x] Sort out what is supposed to happen with `test_update_password`.
- [x] Put the list of permissions to create in a single place if possible.
- [x] Let @jerith finish his review.

hodgestar:
I decided that it's actually a good idea to keep the two lists of permissions to create in separate places because they're separate things. The one in the migration is "what should have originally been set up by django.contrib.auth". The one in `go.base.models` is "what would be set up by auth now". If they get out of sync the correct approach will be to update `go.base.models` and add a data migration.

hodgestar:
I created #768 to deal with the password changing auth token.

PR #753, opened 2013-11-05
Conversation and campaign metrics

justinvdm:
The aim of this ticket is to flesh out the metric infrastructure we need for #750. Once a PR is attached to this ticket and landed, #750 will use the added infrastructure as the new way of collecting and publishing metrics.

### DIFF
--- a/go/vumitools/metrics.py
+++ b/go/vumitools/metrics.py
@@ -1,0 +1,65 @@
+from django.conf import settings
+from twisted.internet.defer import inlineCallbacks
+
+from vumi.blinkenlights.metrics import Metric
+
+
+class GoMetric(object):
+    AGGREGATORS = None
+
+    def __init__(self, name, aggregators=None):
+        if aggregators is None:
+            aggregators = self.AGGREGATORS
+        self.metric = Metric(name, aggregators)
+
+    def get_full_name(self):
+        """
+        This is for constructing the full, prefixed metric name in django land
+        if a manager is not available.
+        """
+        return settings.GO_METRICS_PREFIX + self.metric.name
+
+    def get_value(self):
+        raise NotImplementedError(
+            "ConversationMetric.value() needs to be overriden")
+
+    @inlineCallbacks
+    def oneshot(self, manager, value=None):
+        if value is None:
+            value = self.get_value()
+        manager.oneshot(self.metric, (yield value))
+
+
+class ConversationMetric(GoMetric):
+    METRIC_NAME = None
+
+    def __init__(self, conv):
+        super(ConversationMetric, self).__init__(self.make_name(conv))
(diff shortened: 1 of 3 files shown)


justinvdm:
Ready for review.

go/vumitools/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+    def get_full_name(self):
+        """
+        This is for constructing the full, prefixed metric name in django land
+        if a manager is not available.
+        """
+        return settings.GO_METRICS_PREFIX + self.metric.name
+
+    def get_value(self):
+        raise NotImplementedError(
+            "ConversationMetric.value() needs to be overriden")
```
hodgestar: Should this perhaps be moved onto `ConversationMetric` or have the error message updated?

hodgestar: Either way we should add a docstring describing what `get_value` should return.

go/vumitools/tests/test_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __init__(self, *a, **kw):
+        super(ToyGoMetric, self).__init__(*a, **kw)
+        self.value = None
+
+    def set_value(self, value):
+        self.value = value
+
+    def get_value(self):
+        return self.value
```
hodgestar, line 46: We probably also want to test that we handled returning via `inlineCallbacks`, `deferreds` or `calls_manager`.

go/vumitools/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @inlineCallbacks
+    def oneshot(self, manager, value=None):
```
hodgestar, line 80: I think that perhaps we want a comment here making it clear that `oneshot` and `get_value` should only be called from Twisted (since metrics will also be used in Django land)?

go/vumitools/tests/test_metrics.py
hodgestar, line 46: Ah. I see you handled that differently. Woot.

go/vumitools/metrics.py
hodgestar, line 80: It might even we worthwhile naming methods `tx_oneshot` or something like that so it's totally clear?

hodgestar:
Left a few comments but generally happy for this to land and be fixed up in #750. :+1:

go/vumitools/metrics.py
justinvdm, line 80: I'm +1 on adding a comment to `oneshot`, but I think `get_value` will be used from django land for django metrics, so not sure if it make sense putting a similar comment there.

justinvdm:
Ready for another review.

hodgestar:
:+1:

PR #762, opened 2013-11-07
Harden HTTP API against DNSLookupErrors and missing conversation keys in messages.

hodgestar:


jerith:
:+1:, although I'd like tests for these two cases as well if it's not too hard to do.

jerith:
:+1:

PR #675, opened 2013-11-07
Contact look-up for messages without a transport_type raises an exception.

hodgestar:
Currently any attempt to display a message without a `transport_type` fails because the `delivery_class` look-up relies on a `transport_type` being set and contact look raises an error if the `delivery_class` is `None`.

Outbound messages are expected to not have a `transport_type` set when:
- They have yet to be dispatched towards a transport by the routing dispatcher because they are still in flight.
- Will never reach a transport because they conversation that sent them is not hooked up to one.

### DIFF
--- a/go/conversation/templates/conversation/message_list_table.html
+++ b/go/conversation/templates/conversation/message_list_table.html
@@ -15,10 +15,16 @@
             {% get_contact_for_message request.user_api message message_direction as contact %}
             <td><i class="glyphicon glyphicon-repeat"></i> {{conversation.delivery_class}}</td>
             <td>
+                {% if contact != None %}
                 <h6><a href="{% url 'contacts:person' person_key=contact.key %}">{{contact}}</a> via {{message.transport_type}} - {{message.timestamp}}</h6>
+                {% elif message_direction == 'inbound' %}
+                <h6><i>{{message.user}}</i> via {{message.transport_type}} (unsupported) - {{message.timestamp}}</h6>
+                {% else %}
+                <h6><i>{{message.to_addr}}</i> via {{message.transport_type}} (unsupported) - {{message.timestamp}}</h6>
+                {% endif %}
                 <p>{{message.content}}</p>
             </td>
-            {% if message_direction == 'inbound' and conversation.has_channel_supporting_generic_sends %}
+            {% if message_direction == 'inbound' and contact != None and conversation.has_channel_supporting_generic_sends %}
             <td>
                 <a href="#reply-{{message.message_id}}" class="btn btn-default" title="Reply" data-toggle='modal'><i class="glyphicon glyphicon-repeat"></i> Reply</a>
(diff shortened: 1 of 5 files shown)


go/conversation/templates/conversation/message_list_table.html
```diff
@@ -15,10 +15,16 @@
             {% get_contact_for_message request.user_api message message_direction as contact %}
             <td><i class="glyphicon glyphicon-repeat"></i> {{conversation.delivery_class}}</td>
             <td>
+                {% if contact != None %}
                 <h6><a href="{% url 'contacts:person' person_key=contact.key %}">{{contact}}</a> via {{message.transport_type}} - {{message.timestamp}}</h6>
+                {% elif message_direction == 'inbound' %}
+                <h6><i>{{message.user}}</i> via {{message.transport_type}} (unsupported) - {{message.timestamp}}</h6>
+                {% else %}
+                <h6><i>{{message.to_addr}}</i> via {{message.transport_type}} (unsupported) - {{message.timestamp}}</h6>
+                {% endif %}
```
justinvdm, line 24: I might be misunderstanding this, but shouldn't inbound messages show `message.from_addr` and outbound messages `message.user`?

justinvdm:
Erm, just realised this is probably still in progress. Will wait before continuing on the review.

justinvdm:
Update from IRC: PR is meant to be ready for review :)

go/conversation/templatetags/conversation_tags.py
```diff
@@ -41,6 +42,8 @@ def get_contact_for_message(user_api, message, direction='inbound'):
     # retrieving a contact return something useful for debugging (i.e.
     # the `transport_type` that failed to be looked up).
     delivery_class = user_api.delivery_class_for_msg(message)
+    if not user_api.contact_store.delivery_class_supported(delivery_class):
+        return None
```
justinvdm, line 45: `ContactError` is currently unused. Did you maybe mean to use it here?

I'm wondering if it was a good idea to change contacts to raise an error when we can't retrieve contacts. I probably should've gotten some feedback before making that change a few months ago, so maybe it needs discussion. For a later PR if anything, though.

go/conversation/templates/conversation/message_list_table.html
justinvdm, line 24: Oh right, `message.from_addr == message.user`. In that case, shouldn't it just be `message.user` for both? Feels like I'm missing something, though.

justinvdm, line 24: Ah, nevermind, makes sense. We want to always show the addr for the other end, not ours.

justinvdm:
Looks good, just a comment about the unused `ContactError` (the other comments were just me misunderstanding something, sorry for the noise :)

justinvdm:
:+1:

PR #758, opened 2013-11-08
Allow for a resource managing opt-in/opt-out status

smn:
Currently we have no way of allowing someone to opt back in after an opt-out.
Our complete approach in this area needs a rethink but this is a little step to at least solve a part of the problem.

### DIFF
--- a/go/apps/jsbox/opt_out.py
+++ b/go/apps/jsbox/opt_out.py
@@ -4,12 +4,15 @@
 """Resource for accessing and modifying a contact's opt-out/opt-in
    status from the sandbox"""
 
+from functools import wraps
+
 from twisted.internet.defer import inlineCallbacks, returnValue
 
 from vumi.application.sandbox import SandboxResource
 
 
 def optout_authorized(func):
+    @wraps(func)
     @inlineCallbacks
     def wrapper(self, api, command):
         if not (yield self.is_allowed(api)):
@@ -24,6 +27,7 @@ def optout_authorized(func):
 
 def ensure_params(*keys):
     def decorator(func):
+        @wraps(func)
         def wrapper(self, api, command):
             for key in keys:
                 if key not in command:
@@ -31,8 +35,8 @@ def ensure_params(*keys):
                                       reason='Missing key: %s' % (key,))
 
                 value = command[key]
-                # value is not allowed to be `False`, `None` or an empty
-                # string.
+                # value is not allowed to be `False`, `None`, `0`
+                # or an empty string.
                 if not value:
                     return self.reply(
                         command, success=False,
--- a/go/apps/jsbox/opt_out.py
+++ b/go/apps/jsbox/opt_out.py
@@ -1,4 +1,4 @@
-# -*- test-case-name: go.apps.jsbox.tests.test_optout -*-
+# -*- test-case-name: go.apps.jsbox.tests.test_opt_out -*-
 # -*- coding: utf-8 -*-
 
 """Resource for accessing and modifying a contact's opt-out/opt-in
@@ -43,7 +43,7 @@ def ensure_params(*keys):
     return decorator
 
 
-class OptoutResource(SandboxResource):
+class OptOutResource(SandboxResource):
 
     def get_user_api(self, api):
         return self.app_worker.user_api_for_api(api)
--- a/go/apps/jsbox/opt_out.py
+++ b/go/apps/jsbox/opt_out.py
@@ -35,6 +35,8 @@ def ensure_params(*keys):
                                       reason='Missing key: %s' % (key,))
 
                 value = command[key]
+                # NOTE: This needs to be updated once we have some proper
+                #       tools for validating command input
                 # value is not allowed to be `False`, `None`, `0`
                 # or an empty string.
                 if not value:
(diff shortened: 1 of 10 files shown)


smn:
@hodgestar @justinvdm @jerith @vgeddes ready for review. Will write a management command for setting the `can_manage_optouts` flag  on the `account` if you guys are happy with the current approach.

go/apps/jsbox/optout.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @ensure_params('address_type', 'address_value')
+    @optout_authorized
+    @inlineCallbacks
+    def handle_status(self, api, command):
+        """
+        Accepts an address_type and address_value and
+        retrieves the opt-out entry for it.
+
+        Returns ``None`` if it doesn't exist.
```
hodgestar: We should perhaps document some common `address_type` values here.

go/vumitools/account/models.py
```diff
@@ -42,6 +42,7 @@ class UserAccount(Model):
     event_handler_config = Json(default=list)
     msisdn = Unicode(max_length=255, null=True)
     confirm_start_conversation = Boolean(default=False)
+    can_manage_optouts = Boolean(default=False)
```
hodgestar, line 45: I'm wondering if we want a migration for this? It will probably work without it, but it might be nice to record the change and do a mini-migration anyway so the value is explicitly stored eventually.

go/vumitools/opt_out/models.py
```diff
@@ -61,3 +61,7 @@ def opt_outs_for_addresses(self, addr_type, addresses):
         mr.filter_not_found()
         opt_out_keys = yield mr.get_keys()
         returnValue(opt_out_keys)
+
+    def count(self):
+        return self.opt_outs.index_lookup(
+            'user_account', self.user_account_key).get_count()
```
hodgestar, line 67: We should add a test for this.

hodgestar:
General approach looks good. Left some comments while I was reading through.

go/vumitools/account/models.py
smn, line 45: Added

go/vumitools/opt_out/models.py
smn, line 67: Added

smn:
@hodgestar @jerith @justinvdm ready for re-review

go/apps/jsbox/opt_out.py
```diff
@@ -0,0 +1,185 @@
...
+        def wrapper(self, api, command):
+            for key in keys:
+                if key not in command:
+                    return self.reply(command, success=False,
+                                      reason='Missing key: %s' % (key,))
+
+                value = command[key]
+                # value is not allowed to be `False`, `None` or an empty
+                # string.
+                if not value:
```
justinvdm, line 42: I'm not sure if this will happen for any of the params in this resource's commands, but we will be sending failure replies if the value is `0`, which we might not want.

go/apps/jsbox/opt_out.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @ensure_params('address_type', 'address_value')
+    @optout_authorized
+    @inlineCallbacks
+    def handle_status(self, api, command):
+        """
+        Accepts an address_type and address_value and
+        retrieves the opt-out entry for it.
+
+        Returns ``None`` if it doesn't exist.
```
justinvdm: I'm probably missing something, but this line doesn't sound relevant.

go/apps/jsbox/opt_out.py
```diff
@@ -0,0 +1,185 @@
...
+            - ``address_type``: the type of address to opt-out.
+              At the moment only ``msisdn`` is used.
+            - ``address_value``: the value of the ``address_type`` to opt-out.
+            - ``message_id`` the message_id of the message that triggered
+              the opt-out, for auditing purposes.
+
+        Success reply fields:
+            - ``success``: set to ``true``
+            - ``opted_out``: set to ``true``
+            - ``created_at``: the timestamp of the opt-out (if opted out)
```
justinvdm: _(Epic bikeshed alert)_ I don't think "(if opted out)" makes sense here.

justinvdm:
Looks good, just a few small comments.

go/apps/jsbox/opt_out.py
```diff
@@ -0,0 +1,185 @@
...
+   status from the sandbox"""
+
+from twisted.internet.defer import inlineCallbacks, returnValue
+
+from vumi.application.sandbox import SandboxResource
+
+
+def optout_authorized(func):
+    @inlineCallbacks
+    def wrapper(self, api, command):
```
justinvdm, line 17: This is probably planned, just leaving a note here in case. We should probably use `functools.wrap` here and for `ensure_params` below like we did in #769.

go/apps/jsbox/opt_out.py
smn: You're right, docstring from old implementation. Removing.

go/apps/jsbox/opt_out.py
smn: Thanks! Copy past error.

justinvdm:
:+1:

smn:
thanks!

PR #750, opened 2013-11-08
Allow conversations metrics to be specified as part of the conversation definition

justinvdm:
This will give us a way to know which metrics should be collected,  how they should obtain their values, and will allow us to build up dashboards from the specified metrics.


go/vumitools/metrics.py
```diff
@@ -35,9 +35,17 @@ def get_value(self):
...
+        raise NotImplementedError(
+            "GoMetric.oneshot_with_value() needs to be overriden")
+
     def oneshot(self, *a, **kw):
         """
-        Should do a once-off publish for the metric.
+        Should do a once-off publish for the metric using `get_value()` to
+        calculate the metric's value.
         """
         raise NotImplementedError("GoMetric.oneshot() needs to be overriden")
```
justinvdm: Sometimes `get_value()` needs to take in args, so `oneshot` now proxies those args to `get_value`. This means we needed to create a separate `oneshot_with_value` for publishing values explicitly. Not sure if it would be better to just let `oneshot` take in the manager and a value, and have whatever is calling it first do a `get_value` for cases where that needs to happen. Thoughts?

justinvdm:
Ready for review.

go/apps/subscription/tests/test_definition.py
```diff
@@ -0,0 +1,37 @@
...
+        self.conv_def = ConversationDefinition(self.conv)
+
+    def test_metrics_retrieval(self):
+        [m1, m2, m3, m4, m5, m6] = self.conv_def.get_metrics()
+
+        self.assertTrue(isinstance(m1, MessagesSentMetric))
+        self.assertTrue(isinstance(m2, MessagesReceivedMetric))
+
+        self.assertTrue(m3.get_full_name().endswith('campaign-1.subscribed'))
+        self.assertTrue(isinstance(m3, SubscribedMetric))
```
hodgestar, line 41: Maybe these should rather do `self.assertEqual(m3.metric.name, 'campaign-1.subscribed')`?

go/apps/subscription/tests/test_definition.py
```diff
@@ -0,0 +1,37 @@
...
+
+        self.assertTrue(m4.get_full_name().endswith('campaign-2.subscribed'))
+        self.assertTrue(isinstance(m4, SubscribedMetric))
+
+        self.assertTrue(m5.get_full_name().endswith('campaign-1.unsubscribed'))
+        self.assertTrue(isinstance(m5, UnsubscribedMetric))
+
+        self.assertTrue(m6.get_full_name().endswith('campaign-2.unsubscribed'))
+        self.assertTrue(isinstance(m6, UnsubscribedMetric))
```
hodgestar, line 56: I think we also need a test for `get_value()`?

go/apps/subscription/metrics.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @classmethod
+    def make_metric_name(cls, campaign_name):
+        return "%s.%s" % (campaign_name, cls.METRIC_NAME)
+
+    def get_value(self, vumi_app, user_api):
+        contacts = user_api.contact_store.contacts
+        search = contacts.raw_search(
+            "subscription-%s:%s" %
+            (self.campaign_name, self.METRIC_NAME))
```
hodgestar: I think it'd prefer `METRIC_NAME` to not also be used as the value to look up in contacts. Maybe it could be a separate value on the class?

go/apps/subscription/tests/test_definition.py
hodgestar, line 56: Ah. This is tested in `test_metrics`. Woot.

go/apps/subscription/tests/tests_metrics.py
```diff
@@ -0,0 +1,53 @@
+from twisted.internet.defer import inlineCallbacks
+from go.apps.subscription.metrics import SubscriptionMetric
```
hodgestar: Filename, `tests_metrics.py`, should maybe be `test_metrics.py`?

go/apps/subscription/tests/test_definition.py
justinvdm, line 41: `m3.get_full_name() == settings.GO_METRIC_PREFIX + m3.metric.name`, `'campaign-1.subscribed'` is just the end of `m3.metric.name`. I'll test with the full name.

justinvdm:
Ready for another review.

go/apps/subscription/tests/test_metrics.py
```diff
@@ -0,0 +1,54 @@
+from twisted.internet.defer import inlineCallbacks
+from go.apps.subscription.metrics import SubscriptionMetric
+from go.vumitools.tests.utils import TxMetricTestBase
+from go.vumitools.contact import ContactStore
+
+
+class ToySubscriptionMetric(SubscriptionMetric):
+    METRIC_NAME = 'toy-subscription-metric'
+    CONTACT_LOOKUP_KEY = 'toy-subscription-metric'
```
hodgestar: Maybe these should have different values for the tests?

go/vumitools/metrics.py
hodgestar: This got into a bit of mess. Let me think a bit about how to make it better.

hodgestar:
Other than the minor test comment and the `oneshot` / `oneshot_with_value` craziness, looks good.

go/vumitools/metrics.py
hodgestar: I think `Metric.oneshot` was a bit of a bad idea (it's confusing the metric which is a name and a way to get a value in this case) with the manager / connection (which is a way to publish a metric and a value). I'd rather not add fixing that to this PR though (or even to our short-term todo list). So what if we removed `oneshot_with_value` and made the base signature of `oneshot`:

```
def oneshot(self, manager_or_connection, **kw):
```

If value is `None`, we call `self.get_value(**kw)`, otherwise we use the passed in value. Then we call `self.publish_value(manager_or_connection, value)` (the equivalent of `.oneshot_with_value`) and the Django and Tx classes only need to override `.get_value` and `.publish_value`.

I guess this is a bit like just renaming `oneshot_with_value`  and changing it's signature a bit? Thoughts.

hodgestar:
Reviewed. Hopefully I'm not painting the bikeshed late a night when I can't see the colours properly.

go/vumitools/metrics.py
justinvdm: Sounds like a good approach. A few questions:
- Did you mean to have a `value=None` in the `oneshot` definition example? (just to be sure I'm understanding this)
- In order to have `GoMetric` implement `oneshot`, it would need to handle cases where `get_value()` returns a deferred. For django metrics, that wouldn't make sense, so `DjangoMetric` and `TxMetric` would need to have their own `oneshot`s. That okay?

hodgestar: - Yes to `value=None` being intentional (or at least it seemed like a good
  idea last night :).
- +1 on implementing oneshot separately for Tx and Django metrics (less
  than ideal, but easier than fiddling with @calls_manager).

justinvdm: Cool, thanks :) I'd be happier knowing that `oneshot` is in a better state, so I think I'll make the change in this PR if that is alright.

hodgestar: Fine by me.

justinvdm:
Ready for another review.

hodgestar:
Something I forgot to mention -- `user_api.api` is the `vumi_api` so you don't need to pass the `vumi_api` and the `user_api`.

hodgestar:
Looks good otherwise.

hodgestar:
:+1:

PR #769, opened 2013-11-12
Fix docs for MessageStoreResource

smn:
docstrings going missing because of sloppy decorating.


justinvdm:
:+1: Sorry for not catching that in the original PR.

smn:
thanks!

PR #770, opened 2013-11-13
HTTP API transport should not accept messages where the to_addr or from_addr are not unicode strings.

hodgestar:
Example traceback:

``` python
2013-11-13 15:02:43+0000 [VumiRedis,client] Unhandled Error
        Traceback (most recent call last):
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 421, in errback
            self._startRunCallbacks(fail)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 488, in _startRun
Callbacks
            self._runCallbacks()
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 575, in _runCallb
acks
            current.result = callback(current.result, *args, **kw)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1126, in gotResul
t
            _inlineCallbacks(r, g, deferred)
        --- <exception caught here> ---
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1068, in _inlineC
allbacks
            result = result.throwExceptionIntoGenerator(g)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
          File "/var/praekelt/vumi-go/go/apps/http_api/resource.py", line 161, in handle_PUT
            yield self.handle_PUT_send_to(request, payload)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1068, in _inlineCallbacks
            result = result.throwExceptionIntoGenerator(g)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
          File "/var/praekelt/vumi-go/go/apps/http_api/resource.py", line 215, in handle_PUT_send_to
            to_addr, content, endpoint='default', **msg_options)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1068, in _inlineCallbacks
            result = result.throwExceptionIntoGenerator(g)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
          File "/var/praekelt/vumi-go/ve/src/vumi/vumi/middleware/base.py", line 121, in _handle
            message = yield handler(message, connector_name)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1068, in _inlineCallbacks
            result = result.throwExceptionIntoGenerator(g)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
          File "/var/praekelt/vumi-go/go/vumitools/middleware.py", line 337, in handle_outbound
            yield self.store.add_outbound_message(message, batch_id=batch_id)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1068, in _inlineCallbacks
            result = result.throwExceptionIntoGenerator(g)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
          File "/var/praekelt/vumi-go/ve/src/vumi/vumi/components/message_store.py", line 215, in add_outbound_message
            yield self.cache.add_outbound_message(batch_id, msg)
          File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1070, in _inlineCallbacks
            result = g.send(result)
          File "/var/praekelt/vumi-go/ve/src/vumi/vumi/components/message_store_cache.py", line 141, in add_outbound_message
            yield self.add_to_addr(batch_id, msg['to_addr'], timestamp)
          File "/var/praekelt/vumi-go/ve/src/vumi/vumi/components/message_store_cache.py", line 238, in add_to_addr
            to_addr.encode('utf-8'): timestamp,
        exceptions.AttributeError: 'int' object has no attribute 'encode'
```

### DIFF
--- a/go/apps/http_api/resource.py
+++ b/go/apps/http_api/resource.py
@@ -261,6 +261,10 @@ class MessageStream(StreamResource):
         conversation = yield self.get_conversation(user_account)
 
         msg_options = SendToOptions(payload)
+        if msg_options.error:
+            self.client_error_response(request, msg_options.error)
+            return
+
         helper_metadata = conversation.set_go_helper_metadata()
 
         msg = yield self.worker.send_to(
--- a/go/apps/http_api/resource.py
+++ b/go/apps/http_api/resource.py
@@ -132,6 +132,50 @@ class EventStream(StreamResource):
     routing_key = '%(transport_name)s.stream.event.%(conversation_key)s'
 
 
+class MsgOptions(object):
+    """Helper for sanitizing msg options from clients."""
+
+    WHITELIST = {}
+
+    def __init__(self, payload):
+        self.error = None
+        for key, checker in self.WHITELIST.iteritems():
+            value = payload.get(key)
+            if not checker(value):
+                self.error = (
+                    "Invalid or missing value for payload key %r" % (key,))
+                break
+            setattr(self, key, value)
+
+
+class MsgCheckHelpers(object):
+    @staticmethod
+    def is_unicode_or_none(value):
+        return (value is None) or (isinstance(value, unicode))
+
+    @staticmethod
+    def is_session_event(value):
+        return value in TransportUserMessage.SESSION_EVENTS
+
+
+class SendToOptions(MsgOptions):
+    """Payload options for messages sent with `.send_to(...)`."""
+
+    WHITELIST = {
+        'content': MsgCheckHelpers.is_unicode_or_none,
+        'to_addr': MsgCheckHelpers.is_unicode_or_none,
+    }
+
+
+class ReplyToOptions(MsgOptions):
+    """Payload options for messages sent with `.reply_to(...)`."""
+
+    WHITELIST = {
+        'content': MsgCheckHelpers.is_unicode_or_none,
+        'session_event': MsgCheckHelpers.is_session_event,
+    }
+
+
 class MessageStream(StreamResource):
 
     message_class = TransportUserMessage
@@ -156,13 +200,6 @@ class MessageStream(StreamResource):
         if load_balancer is not None:
             return {'load_balancer': copy.deepcopy(load_balancer)}
         return {}
-
-    def get_msg_options(self, payload, white_list=[]):
-        raw_payload = copy.deepcopy(payload.copy())
-        msg_options = dict((key, value)
-                           for key, value in raw_payload.items()
-                           if key in white_list)
-        return msg_options
 
     def get_conversation_tag(self, conversation):
         return (conversation.delivery_tag_pool, conversation.delivery_tag)
@@ -202,16 +239,18 @@ class MessageStream(StreamResource):
             self.client_error_response(request, 'Invalid in_reply_to value')
             return
 
-        msg_options = self.get_msg_options(payload,
-                                           ['session_event', 'content'])
-        content = msg_options.pop('content')
-        continue_session = (msg_options.pop('session_event', None)
+        msg_options = ReplyToOptions(payload)
+        if msg_options.error:
+            self.client_error_response(request, msg_options.error)
+            return
+
+        continue_session = (msg_options.session_event
                             != TransportUserMessage.SESSION_CLOSE)
         helper_metadata = conversation.set_go_helper_metadata()
         helper_metadata.update(self.get_load_balancer_metadata(payload))
 
         msg = yield self.worker.reply_to(
-            reply_to, content, continue_session,
+            reply_to, msg_options.content, continue_session,
             helper_metadata=helper_metadata)
 
         self.successful_send_response(request, msg)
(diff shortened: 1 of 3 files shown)


go/apps/http_api/resource.py
```diff
@@ -110,6 +132,50 @@ class EventStream(StreamResource):
...
+
+class ReplyToOptions(MsgOptions):
+    """Payload options for messages sent with `.reply_to(...)`."""
+
+    WHITELIST = {
+        'content': MsgCheckHelpers.is_unicode_or_none,
+        'session_event': MsgCheckHelpers.is_session_event,
+    }
+
+
```
justinvdm, line 191: I'm wondering if these helpers should live in vumi, might be cool to have them be reusable for other cases where we might be getting messages from outside of vumi.

go/apps/http_api/resource.py
```diff
@@ -110,6 +132,50 @@ class EventStream(StreamResource):
...
+    WHITELIST = {}
+
+    def __init__(self, payload):
+        self.error = None
+        for key, checker in self.WHITELIST.iteritems():
+            value = payload.get(key)
+            if not checker(value):
+                self.error = (
+                    "Invalid or missing value for payload key %r" % (key,))
+                break
```
justinvdm: I've seen some http apis carry on checking other fields and responding with a list of all the errors instead of short circuiting on the first error found. The current approach looks good, but thought I should bring it up. Thoughts?

go/apps/http_api/resource.py
```diff
@@ -180,43 +236,42 @@ def handle_PUT_in_reply_to(self, request, payload, in_reply_to):
                         ' key' % (reply_to,))
             msg_conversation_key = None
         if msg_conversation_key != conversation.key:
-            request.setResponseCode(http.BAD_REQUEST)
-            request.write('Invalid in_reply_to value')
```
justinvdm, line 184: Our responses body used to be a plain string, but we've changed it to be JSON response now. Just a bit a bit worried that clients are checking the response body in a way that will break when this change is introduced.

justinvdm, line 184: Based on RL discussion: The plan is to keep the change, the likelihood of a client doing things with the response isn't high.

justinvdm:
We have tests for everything that uses `MsgOptions` and friends, but not for `MsgOptions` itself. We probably don't need this if we don't move `MsgOptions` into vumi, though.

go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -300,40 +381,50 @@ def test_invalid_in_reply_to_with_missing_conversation_key(self):
...
+            'content': 'foo',
+            'in_reply_to': inbound_msg['message_id'],
+            'message_id': 'evil_id'
+        }
+
+        url = '%s/%s/messages.json' % (self.url, self.conversation.key)
+        response = yield http_request_full(url, json.dumps(msg),
+                                           self.auth_headers, method='PUT')
+
+        self.assertEqual(response.code, http.OK)
```
justinvdm, line 422: Just to be sure, do we send a 200 response for messages with evil message ids because we are just sending back the bad message that was sent? Just wondering why we don't send back a 'bad request' response for this case.

justinvdm:
Looks good, just a few comments, mostly me making sure I'm understanding things.

go/apps/http_api/resource.py
hodgestar, line 191: Broadly I think I agree -- it would be nice to have better tools for dealing with message payloads in Vumi. Not quite sure how to structure it right now though (i.e. what a minimal useful feature set is).

go/apps/http_api/resource.py
hodgestar: This is a good idea. Let me not be lazy and just implement it quickly.

go/apps/http_api/resource.py
justinvdm, line 191: Cool. I guess it makes sense to leave these here then, and replace them once those structures are in place in Vumi.

hodgestar:
Tests added for `MsgOptions`.

go/apps/http_api/tests/test_vumi_app.py
hodgestar, line 422: Bad message ids are just ignored (as are any other unexpected fields). I think this was an intentional design choice (be liberal in what one accepts and all). There's a specific test for ignoring the message id because using a message id passed in from outside would be a security hole.

hodgestar:
I think it's ready for review again.

justinvdm:
:+1:

PR #774, opened 2013-11-16
MetricsMiddleware doesn't expire keys.

smn:
Leaves _lots_ of keys lying around.

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -193,6 +193,9 @@ class MetricsMiddleware(BaseMiddleware):
         received its `message_id` is stored and when a reply for the given
         `message_id` is sent out, the timestamps are compared and a averaged
         metric is published.
+    :param int max_lifetime:
+        How long to keep a timestamp for. Anything older than this is trashed.
+        Defaults to 60 seconds.
     :param dict redis_manager:
         Connection configuration details for Redis.
     :param str op_mode:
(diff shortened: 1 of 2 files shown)


smn:
ohai memory!

```
$ redis-cli keys "*metrics*" | wc -l
7253232
```

go/vumitools/middleware.py
```diff
@@ -263,7 +267,9 @@ def key(self, transport_name, message_id):
 
     def set_inbound_timestamp(self, transport_name, message):
         key = self.key(transport_name, message['message_id'])
-        return self.redis.set(key, repr(time.time()))
+        d = self.redis.set(key, repr(time.time()))
+        d.addCallback(lambda _: self.redis.expire(key, self.max_lifetime))
+        return d
```
jerith: Maybe use `setex()` here?

jerith:
Nice catch.

go/vumitools/tests/test_middleware.py
```diff
@@ -306,6 +306,15 @@ def metric_name(status):
...
         self.assertEqual(unspecified[1], 1)
 
+    @inlineCallbacks
+    def test_expiry(self):
+        mw = yield self.get_middleware({'max_lifetime': 10})
+        msg1 = self.msg_helper.make_inbound('foo', transport_name='endpoint_0')
+        yield mw.handle_inbound(msg1, 'dummy_endpoint')
+        key = mw.key('dummy_endpoint', msg1['message_id'])
+        ttl = yield mw.redis.ttl(key)
+        self.assertTrue(0 < ttl < 60)
```
hodgestar: Should this check perhaps be `0 < ttl <= 10` since the `max_lifetime` is set to `10` in the config we pass in?

hodgestar:
Other than the minor comment on the test and jerith's suggestion to use `setex`, :+1:.

Once this lands we need to set an expiry time on all these keys in our production system.

PR #752, opened 2013-11-18
Add credit balance and billing statements to Vumi Go UI

andreit:


go/account/templates/account/billing.html
```diff
@@ -8,18 +10,31 @@
...
+                        <th></th>
+                        <th>
+                            {% if o == '-month,-year' %}
+                                <a href="?{% add_params request o='year,month'|split:',' %}">Date</a>
+                                <span class="glyphicon glyphicon-chevron-up"></span>
+                            {% elif o == 'month,year' %}
+                                <a href="?{% add_params request o='-year,-month'|split:',' %}">Date</a>
+                                <span class="glyphicon glyphicon-chevron-down"></span>
+                            {% else %}
+                                <a href="?{% add_params request o='year,month'|split:',' %}">Date</a>
```
hodgestar: I think this needs to be cleaned up a bit (maybe by setting something in the view function?). As written it's a bit mad (three copies of the same `<a>` tag just to add some parameters) and error prone.

go/account/templates/account/details.html
```diff
@@ -15,31 +15,29 @@
...
+                    <p>View a list of monthly statements</p>
+                </li>
+                {% comment %}
+                <li>
+                    <a href="{% url "account:user_list" %}">
+                        Users and permissions<br>
+                        <span>Add and edit users and access rights</span>
+                    </a>
+                </li>
+                {% endcomment %}
```
hodgestar: Can you explain a bit what the commented out section is for? I thought this was just the per-account billing page and I'm not sure what user permissions would need to be set here?

go/account/templates/account/billing.html
andreit: Agreed. I am actually looking for a better way to do this.

go/account/templates/account/details.html
hodgestar: Oh, I see. This came from the previous code. Unless you know what it's supposed to do, maybe just delete it.

andreit: yeah, wasn't sure what it was, so i left it there... will remove it.

go/account/views.py
```diff
@@ -16,6 +19,10 @@
 from go.account.tasks import update_account_details
 from go.base.models import UserProfile
 from go.token.django_token_manager import DjangoTokenManager
+from go.billing.models import MonthlyStatement
+
+
+STATEMENTS_PER_PAGE = 12
```
hodgestar: Maybe this should go into `go.billing.settings`?

go/account/views.py
```diff
@@ -171,17 +178,55 @@ def user_detail(request, user_id=None):
...
-    # TODO: Complete once billing data exists.
+@login_required
+def billing(request, template_name='account/billing.html'):
+    """Display a list of available statements for the logged in user"""
+    try:
+        order_by = request.GET.getlist('o', ['-year', '-month'])
+        account = request.user.account_set.all()[0]
+        statement_list = MonthlyStatement.objects\
+            .filter(account=account)\
+            .order_by(*order_by)
```
hodgestar: We definitely need to sanity check the value passed in to `o` before passing it to `order_by`.

go/account/views.py
andreit: Cool, it was done like that in the dashboard view. I thought it was the convention. :)

hodgestar: We just weren't awesome enough to add proper settings for the dashboard. :)

go/account/views.py
```diff
@@ -171,17 +178,55 @@ def user_detail(request, user_id=None):
...
+
+    except IndexError:
+        statement_list = []
```
hodgestar, line 202: I would rather do an `if account_set` check than wrap all these statements in an `IndexError` (too easy for one of them to call something that raises an `IndexError`).

go/account/views.py
```diff
@@ -171,17 +178,55 @@ def user_detail(request, user_id=None):
...
+    paginator = Paginator(statement_list, STATEMENTS_PER_PAGE)
+    try:
+        page = paginator.page(request.GET.get('p', 1))
+    except PageNotAnInteger:
+        page = paginator.page(1)
+    except EmptyPage:
+        page = paginator.page(paginator.num_pages)
+
+    context = {
+        'statement_list': statement_list,
```
hodgestar: Maybe we don't need to pass in `statement_list` any more?

andreit: Yip, will take out.

go/base/templatetags/go_tags.py
```diff
@@ -5,3 +7,44 @@
...
+def get_param(request, name, default=''):
+    """Get the URL parameter value.
+
+    If the value is list return a CSV string.
+    """
+    value = request.GET.getlist(name)
+    if value:
+        value.sort()
+        return ','.join(value)
+    return default
```
hodgestar: `add_params` seems useful but I'm a bit unsure about `get_param` -- shouldn't parameter fetching should be done in the view function?

andreit: I needed a way to get parameters which are lists in a template, but as I refactor the ugly bit above that outputs the sortable header this will probably fall away.

go/billing/api.py
```diff
@@ -174,11 +178,24 @@ class AccountResource(BaseResource):
...
+            month = request.args.get('month', None)
+            if year and month:
+                d = self.get_account_statement(account_number, year[0],
+                                               month[0])
+
+                d.addCallbacks(self._render_to_json, self._handle_error,
+                               callbackArgs=[request], errbackArgs=[request])
+            else:
+                self._handle_bad_request(request)
+
```
hodgestar: This `render_GET` is now serving too many different things, I think. Maybe we need sub-resource and to change how we select whether to server statements or account information / lists?

andreit: I can add a `Statement` resource for that and hook it up to `/statements/` or something?

hodgestar: A statements resource sounds good.

go/billing/models.py
```diff
@@ -114,3 +114,30 @@ class Transaction(models.Model):
...
     def __unicode__(self):
         return unicode(self.pk)
+
+
+class MonthlyStatement(models.Model):
+    """Monthly account statement data"""
+
+    account = models.ForeignKey(Account, related_name='monthly_statements')
+    year = models.PositiveIntegerField()
+    month = models.PositiveIntegerField()
```
hodgestar: I can see us needing statements that aren't related to a particular month (e.g. weekly statements for high traffic accounts or adhoc statements). Those don't need to be supported right now but could we perhaps change this model to rather than `title`, `start_time` and `end_time` fields instead of `year` and `month`?

go/billing/models.py
```diff
@@ -114,3 +114,30 @@ class Transaction(models.Model):
...
+        return u"%s (%d/%d)" % (self.account, self.year, self.month)
+
+
+class LineItem(models.Model):
+    """Monthly statement line item"""
+
+    statement = models.ForeignKey(MonthlyStatement, related_name='line_items')
+    tag_pool_name = models.CharField(max_length=100, blank=True)
+    tag_name = models.CharField(max_length=100, blank=True)
+    message_direction = models.CharField(max_length=20, blank=True)
```
hodgestar: I think we're soon going to have LineItems that aren't messaging costs. Any thoughts on how we should support those?

go/billing/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+@inlineCallbacks
+def _generate_monthly_account_statements():
+    try:
+        billing_api = BillingApi(settings.GO_BILLING_API_URL)
+        today = date.today()
+        last_month = today - relativedelta(months=1)
+        account_list = Account.objects.all()
+        for account in account_list:
+            monthly_statement, created = account.monthly_statements\
+                .get_or_create(year=last_month.year, month=last_month.month)
```
hodgestar: This should always be an object creationg -- retrieving an existing statement and adding line items to it would be bad.

hodgestar: Ah, I see there is the `if created` below. Maybe we should just always create one?

go/billing/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+
+            if created:
+                result = yield billing_api.get_account_statement(
+                    account.account_number, last_month.year, last_month.month)
```
hodgestar: Should get account statement really be an API thing? It seems more like a database operation that could just be done in Django? Although being able to retrieve line items within a particular period seems useful? Also a bit worried about naming `get_account_statement` sounds like it should return a billing statement when what it actually does is return line items within a date range.

go/billing/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+    except Exception as error:
+        log.err(error)
+    finally:
+        reactor.stop()
+
+
+@task(ignore_result=True)
+def generate_monthly_account_statements(interval):
+    reactor.callWhenRunning(_generate_monthly_account_statements)
+    reactor.run()
```
hodgestar: Use of reactor inside a celery task is interesting. Any idea where there are any things to aware of when doing this?

andreit: I was hoping you can give me some advice on this. Wasn't 100% comfortable doing this either. :)

hodgestar: We've tended to stick to plain Django methods inside celery tasks (which you're welcome to do in this case too although it means restructuring things a bit). You might also consider spawning sub-tasks for each account so that you don't end up with one task that takes forever when there are lots of accounts.

go/billing/tasks.py
andreit: I was also in 2 minds about this. Perhaps the task should do all the work of generating the statement. The task itself is located under the `go.billing.tasks` module, so logically its all on one place.

go/billing/tasks.py
hodgestar: Out of interest, does this Twisted approach actually work?

go/billing/tasks.py
hodgestar: I'm fine with the task doing all of the work.

go/billing/tasks.py
andreit: Yeah it does, but I would feel a lot more comfortable if the task just interacted with the DB directly.

go/billing/tasks.py
andreit: I did that to avoid creating multiple statements for the same month, so if the query found a statement for the given month/year it won't try to create another one.

go/billing/models.py
andreit: If I am going to add `title`, `from_date` and `to_date` to `MonthlyStatement` maybe we should rename `MonthlyStatement` to just `Statement` and the `LineItem` to something like `MessageCostLineItem`?

go/billing/tasks.py
andreit: Cool, do you still want me to implement the `Statement` resource in the API then? or shall we leave it for when it becomes necessary?

go/billing/models.py
hodgestar: Happy with just `Statement`. Happy for `LineItem` to stay with the same
name, unless you have a better idea for supporting other line items than
adding fields to `LineItem`.

andreit:
@hodgestar Can we upgrade celery to a version that supports `day_of_month` in the `crontab()` setup? I would like to run the statement task on the 1st of every new month.

hodgestar:
Happy for you to update the version in requirements.pip in this branch and
we can see if anything breaks?

andreit:
@hodgestar Not sure why the build is failing now. Can't seem to find `go.testsettings`. Do you have any idea why it could be happening?

hodgestar:
@andreit My guess is that `go.testsettings` raises an error while being imported.

andreit:
@hodgestar django-celery and celery version 3.0.23 seems to work and tests pass. Is it possible to deploy this branch on a test instance to verify that all is well?

hodgestar:
I think if tests pass that's fine -- we will just deploy to production
carefully.

andreit:
@hodgestar Could you please review the new solution of doing the work in the celery task? Thanks.

hodgestar:
Can we make `title = models.CharField(max_length=100)` a bit bigger? Say 255? Titles can easily get a bit long-winded.

hodgestar:
Doing the work in the celery task looks good -- much cleaner. Some nice to haves:
- [ ] splitting the celery task up into sub-tasks (so have a main task that spawns a sub-task for each account).
- [ ] adding a `type` field to the statement model (so that if we add adhoc statements later we can identify them).

hodgestar:
Otherwise looks good.

andreit:
@hodgestar Changes done. Is there anything else we need to land this one?

hodgestar:
Looks good. Last thing:
- I don't think we want to use ignore_result. It would be better to have the parent task return `group(list_of_results)`. I'm assuming the celery beat worker does the right thing and waits to process tasks with results. Our tests should then also check the results returned.

Otherwise looks good.

andreit:
@hodgestar Updated tasks as requested. Are you referring to the result of the individual account task that should be checked?

hodgestar:
@andreit I think what you committed looks good.

andreit:
@hodgestar Cool, can we merge this pull request?

andreit:
@hodgestar Is there something outstanding on this PR?

hodgestar:
Nope. I've just been getting other things out the way. Will try land today
(although I need to negotiate with @jerith a bit around whether we need to
land #783 and #786 first).

andreit:
Ok. Thanks.

PR #779, opened 2013-11-19
500 when trying to an account to billing Admin section

hodgestar:
When trying to add an account to the billing section: http://qa-vumi.za.prk-host.net/admin/billing/account/add/

Once I have entered all the info and clicked 'save' I get a 500

screen shot 2013-11-18 at 11 38 41 am


justinvdm:
:+1:

PR #780, opened 2013-11-19
Fake server test utility

justinvdm:
We need this for faking diamondash's api responses.


go/base/tests/utils.py
```diff
@@ -356,3 +360,99 @@ def assert_command_error(self, regexp, *command, **options):
     def assert_command_output(self, expected_output, *command, **options):
         self.call_command(*command, **options)
         self.assertEqual(expected_output, self.command.stdout.getvalue())
+
+
+class FakeResponse(requests.Response):
+    def __init__(self, content=None, data=None, code=200):
+        super(FakeResponse, self).__init__()
+        self.status_code = code
```
hodgestar, line 368: Could you remind me why it was necessary to inherit from requests.Response?

go/base/tests/utils.py
```diff
@@ -356,3 +360,99 @@ def assert_command_error(self, regexp, *command, **options):
...
+            self._content = json.dumps(data)
+        else:
+            self._content = ""
+
+    @property
+    def reason(self):
+        return self._content
+
+
+class FakeRpcResponse(FakeResponse):
```
hodgestar, line 382: Maybe this should be called FakeJsonRpcResponse for clarity?

go/base/tests/utils.py
```diff
@@ -356,3 +360,99 @@ def assert_command_error(self, regexp, *command, **options):
...
+            kwargs['method'] = method
+
+            if 'data' in kwargs:
+                kwargs['data'] = json.loads(kwargs['data'])
+
+            self.requests.append(kwargs)
+            resp = self.response
+        else:
+            original = self.originals[method]
+            resp = original(url, **kwargs)
```
hodgestar, line 453: Should we perhaps raise an error here instead?

hodgestar:
I'm a little concerned that we're mocking out a lot of the requests library in a very broad fashion.

go/base/tests/utils.py
justinvdm, line 368: `FakeResponse` is meant to be given as a response to requests issues using `requests`. For example, `FakeServer` stubs calls to `requests` and sends back a `FakeResponse` instance. The code we are testing is expecting a `Response` object, so we have two choices: mock out everything `Response` does, or subclass `Response` and give it only the things it needs.

I think the latter is a better approach because we can expect `Response` to follow its usual behaviour almost all the time, instead of needing to mimic it. Using the first approach, we would need to add stubbed methods every time we need a particular method in the code we are testing.

justinvdm:
I understand. I'm up for alternatives.

go/base/tests/utils.py
justinvdm, line 453: This is here so we can use multiple `FakeServer`s in the same test, if need be.

hodgestar, line 453: Ah! Cool. Maybe add a comment about that?

justinvdm:
Closing this in favor of #785

PR #782, opened 2013-11-20
Move VUMI_INSTALLED_APPS into a separate file so it's accessible to Twisted workers.

hodgestar:
This is a temporary solution until we have a more complete share configuration framework.

### DIFF
--- a/go/base/utils.py
+++ b/go/base/utils.py
@@ -8,8 +8,11 @@ from django import forms
 from django.http import Http404
 from django.conf import settings
 
-from go.base.amqp import connection
 from go.errors import UnknownConversationType, UnknownRouterType
+from go.config import (
+    get_conversation_pkg, get_router_pkg,
+    obsolete_conversation_types, obsolete_router_types)
+from go.base.amqp import connection
 from go.vumitools.api import VumiApi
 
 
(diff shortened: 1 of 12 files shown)


jerith:
Looks reasonable. :+1: once Travis is happy.

go/base/utils.py
```diff
@@ -187,7 +148,18 @@ def get_conversation_view_definition(conversation_type, conv=None):
...
     # Scoped import to avoid circular deps.
     from go.router.view_definition import RouterViewDefinitionBase
-    router_pkg = get_router_pkg(router_type, ['definition', 'view_definition'])
+    try:
+        router_pkg = get_router_pkg(
+            router_type, ['definition', 'view_definition'])
+    except UnknownRouterType:
+        # To handle obsolete routers that are still viewable
+        if router_type not in obsolete_router_types():
+            raise
```
justinvdm, line 157: Shouldn't we be raising something here?

justinvdm, line 157: Could we maybe add a test for this case, and a similar test for the case of obsolete conversation types?

hodgestar, line 157: A "raise" by itself re-raises the current exception.

hodgestar, line 157: Test was added. :)

justinvdm, line 157: Ah, I see :) /me goes back to python school

justinvdm:
Looks good. :+1: from me once Travis is happy.

jerith:
Lots of `AttributeError: 'Settings' object has no attribute 'VUMI_INSTALLED_ROUTERS'` in the Travis builds, so :-1: until those are fixed.

jerith:
I'm happy with this once we have a successful build.

PR #783, opened 2013-11-21
Switch to test helpers instead of TestCase classes

jerith:
This is the bulk of the test cleanup work. There are still a few things that could use improving, but we can probably do those incrementally during other work.

I'm breaking this up into smaller bits to make reviewing it possible:
- [x] #786


go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+            self._cleanup_funcs = []
+        self._cleanup_funcs.append((func, args, kw))
+
+
+class DjangoVumiApiHelper(object):
+    is_sync = True  # For when we're being treated like a VumiApiHelper.
+
+    def __init__(self, use_riak=True):
+        self.use_riak = use_riak
+        self._vumi_helper = VumiApiHelper(is_sync=True, use_riak=use_riak)
```
hodgestar, line 62: Seems a bit odd to set `is_sync=True` here and on the class? Maybe `is_sync` should be an argument too like it is for `VumiApiHelper`? I think `use_riak` should go onto `VumiApiHelper` the same way `is_sync` does -- it seems generally useful to be able to tell what sort of vumi_helper one has. Maybe `DjangoVumiApiHelper` could have properties that proxy to `vumi_helper` for these two attributes if needed (although they don't appear to be used in the class itself).

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        self.restore_django_bits()
+        for patch in reversed(self._settings_patches):
+            patch.disable()
+
+    def replace_django_bits(self):
+        # We do this redis manager hackery here because we might use it from
+        # Django-land before setting (or without) up a vumi_api.
+        # TODO: Find a nicer way to give everything the same fake redis.
+        pcfg = self._vumi_helper._persistence_helper._config_overrides
+        pcfg['redis_manager']['FAKE_REDIS'] = self.get_redis_manager()
```
hodgestar, line 94: Would it make sense for this to be the default behaviour for the persistence helper? I.e. to store the created redis instance for everyone? Alternatively, perhaps we should make this feature an optional argument to the persistence_helper constructor? E.g. `PersistenceHelper(share_redis=True)`?

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+
+        vumi_config = settings.VUMI_API_CONFIG.copy()
+        vumi_config.update(self.mk_config({}))
```
hodgestar: What about just `vumi_config = self.mk_config(settings.VUMI_API_CONFIG)`?

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        self.patch_settings(VUMI_API_CONFIG=vumi_config)
+
+        has_listeners = lambda: post_save.has_listeners(get_user_model())
+        assert has_listeners(), "User model has no listeners. Aborting."
+        post_save.disconnect(
+            sender=get_user_model(),
+            dispatch_uid='go.base.models.create_user_profile')
+        assert not has_listeners(), (
+            "User model still has listeners. Make sure DjangoVumiApiHelper"
+            " is cleaned up properly.")
```
hodgestar: Are these asserts here to make sure set up and tear down are happening properly? If so, perhaps the error messages should say something like:

```
User model has no post_save listeners. Check that previous tests restored them properly. Aborting.
```

and 

```
User model has unexpected post_save listeners. Check that previous tests removed the stubbed listeners. Aborting.
```

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        self._settings_patches = []
+        self.replace_django_bits()
+
+    def cleanup(self):
+        self._vumi_helper.cleanup()
+        self.restore_django_bits()
+        for patch in reversed(self._settings_patches):
+            patch.disable()
+
+    def replace_django_bits(self):
```
hodgestar, line 85: I would maybe split this function up into more clearly named pieces (e.g. `replace_post_save_hooks`, `replace_settings`). Not a big issue -- just flagging for consideration.

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        post_save.connect(
+            base_models.create_user_profile,
+            sender=get_user_model(),
+            dispatch_uid='go.base.models.create_user_profile')
+
+    @proxyable
+    def get_client(self):
+        client = Client()
+        client.login(username='user@domain.com', password='password')
+        return client
```
hodgestar, line 128: I've often wanted `username` and `password` to be keyword arguments so I can override them to login as different users.

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        patch = override_settings(**kwargs)
+        patch.enable()
+        self._settings_patches.append(patch)
+
+    @proxyable
+    def make_django_user(self):
+        user = get_user_model().objects.create_user(
+            email='user@domain.com', password='password')
+        user.first_name = "Test"
+        user.last_name = "User"
```
hodgestar: Might be useful is some of these were arguments too.

go/base/tests/helpers.py
jerith, line 62: `DjangoVumiApiHelper` is a wrapper around `VumiApiHelper` and therefore needs to have both `is_sync` and `use_riak` properties. However, since it's only useful in Django-based tests, `is_sync` will _always_ be `True` on `DjangoVumiApiHelper`. I can move that to the constructor instead of making it a class attribute if you'd prefer, but it doesn't make sense to have a constructor parameter for a thing that can only ever have one value.

hodgestar, line 62: My vote would be to have True passed to `VumiApiHelper` in the constructor
(with a comment that this is the only allowed value) and then `.is_sync`
and `.use_riak` proxied to the wrapped `VumiApiHelper` (since the intention
is to make available the attributes on the wrapped instance).

go/base/tests/helpers.py
jerith, line 94: Agreed. I'd rather do that in a separate PR, though, because we need to be careful about how we construct and cache FakeRedis instances for the cases where we actually need two different Redis instances (which we don't test yet but probably should).

go/base/tests/helpers.py
jerith, line 128: Agreed. I'd rather do this in a new branch unless there are places in this branch where it can help us.

go/base/tests/helpers.py
jerith, line 62: That works for me.

go/base/tests/helpers.py
hodgestar, line 128: Can we just do the one-line change in this branch and be happy?

go/apps/jsbox/tests/test_message_store.py
```diff
@@ -98,6 +87,7 @@ def test_handle_count_sent_messages(self):
     @inlineCallbacks
     def test_handle_count_inbound_uniques(self):
         reply = yield self.dispatch_command('count_inbound_uniques')
+        print reply
```
hodgestar: Stray print.

go/apps/multi_surveys/tests/test_views.py
```diff
@@ -4,16 +4,16 @@
...
-    TEST_CONVERSATION_TYPE = u'multi_survey'
+class TestMultiSurveyViews(GoDjangoTestCase):
 
     def setUp(self):
-        super(MultiSurveyTestCase, self).setUp()
-        self.patch_settings(
-            VXPOLLS_REDIS_CONFIG=self._persist_config['redis_manager'])
+        self.app_helper = AppViewsHelper(u'multi_survey')
+        self.add_cleanup(self.app_helper.cleanup)
+        self.client = self.app_helper.get_client()
```
hodgestar, line 15: What happened to the patching of `VXPOLLS_REDIS_CONFIG` here?

go/apps/opt_out/tests/test_views.py
```diff
@@ -1,23 +1,26 @@
...
         """
         Test showing the conversation
         """
-        self.setup_conversation()
-        response = self.client.get(self.get_view_url('show'))
-        conversation = response.context[0].get('conversation')
-        self.assertEqual(conversation.name, self.TEST_CONVERSATION_NAME)
+        conv_helper = self.app_helper.create_conversation(name=u"myconv")
+        response = self.client.get(conv_helper.get_view_url('show'))
+        self.assertContains(response, u"<h1>myconv</h1>")
```
hodgestar, line 17: Any particular reason for changing the assert here?

go/apps/opt_out/tests/test_views.py
```diff
@@ -1,23 +1,26 @@
...
         Test showing the conversation
         """
-        self.setup_conversation(started=True)
-        response = self.client.get(self.get_view_url('show'))
-        conversation = response.context[0].get('conversation')
-        self.assertEqual(conversation.name, self.TEST_CONVERSATION_NAME)
+        conv_helper = self.app_helper.create_conversation(
+            name=u"myconv", started=True)
+        response = self.client.get(conv_helper.get_view_url('show'))
+        self.assertContains(response, u"<h1>myconv</h1>")
```
hodgestar, line 26: Same question about changing the assert here.

hodgestar, line 26: There are a bunch of places where this happens but I won't flag the others unless I don't think `ack-grep 'response,.*myconv'` will find them.

go/apps/surveys/tests/test_vumi_app.py
```diff
@@ -118,7 +67,7 @@ def create_survey(self, conversation, questions=None, end_response=None):
         config = yield self.pm.get_config(poll_id)
         config.update({
             'poll_id': poll_id,
-            'transport_name': self.transport_name,
+            # 'transport_name': self.transport_name,
```
hodgestar: Stray commented out line.

go/apps/tests/helpers.py
```diff
@@ -0,0 +1,170 @@
...
+        yield user_helper.user_api.acquire_specific_tag((u"pool", u"tag"))
+        channel = yield user_helper.user_api.get_channel((u"pool", u"tag"))
+        returnValue(channel)
+
+    @proxyable
+    @maybe_async
+    def get_conversation(self, conversation_key):
+        user_helper = yield self.vumi_helper.get_or_create_user()
+        conversation = yield user_helper.get_conversation(conversation_key)
+        returnValue(conversation)
```
hodgestar, line 92: I found it a bit surprising that `create_conversation` calls `self.conversation_wrapper` but `get_conversation` does not.

go/apps/tests/helpers.py
```diff
@@ -0,0 +1,170 @@
...
+from go.vumitools.api import VumiApiCommand, VumiApiEvent
+from go.vumitools.tests.helpers import GoMessageHelper, VumiApiHelper
+
+
+class ApplicationHelper(object):
+    def __init__(self, conversation_type, vumi_helper):
+        self.is_sync = vumi_helper.is_sync
+        self._conversation_type = conversation_type
+        self.vumi_helper = vumi_helper
+        self.conversation_wrapper = None
```
hodgestar: Should this maybe be a parameter? I would set it to `lambda x: x` if the parameter is `None` -- it documents a bit what the signature is supposed to be and saves some checks later.

go/apps/tests/helpers.py
```diff
@@ -0,0 +1,170 @@
...
+        conversation = yield user_helper.get_conversation(conversation_key)
+        returnValue(conversation)
+
+
+class AppWorkerHelper(object):
+    def __init__(self, worker_class, **msg_helper_args):
+        self._worker_class = worker_class
+        msg_helper_kw = {}
+        if msg_helper_args is not None:
+            msg_helper_kw.update(msg_helper_args)
```
hodgestar: I think we can delete these three lines and just do `**msg_helper_args` lower down? `msg_helper_args` can't be `None` and `msg_helper_kw` doesn't contain anything.

go/apps/tests/helpers.py
```diff
@@ -0,0 +1,170 @@
...
+        generate_proxies(self, self.worker_helper)
+        generate_proxies(self, self.dispatch_helper)
+
+    def _worker_name(self):
+        return self._worker_class.worker_name
+
+    def _conversation_type(self):
+        # This is a guess based on worker_name.
+        # We need a better way to do this.
+        return self._worker_name().rpartition('_')[0].decode('utf-8')
```
hodgestar, line 124: Probably we need a TODO for this somewhere if we don't have one already.

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        user_helper = self.make_user(
+            unicode(instance.email), enable_search=False)
+        base_models.UserProfile.objects.create(
+            user=instance, user_account=user_helper.account_key)
+        # We add this to the helper instance rather than subclassing or
+        # wrapping it because we only need the one thing.
+        user_helper.get_django_user = lambda: (
+            get_user_model().objects.get(pk=instance.pk))
```
hodgestar: Maybe `UserApiHelper` should just have this method and raise an error if it's called when `is_sync == True` or if some flag on `user_helper` is `False`? Just worried that figure out how `.get_django_user` works is currently a bit crazy (it's stashed on object that is magically held in a cache and then retrieved via the cache later).

go/base/tests/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @property
+    def amqp_connection(self):
+        # This is a property so that we get the patched version.
+        return base_utils.connection
```
hodgestar: What actually does the patching?

go/base/tests/test_go_import_contacts.py
```diff
@@ -3,38 +3,38 @@
...
-        self.profile = self.user.get_profile()
+        self.vumi_helper = DjangoVumiApiHelper()
+        self.add_cleanup(self.vumi_helper.cleanup)
+        self.vumi_helper.setup_vumi_api()
+        self.user_helper = self.vumi_helper.make_django_user()
+        self.contact_store = self.user_helper.user_api.contact_store
+
+        self.command = go_import_contacts.Command()
+        self.command.stdout = StringIO()
+        self.command.stderr = StringIO()
```
hodgestar, line 19: I think it was better when `command` was just created by `invoke_command` -- fewer attributes to worry about.

hodgestar, line 19: This applies in a few places too.

go/base/tests/test_middleware.py
```diff
@@ -1,24 +1,21 @@
...
+class TestVumiUserApiMiddleware(GoDjangoTestCase):
 
-class VumiUserApiMiddlewareTestCase(VumiGoDjangoTestCase):
     def setUp(self):
-        super(VumiUserApiMiddlewareTestCase, self).setUp()
-        self.setup_api()
-        self.user = self.mk_django_user()
+        self.vumi_helper = DjangoVumiApiHelper()
+        self.add_cleanup(self.vumi_helper.cleanup)
+        self.vumi_helper.setup_vumi_api()
```
hodgestar: I think it's worth cleaning up this

``` python
        self.vumi_helper = DjangoVumiApiHelper()
        self.add_cleanup(self.vumi_helper.cleanup)
        self.vumi_helper.setup_vumi_api()
```

dance because we do it in a lot of places. The setup_vumi_api can maybe be an argument. Maybe we can have an .add_helper method that takes the helper class and params for **init** and returns an instance already registerd for clean-up?

go/base/tests/test_utils.py
```diff
@@ -26,13 +30,18 @@ def test_get_conversation_view_definition_unknown_conv_type(self):
...
+            lambda: set(['old_conv']))
         view_def = get_conversation_view_definition('old_conv')
         self.assertEqual(view_def._conv_def.conversation_type, 'old_conv')
 
 
-class RouterDefinitionHelpersTestCase(VumiGoDjangoTestCase):
+class TestRouterDefinitionHelpers(GoDjangoTestCase):
+    def setUp(self):
+        self.patch_helper = PatchHelper()
+        self.add_cleanup(self.patch_helper.cleanup)
```
hodgestar: Maybe `GoDjangoTestCase` should just have a `self.patch_helper`? Seems generically useful to have around.

go/conversation/tests.py
```diff
@@ -303,8 +301,13 @@ def test_post_new_conversation_extra_endpoints(self):
 
 
 class TestConversationViews(BaseConversationViewTestCase):
+    def setUp(self):
+        super(TestConversationViews, self).setUp()
+        self.msg_helper = GoMessageHelper(
+            mdb=self.vumi_helper.get_vumi_api().mdb)
```
hodgestar: It would also be good to simplify this fishing for `.mdb` a bit if we could. Maybe `GoMessageHelper` could alternatively take a `vumi_api` or a `vumi_helper` as an argument?

go/conversation/tests.py
```diff
@@ -726,75 +735,6 @@ def test_send_one_off_reply(self):
...
-                        [t.name for t in response1.templates])
-        self.assertEqual(response1.context['delay'], 1.1 * 100)
-        # Second time it should still render the messages
-        self.assertTrue('generic/includes/message-list.html' in
-                        [t.name for t in response2.templates])
-        self.assertEqual(response1.context['token'], fake_client.token)
-        # Second time we should list the matching messages
-        self.assertEqual(response2.context['token'], fake_client.token)
-        self.assertEqual(
-            len(response2.context['message_page'].object_list), 10)
```
hodgestar, line 796: I'm a bit -1 on removing these tests because they're a reminder to add back message search and of how message search worked. Perhaps we should just create a detailed ticket though.

go/dashboard/tests/test_dashboard.py
```diff
@@ -78,9 +80,11 @@ def get_config_before_sync(self):
         self.assertRaises(self.dashboard.get_config, DashboardError)
 
 
-class TestDashboardLayout(VumiGoDjangoTestCase):
+class TestDashboardLayout(GoDjangoTestCase):
     def setUp(self):
-        super(TestDashboardLayout, self).setUp()
+        self.vumi_helper = DjangoVumiApiHelper()
+        self.add_cleanup(self.vumi_helper.cleanup)
+        self.vumi_helper.setup_vumi_api()
```
hodgestar: Does this test case use any of these?

go/routers/keyword/tests/test_vumi_app.py
```diff
@@ -1,48 +1,49 @@
...
 
-class TestKeywordRouter(RouterWorkerTestCase):
+class TestKeywordRouter(VumiTestCase):
 
     router_class = KeywordRouter
 
     @inlineCallbacks
     def setUp(self):
-        super(TestKeywordRouter, self).setUp()
+        self.app_helper = RouterWorkerHelper(KeywordRouter)
```
hodgestar: Not sure if naming these `app_helper` is a good idea.

go/routers/tests/helpers.py
```diff
@@ -0,0 +1,198 @@
...
+from go.vumitools.api import VumiApiCommand, VumiApiEvent
+from go.vumitools.tests.helpers import GoMessageHelper, VumiApiHelper
+
+
+class RouterHelper(object):
+    def __init__(self, router_type, vumi_helper):
+        self.is_sync = vumi_helper.is_sync
+        self._router_type = router_type
+        self.vumi_helper = vumi_helper
+        self.router_wrapper = None
```
hodgestar: Same comments about `router_wrapper`, `create_router` and `get_router` as for the ConversationHelper equivalents.

go/routers/tests/helpers.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def cleanup(self):
+        pass
```
hodgestar, line 24: Is this just for future proofing? Are all helpers expected to have `cleanup` methods?

go/routers/tests/helpers.py
```diff
@@ -0,0 +1,198 @@
...
+        # We need versions of these for both inbound and outbound so we can't
+        # generate proxies automagically.
+        # TODO: Build some mechanism for grouping sets of proxies together.
+        self._ri_worker_helper = self.vumi_helper.get_worker_helper(
+            self.ri_connector_name)
+        self._ri_dispatch_helper = MessageDispatchHelper(
+            self.msg_helper, self._ri_worker_helper)
+        # Grab all proxies for these ones. We'll override the RO versions.
+        generate_proxies(self, self._ri_worker_helper)
+        generate_proxies(self, self._ri_dispatch_helper)
```
hodgestar: How about we just write a simple class that groups the necessary things and have one set to `self.ri` and other to `self.ro`?

go/routers/tests/view_helpers.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def get_api_commands_sent(self):
+        return base_utils.connection.get_commands()
+
+
+class RouterViewHelper(object):
+    def __init__(self, router_views_helper, router_key):
+        self.router_key = router_key
+        self.router_type = router_views_helper.router_type
+        self.app_helper = router_views_helper
```
hodgestar: Maybe this should be `router_helper`?

go/token/tests/test_django_token_manager.py
```diff
@@ -4,32 +4,32 @@
...
-    use_riak = False
+class TestDjangoTokenManager(GoDjangoTestCase):
 
     def setUp(self):
-        super(DjangoTokenManagerTestCase, self).setUp()
+        self.vumi_helper = DjangoVumiApiHelper()
+        self.add_cleanup(self.vumi_helper.cleanup)
+        self.vumi_helper.setup_vumi_api()
+        self.user_helper = self.vumi_helper.make_django_user()
+        self.user_pk = self.user_helper.get_django_user().pk
```
hodgestar, line 17: Was there a point to storing `user_pk` rather than `user`?

hodgestar:
Done with initial review of everything not in #786. Some more general questions:
- [x] Perhaps the VumiApiHelper should take a `vumi_api` instance as a parameter? I think would make life easier in a bunch of places.
- [ ] ~~It would be good to have documentation for the helpers because they're quite complicated to use (probably in a separate PR).~~ Created praekelt/vumi-go#824 and praekelt/vumi#678 for this.
- [ ] ~~It would be good to have tests for the helpers.~~ Created praekelt/vumi-go#824 and praekelt/vumi#678 for this.
- [x] I think it would be useful to have a `supports` argument to `create_channel` to make declaring what a channel supports easier.
- [ ] ~~I would be useful to split `user_helper` into things that are actually user related and things that are really campaign related and have a separate `campaign_helper`.~~ Happy to leave this for later.
- [ ] ~~`track_event` and the event helper seem to be connected up in a somewhat convoluted way -- would be good to clean that up a bit.~~ (moving to new list further down)

go/apps/opt_out/tests/test_views.py
jerith, line 17: The new one is better than the old one because it asserts that the conversation name is actually in the response body rather than on an object we're passing around in the template rendering context, especially since it means we no longer have to grovel around in the context at all anymore.

go/apps/opt_out/tests/test_views.py
jerith, line 26: See above.

go/apps/tests/helpers.py
jerith: Indeed. `msg_helper_args` used to be a dict that was passed in rather than **kwargs.

go/apps/tests/helpers.py
jerith, line 92: This bit could definitely use some reworking. The only reason `self.conversation_wrapper` exists is that we want to get a helper object rather than a conversation object in the view tests. I must have been particularly tired when I decided this way of doing that was a good idea. :-/

go/apps/tests/helpers.py
jerith: As I mentioned above, it's actually a really confusing way to do the thing I invented it for. I'd rather get rid of it completely.

go/base/tests/test_go_import_contacts.py
jerith, line 19: We have almost as many different ways of testing management commands as we have management commands. The real solution to this is to refactor all the management commands and put most of their functionality behind an API, but that's out of scope for this branch.

go/conversation/tests.py
jerith: I did it that way initially and then there were a couple of places where I had a bare `mdb` with no `vumi_helper`. I think those places are gone now, though.

go/conversation/tests.py
jerith, line 796: This definitely belongs in a ticket instead of a broken and obsolete test that refers to methods that no longer exist. I spent a couple of minutes trying to get pyflakes/pep8 warnings out of them and then gave up and jut killed them instead.

go/routers/keyword/tests/test_vumi_app.py
jerith: Probably not.

go/dashboard/tests/test_dashboard.py
jerith: Possibly not. I was copying piles of boilerplate around to fix merge conflicts when I did this.

go/routers/tests/helpers.py
jerith, line 24: Partly future proofing, partly reducing the cognitive load of having to remember which helpers have cleanup and which don't. I think it's worth mandating cleanup methods on all helpers, but that needs to happen in vumi first.

go/base/tests/helpers.py
jerith: There was a reason I didn't do it that way, but I think that reason is obsolete.

go/base/tests/helpers.py
jerith: `VumiApiHelper._django_amqp_setup()` does. We only need this for various calls to `.get_metrics()` and `.get_commands()` on the fake connection, so it would probably be better to find alternate ways to get at those things and simplify the fake.

jerith:
> I would be useful to split `user_helper` into things that are actually user related and things that are really campaign related and have a separate `campaign_helper`.

This isn't easy (or particularly useful) until we have a meaningful distinction between users and campaigns. I'd rather defer the split until then.

jerith:
> Perhaps the VumiApiHelper should take a vumi_api instance as a parameter? I think would make life easier in a bunch of places.

I think new `.add_helper()` stuff deals with this issue.

> It would be good to have documentation for the helpers because they're quite complicated to use (probably in a separate PR).

Separate branch, I think. I don't want to delay this for too long.

> It would be good to have tests for the helpers.

Sounds reasonable, but I'd prefer to do it in a new branch. Maybe with the documentation.

> I think it would be useful to have a supports argument to create_channel to make declaring what a channel supports easier.

That involves tagpool metadata which is managed in a different place. Two channels backed by the same tagpool can't have different supported things, for example. I think there are too many edge cases to try do this now.

> I would be useful to split user_helper into things that are actually user related and things that are really campaign related and have a separate campaign_helper.

This isn't easy (or particularly useful) until we have a meaningful distinction between users and campaigns. I'd rather defer the split until then.

> track_event and the event helper seem to be connected up in a somewhat convoluted way -- would be good to clean that up a bit.

Possibly, but I'd rather try solve that problem in a new branch instead of blocking this one.

hodgestar:
I've created praekelt/vumi-go#824 for writing tests and documentation for the Vumi Go helpers and praekelt/vumi#678 for the Vumi ones.

hodgestar:
> > I think it would be useful to have a supports argument to create_channel to make declaring what a channel 
> > supports easier.
> 
> That involves tagpool metadata which is managed in a different place. Two channels backed by the same 
> tagpool can't have different supported things, for example. I think there are too many edge cases to try do this 
> now.

`create_channel` already takes a `metadata` argument and declares tags so I'm not really sure what new possibility for confusion a `supports` argument would add?

hodgestar:
> > track_event and the event helper seem to be connected up in a somewhat convoluted way -- would be good to 
> > clean that up a bit.
> 
> Possibly, but I'd rather try solve that problem in a new branch instead of blocking this one.

Maybe this can be cleaned up a bit by moving `setup_event_dispatcher` into `setup` and passing `name`, `cls` and `config` into the `EventHandlerHelper` constructor?

go/base/tests/helpers.py
hodgestar, line 62: `.use_riak` isn't proxied currently -- it's just set as an attribute.

go/base/tests/helpers.py
hodgestar, line 94: Could you create a suitable issue for this?

go/apps/multi_surveys/tests/test_views.py
hodgestar, line 15: Still question about what happened to the patching of `VXPOLLS_REDIS_CONFIG` here? Note that `go/apps/surveys/tests/test_views.py` still does the patching.

go/base/tests/helpers.py
hodgestar: Resolved.

go/base/tests/helpers.py
hodgestar: Resolved.

go/base/tests/helpers.py
hodgestar, line 85: Resolved.

go/base/tests/helpers.py
hodgestar, line 128: Resolved.

go/base/tests/helpers.py
hodgestar: Resolved.

go/apps/jsbox/tests/test_message_store.py
hodgestar: Resolved.

go/apps/opt_out/tests/test_views.py
hodgestar, line 17: I don't think checking the context passed to the template is any way inferior to checking what's in the rendered text. Probably both are useful. Don't care enough to fight this though.

Resolved.

go/apps/opt_out/tests/test_views.py
hodgestar, line 26: Resolved (see above).

go/apps/surveys/tests/test_vumi_app.py
hodgestar: Resolved.

go/apps/tests/helpers.py
hodgestar: Resolved.

go/base/tests/helpers.py
jerith, line 62: It isn't proxied because it isn't an attribute on `VumiApiHelper`. It's only an attribute on on `DjangoVumiApiHelper` because we need to look at it in another method.

go/apps/tests/helpers.py
hodgestar, line 124: Resolved.

go/base/tests/helpers.py
hodgestar, line 62: Ah, I see.

Resolved.

go/apps/opt_out/tests/test_views.py
jerith, line 17: The context is an implementation detail that isn't obvious at all if you only have a passing familiarity with Django's templating stuff and it doesn't guarantee that the thing you're checking will be rendered at all. Checking in the rendered text is a little messier, but it's asserting on what's actually being shown to the client.

go/apps/multi_surveys/tests/test_views.py
jerith, line 15: It probably got lost accidentally. I'll put it back.

go/base/tests/helpers.py
hodgestar: In that case could we proxy `amqp_connection` from `VumiApiHelper` (and maybe stash the amqp connection created by VumiApiHelper on the helper itself somewhere?).

jerith:
> create_channel already takes a metadata argument and declares tags so I'm not really sure what new possibility for confusion a supports argument would add?

Whoops, I must be misremembering. I'll take another look.

go/base/tests/test_middleware.py
hodgestar: Resolved.

jerith:
:+1: on b32f201

go/apps/opt_out/tests/test_views.py
hodgestar, line 17: The whole code base is an implementation detail. :P

go/base/tests/test_go_import_contacts.py
hodgestar, line 19: Probably we need to clean this up with a command test helper.

Resolved.

go/dashboard/tests/test_dashboard.py
hodgestar: In that case let's remove `self.vumi_helper` here?

go/routers/keyword/tests/test_vumi_app.py
hodgestar: Can we rename it?

go/routers/tests/helpers.py
hodgestar, line 24: Resolved.

go/base/tests/helpers.py
jerith, line 94: praekelt/vumi#679

go/apps/multi_surveys/tests/test_views.py
jerith, line 15: Resolved.

go/dashboard/tests/test_dashboard.py
jerith: Resolved.

go/base/tests/helpers.py
hodgestar, line 94: Thanks.

Resolved.

go/base/tests/helpers.py
jerith: Resolved.

go/routers/keyword/tests/test_vumi_app.py
jerith: Resolved.

go/routers/tests/view_helpers.py
jerith: Resolved.

go/base/tests/helpers.py
jerith: Resolved.

go/token/tests/test_django_token_manager.py
jerith, line 17: We use `user_pk` in a bunch of places and it's the only thing we need from the user object.

go/conversation/tests.py
hodgestar, line 796: I created #825 for this.

Resolved.

go/token/tests/test_django_token_manager.py
hodgestar, line 17: I would have preferred storing `user` and then doing `self.user.pk` since it's no extra typing and it's more generally useful, but happy for this to stay as it it.

Resolved.

hodgestar:
Things remaining:
- [x] Patch helper for base class (see https://github.com/praekelt/vumi-go/pull/783#discussion_r8162802).
- [x] `GoMessageHelper(mdb=self.vumi_helper.get_vumi_api().mdb)` discussion (see https://github.com/praekelt/vumi-go/pull/783#discussion_r8163394).
- [ ] Adding router connection helpers as `self.ri` and `self.ro` to make distinguishing them cleaner (see https://github.com/praekelt/vumi-go/pull/783#discussion_r8165246).
- [ ] ~~Deconvolve event helper (was: `track_event` and the event helper seem to be connected up in a somewhat convoluted way -- would be good to clean that up a bit).~~ (Happy to leave it for another time -- this event stuff needs a big overhaul anyway).

jerith:
Can we do the last two items on that list separately? Both are fairly convoluted and I don't want to hold up everything else for them.

hodgestar:
@jerith Happy to leave event handler helper deconvolving for another time. I pushed a commit that separates `ro` and `ri` helpers. If you're happy with that commit, I'm :+1: on the rest.

jerith:
:+1: to 5b10f4b once Travis is happy.

hodgestar:
:+1: on e4b6138.

PR #781, opened 2013-11-21
AccountRoutingTableDispatcher should reply with a SESSION_CLOSE message for unroutable USSD requests

hodgestar:
Currently if we get an unroutable USSD request we simply timeout, if we see the message is a USSD message then we should send a default closing reply so the connection to the phone is terminated instead of timing out.


go/vumitools/routing.py
```diff
@@ -160,6 +160,10 @@ class AccountRoutingTableDispatcherConfig(RoutingTableDispatcher.CONFIG_CLASS,
         static=True, required=False)
     user_account_key = ConfigText(
         "Key of the user account the message is from.")
+    unroutable_inbound_session_reply = ConfigText(
+        "Text to send on encounter unroutable inbound messages"
+        " that are part of a session.",
```
jerith: Clumsy wording here. Perhaps "Text to send in response to unroutable inbound messages that are part of a session." instead?

jerith:
Travis is unhappy with this and it looks like the failures are related to the changes in this branch.

jerith:
I'm not hugely enthusiastic about a single static response to all the various flavours of sessions we may encounter. Perhaps the copy belongs in tagpool metadata with a fallback configured in the dispatcher?

hodgestar:
Travis is unhappy because the Vumi side only just landed in Vumi develop, I
think?

hodgestar:
I like the suggestion of being able to configure the message
per-tagpool. Will add.

jerith:
Maybe tagpool metadata should also be allowed to override whether the replies are sent at all?

hodgestar:
I'm struggling to think of a great use case for that off the top of my head
but I can see how it's useful to have theoretically. Bit wary of adding too
many knobs to tweak at the start. Would prefer to leave adding it as a
feature later, but happy to add it now if you feel it's really important.

hodgestar:
:+1:

PR #786, opened 2013-11-22
Switch go.vumitools to test helpers

jerith:
This is a smaller chunk of #783 that should be rather more reviewable.


go/vumitools/channel/tests/test_models.py
```diff
@@ -35,16 +36,18 @@ def test_supports_replies(self):
...
 
     @inlineCallbacks
     def setUp(self):
-        super(TestChannelStore, self).setUp()
-        self.manager = self.get_riak_manager()
-        self.account_store = AccountStore(self.manager)
-        self.account = yield self.mk_user(self, u'user')
-        self.channel_store = ChannelStore.from_user_account(self.account)
+        self.vumi_helper = VumiApiHelper(is_sync=self.is_sync)
+        print self.vumi_helper.is_sync
```
hodgestar: Stray print.

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+    @proxyable
     def make_stored_inbound(self, conv, content, **kw):
         msg = self.make_inbound(content, conv=conv, **kw)
         d = self.store_inbound(conv, msg)
-        return d.addCallback(lambda _: msg)
+        if isinstance(d, Deferred):
+            # We're async, return a deferred
+            return d.addCallback(lambda _: msg)
+        else:
+            return msg
```
hodgestar: We should be using `maybe_async` here (and it all the other places).

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+            [(timestamp, value)]))
+        return self.publish_metric_message(metric_msg)
+
+
+class VumiApiHelper(object):
+    # TODO: Clear bucket properties.
+    def __init__(self, is_sync=False, use_riak=True):
+        self.is_sync = is_sync
+        self._patch_helper = PatchHelper()
+        generate_proxies(self, self._patch_helper)
```
hodgestar, line 268: My instinct is that this isn't the correct place to proxy `PatchHelper` -- `VumiApiHelper` should be for manipulating the VumiApi, not patching adhoc things (if it needs its own `PatchHelper`, that's cool).

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+        assert self._vumi_api is not None, "No vumi_api provided."
+        return self._vumi_api
+
+    @proxyable
+    def set_vumi_api(self, vumi_api):
+        assert self._vumi_api is None, "Can't override existing vumi_api."
+        self._vumi_api = vumi_api
+        # TODO: Find a nicer way to give everything the same fake redis.
+        pcfg = self._persistence_helper._config_overrides
+        pcfg['redis_manager']['FAKE_REDIS'] = vumi_api.redis
```
hodgestar, line 324: See comments in #783 about passing in a `vumi_api` as an argument (perhaps `None` by default) and about perhaps making this behaviour the default or an optional flag for the persistence helper.

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+    def publish_metric(self, metric_name, aggregators, value, timestamp=None):
+        metric_msg = MetricMessage()
+        metric_msg.append((metric_name,
+            tuple(sorted(agg.name for agg in aggregators)),
+            [(timestamp, value)]))
+        return self.publish_metric_message(metric_msg)
+
+
+class VumiApiHelper(object):
+    # TODO: Clear bucket properties.
```
hodgestar, line 254: Is this `TODO` a `PersistenceHelper` thing? If so, should it not go there?

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+
+class UserApiHelper(object):
+    def __init__(self, vumi_helper, account_key):
+        self.is_sync = vumi_helper.is_sync
+        self._vumi_helper = vumi_helper
+        self.account_key = account_key
+        self.user_api = vumi_helper.get_vumi_api().get_user_api(account_key)
+
+        # Easier access to these stores is useful.
+        self.contact_store = self.user_api.contact_store
```
hodgestar, line 395: Any reason not to add the conversation and router stores here too?

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+        yield permission.save()
+        account = yield self.get_user_account()
+        account.tagpools.add(permission)
+        yield account.save()
+
+    @proxyable
+    @maybe_async
+    def add_app_permission(self, application):
+        account_store = self.user_api.api.account_store
+        permission = account_store.application_permissions(
```
hodgestar, line 423: This probably needs a `TODO` to add this to the API too (like `add_tagpool_permission` above).

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+        return self.vumi_helper.cleanup()
+
+    @inlineCallbacks
+    def setup_event_dispatcher(self, name, cls, config):
+        app_config = self.vumi_helper.mk_config({
+            'event_handlers': {
+                name: "%s.%s" % (cls.__module__, cls.__name__),
+            },
+            name: config,
+            'transport_name': 'sphex',  # TODO: Kill this when possible.
```
hodgestar: Could this comment be update to explain roughly what has to happen before it's possible?

go/vumitools/tests/helpers.py
```diff
@@ -1,72 +1,509 @@
...
+
+        self.event_dispatcher = yield self.worker_helper.get_worker(
+            EventDispatcher, app_config)
+        self.vumi_helper.set_vumi_api(self.event_dispatcher.vumi_api)
+
+        self.user_helper = yield self.vumi_helper.make_user(u'acct')
+        yield self.vumi_helper.setup_tagpool(u"pool", [u"tag1", u"tag2"])
+        yield self.user_helper.add_tagpool_permission(u"pool")
+        self.conversation = yield self.user_helper.create_conversation(
+            u'bulk_message')
```
hodgestar, line 504: Is there a reason to create all these things in EventHandlerHelper? We seem to be steering away from magically creating lots of things in the other helpers (which I like).

go/vumitools/tests/test_api.py
```diff
@@ -348,20 +347,16 @@ class TestVumiUserApi(TestTxVumiUserApi):
...
-            self.vumi_api = VumiApi.from_config_sync(
-                self.mk_config({}), FakeAmqpConnection(self._amqp))
-        else:
-            self.vumi_api = yield VumiApi.from_config_async(
-                self.mk_config({}), get_fake_amq_client(self._amqp))
-        self.user_account = yield self.mk_user(self.vumi_api, u'Buster')
-        self.user_api = VumiUserApi(self.vumi_api, self.user_account.key)
+        self.vumi_helper = VumiApiHelper(is_sync=self.is_sync)
+        self.add_cleanup(self.vumi_helper.cleanup)
+        yield self.vumi_helper.setup_vumi_api()
```
hodgestar: Same comment as in #783 -- I'd like to clean up this `vumi_helper.setup_vumi_api()` dance since we do it a lot.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -141,59 +141,68 @@ def test_create_transaction_error(self):
...
-        self.dispatcher = yield self.get_dispatcher()
-        self.vumi_api = self.dispatcher.vumi_api
+        self.vumi_helper = VumiApiHelper()
+        self.add_cleanup(self.vumi_helper.cleanup)
+        yield self.vumi_helper.setup_vumi_api()
+        self.msg_helper = GoMessageHelper()
+        self.ri_helper = self.vumi_helper.get_worker_helper(
+            "billing_dispatcher_ri")
+        self.ro_helper = self.vumi_helper.get_worker_helper(
+            "billing_dispatcher_ro")
```
hodgestar, line 153: This makes it look like `get_worker_helper` is really `get_connection_helper`.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -141,59 +141,68 @@ def test_create_transaction_error(self):
...
         config = {
-            "receive_inbound_connectors": [
-                "billing_dispatcher_ri"
-            ],
-            "receive_outbound_connectors": [
-                "billing_dispatcher_ro"
-            ],
+            "receive_inbound_connectors": ["billing_dispatcher_ri"],
+            "receive_outbound_connectors": ["billing_dispatcher_ro"],
             "api_url": "http://127.0.0.1:9090/",
```
hodgestar, line 160: What is `api_url` for?

hodgestar, line 160: Oh, nm. This is specifically the billing dispatcher.

go/vumitools/tests/test_contact.py
```diff
@@ -4,29 +4,30 @@
...
-        self.store_alt = ContactStore.from_user_account(self.account_alt)
-        yield self.store.contacts.enable_search()
-        yield self.store_alt.contacts.enable_search()
+        self.vumi_helper = VumiApiHelper()
+        self.add_cleanup(self.vumi_helper.cleanup)
+        yield self.vumi_helper.setup_vumi_api()
+
+        self.user_helper = yield self.vumi_helper.make_user(u'user')
+        user_account = yield self.user_helper.get_user_account()
+        self.store = ContactStore.from_user_account(user_account)
```
hodgestar, line 24: Doesn't `user_helper` already have a user_api with the contact store on it?

go/vumitools/tests/helpers.py
jerith, line 268: We need our own patch helper here. Since we have one already, it kind of makes sense to use it in other places as well instead of creating a whole pile of different patchers all with their own cleanup that happens in different places.

go/vumitools/tests/test_metrics_worker.py
```diff
@@ -49,45 +53,20 @@ def dummy_bucket_for_conversation(self, conv_key):
...
-        self.assertEqual(bucket, hash(conversations["conv1"].key) % 60)
+        user_helper = yield self.vumi_helper.make_user(u'acc1')
+        conv1 = yield self.make_conv(user_helper, u'conv1')
+
+        bucket = worker.bucket_for_conversation(conv1.key)
+        self.assertEqual(bucket, hash(conv1.key) % 60)
 
     def assert_conversations_bucketed(self, worker, expected):
-        expected = copy.deepcopy(expected)
+        expected = expected.copy()
```
hodgestar, line 67: Any idea why this was changed?

go/vumitools/tests/test_routing.py
```diff
@@ -247,51 +266,13 @@ def setup_routing_table_dispatcher_test(self):
...
-                "router": "router_ri",
-            },
-            "opt_out_connector": "optout",
-        }
-        config.update(config_extras)
-        dispatcher = yield self.get_worker(
-            self.mk_config(config), AccountRoutingTableDispatcher)
-        returnValue(dispatcher)
 
     def with_md(self, msg, user_account=None, conv=None, router=None,
```
hodgestar, line 268: Maybe this should be renamed to `add_md` too if we're that keen on renaming it?

go/vumitools/tests/helpers.py
jerith, line 324: There are only three places outside this class that call `set_vumi_api()` and none of them are in this branch. Can you point out a thing that would be easier if `VumiApiHelper` took a `vumi_api` parameter?

(I think I originally had the parameter you suggested, but removed it because nothing was actually passing it in.)

go/vumitools/tests/helpers.py
jerith, line 254: It's mostly a `PersistenceHelper` thing, but the `TODO` is here because the only way we can find the buckets that need their properties removed is by looking at the accounts we created.

go/vumitools/tests/test_routing.py
```diff
@@ -814,28 +851,32 @@ def test_event_routing_to_app1(self):
 
 class TestUnroutableSessionResponse(RoutingTableDispatcherTestCase):
 
-    @inlineCallbacks
-    def setUp(self):
-        yield super(TestUnroutableSessionResponse, self).setUp()
-        yield self.setup_routing_table_dispatcher_test()
+    def get_routing_table(self):
+        return RoutingTable({})
```
hodgestar, line 852: Just flagging this method because it seems odd and I'm not entirely sure what it's here for.

go/vumitools/tests/helpers.py
jerith, line 395: It's probably better to add higher-level helper functions, but I'll take a look around and see if adding the other stores will be useful.

go/vumitools/tests/helpers.py
hodgestar, line 268: I don't see that as adding a lot of overhead. Would be cool to have a PatchHelper available on the base test class.

go/vumitools/tests/helpers.py
jerith: I don't actually know what needs to happen there. I vaguely remember deciding it shouldn't be necessary but discovering it was hard to remove.

hodgestar: Can we perhaps figure it out before this lands?

go/vumitools/tests/helpers.py
hodgestar, line 324: One is in this branch (in `EventHandlerHelper`).  Mostly it just all seems a bit clumsy somehow and I'd like us to try fix it.

go/vumitools/tests/helpers.py
hodgestar, line 254: Ah. Can we expand the TODO to explain that?

go/vumitools/tests/helpers.py
hodgestar, line 395: Cool. Tx.

hodgestar:
First pass of review done. One general comment is that I'm a bit worried that all the proxying has left us a bit in the same place that inheritance did -- i.e. with big objects whose methods come from all sorts of places. Would like to fix that somehow but not entirely sure how.

go/vumitools/tests/helpers.py
jerith, line 504: `EventHandlerHelper` is a bit of a strange beast. It's really a specialised collection of helpers (like `ApplicationHelper` and friends in #783), but it's used in the tests for `go.vumitools.subscription` as well and therefore can't really live in `go.apps.tests.helpers`.

The magic setup stuff is all the common boilerplate extracted from the various event handler tests that use it because it's mostly irrelevant to the thing being tested but still needs to be there for stuff to work.

go/vumitools/tests/test_api.py
jerith: I couldn't find a clean way to do it that didn't build a bunch of helper-specific behaviour into the various test classes.

I suppose some kind of `.get_helper(helper_class, *args, **kw)` method that constructed the helper and added its cleanup would be a good things to have. That drops two lines of boilerplate down to one (or three to two in this case) at the cost of making sure all helper classes have a `.cleanup()` method which is fairly sensible anyway.

I'll have to think about ways to get rid of the `.setup_vumi_api()` call. That's tricky because (a) it's potentially async and (b) it's not always desired.

go/vumitools/tests/test_billing_worker.py
jerith, line 153: By method count, `WorkerHelper` is almost entirely `ConnectionHelper` (and it was originally called `AmqpHelper`). However, a critical part of its functionality is `.get_worker()` and you suggested I rename the class back when it first appeared in vumi. I couldn't find a good way to split the two bits, but I think they're kind of worth splitting.

go/vumitools/tests/test_contact.py
jerith, line 24: Yes, but we're specifically testing `ContactStore` here and we want to build the store ourselves in case the helper does something nefarious behind our backs, like replacing it with a test double. (That's a thing I'd really like to do once this stuff has all landed and we've done the "put all sorts of things behind an API" work we've been talking about.)

go/vumitools/tests/test_metrics_worker.py
jerith, line 67: The `deepcopy()` isn't necessary because we're not modifying anything past the top of the dict. Also, it was exploding all over my face when it tried to deepcopy conversation objects full of ~~stars~~ riak models.

The `deepcopy()` probably isn't necessary for the next line either, now that I've looked at the code again.

jerith:
> One general comment is that I'm a bit worried that all the proxying has left us a bit in the same place that inheritance did -- i.e. with big objects whose methods come from all sorts of places.

Short of either reimplementing all sorts of things in all sorts of places or `violating.the.law.of.Demeter.in.very.long.lines()`, there's not much we can do about that.

This does leave us in a _much_ better position than inheritance did, though, because we can use all the lower-level helpers independently without dragging along piles of unnecessary baggage.

go/vumitools/tests/helpers.py
jerith, line 268: Twisted-based tests already have one. `GoDjangoTestCase` (which isn't in this branch) initially had its own but then I found I was almost never using it without already having a `VumiApiHelper` handy anyway, so I dropped it. I can look at putting it back, but I'd rather not do that unless it's actually used in a bunch of places.

go/vumitools/tests/helpers.py
jerith, line 324: I'm open to suggestions. I rewrote this stuff four or five times before settling on what I ended up with here because everything else I could think of was worse. :-/

go/vumitools/tests/test_routing.py
jerith, line 852: It overrides `RoutingTableDispatcherTestCase.get_routing_table()` which is used in `.setUp()`. All the other tests need a full routing tables, this one needs an empty one.

go/vumitools/tests/test_api.py
jerith: See praekelt/vumi#668 for this.

go/vumitools/tests/helpers.py
jerith, line 254: Done, but all the changes were below this line so github didn't obsolete this set of comments.

go/vumitools/tests/helpers.py
jerith, line 395: Turns out this is only actually used in one place (and that's only in #783). It's probably worth tracking down all the places we go through `user_helper.user_api` to get at these stores and either add the stores here or add higher-level helpers. I'd rather do that in a new branch than spend half a day running around this one, though.

go/vumitools/tests/test_metrics_worker.py
hodgestar, line 67: Can we just not change this for now?

go/vumitools/tests/test_routing.py
hodgestar, line 852: Cool.

go/vumitools/tests/test_metrics_worker.py
jerith, line 67: Only if you want four tests to fail with:

```
Traceback (most recent call last):
  File "/Users/jerith/.virtualenvs/vumi-go/lib/python2.7/site-packages/twisted/internet/defer.py", line 1070, in _inlineCallbacks
    result = g.send(result)
  File "/Users/jerith/code/vumi-go/go/vumitools/tests/test_metrics_worker.py", line 249, in test_metrics_loop_func_bucket_nonzero
    3: [conv3],
  File "/Users/jerith/code/vumi-go/go/vumitools/tests/test_metrics_worker.py", line 67, in assert_conversations_bucketed
    expected = copy.deepcopy(expected)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 163, in deepcopy
    y = copier(x, memo)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 257, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)

  ...

  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 163, in deepcopy
    y = copier(x, memo)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 257, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/jerith/.virtualenvs/vumi-go/lib/python2.7/site-packages/_pytest/main.py", line 273, in __hash__
    return hash((self.name, self.parent))
AttributeError: 'UnitTestCase' object has no attribute 'name'
```

jerith, line 67: Also, I was misremembering. It wasn't riak models, it was test infrastructure.

go/vumitools/tests/test_routing.py
jerith, line 268: It was renamed in the billing tests because that was just adding the metadata to a message. This also returns the message with the metadata added, so I think `with_md()` is a more appropriate name here.

go/vumitools/tests/test_metrics_worker.py
hodgestar, line 67: Tests on my machine fail somewhat differently:

```
  File "/usr/lib/python2.7/copy.py", line 190, in deepcopy
    y = _reconstruct(x, rv, 1, memo)
  File "/usr/lib/python2.7/copy.py", line 329, in _reconstruct
    y = callable(*args)
  File "/home/simon/projects/repos-praekelt/vumi-go/ve/local/lib/python2.7/site-packages/twisted/persisted/styles.py", line 75, in unpickleModule
    return __import__(name,{},{},'x')
ImportError: No module named amqp80
```

The issue seems to be that the conversations in expected were made into ConversationWrappers by this branch which means they have all sorts of crazy attributes hidden on them.

hodgestar:
:+1: -- we can discuss the `set_vumi_api` issue in #783.

PR #766, opened 2013-11-22
Dashboards for displaying conversation metrics

justinvdm:


justinvdm:
We still need to do a release for diamondash and include it as a bower dependency in Go, and there may still be some tests to fix, but otherwise, ready for review.

justinvdm:
Actually ready for review now.

justinvdm:
Based on RL discussion, did some name changes from 'conversation dashboard' to to 'conversation reports'. Ready for review for real this time.

go/conversation/view_definition.py
```diff
@@ -616,6 +617,103 @@ def put(self, request, conversation):
...
+        """
+        logger.error(e)
+
+    def get(self, request, conversation):
+        dashboard = None
+
+        try:
+            # build the dashboard
+            layout = self.build_layout(conversation)
+            dashboard = Dashboard(conversation.key, conversation.name, layout)
```
hodgestar: Maybe `Dashboard` should be called `ConversationDashboard` since it seems to be specific to a dashboard generated from a conversation?

justinvdm: `Dashboard`s are re-usable, its the layout that is specific to a certain thing, like conversations. `build_layout` should return a `ConversationReportsLayout`, so that's where the converstation specific stuff goes.

hodgestar: Oh, I see that `conversation.key` is the name and `conversation.name` is the title. I wonder if we should namespace the name and bit and maybe change the title to something like `Report for %s`?

justinvdm: As it turns out we don't actually use the dashboard title from inside Go, since the client side doesn't render the title. I think it'd be best to remove `title` from `Dashboard`.

As for the name, what kind of namespacing do you have in mind?

hodgestar: Maybe something like `conversation.<conversation-id>`?

justinvdm: Ah ok, good idea, will do.

go/dashboard/dashboard.py
```diff
@@ -0,0 +1,156 @@
...
+        Ensures the dashboard exists on diamondash's side
+        """
+        try:
+            raw_config = self._raw_serialize()
+            self.config = self.diamondash_api.replace_dashboard(raw_config)
+        except Exception, e:
+            raise DashboardSyncError("Dashboard sync failed: %s" % e)
+
+    def serialize(self):
+        return self.config
```
hodgestar, line 103: Wondering whether this should raise an error if `self.config` is still `None`? And whether maybe it should be named something else? What is returned by `replace_dashboard`?

justinvdm, line 103: The parsed dashboard config is returned by `replace_dashboard`. Raising an error if it is `None` is a good idea, will do.

go/dashboard/tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+        raise requests.exceptions.HTTPError(self.content)
+
+
+class TestDashboardApiClient(VumiGoDjangoTestCase):
+    def test_raw_request(self):
+        resp = FakeResponse('spam', 201)
+
+        def stubbed_request(method, url, data):
+            self.assertEqual(method, 'put')
+            self.assertEqual(url, 'http://localhost:7115/api/foo')
```
hodgestar: There is a `.patch_settings` method (and now a `.patch_config` method) on the VumiGoDjangoTestCase class to patch settings if we want to explicitly set the Diamondash URL in the tests.

justinvdm: Will do.

go/dashboard/tests/test_dashboard.py
```diff
@@ -0,0 +1,302 @@
...
+                    'visited': True,
+                    'spam': {
+                            # pep8 or pyflakes somehow isn't happy with this
+                            # being indented properly
+                            'visited': True,
+                            'lerp': 'larp',
+                    },
+                }]
+            }
+        }])
```
hodgestar, line 168: I think these should be separate tests (with `traverse` maybe a helper function on the test case).

go/dashboard/tests/test_client.py
```diff
@@ -0,0 +1,101 @@
...
+        self.monkey_patch(requests, 'request', lambda *a, **kw: resp)
+
+        client = DiamondashApiClient()
+        self.assertRaises(
+            DiamondashApiError,
+            client.raw_request,
+            'put', 'foo', 'bar')
+
+    def test_request(self):
+        self.patch_settings(DIAMONDASH_API_URL='http://diamondash.moc/api')
```
hodgestar: I was thinking this should just be done in `.setUp()` since a lot of tests can potentially use it and it doesn't really cost any time.

go/dashboard/tests/test_dashboard.py
```diff
@@ -0,0 +1,304 @@
...
+                'name': 'foo',
+            }),
+            "go.campaigns.%s.conversations.%s.foo.avg" %
+            (self.conv.user_account.key, self.conv.key))
+
+    def test_conversation_metric_handling_for_missing_fields(self):
+        self.layout.handle_metric({
+            'metric_type': 'conversation',
+            'name': 'foo',
+        })
```
hodgestar: Can the three lines above just be removed?

go/dashboard/tests/test_dashboard.py
```diff
@@ -0,0 +1,304 @@
...
+            }),
+            "go.campaigns.%s.stores.red.foo.avg" %
+            (self.conv.user_account.key))
+
+    def test_account_metric_handling_for_missing_fields(self):
+        self.layout.handle_metric({
+            'metric_type': 'account',
+            'store': 'red',
+            'name': 'foo',
+        })
```
hodgestar: I think these can be removed too since they test the same thing as `test_account_metric_handling`?

go/dashboard/tests/test_dashboard.py
justinvdm: Over here I'm trying to assert that `handle_metric` doesn't complain with the correct parameters. I suppose it can be removed.

go/dashboard/views.py
```diff
@@ -0,0 +1,26 @@
...
+    proxying solution should be used in production (eg. haproxy).
+    """
+    api = client.get_diamondash_api()
+    _, url = request.path.split('/diamondash/api', 1)
+    response = api.raw_request(request.method, url, content=request.body)
+
+    return HttpResponse(
+        response['content'],
+        status=response['code'],
+        content_type='application/json')
```
hodgestar, line 29: Do we need a TODO here to check that the metrics requested are allowed for the given account?

go/dashboard/tests/test_dashboard.py
justinvdm: Ah, as you said in the comment below, these are already tested. Will remove.

go/dashboard/views.py
justinvdm, line 29: Good idea, will do.

go/vumitools/metrics.py
```diff
@@ -121,6 +130,44 @@ def make_name(cls, conv, metric_name):
...
 
+class MetricSet(object):
+    def __init__(self, metrics=None):
+        self.metrics = []
+        self.metrics_by_name = {}
+        self.extend(metrics or [])
+
+    def __iter__(self):
+        for metric in self.metrics:
+            yield metric
```
hodgestar: I think this could also be just `return iter(self.metrics)`?

hodgestar:
:+1: once I've figured out why our production metrics are `last` metrics.

justinvdm:
Thanks for reviews :)

hodgestar:
:+1:

PR #787, opened 2013-11-24
Avoid importing celery task in survey conversation definition

jerith:
This just moves the import to method-scope so the module can be imported in vumi-land for metrics stuff.


smn:
:+1: when Travis is happy

hodgestar:
:+1:

PR #790, opened 2013-11-25
Styling overrides for the diamondash dashboards

justinvdm:
Once #766 lands, we will have diamondash dashboards inside Go. These dashboards will be using diamondash's styles, which clearly don't fit with Go's styling. We need to make these dashboards' styles more consistent with the rest of Go.


hodgestar:
:+1:

PR #795, opened 2013-11-27
Add authentication for diamondash api client requests

justinvdm:


go/dashboard/client.py
```diff
@@ -17,8 +17,17 @@ def make_api_url(self, path):
             p.strip('/')
             for p in [settings.DIAMONDASH_API_URL, path])
 
+    def get_api_auth(self):
+        return (
+            settings.DIAMONDASH_API_USERNAME,
+            settings.DIAMONDASH_API_PASSWORD)
```
hodgestar: Maybe we should return `None` if this is not set? That way AUTH is optional.

hodgestar:
Other than the suggestion to make auth optional, +1.

go/dashboard/client.py
justinvdm: Good idea

justinvdm:
Ready for review.

justinvdm:
Ready for another review.

hodgestar:
:+1:

PR #791, opened 2013-11-27
Allow sandbox apps to configure their dashboards from their config

justinvdm:
Since each sandbox app chooses the metrics it fires, allowing the app's conversation dashboard to be configured using the sandbox config will allow each app to choose which metrics to display and how to display them.

### DIFF
--- a/go/apps/jsbox/view_definition.py
+++ b/go/apps/jsbox/view_definition.py
@@ -1,9 +1,31 @@
+from go.dashboard.dashboard import ConversationReportsLayout
 from go.conversation.view_definition import (
-    ConversationViewDefinitionBase, ConversationTemplateView,
-    EditConversationView)
+    ConversationViewDefinitionBase, ConversationReportsView,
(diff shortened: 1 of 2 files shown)


go/apps/jsbox/view_definition.py
```diff
@@ -46,4 +68,5 @@ class ConversationViewDefinition(ConversationViewDefinitionBase):
 
     extra_views = (
         JSBoxLogsView,
+        JSBoxReportsView,
```
justinvdm, line 71: This relies on the fact that `extra_views` are added to the conversation view definition after the default views, so it 'overwrites' the default reports view since they both have the same path suffix. This feels like a bad idea, ideas on an alternative?

justinvdm:
Ready for review.

go/apps/jsbox/view_definition.py
hodgestar, line 71: Maybe we need to document this as a feature?

hodgestar:
:+1:

PR #802, opened 2013-12-03
Cache get_wrapped_conversation() calls in HTTP API

smn:


smn:
@jerith's much better work obsoletes this. Closing.

PR #803, opened 2013-12-03
Cache conversation object in helper metadata

jerith:
We currently load the conversation at least twice for each message processed (since we're using the message storing middleware) and we only ever read it there.


hodgestar:
:+1:

PR #807, opened 2013-12-04
API call metrics for Wikipedia app

jerith:
This builds on top of praekelt/vumi-wikipedia#26.


smn:
:+1: once Travis confirms all is ok

hodgestar:
:+1: since this is breaking other builds but lets sort this out today.

PR #809, opened 2013-12-05
search on contacts

smn:
please add the ability to search on contact details and/or contact extras stored in the database.

use case:
A persons details are recorded and details saved, including their id number.  A user must be able to retrieve that persons details using their id number as search criteria.


go/apps/jsbox/contacts.py
```diff
@@ -419,6 +419,59 @@ def handle_save(self, api, command):
...
+        """
+        try:
+            contact_store = self._contact_store_for_api(api)
+            keys = yield contact_store.contacts.raw_search(
+                command['query']).get_keys()
+            contacts = []
+            for contact_bunch in contact_store.contacts.load_all_bunches(keys):
+                contacts.extend((yield contact_bunch))
+
+        except (SandboxError,) as e:
```
vgeddes, line 532: Single letter variable names such as `e` make it hard to search for variables. I would suggest changing it to `ex`. If you agree, I can change this for you since you're very busy with the http_api.

jerith, line 532: Most of our exception handlers use `e` as their variable. The scope is sharply limited, so I think it's okay here.

go/apps/jsbox/contacts.py
```diff
@@ -419,6 +419,59 @@ def handle_save(self, api, command):
...
+                },
+                function(reply) { api.log_info(reply.contacts); });
+        """
+        try:
+            contact_store = self._contact_store_for_api(api)
+            keys = yield contact_store.contacts.raw_search(
+                command['query']).get_keys()
+            contacts = []
+            for contact_bunch in contact_store.contacts.load_all_bunches(keys):
+                contacts.extend((yield contact_bunch))
```
jerith: This could be problematic if the search returns a million contacts. We should probably limit the number of results we return, similar to what we do in contacts views.

jerith:
Reviewed. Looks good apart from the search result limit.

hodgestar:
I would limit the `SandboxError` to around the specific things that can
throw `SandboxError`.

go/apps/jsbox/contacts.py
vgeddes: Good catch. I just checked, and for Botle Buhle, there at least 23000 contacts, which is probably large enough to cause slight performance issues.

smn: I'll add pagination

smn: actually I'm not going to get to this @vgeddes since you're working with @ulrichenslin on this stuff, can you add pagination? Probably store keys in redis with an expire of a few minutes so they're cached.

go/apps/jsbox/contacts.py
```diff
@@ -1,10 +1,15 @@
 # -*- test-case-name: go.apps.jsbox.tests.test_contacts -*-
 # -*- coding: utf-8 -*-
 
+import hashlib
+import uuid
+import json
+
```
hodgestar: These are all unused now by the look of things?

go/apps/jsbox/contacts.py
```diff
@@ -19,6 +24,14 @@ class ContactsResource(SandboxResource):
...
     """
 
+    @inlineCallbacks
+    def setup(self):
+        redis_config = self.config.get('redis_manager', {})
+        self.redis = yield TxRedisManager.from_config(redis_config)
+
+    def teardown(self):
+        return self.redis.close_manager()
+
```
hodgestar: Doesn't look like we use Redis any more?

go/apps/jsbox/contacts.py
```diff
@@ -419,6 +432,113 @@ def handle_save(self, api, command):
...
+                     query: 'name:"My Name"',
+                },
+                function(reply) { api.log_info(reply.contacts); });
+
+        """
+        try:
+            contact_store = self._contact_store_for_api(api)
+            keys = yield contact_store.contacts.raw_search(
+                command['query']).get_keys()
+            keys = keys[:self.config['max_search_result_keys']]
```
hodgestar: This doesn't appear to be defined anywhere and I think `max_keys` can just be a parameter to this command with a default (say `10`?).

smn: I'd still prefer to put a config enforced limit in place. So we can guarantee how many contact keys at most will be returned. If the command specifies fewer than that then we can return a subset.

hodgestar: I don't particularly mind if there is one, but we're getting all the keys
into memory anyway and the sandbox will be killed anyway if it tries to
read too many keys.

go/apps/jsbox/contacts.py
```diff
@@ -419,6 +419,109 @@ def handle_save(self, api, command):
...
+
+            contact_store = self._contact_store_for_api(api)
+            contact = (yield contact_store.get_contact_by_key(key)).get_data()
+
+        except (KeyError, ContactNotFoundError, SandboxError,) as e:
+            log.warning(str(e))
+            returnValue(self.reply(command, success=False, reason=unicode(e)))
+        except (Exception,) as e:
+            log.warning(str(e))
+            returnValue(self.reply(command, success=False, reason=unicode(e)))
```
jerith, line 467: We don't need both of these `except` blocks here.

go/apps/jsbox/contacts.py
vgeddes: That check is mainly there so we can inform users in the documentation that there is a hard limit to the number of results that can be returned.

I'm not sure there is a useful case for letting the user supply a limit on the number of results they want. It would certainly be useful if we had paging, but we've just removed that.

go/apps/jsbox/contacts.py
```diff
@@ -419,6 +419,109 @@ def handle_save(self, api, command):
...
+            api.request(
+                'contacts.get_by_key', {
+                     key: '391cea45-ae51-441c-b972-5de765c7a0dc',
+                },
+                function(reply) { api.log_info(reply.contact); });
+
+        """
+        try:
+            if 'key' not in command:
+                raise KeyError("Expected 'key' field in request")
```
jerith: We should probably reply with the appropriate failure response instead of raising an exception here.

go/apps/jsbox/contacts.py
```diff
@@ -419,6 +419,109 @@ def handle_save(self, api, command):
...
+                'contacts.search', {
+                     query: 'name:"My Name"',
+                },
+                function(reply) { api.log_info(reply.contacts); });
+
+        """
+        try:
+            contact_store = self._contact_store_for_api(api)
+            keys = yield contact_store.contacts.raw_search(
+                command['query']).get_keys()
```
jerith, line 528: We should check for the `query` key in the command and return an appropriate failure response if it isn't there.

go/apps/jsbox/contacts.py
vgeddes: There's an exception handler for `KeyError` which makes sure an appropriate response is returned.

go/apps/jsbox/contacts.py
jerith: I'm with @hodgestar. We already have all the keys, so there isn't much point in not passing them along. (We need to handle the case where Riak gives us a billion keys closer to the source, and that probably needs paging support.)

go/apps/jsbox/contacts.py
jerith: That exception handler also logs a warning, which will turn into Sentry noise. I'd rather we just told the client they did it wrong and bailed out.

jerith:
We need tests for API calls with missing parameters as well.

go/apps/jsbox/contacts.py
vgeddes: If we pass a million keys to the sandbox, it will also blow up before it has a chance to do logging or any other cleanups. I'm not sure if the logging and contact resource share the same worker process, but if they do, then that's an issue as they should not share the same fate on most failures.

hodgestar: Note the Vumi worker won't blow up -- it'll just terminate the running
sandbox.

vgeddes: ah, cool, good to know.

go/apps/jsbox/tests/test_contacts.py
```diff
@@ -489,6 +489,112 @@ def test_handle_save_for_unicode_chars(self):
...
+                msisdn=u'+27831234567',
+                groups=[u'group-a', u'group-b'])
+            keys.add(contact.key)
+
+        # subset
+        reply = yield self.dispatch_command('search',
+                                            query=u'surname:Jack*',
+                                            max_keys=3)
+        self.assertTrue(reply['success'])
+        self.assertTrue(set(reply['keys']).issubset(keys))
```
smn, line 546: Not a major issue but just curious why you went this route instead of 

``` python
self.assertEqual(len(reply['keys']), 3)
```

smn:
I'm sure there's a way to do the `max_keys` limiting with less code but let's leave that, this has gone on for long enough, :+1: from me.

go/apps/jsbox/contacts.py
```diff
@@ -419,6 +419,133 @@ def handle_save(self, api, command):
...
+                    command,
+                    success=False,
+                    reason=u"Expected 'query' field in request"))
+            max_keys = None
+            if 'max_keys' in command:
+                try:
+                    value = int(command['max_keys'])
+                    if value >= 0:
+                        max_keys = value
+                except ValueError:
```
hodgestar: You also need to catch `TypeError` here.

hodgestar: And there's technically no reason to allow strings -- ints are passed through in JSON fine.

hodgestar: This also won't send an error if `value < 0` -- not sure if that was the intention.

go/apps/jsbox/tests/test_contacts.py
hodgestar, line 546: The subset check is cool (i.e. that the right keys are returned) but a length check is important too to catch out-by-one errors.

go/apps/jsbox/tests/test_contacts.py
```diff
@@ -489,6 +489,112 @@ def test_handle_save_for_unicode_chars(self):
...
+        reply = yield self.dispatch_command('search',
+                                            query=u'surname:Jack*',
+                                            max_keys="Haha!")
+        self.assertFalse(reply['success'])
+        self.assertTrue(
+            "Value for parameter 'max_keys' is invalid" in reply['reason']
+        )
+
+
+
```
hodgestar: Looks like there are too many blank lines here.

smn:
:+1: from me, @hodgestar the last commit addresses your comments, can you also have a look?

hodgestar:
:+1:

PR #808, opened 2013-12-05
POST-only HTTP API app

jerith:
We need a way to run multiple HTTP API workers concurrently, and a new app/conversation type is the quickest way to make that happen without breaking streaming clients.

### DIFF
--- a/go/apps/http_api_nostream/auth.py
+++ b/go/apps/http_api_nostream/auth.py
@@ -1,0 +1,42 @@
+from zope.interface import implements
+
+from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.web import resource
+from twisted.cred import portal, checkers, credentials, error
+
+
+class ConversationRealm(object):
+    implements(portal.IRealm)
+
+    def __init__(self, resource):
+        self.resource = resource
+
+    def requestAvatar(self, user, mind, *interfaces):
+        if resource.IResource in interfaces:
+            return (resource.IResource, self.resource, lambda: None)
+        raise NotImplementedError()
+
+
+class ConversationAccessChecker(object):
+    implements(checkers.ICredentialsChecker)
+    credentialInterfaces = (credentials.IUsernamePassword,)
+
+    def __init__(self, vumi_api, conversation_key):
+        self.vumi_api = vumi_api
+        self.conversation_key = conversation_key
+
+    @inlineCallbacks
+    def requestAvatarId(self, credentials):
+        username = credentials.username
+        token = credentials.password
+        user_exists = yield self.vumi_api.user_exists(username)
+        if user_exists:
+            user_api = self.vumi_api.get_user_api(username)
+            conversation = yield user_api.get_wrapped_conversation(
+                self.conversation_key)
+            if conversation is not None:
+                tokens = conversation.config.get('http_api_nostream', {}).get(
+                    'api_tokens', [])
+                if token in tokens:
+                    returnValue(username)
+        raise error.UnauthorizedLogin()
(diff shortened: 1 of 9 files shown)


go/apps/http_api_nostream/resource.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def get_conversation_tag(self, conversation):
+        return (conversation.delivery_tag_pool, conversation.delivery_tag)
+
+    @inlineCallbacks
+    def handle_PUT(self, request):
+        try:
+            payload = json.loads(request.content.read())
+        except ValueError:
+            self.client_error_response(request, 'Invalid Message')
```
vgeddes, line 163: I would include make the reason `"Invalid Message: %s " % str(e)`. I can make the necessary changes if you want.

go/apps/http_api_nostream/resource.py
```diff
@@ -0,0 +1,357 @@
...
+            'metrics.json': MetricResource,
+        }
+        resource_class = class_map.get(path)
+
+        if resource_class is None:
+            returnValue(resource.NoResource())
+
+        user_id = request.getUser()
+        config = yield self.get_worker_config(user_id)
+        if (yield self.is_allowed(config, user_id)):
```
vgeddes, line 321: I would swap the order of the branches in this `if` statement for better readability.

go/apps/http_api_nostream/resource.py
```diff
@@ -0,0 +1,357 @@
...
+
+            yield self.track_request(user_id)
+            returnValue(resource_class(self.worker, self.conversation_key))
+        returnValue(resource.ErrorPage(http.FORBIDDEN, 'Forbidden',
+                                       'Too many concurrent connections'))
+
+
+class AuthorizedResource(resource.Resource):
+
+    resource_class = ConversationResource
```
vgeddes, line 335: A minor thing, but I would rather make `AuthorizedResource` parameterizable. By this I mean passing in the resource_class as an argument to the constructor. I've seen similar hardcoded AuthorizedResource resources around the Go codebase, which makes it harder to make this class reusable in future.

vgeddes, line 335: Also, this would make the intent of `AuthorizedResource` clearer in `NoStreamingHTTPWorker`.

go/apps/http_api_nostream/resource.py
jerith, line 163: This was copied verbatim from `http_api`. I'd rather leave it as-is for now to minimise the possibility of breaking things.

go/apps/http_api_nostream/resource.py
jerith, line 321: Agreed, but this was copied verbatim from `http_api` and I'd rather leave it as-is for now to minimise the possibility of breaking things.

go/apps/http_api_nostream/resource.py
jerith, line 335: Agreed, but this was copied verbatim from `http_api` and I'd rather leave it as-is for now to minimise the possibility of breaking things.

go/apps/http_api_nostream/auth.py
```diff
@@ -0,0 +1,42 @@
...
+    def requestAvatarId(self, credentials):
+        username = credentials.username
+        token = credentials.password
+        user_exists = yield self.vumi_api.user_exists(username)
+        if user_exists:
+            user_api = self.vumi_api.get_user_api(username)
+            conversation = yield user_api.get_wrapped_conversation(
+                self.conversation_key)
+            if conversation is not None:
+                tokens = conversation.config.get('http_api_nostream', {}).get(
```
smn, line 38: Can we not pass this in in `__init__` and save the copy pasting?

auth py_vs _auth py

smn, line 38: Or even just reference the conversation objects' `conversation_type`?

smn:
:+1:, will start on drying this up a lot in #811.

PR #814, opened 2013-12-06
The NoStreamingHTTPWorker should validate conversation configuration fields.

jerith:
It seems the the worker choked on a conversation that had empty message_push, event_push urls. 

Here's the log of what happened:
https://gist.github.com/vgeddes/1bc8f87f2063fdb9205b

Until we fix this, we're just going to make sure that conversations are well-defined when used with this transport.

### DIFF
--- a/go/apps/http_api_nostream/vumi_app.py
+++ b/go/apps/http_api_nostream/vumi_app.py
@@ -73,6 +73,9 @@ class NoStreamingHTTPWorker(GoApplicationWorker):
 
         push_message_url = self.get_api_config(conversation,
                                                'push_message_url')
+        if push_message_url is None:
+            log.warning("URL not configured: push_message_url")
+            return
         yield self.push(push_message_url, message)
 
     @inlineCallbacks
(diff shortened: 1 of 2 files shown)


hodgestar:
I'd have preferred something that made use of `definition.is_config_valid`? Admittedly that needs a little bit of implementing too.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -73,6 +73,9 @@ def consume_user_message(self, message):
 
         push_message_url = self.get_api_config(conversation,
                                                'push_message_url')
+        if push_message_url is None:
+            log.warning("URL not configured: push_message_url")
+            return
```
hodgestar, line 80: Maybe add something on which conversation it was for somehow? I think this counts as an error (certainly it will as soon as we validate the conversation config).

hodgestar:
:+1: on landing but see other comments.

jerith:
> I'd have preferred something that made use of `definition.is_config_valid`?

Nothing uses that yet and we'd need to consider the UI and state change implications as part of the work. I'll look for an existing ticket and create one if necessary.

go/apps/http_api_nostream/vumi_app.py
jerith, line 80: Good point. I'll do that.

hodgestar:
:+1: after finding / creating the is_config_valid ticket.

jerith:
See #815 for solving the broader config validation problem.

PR #805, opened 2013-12-09
Ability to migrate conversations and smartgroups from 1 account to another

smn:
For MAMA, we have a need migrate conversations and smartgroups from a QA account to the live one, to avoid having to recreate all conversations manually.

### DIFF
--- a/go/base/management/commands/go_manage_conversation.py
+++ b/go/base/management/commands/go_manage_conversation.py
@@ -24,6 +24,8 @@ class Command(BaseGoAccountCommand):
         make_option('--conversation-key',
                     dest='conversation_key',
                     help='The conversation key'),
+        make_option(
+            '--file', dest='file', help='The file to import')
     )
 
     def get_conversation(self, options):
@@ -87,4 +89,24 @@ class Command(BaseGoAccountCommand):
 
     def handle_command_export(self, *apps, **options):
         conversation = self.get_conversation(options)
+        self.stdout.write(json.dumps(conversation.get_data()))
+
+    def load_file(self, options):
+        if 'file' not in options:
+            raise CommandError('Please specify a file to load.')
+        file_name = options['file']
+        return open(file_name, 'r')
+
+    def handle_command_import(self, *apps, **options):
+        raw_conv_data = json.load(self.load_file(options))
+        allowed_keys = [
+            'conversation_type',
+            'description',
+            'name',
+            'config',
+            ]
+
+        new_conv_data = dict([(key, raw_conv_data[key]) for key in
+                              allowed_keys])
+        conversation = self.user_api.new_conversation(**new_conv_data)
         self.stdout.write(json.dumps(conversation.get_data()))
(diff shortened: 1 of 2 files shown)


smn:
@miltontony @jerith ready for review

miltontony:
@smn Does this also include smart groups?

smn:
@miltontony no it doesn't, only conversations.

smn:
Ah I'm seeing the smartgroups in the ticket header now. How many smart groups are there and is that worth the effort?

miltontony:
Sorry, didn't get any notifications. Not sure why..
We have 12 smart groups to copy over. It's not too big a deal though.
:+1:

smn:
thanks @miltontony, @jerith @hodgestar thoughts on this?

hodgestar:
:+1: from me but I'd like @jerith to look too since he has opinions about conversation copying.

One architectural comment -- it might be worthwhile making `export_conversation` and `import_conversation` utility methods to facilitate conversation copying elsewhere in Vumi Go (e.g. from the UI).

I might also make sense to sanity check the conversation type -- we don't want new conversations being created that have old / invalid conversation types.

go/base/management/commands/go_manage_conversation.py
```diff
@@ -82,3 +86,27 @@ def handle_command_archive(self, *apps, **options):
...
+        raw_conv_data = json.load(self.load_file(options))
+        allowed_keys = [
+            'conversation_type',
+            'description',
+            'name',
+            'config',
+            ]
+
+        new_conv_data = dict([(key, raw_conv_data[key]) for key in
+                              allowed_keys])
```
jerith, line 111: I'd prefer to throw an error here if we have unexpected keys rather than silently discarding them.

jerith:
> One architectural comment -- it might be worthwhile making `export_conversation` and `import_conversation` utility methods to facilitate conversation copying elsewhere in Vumi Go (e.g. from the UI).

This kind of thing belongs in an internal API layer, which we've discussed a bit in the past. Ideally, almost all our management commands would be reduced to thin wrappers around an external interface to this API layer.

I'm inclined to leave this as-is for now and do it systematically in the (fairly near) future.

jerith:
:+1:, although I'd like to see the comments about conversation type (from @hodgestar) and unexpected keys addressed either here or soon after this lands.

smn:
@jerith so you'd want us to manually edit the JSON files before importing them? I don't see how else you'd avoid getting unexpected keys and it raising errors at every import.

smn:
Added check for invalid `conversation_type`s

go/base/management/commands/go_manage_conversation.py
hodgestar, line 111: We have to discard the keys because most of the conversation fields are not
imported (e.g. `key`, `created_at`, etc). We could add another layer of
checking and throw an error if the field is not in the list of fields on
the Conversation model class (or one of the special fields like $VERSION).

jerith, line 111: Whoops, I missed that. Sorry.

jerith:
> manually edit the JSON files before importing them? I don't see how else you'd avoid getting unexpected keys and it raising errors at every import.

I missed the fact that the export contains a bunch of keys that aren't imported. Silently ignoring them makes much more sense now.

PR #817, opened 2013-12-10
Ensure that unmapped CSV fields are imported into the 'extras' dict.

vgeddes:
**Still in progress**

Currently, Go's CSV importer has a bug where if you don't select a mapping for a CSV field, it doesn't get put into the 'extras' dict for the imported contacts.

The cleanest solution would to add a mapping type "Other" or "Extra" so that the user must explicitly decide whether to import a certain field that we can't auto-detect a mapping for. 

Here's the actual cause of the bug, where we basically end up nilling out the field name for columns which the user doesn't map.
https://github.com/praekelt/vumi-go/blob/develop/go/contacts/templates/contacts/includes/tools.html#L132

### DIFF
--- a/go/contacts/parsers/base.py
+++ b/go/contacts/parsers/base.py
@@ -1,5 +1,5 @@
 import os.path
-from collections import NamedTuple
+from collections import namedtuple
 
 from django.core.files.storage import default_storage
 
@@ -7,13 +7,13 @@ from vumi.utils import load_class, normalize_msisdn
 
 
 # Information required to import a specific CSV column
-# `field_name`         - The original header field in the CSV or XSL data
-# `contact_field_name` - the field to map to in the Contact dict
-# `normalizer`         - The name of the normalizer function to apply
-FieldInfo = NamedTuple('FieldInfo',
+# field              - The original field name in the tabular data
+# contact_field      - the field to map to in the Contact dict
+# normalizer         - The name of the normalizer function to apply
+FieldInfo = namedtuple('FieldInfo',
                        [
-                           'field_name'
-                           'contact_field_name',
+                           'field',
+                           'contact_field',
                            'normalizer'
                        ])
 
@@ -231,16 +231,15 @@ class ContactFileParser(object):
         SETTABLE_ATTRIBUTES list, which defaults to the DEFAULT_HEADERS keys.
         """
 
-        # get the actual field_names which were previously detected.
-        field_names = [field.field_name for field in fields]
-
-        # build a dictionary of field_names to FieldInfo objects
-        field_map = dict([(field.field_name, field) for field in fields])
+        # build a dictionary of field names to FieldInfo objects
+        field_map = dict([(info.field, info) for info in fields])
 
         # We're expecting a generator so loop over it and save as contacts
         # in the contact_store, normalizing anything we need to
         data_dictionaries = self.read_data_from_file(
-            file_path, field_names, has_header
+            file_path,
+            [info.field for info in fields],
+            has_header
         )
         for data_dictionary in data_dictionaries:
 
@@ -248,9 +247,9 @@ class ContactFileParser(object):
             # contact to be saved
             contact_dictionary = {}
             for key, value in data_dictionary.items():
-                if field_map[key].contact_field is None:
-                    continue
                 contact_field = field_map[key].contact_field
+                if contact_field is None:
+                    continue
                 normalizer = field_map[key].normalizer
                 value = self.normalizer.normalize(normalizer, value)
 
@@ -268,7 +267,7 @@ class ContactFileParser(object):
                 # programmer error, rather than runtime errors.
                 assert contact_field in self.SETTABLE_ATTRIBUTES
 
-                if contact_field:
+                if contact_field not in ('extra',):
                     contact_dictionary[contact_field] = value
                 else:
                     extra = contact_dictionary.setdefault('extra', {})
--- a/go/contacts/parsers/base.py
+++ b/go/contacts/parsers/base.py
@@ -264,7 +264,10 @@ class ContactFileParser(object):
                 if value is None or value == '':
                     continue
 
+                # This is a static-style assertion to guard against
+                # programmer error, rather than runtime errors.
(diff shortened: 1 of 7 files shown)


smn:
There's a `leave as is` mapping that does this I think?

vgeddes:
It doesn't. The intent of "Leave as is" is to refrain from performing any kind of normalization whatsoever. Looking through the code and UI, I think "Leave as is" is redundant, given that there is already a 'Plain Text" normalization option.

Personally, I was confused when I tried to upload a CSV and had to choose between "Leave as is" or "Plain text", since "Plain text" is generally what other importers use by default.

vgeddes:
When I was getting started on this PR, I had some trouble understanding the workflow of the importer, so I scratched an itch and added a namedtuple `go.contacts.parser.base.FieldInfo`, which describes compactly what actions to take on a specific CSV/XLS column.

go/contacts/parsers/base.py
```diff
@@ -1,10 +1,23 @@
...
+# Information required to import a specific CSV column
+# field              - The original field name in the tabular data
+# contact_field      - the field to map to in the Contact dict
+# normalizer         - The name of the normalizer function to apply
+FieldInfo = namedtuple('FieldInfo',
+                       [
+                           'field',
+                           'contact_field',
+                           'normalizer'
+                       ])
```
hodgestar: I'd rather this was just an ordinary class since that allows us to add a docstring instead of a comment.

hodgestar: If we're going this route I think the list of fields should be a list of `FieldInfo` objects -- no point in constructing that later.

hodgestar: And `normalizer` should probably just be the normalizer function object itself.

go/contacts/parsers/base.py
```diff
@@ -209,28 +223,38 @@ def read_data_from_file(self, file_path, field_names, has_header):
...
-        # and decoded as JSON
-        field_names = [field[0] for field in fields]
-        field_map = dict(fields)
+
+        # build a dictionary of field names to FieldInfo objects
+        field_map = dict([(info.field, info) for info in fields])
+
         # We're expecting a generator so loop over it and save as contacts
         # in the contact_store, normalizing anything we need to
+        print has_header
```
hodgestar: Stray print.

go/contacts/parsers/base.py
```diff
@@ -209,28 +223,38 @@ def read_data_from_file(self, file_path, field_names, has_header):
...
+            has_header
+        )
         for data_dictionary in data_dictionaries:
 
             # Populate this with whatever we'll be sending to the
             # contact to be saved
             contact_dictionary = {}
             for key, value in data_dictionary.items():
-                value = self.normalizer.normalize(field_map[key], value)
+                print "%s %s\n" % (key, value)
```
hodgestar: Another stray print.

go/contacts/parsers/base.py
```diff
@@ -209,28 +223,38 @@ def read_data_from_file(self, file_path, field_names, has_header):
...
+                contact_field = field_map[key].contact_field
+                if contact_field is None:
+                    continue
```
hodgestar, line 332: When does a `FieldInfo` have a `None` `contact_field`? What if `key` is not in `field_map`?

go/vumitools/metrics.py
```diff
@@ -23,7 +23,10 @@ def get_full_name(self):
         This is for constructing the full, prefixed metric name in
         *Django land* if a manager is not available.
         """
-        return settings.GO_METRICS_PREFIX + self.get_name()
+
+        # FIXME: I had to do this to get the tests to run (haha!). I'm not
+        #        sure how this bit of code is related to my changes.
+        return settings.GO_METRICS_PREFIX + (self.get_name() or 'haha')
```
hodgestar, line 29: Not sure why you needed this -- maybe because you were running without `go.testsettings`? Tests run okay on other machines (e.g. Travis). Anyway, stuff like this should ideally never even reach PRs.

hodgestar:
Reviewed. Lots of minor things to clean up.

As mentioned in discussion with @smn on IRC last night, I think it was possible to get things into `extras` but only if the .csv file had a header row (I could be wrong here -- I just read the code carefully). Note that it makes no sense to put anything into extras without a header row because there's no other way to specify what to call it.

I'm in favour of making everything explicit and in particular:
- remove `Leave as is` from the normalizers in favour of `Plain text`.
- make it more clear which fields will end up in extras (e.g. by showing this in the UI somehow)
- have a way of explicitly ignoring fields.

This can probably be achieved with much simpler changes than are currently present in this PR, i.e. removing the `Leave as is` normalizer option and adding something to the UI to indicate which header fields will end up in extras.

I do like the idea of creating a class to hold field information though (we can stash information about whether this is a default or an extra field on this).

vgeddes:
The CSV file which motivated this PR actually did have headers. In the current deployed Go code, there is a single test case for putting things into `extra` using a headerless CSV (1). It is slightly incorrect in the sense that the column specification it POSTs isn't what actual users will be able to specify. For example `columns["column-2"]` must be a member of `parser.base.SETTABLE_ATTRIBUTES`, and `"integer"` is certainly not one of those. So the test case actually ends up not really testing CSVs with or without headers.

I'm gonna sort out all the issues you brought up. thanks for the review!
1. https://github.com/praekelt/vumi-go/blob/develop/go/contacts/tests.py#L445

smn:
I've tested this with [this sample CSV](https://gist.github.com/smn/95ac1a04624a22fd98e2) and it works fine.

screenshot_2013_12_11__4_13_pm
screenshot_2013_12_11__4_13_pm
screenshot_2013_12_11__4_14_pm

The problem is that the CSV file didn't have a header row with column values that trigger [the header parsing](https://github.com/praekelt/vumi-go/blob/develop/go/contacts/parsers/base.py#L159) in Go's importer

hodgestar:
@vgeddes Thoughts on the current status of this?

hodgestar:
Closing because this works as is and we'd probably need an entirely different PR to improve things.

PR #777, opened 2013-12-10
Cannot add a decimal when defining messaging costs

andreit:
When adding a new messaging cost: http://qa-vumi.za.prk-host.net/admin/billing/messagecost/add/

I am not able to enter a decimal. For e.g. Vumi's per message storage cost is .001 which I currently cannot add.

screen shot 2013-11-18 at 11 34 52 am


andreit:
@hodgestar How should I handle a decimal quantization that results in `Decimal('0')`? Let the exception propagate through and log it?

hodgestar:
That sounds good. Ideally we should try pick this up when costs are updated
via the admin _and_ complain loudly (i.e. have the exception raised and
logged) while calculating costs for transactions.

andreit:
Cool, I will add validation to the MessageCost form in the admin.

go/billing/api.py
```diff
@@ -44,9 +44,14 @@ def _render_to_json(self, result, request, *args, **kwargs):
...
-            request.setHeader('Content-Type', 'application/json')
-            request.write(json.dumps(result, cls=JSONEncoder))
+            try:
+                data = json.dumps(result, cls=JSONEncoder)
+                request.setResponseCode(200)  # OK
+                request.setHeader('Content-Type', 'application/json')
+                request.write(data)
+            except JSONEncoder.EncodeError:
+                log.err()
+                request.setResponseCode(500)  # Internal Server Error
```
hodgestar: Can't we just let Twisted's higher-level stuff log this error?

go/billing/api.py
```diff
@@ -60,8 +65,11 @@ def _parse_json(self, request):
         """
         content_type = request.getHeader('Content-Type')
         if request.method == 'POST' and content_type == 'application/json':
-            return json.loads(request.content.read(), cls=JSONDecoder)
-
+            try:
+                content = request.content.read()
+                return json.loads(content, cls=JSONDecoder)
+            except JSONDecoder.DecodeError:
+                log.err()
```
hodgestar: Same comment here.

go/billing/forms.py
```diff
@@ -1,10 +1,39 @@
...
+from go.billing.models import Account, MessageCost, Transaction
+
+
+class MessageCostForm(ModelForm):
+
+    class Meta:
+        model = MessageCost
+
+    def clean(self):
+        """Make sure the resulting credit cost does not result in 0"""
```
hodgestar: I would write this as "does not underflow to zero".

go/billing/migrations/0006_auto__chg_field_messagecost_message_cost__chg_field_transaction_credit.py
```diff
@@ -0,0 +1,135 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+
+        # Changing field 'MessageCost.message_cost'
+        db.alter_column(u'billing_messagecost', 'message_cost', self.gf('django.db.models.fields.DecimalField')(max_digits=10, decimal_places=3))
```
hodgestar, line 13: How come this one is going to `3` places while the others are going to `6`? (Note suggesting this should change, just checking it).

go/billing/utils.py
```diff
@@ -17,9 +17,21 @@
...
-            return float(obj.quantize(decimal.Decimal('.01')))
+            context = decimal.Context()
+            value = obj.quantize(decimal.Decimal('.000001'),
+                                 context=context)
+
+            if (context.flags[decimal.Inexact]
+                    and value == decimal.Decimal('0.0')):
+                raise self.EncodeError("Decimal quantization resulted in 0")
+
+            return float(value)
```
hodgestar, line 35: I think we should probably make this quantization and value check a class method on the model that the field comes from (or at least a utility method in the models) because it's closely tied to the precision declared on the model fields (and we're putting '.000001' a bit all over the place).

Obviously the error raising still needs to happen here.

go/billing/tests/test_api.py
```diff
@@ -232,11 +228,17 @@ def runTest(self):
...
         self.assertTrue('credit_amount' in base_cost)
-        credit_factor = float(app_settings.CREDIT_CONVERSION_FACTOR)
-        credit_amount = round((90 + (90 * 20.0 / 100.0)) * credit_factor)
-        self.assertEqual(base_cost.get('credit_amount'), int(credit_amount))
+        credit_factor = decimal.Decimal(app_settings.CREDIT_CONVERSION_FACTOR)
+        message_cost = decimal.Decimal('0.9')
+        markup_percent = decimal.Decimal('20.0')
+        credit_amount = (message_cost
+                         + (message_cost * markup_percent
+                            / decimal.Decimal('100.0'))) * credit_factor
```
hodgestar: This calculation should probably also be a function somewhere so that we don't have to repeat it everywhere.

go/billing/api.py
```diff
@@ -604,9 +610,9 @@ def get_cost(self, account_number, tag_pool_name, message_direction):
         query = """
             SELECT t.account_number, t.tag_pool_name, t.message_direction,
                    t.message_cost, t.markup_percent,
-                   CAST((t.message_cost +
-                         (t.message_cost * t.markup_percent / 100.0))
-                         * %(credit_factor)s AS INT) AS credit_amount
+                   (t.message_cost +
+                    (t.message_cost * t.markup_percent / 100.0))
+                   * %(credit_factor)s AS credit_amount
```
hodgestar: I wondering whether it really makes sense to have the database do these calculations or whether it would be better (for consistency) to have them all happen in Python using the function I suggested (in another comment) we write?

hodgestar:
First review done.

go/billing/api.py
andreit: Do you mean `log.msg`?

go/billing/forms.py
andreit: Cool, will update.

go/billing/migrations/0006_auto__chg_field_messagecost_message_cost__chg_field_transaction_credit.py
andreit, line 13: This is the message cost in cents. Used 3 decimal places as per Ben's request. The credit amounts are at 6 because when we multiply the cost in cents by the credit conversion factor we need more precision for the result.

go/billing/utils.py
andreit, line 35: I was thinking of declaring the quantization exponent in `go.billing.settings`. I like this idea of adding a class method to `go.billing.models.MessageCost`.

go/billing/api.py
hodgestar: No, I meant not catching the exception at all here and just letting
Twisted's web resource stuff handle the error.

andreit:
@hodgestar Made the suggested changes. Let me know if there is anything else.

hodgestar:
Looks good. :+1:

PR #806, opened 2013-12-11
Create Airtime Vouchers and Unique Codes pages

andreit:


andreit:
@hodgestar Could you please give me some feedback on the overall solution for this? Thanks.

go/vouchers/forms.py
```diff
@@ -0,0 +1,306 @@
...
+
+class BaseVoucherPoolForm(forms.ModelForm):
+
+    EXCEL_CONTENT_TYPES = (
+        'application/vnd.ms-excel',
+        'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+    )
+
+    CSV_CONTENT_TYPES = (
+        'text/csv'
```
hodgestar: I think this is missing a comma.

go/vouchers/forms.py
```diff
@@ -0,0 +1,306 @@
...
+from go.vouchers import settings
+from go.vouchers.models import VoucherPool
+from go.vouchers.services import AirtimeVoucherService
+
+
+class BaseVoucherPoolForm(forms.ModelForm):
+
+    EXCEL_CONTENT_TYPES = (
+        'application/vnd.ms-excel',
+        'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
```
hodgestar: Trailing comma would be good here too (although less vital than the one for CSV_CONTENT_TYPES).

go/vouchers/forms.py
```diff
@@ -0,0 +1,306 @@
...
+
+        return pool_name
+
+    def _make_ext_pool_name(self, pool_name):
+        """Return a unique name for the pool to be used in the external
+        voucher service.
+        """
+        pool_name = pool_name.strip().lower()
+        return "user_%d_%s" % (self.user.id,
+                               re.sub(r'\W+', '_', pool_name))
```
hodgestar: I think we should follow a similar naming pattern to that used by metrics, e.g. `go.campaigns.<account-id>.vouchers.<pool-name>`.

hodgestar: We've traditionally made the `go.` part a prefix that can be specified in settings, e.g. `go.settings.GO_VOUCHER_POOL_PREFIX`.

go/vouchers/forms.py
```diff
@@ -0,0 +1,306 @@
...
+            self._import_excel_file(voucher_pool)
+
+
+class AirtimeVoucherPoolForm(BaseVoucherPoolForm):
+    """A form used to create a new *Airtime* `VoucherPool`"""
+
+    VOUCHERS_FILE_LABEL = _("Select a file from your hard drive containing "
+                            "the airtime vouchers")
+
+    VOUCHERS_FILE_HELP_TEXT = """
```
hodgestar: Maybe this should be called `_HELP_HTML` instead of `_HELP_TEXT`?

hodgestar: Since we're making this translatable (which is awesome, btw) is there a good way to translate this?

go/vouchers/forms.py
```diff
@@ -0,0 +1,306 @@
...
+            csvfile = StringIO.StringIO(vouchers_file.read())
+            try:
+                reader = csv.reader(csvfile)
+                headings = reader.next()
+            finally:
+                csvfile.close()
+        if (len(headings) != len(settings.AIRTIME_VOUCHER_FILE_FORMAT)
+                or headings[0] != settings.AIRTIME_VOUCHER_FILE_FORMAT[0]
+                or headings[1] != settings.AIRTIME_VOUCHER_FILE_FORMAT[1]
+                or headings[2] != settings.AIRTIME_VOUCHER_FILE_FORMAT[2]):
```
hodgestar: I think this could just be something like:

```
if headings != settings.AIRTIME_VOUCHER_FILE_FORMAT:
    ...
```

hodgestar: Possibly with `list(...)` around one or both sides.

hodgestar:
@andreit: I've gone through the branch now. Generally looking good. There are some fairly big structural changes I'd like made in order to fit in with our future plans to make these services more "component like". Broadly speaking the plan is to have the services look a bit like conversation and router typse currently do -- i.e. a Riak model for data and a view_definition object describing the UI. So, the plan now would be to:
- [ ] Split the unique codes and airtime services into separate packages under `go.services.unique_codes` and `go.services.airtime`. Each package should have a `model.py` (with a Riak model), and a `view_definition.py` with the view defintion.
- [ ] A global `views.py` and `urls.py` in `go.services` can dispatch to the individual view definitions based on service type (see `go.conversations.{views,urls}` for what this looks like).
- [ ] Convert the Django airtime model into a Riak model. You can look at `go.vumitools.conversation.models` for what this should look like.

This will result in a two separate services, each with its own view definition and models. If you need to share utilities between the two, feel free to put them in a `go.services.voucher_utils` package for now.

I know these are quite big changes -- shout if you'd like to discuss things over Skype or Hangout.

go/vouchers/forms.py
andreit: The names are used as prefixes for the table names in the voucher services. I am not sure if dots are allowed in the table names and whether there would be any length restrictions?

go/vouchers/forms.py
andreit: I was thinking we set `VOUCHERS_FILE_HELP_TEXT` to a list of translatable strings, then have a utility function convert it into the HTML assigned to the field's `help_text` property?

hodgestar: After reading up and thinking a bit, I think best practice is just to wrap the whole text block in `_()` and let the translator figure it out -- they can add formatting and emphasis as appropriate to their language.

go/vouchers/forms.py
hodgestar: Hmm. That's a bit annoying. Let's go with something like `go_<acc-key>` then. Having to use Django model user IDs will be frustrating from many parts of vumi-go.

andreit:
@hodgestar Refactored as suggested, please check it out when you have some time... i am starting to feel like this:
post-10513-code-refactoring-cat-in-bathtu-yrzt

go/services/airtime/forms.py
```diff
@@ -0,0 +1,138 @@
...
+            sheets = book.sheets()
+            if not sheets:
+                raise forms.ValidationError(
+                    "The file is empty.")
+
+            sheet = sheets[0]
+            if sheet.ncols != len(service_settings.FILE_FORMAT):
+                raise forms.ValidationError("Invalid file format.")
+
+            headings = [sheet.cell(0, 0).value, sheet.cell(0, 1).value,
```
hodgestar: This needs to be something like:

```
headings = [sheet.cell(0, i).value for i in range(sheet.ncols)]
```

otherwise adding or removing items from `settings.AIRTIME_VOUCHER_FILE_FORMAT` will break the check. That will probably allow you to combine these two checks into one.

I would also improve the error messages, e.g. `"Invalid file format. Headers should be %r" % service_settings.FILE_FORMAT`.

go/base/models.py
```diff
@@ -69,6 +69,16 @@ def get_short_name(self):
...
 
+    def get_account(self):
+        """Return the user's account or `None` if the user has no accounts.
+
+        If the user has multiple accounts, return the first one.
+        """
+        try:
+            return self.account_set.all()[0]
+        except IndexError:
+            return None
```
hodgestar: What is `self.account_set` / where does it come from?

hodgestar:
I've asked @jerith to take a look on Monday and give feedback since he also has ideas about how the service framework should look.

go/base/models.py
andreit: It comes from the `go.billing.Account` model. Added this when I was looking for a way to get the logged in user's account number but realised I can get it from the `request.user_api`. So this can be removed.

jerith:
I think the voucher upload handling should probably live in a background task rather the Django form, similar to how we handle contact imports in `go.contacts.views` and `go.contacts.tasks` (although hopefully with less messiness).

jerith:
There's some general reorganisation that I think would improve things:
- `VoucherService` and `BaseVoucherService` probably belong in a new `go.services.airtime.service` module instead of various `utils` modules.
- The form stuff in `go.services.voucher_utils` probably belongs in `go.services.airtime.forms` instead, at least for now.
- Templates specific to particular services should live in the service packages instead of `go.services.templates`. I don't know how this interacts with Django's template-finding stuff, but we do something similar for `go.apps.dialogue`.
- If we're going to share a bunch of code between unique code and airtime voucher things, I'd rather do one or more of the following:
  - Make it generic enough for a wider range of services to use.
  - If that's not possible, I'd prefer to put both services in the same package (perhaps `go.services.vouchers.airtime` and `go.services.vouchers.unique_codes` or something).
  - If a chunk of the code is just a client library for aludel-based things, it would actually make more sense to move that to aludel and/or the two service repositories.

This branch seems to be heading in the right direction, but it's the first prototype of a kind of thing we'll be using for a lot more features in the future so it's likely to need more restructuring than most PRs. Sorry. :-/

andreit:
Thanks for the feedback. I will make the suggested changes.

andreit:
@jerith Did some restructuring based on your input. Let me know if this is what you had in mind. Also moved the template tags inside the service packages, as in `go.services.vouchers.airtime.templatetags.py`, and have them registered in `go.services.vouchers.templatetags.services_tags.py`. Let me know your thoughts. Thanks.

jerith:
Thanks, @andreit. The new structure looks good for now, but we may need to move things around again later when the code's a little more complete and I have a better idea of how all the bits fit together. That's likely to happen in a new branch, though.

I've noticed some failing tests in the Travis builds that need to be fixed. Also, some tests for the new code in this branch would also be a good idea.

jerith:
I've merged develop into this branch, because quite a lot has happened since the last merge. In particular, a lot of the testing infrastructure has been improved.

andreit:
@jerith Thanks for the feedback. I am busy with the tests now. Will implement the Unique Codes service in a similar way then we can see if we can clean things up even more.

andreit:
@jerith @hodgestar Hey guys, sorry for the huge pull request. Could you please do the first round of reviews. Thanks

jerith:
Nothing's happened here for many months, and this code is probably obsolete. If we need this sort of thing, we can reopen or start again.

PR #776, opened 2013-12-11
Add dropdown w/ transport names to add tag pools

andreit:
When adding a tag pool to admin: http://qa-vumi.za.prk-host.net/admin/billing/tagpool/add/

Instead of presenting a text field to the user where anything can be inputted, replace the text field with a dropdown populated with the transport names (transport_name) from tagpools.yaml: https://github.com/praekelt/puppet/blob/develop/modules/vumigo/files/tagpools.yaml

screen shot 2013-11-18 at 12 00 12 pm

### DIFF
--- a/go/billing/admin.py
+++ b/go/billing/admin.py
@@ -8,12 +8,15 @@ from django.contrib import admin
 from django.contrib import messages
 
 from go.billing.models import TagPool, Account, MessageCost, Transaction
-from go.billing.forms import CreditLoadForm, BaseCreditLoadFormSet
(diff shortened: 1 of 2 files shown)


andreit:
@hodgestar Is it safe to access the `tagpools.yml` file like so: `os.path.join(settings.PROJECT_ROOT, '..', 'setup_env', 'tagpools.yml')`

hodgestar:
Nope. You want to go via the TagpoolManager on the VumiApi (i.e.
vumi_api.tpm).

hodgestar:
:+1:

PR #819, opened 2013-12-12
Use MessageMetadataHelper caching more widely

jerith:
There are a bunch of redis/riak lookups we can cache in `MessageMetadataHelper` to avoid unnecessary work.

I'm going to start with middlewares and the routing table dispatcher because that's what's hurting us the most right now.

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -76,9 +76,9 @@ class OptOutMiddleware(BaseMiddleware):
     @inlineCallbacks
     def handle_inbound(self, message, endpoint):
         optout_disabled = False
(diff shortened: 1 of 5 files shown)


jerith:
Turns out there was less to do here than I thought. We can probably improve things a little more by stashing the conversation object on outbound messages (where applicable) as well, but that's a little trickier to arrange.

hodgestar:
Looks good. Can we get tests for `get_tag_info` and `get_tagpool_metadata`?

hodgestar:
:+1:

smn:
:+1:

PR #818, opened 2013-12-12
Kill or fix DebitAccountMiddleware

jerith:
It doesn't have any tests and nothing uses it. It also seems to have major sync/async issues (based on a brief glance at its code).


hodgestar:
I'm a bit surprised `DebitAccountMiddleware` doesn't have tests because I'm sure we wrote them, but it can definitely just be killed.

go/base/context_processors.py
```diff
@@ -15,11 +15,10 @@ def standard_forms(request):
...
 def credit(request):
+    # TODO: Fill this in with real data.
     if request.user.is_authenticated() and hasattr(request, 'user_api'):
-        profile = request.user.get_profile()
-        api = request.user_api.api
         return {
-            'account_credits': api.cm.get_credit(profile.user_account) or 0,
+            'account_credits': 0,
         }
     return {}
```
hodgestar: This can just be deleted -- there is now `go.billing.templatetags.billing_tags.credit_balance` (https://github.com/praekelt/vumi-go/blob/develop/go/billing/templatetags/billing_tags.py?source=cc#L8)

hodgestar:
Otherwise looks good.

hodgestar:
I've restarted the Travis build that failed just to be sure -- the 2.6 build failed in a bunch of places that are probably unrelated.

hodgestar:
:+1:

PR #821, opened 2013-12-13
Unified message storing middleware

jerith:
This will replace our three different message storing middlewares (the base one from vumi, `ConversationStoringMiddleware` and `RouterStoringMiddleware`) with a single middleware that handle all the batches a message needs to be stored in.

This depends on praekelt/vumi#672.


go/vumitools/middleware.py
```diff
@@ -327,33 +327,41 @@ def teardown_middleware(self):
...
+        batch_ids = []
+        mdh = MessageMetadataHelper(self.vumi_api, msg)
+
+        if mdh.get_conversation_info():
+            conversation = yield mdh.get_conversation()
+            batch_ids.append(conversation.batch.key)
+
+        if mdh.get_router_info():
+            router = yield mdh.get_router()
+            batch_ids.append(router.batch.key)
```
hodgestar: Bit concerned that this now means that everything that uses this middleware now retrieves any set conversation and router from Riak regardless of whether this is actually necessary?

hodgestar:
The code looks fine but I think I'm -1 on indiscriminately loading routers and conversations from Riak (especially since this is meant to be a performance improvement). The current behaviour of looking up the conversation and router on conversation or router workers actually makes quite a bit of sense since the conversation / router needs to be looked up there anyway.

jerith:
Inbound messages only have as many unnecessary object lookup as there are routers in the chain (which is none for most of our higher-volume projects) but we're moving a necessary lookup to a more sensible place and removing an unnecessary message store (one lookup, one store, two redis ZADDs) per message.

Outbound messages have more unnecessary lookups, which is a thing I hadn't really considered. I could switch the two old middlewares back to their previous behaviour (which means we only have unnecessary lookups on the routing table dispatcher instead of all app/router workers) but that only buys us one fewer lookup per router.

I'm not that keen on adding config options or whatever for which things need to be looked up where, because that gets complicated quickly in the routing table dispatcher -- I'd rather figure out the right thing to do in the middleware than have a pile of config that we can screw up.

Thoughts for improving this in future:
- cache the `batch_id` for each object in `helper_metadata` somewhere so we don't have to look it up more than once
- stash the relevant object(s) on the messages we're publishing in workers for publish-side middleware to look at

jerith:
As per discussion on IRC, we've decided that this might not be the best approach and we need to give the problem a little more thought. Once we've done that we'll either close this PR or finish it.

hodgestar:
An alternative approach to the one taken here is to not store messages at the routing table dispatcher (since all the things it connects to should already store messages).

jerith:
This approach turned out to be problematic. Closing.

PR #811, opened 2013-12-17
DRY the http_api_nostream

jerith:
mop

Probably 80% has been copied wholesale from `http_api` and should either just be imported or things should be refactored a bit to remove the duplication.


go/apps/http_api_nostream/resource.py
```diff
@@ -329,29 +322,8 @@ def getDeferredChild(self, path, request):
...
-            protected_resource = HTTPAuthSessionWrapper(p, [factory])
-
-            return protected_resource
-        else:
-            return resource.NoResource()
+    def get_child_resource(self, path):
+        return {
+            'messages.json': MessageResource,
+            'metrics.json': MetricResource,
+        }.get(path)
```
hodgestar, line 329: No directly related to this PR, but should events no be accessible via http_api_nostream too?

hodgestar, line 329: Wait, those are only outgoing so they're probably somewhere else. /me looks.

hodgestar:
:+1:

go/apps/http_api_nostream/resource.py
jerith, line 329: `http_api_nostream` only needs resources for things the client can send, which doesn't include events.

PR #823, opened 2013-12-18
Split up MessageMetadataHelper and use it in more places

jerith:
This is extracted from #821 because it's useful on its own.

### DIFF
--- a/go/vumitools/conversation/utils.py
+++ b/go/vumitools/conversation/utils.py
@@ -9,7 +9,7 @@ from twisted.internet.defer import returnValue
 from vumi.persist.model import Manager
 
 from go.vumitools.opt_out import OptOutStore
-from go.vumitools.utils import MessageMetadataHelper
+from go.vumitools.utils import MessageMetadataDictHelper
 
 
 class ConversationWrapper(object):
@@ -149,10 +149,8 @@ class ConversationWrapper(object):
     def set_go_helper_metadata(self, helper_metadata=None):
         if helper_metadata is None:
             helper_metadata = {}
-        go_metadata = helper_metadata.setdefault('go', {})
-        go_metadata['user_account'] = self.user_account.key
-        go_metadata['conversation_type'] = self.conversation_type
-        go_metadata['conversation_key'] = self.key
+        mdh = MessageMetadataDictHelper(helper_metadata)
+        mdh.add_conversation_metadata(self)
         return helper_metadata
 
     @Manager.calls_manager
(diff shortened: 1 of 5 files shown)


go/vumitools/conversation/utils.py
```diff
@@ -249,7 +249,7 @@ def collect_messages(self, keys, proxy, include_sensitive, scrubber):
         for message in messages:
             # vumi message is an attribute on the inbound message object
             msg = message.msg
-            msg_mdh = MessageMetadataHelper(self.api, msg)
+            msg_mdh = MessageMetadataDictHelper(msg.get('helper_metadata', {}))
```
hodgestar: Not sure why this changed from MessageMetadataHelper to MessageMetadataDictHelper? We have a msg object here.

go/vumitools/tests/test_utils.py
```diff
@@ -1,13 +1,164 @@
...
 
-from go.vumitools.utils import MessageMetadataHelper
+from go.vumitools.utils import MessageMetadataDictHelper, MessageMetadataHelper
 from go.vumitools.tests.helpers import VumiApiHelper
 
 
+class TestMessageMetadataDictHelper(VumiTestCase):
+
+    def mk_msg(self, go_metadata=None, optout_metadata=None):
+        helper_metadata = {}
```
hodgestar, line 15: Should we not be using a msg_helper here instead of implementing `mk_msg` directly on top of `TransportUserMessage`?

go/vumitools/tests/test_utils.py
```diff
@@ -181,8 +332,50 @@ def test_get_router_key(self):
         md = self.mk_md(go_metadata={'router_key': 'router-1'})
         self.assertEqual(md.get_router_key(), 'router-1')
 
+    @inlineCallbacks
     def test_get_router(self):
-        raise SkipTest("Add test once VumiUserApi has a get_router method")
```
hodgestar, line 185: Yay for removing a skip!

go/vumitools/conversation/utils.py
jerith: To avoid the extra import, which is a stupid reason. I'll change it back.

go/vumitools/tests/test_utils.py
jerith, line 15: Possibly. I don't want to use GoMessageHelper, though, because that uses the thing we're trying to test here. I just stuck with what was there before.

hodgestar, line 15: Maybe we could just use the Vumi msg_helper?

hodgestar:
Looks good otherwise!

hodgestar:
:+1:

PR #696, opened 2013-12-19
RapidSMS integration (relay)

hodgestar:
Part of the UNICEF U-report project.


jerith:
Can we split the config changes out into their own PR?

hodgestar:
Created #833 for the config stuff.

hodgestar:
#833 has landed so this is ready for review again.

go/apps/rapidsms/view_definition.py
```diff
@@ -0,0 +1,122 @@
...
+    auth_token = forms.CharField(
+        help_text='The access token for this RapidSMS conversation.',
+        required=True)
+
+    @staticmethod
+    def initial_from_config(data):
+        initial = {}
+        for field in ("rapidsms_url", "rapidsms_username",
+                      "rapidsms_password", "rapidsms_auth_method",
+                      "rapidsms_http_method"):
```
justinvdm: Sorry if this is a little bikesheddy. Could we put this list at the class level so we don't need to change it both here and in `to_config`?

go/apps/rapidsms/view_definition.py
```diff
@@ -0,0 +1,122 @@
...
+            if field in data:
+                initial[field] = data[field]
+        allowed_endpoints = data.get('allowed_endpoints', ["default"])
+        initial["allowed_endpoints"] = u",".join(allowed_endpoints)
```
justinvdm: It looks like we allow the separator to be configurable for `EndpointsField`. Just checking if it is okay for us to just use `","` directly here.

go/apps/rapidsms/tests/test_vumi_app.py
```diff
@@ -0,0 +1,69 @@
...
+from vumi.tests.utils import LogCatcher
+from vumi.tests.helpers import VumiTestCase
+
+from go.apps.tests.helpers import AppWorkerHelper
+from go.apps.rapidsms.vumi_app import RapidSMSApplication, RapidSMSConfig
+
+
+class RapidSMSConfigTestCase(VumiTestCase):
+    # TODO: implement
+    pass
```
justinvdm: Any reason we can't do this now?

go/apps/rapidsms/tests/test_views.py
```diff
@@ -0,0 +1,54 @@
...
+        self.assertContains(response, u"<h1>myconv</h1>")
+
+    def test_show_running(self):
+        """
+        Test showing the conversation
+        """
+        conv_helper = self.app_helper.create_conversation_helper(
+            name=u"myconv", started=True)
+        response = self.client.get(conv_helper.get_view_url('show'))
+        self.assertContains(response, u"<h1>myconv</h1>")
```
justinvdm, line 27: It feels like we are testing the same thing in this test and the test above. Is there a subtle difference I am missing?

go/apps/rapidsms/vumi_app.py
```diff
@@ -0,0 +1,60 @@
...
+    def teardown_application(self):
+        yield super(RapidSMSApplication, self).teardown_application()
+        yield self._go_teardown_worker()
+
+    def get_config_data_for_conversation(self, conversation):
+        return GoWorkerConfigData(
+            self.config, conversation.config.get('rapidsms', {}))
+
+    def get_config(self, msg):
+        return self.get_message_config(msg)
```
justinvdm, line 45: Just for my understanding, we do this here in order to get the dynamic conversation config (as opposed to the static config like in the `RapidSMSRelay` vumi app), right?

go/apps/rapidsms/tests/test_vumi_app.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+from twisted.internet.defer import inlineCallbacks
+
+from vumi.tests.utils import LogCatcher
+from vumi.tests.helpers import VumiTestCase
+
+from go.apps.tests.helpers import AppWorkerHelper
+from go.apps.rapidsms.vumi_app import RapidSMSApplication, RapidSMSConfig
```
justinvdm, line 9: I'm wondering if we need to differentiate these names more from `RapidSMSRelay` and `RapidSMSRelayConfig` in a way that shows these are specific to vumi go. That said, I'm happy leaving these names as is for sake of simplicity. What do you think?

justinvdm:
Looks good, just a few small comments and questions.

go/apps/rapidsms/view_definition.py
hodgestar: Done.

go/apps/rapidsms/view_definition.py
hodgestar: Good idea. Done.

go/apps/rapidsms/tests/test_vumi_app.py
hodgestar: I changed how this worked quite a bit (hopefully for the better). Test added too although there's not much left to test now.

go/apps/rapidsms/tests/test_views.py
hodgestar, line 27: The difference is the `started=True` in the second test.

go/apps/rapidsms/vumi_app.py
hodgestar, line 45: Yep. That we have to implement this here rather than having it on a base class is a bit of an annoying compromise resulting from using mixins (`get_config` is already implemented on the base class so we don't want to override it on the Vumi Go application worker mixin so we implement it a bit all over the place).

go/apps/rapidsms/tests/test_vumi_app.py
hodgestar, line 9: I think it's okay since they're only really used from config where the full dotted path is available and it's consistent with our other Go application names. Probably a good idea to make having 'Go' in the name somewhere a convention for conversation workers that have Go-specific overrides at some point though.

hodgestar:
Thanks for all the comments. Ready for re-review.

go/apps/rapidsms/tests/test_view_definition.py
```diff
@@ -0,0 +1,122 @@
...
+            'rapidsms_url': u'http://www.example.com/',
+            'rapidsms_username': u'rapid-user',
+            'rapidsms_password': u'rapid-pass',
+            'rapidsms_auth_method': u'basic',
+            'rapidsms_http_method': u'POST',
+            'allowed_endpoints': ['default', 'extra'],
+        })
+
+
+class TestRapidSmsForm(VumiTestCase):
```
justinvdm: Did you maybe mean `TestAuthTokensForm` here?

justinvdm:
Other than the small additional comment, :+1:

go/apps/rapidsms/tests/test_view_definition.py
hodgestar: I did. Thanks. Not sure why my editor didn't complaing about the name clash.

justinvdm:
:+1:

hodgestar:
I added a fix for an issue revealed by ad8ead3 (the name clash in the test classes meant the RapidSMSForm tests were no longer running).

justinvdm:
:+1: again :)

PR #832, opened 2014-01-02
Make message export not rely on message keys being in the MessageStoreCache.

smn:
This assumption goes away with the work going on in praekelt/vumi#675. Currently the message store cache is being used as a convenient way of retrieving message keys and timestamps, an alternative to that needs to be found.

### DIFF
--- a/go/conversation/tasks.py
+++ b/go/conversation/tasks.py
@@ -1,5 +1,3 @@
-import warnings
-
 from StringIO import StringIO
 from zipfile import ZipFile, ZIP_DEFLATED
 
--- a/go/conversation/tasks.py
+++ b/go/conversation/tasks.py
@@ -147,10 +147,12 @@ def export_conversation_messages_sorted(account_key, conversation_key):
     # limiting to 0 results in getting the full set, creating lists of tuples
     # with direction sort we can sort sent & received messages in a somewhat
     # threaded fashion further down.
-    sent_messages = [('sent', msg)
-                     for msg in conversation.sent_messages(limit=0)]
-    received_messages = [('received', msg)
-                         for msg in conversation.received_messages(limit=0)]
+    sent_messages = [
+        ('sent', msg)
+        for msg in conversation.sent_messages_in_cache(limit=0)]
+    received_messages = [
+        ('received', msg)
+        for msg in conversation.received_messages_in_cache(limit=0)]
 
     def sort_by_addr_and_timestap(entry):
         """
--- a/go/conversation/tasks.py
+++ b/go/conversation/tasks.py
@@ -45,7 +45,8 @@ def load_messages_in_chunks(conversation, direction=None, size=20,
     :param int size:
         How big the chunks should be. Default 20.
     :param bool include_sensitive:
-        If ``true`` then all messages marked as `sensitive` are skipped.
+        If ``False`` then all messages marked as `sensitive` are skipped.
+        Defaults to ``False``.
     :param callable scrubber:
         If provided, this is called for every message allowing it to be
         modified on the fly.
--- a/go/conversation/tasks.py
+++ b/go/conversation/tasks.py
@@ -31,6 +31,40 @@ def write_messages(writer, messages):
     return messages
 
 
+def load_messages_in_chunks(conversation, direction=None, size=20,
+                            include_sensitive=False, scrubber=None):
+    """
+    Load the conversation's messages in chunks of `size`.
+    Uses `proxy.load_all_bunches()` lower down but allows skipping and/or
+    scrubbing of messages depending on `include_sensitive` and `scrubber`.
+
+    :param Conversation conv:
+        The conversation.
+    :param str direction:
+        The direction, either ``'inbound'`` or ``'outbound'``.
+    :param int size:
+        How big the chunks should be. Default 20.
+    :param bool include_sensitive:
+        If ``true`` then all messages marked as `sensitive` are skipped.
+    :param callable scrubber:
+        If provided, this is called for every message allowing it to be
+        modified on the fly.
+    """
+    direction = ('inbound' if direction == 'inbound' else 'outbound')
+    keys_func = getattr(conversation, '%s_keys' % (direction,))
+    proxy = getattr(conversation.mdb, '%s_messages' % (direction,))
+
+    keys = keys_func()
+    for chunk in grouper(keys, size):
+        # grouper() pads with `None` if less than `size` available,
+        # we unpad here.
+        chunk = filter(None, chunk)
+        messages = conversation.collect_messages(
+            chunk, proxy, include_sensitive=include_sensitive,
+            scrubber=scrubber)
+        yield messages
+
+
 def email_export(user_profile, conversation, io):
     zipio = StringIO()
     zf = ZipFile(zipio, "a", ZIP_DEFLATED)
@@ -60,25 +94,15 @@ def export_conversation_messages_unsorted(account_key, conversation_key):
     api = VumiUserApi.from_config_sync(account_key, settings.VUMI_API_CONFIG)
     user_profile = UserProfile.objects.get(user_account=account_key)
     conversation = api.get_wrapped_conversation(conversation_key)
-    mdb = conversation.mdb
-    io = StringIO()
 
+    io = StringIO()
     writer = UnicodeCSVWriter(io)
     writer.writerow(conversation_export_field_names)
 
-    inbound_keys = conversation.inbound_keys()
-    outbound_keys = conversation.outbound_keys()
-
-    for keys_chunk in grouper(inbound_keys, 20):
-        messages = conversation.collect_messages(
-            message_keys, conversation.messages.inbound_messages,
-            include_sensitive=False, scrubber=None)
+    for messages in load_messages_in_chunks(conversation, 'inbound'):
         write_messages(writer, messages)
 
-    for keys_chunk in grouper(outbound_keys, 20):
-        messages = conversation.collect_messages(
-            message_keys, conversation.messages.outbound_messages,
-            include_sensitive=False, scrubber=None)
+    for messages in load_messages_in_chunks(conversation, 'outbound'):
         write_messages(writer, messages)
 
     email_export(user_profile, conversation, io)
--- a/go/conversation/tasks.py
+++ b/go/conversation/tasks.py
@@ -116,14 +116,6 @@ def export_conversation_messages_unsorted(account_key, conversation_key):
 
 
 @task(ignore_result=True)
-def export_conversation_messages(account_key, conversation_key):
-    warnings.warn('export_conversation_messages() is deprecated. '
-                  'Please use export_conversation_messages_sorted() instead',
-                  category=DeprecationWarning)
-    return export_conversation_messages_sorted(account_key, conversation_key)
-
-
-@task(ignore_result=True)
 def export_conversation_messages_sorted(account_key, conversation_key):
     """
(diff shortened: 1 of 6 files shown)


smn:
@hodgestar @justinvdm ready for review

go/conversation/tasks.py
```diff
@@ -8,29 +10,124 @@
...
+def email_export(user_profile, conversation, io):
+    zipio = StringIO()
+    zf = ZipFile(zipio, "a", ZIP_DEFLATED)
+    zf.writestr("messages-export.csv", io.getvalue())
+    zf.close()
+
+    email = EmailMessage(
+        'Conversation message export: %s' % (conversation.name,),
+        'Please find the messages of the conversation %s attached.\n' % (
+            conversation.name),
```
justinvdm, line 75: If `email_export` is a reusable method for sending out emails, should we maybe have the email subject and contents be params instead? The current subject and contents look specific to message exports.

smn, line 75: Happy for suggestions. I did consider that but I decided to keep this usable within the context of these tasks, trying to make it more generic only gave me a very thin layer on top of the already existing email tools that Django gives us with little added benefit. Perhaps I should rename it or prefix it with an underscore?

justinvdm, line 75: Ah, looks like all of the tasks are actually message related, happy with it staying the way it is.

go/conversation/tests.py
```diff
@@ -839,3 +842,62 @@ def bad_add_entity(*a, **kw):
...
+            email.recipients(), [self.user_helper.get_django_user().email])
+        self.assertTrue(conv.name in email.subject)
+        self.assertTrue(conv.name in email.body)
+        [(file_name, zipcontent, mime_type)] = email.attachments
+        self.assertEqual(file_name, 'messages-export.zip')
+        zipfile = ZipFile(StringIO(zipcontent), 'r')
+        content = zipfile.open('messages-export.csv', 'r').read()
+        # 1 header, 5 sent, 5 received, 1 trailing newline == 12
+        self.assertEqual(12, len(content.split('\n')))
+        self.assertEqual(mime_type, 'application/zip')
```
justinvdm: I'm wondering if it would be helpful to check that we are exporting the messages we are expecting. Checking based on lines in the file is probably enough, but checking against the actual messages feels like a stronger assertion. Could we not do something similar to the `export_conversation_messages_sorted` test, but sort the result rows when doing the assertion?

justinvdm:
Other than the really minor comment about the `export_conversation_messages_unsorted` test, looks good. I do think someone else should take a look though, I'm probably not familiar enough with the message store to know if things are being done right/wrong.

smn:
Ready for re-review.

justinvdm:
I'm :+1: on this landing, but I'm not too confident on my knowledge of the message store. Your call if you think someone else needs to review it.

smn:
@justinvdm thanks! @jerith @hodgestar for when you have time.

go/conversation/tasks.py
```diff
@@ -8,29 +10,124 @@
...
+    scrubbing of messages depending on `include_sensitive` and `scrubber`.
+
+    :param Conversation conv:
+        The conversation.
+    :param str direction:
+        The direction, either ``'inbound'`` or ``'outbound'``.
+    :param int size:
+        How big the chunks should be. Default 20.
+    :param bool include_sensitive:
+        If ``true`` then all messages marked as `sensitive` are skipped.
```
jerith: The param name and description say opposite things. I think the description is wrong, unless `collect_messages()` (which we pass the value to) also uses a bad name.

go/conversation/tasks.py
```diff
@@ -8,29 +10,124 @@
...
+    :param callable scrubber:
+        If provided, this is called for every message allowing it to be
+        modified on the fly.
+    """
+    direction = ('inbound' if direction == 'inbound' else 'outbound')
+    keys_func = getattr(conversation, '%s_keys' % (direction,))
+    proxy = getattr(conversation.mdb, '%s_messages' % (direction,))
```
jerith: The direction should be either `'inbound'` or `'outbound'` here and we should throw a `ValueError` for anything that isn't. I'd also prefer to explicitly name the attributes we're getting (in an `if` block or something) instead of constructing their names and using `getattr()`.

jerith: Actually, since we're calling `keys_func` exactly once just below this, it might make sense to just call the right thing instead of sticking it in a local variable first.

go/conversation/tasks.py
smn: Thanks, fixed.

go/conversation/tasks.py
```diff
@@ -8,29 +10,124 @@
...
+
+    keys = keys_func()
+    for chunk in grouper(keys, size):
+        # grouper() pads with `None` if less than `size` available,
+        # we unpad here.
+        chunk = filter(None, chunk)
```
jerith: Perhaps it would make sense to skip `grouper()` and just use `izip_longest()` here directly without worrying about the padding we're adding and then removing.

go/conversation/tasks.py
smn: You're right, fixed.

go/conversation/tasks.py
```diff
@@ -8,29 +10,130 @@
...
+    """
+    if direction == 'inbound':
+        keys = conversation.inbound_keys()
+        proxy = conversation.mdb.inbound_messages
+    elif direction == 'outbound':
+        keys = conversation.outbound_keys()
+        proxy = conversation.mdb.outbound_messages
+    else:
+        raise ValueError('Invalid value received for `direction`. '
+                         'Only `inbound` and `outbound` are allowed.')
```
jerith: Perhaps put the value we got in the exception? That might be handy for debugging. Also, the `direction` param should now either be mandatory or have a valid default value.

go/conversation/tasks.py
smn: `izip_longest()` defaults to using `None` as the `fillvalue` and doesn't provide an easy means of grouping not seeing how using it directly here would make this cleaner?

go/conversation/tasks.py
```diff
@@ -8,29 +10,130 @@
...
+
+    for chunk in grouper(keys, size):
+        # grouper() pads with `None` if less than `size` available,
+        # we unpad here.
+        chunk = filter(None, chunk)
+        messages = conversation.collect_messages(
+            chunk, proxy, include_sensitive=include_sensitive,
+            scrubber=scrubber)
+        yield messages
```
jerith: Instead of doing all this here, it might be better to factor the message processing out of `collect_messages()` and just use `load_all_bunches()` on the proxy object directly. We should actually get rid of `collect_messages()` completely, because it always loads all messages in a conversation into memory and that's going to hurt us no matter where it's used.

go/conversation/tasks.py
```diff
@@ -8,29 +10,131 @@
...
+
+    email_export(user_profile, conversation, io)
+
+
+@task(ignore_result=True)
+def export_conversation_messages(account_key, conversation_key):
+    warnings.warn('export_conversation_messages() is deprecated. '
+                  'Please use export_conversation_messages_sorted() instead',
+                  category=DeprecationWarning)
+    return export_conversation_messages_sorted(account_key, conversation_key)
```
jerith: Do we need to deprecate this if we find all the places that call it and change them?

go/conversation/tasks.py
smn: `collect_messages()` is useful for the UI but needs to be used with caution, if you give it all the keys it will load them all into memory. The point is not to do that. It is also used by `export_conversation_messages_sorted()` which _is_ currently loading all messages in memory for sorting (which is why I'm doing this work). I'm happy to make that go away in a separate PR.

go/conversation/tasks.py
smn: My thinking is that I'm not sure who else is currently using this task and where, out side of us.

go/conversation/tests.py
```diff
@@ -839,3 +842,65 @@ def bad_add_entity(*a, **kw):
 
         self.assertEqual(self.error_log, [':('])
         self.assertEqual(response.context['dashboard_config'], None)
+
+
+class TestConversationTasks(GoDjangoTestCase):
+    def setUp(self):
+        self.vumi_helper = self.add_helper(
+            DjangoVumiApiHelper(), setup_vumi_api=False)
+        self.vumi_helper.setup_vumi_api()
```
jerith: Why pass `setup_vumi_api=False` and then immediately call `setup_vumi_api()`?

go/vumitools/conversation/utils.py
```diff
@@ -258,9 +259,17 @@ def collect_messages(self, keys, proxy, include_sensitive, scrubber):
...
         returnValue(collection)
 
-    @Manager.calls_manager
     def received_messages(self, start=0, limit=100, include_sensitive=False,
                           scrubber=None):
+        warnings.warn('received_messages() is deprecated. Please use '
+                      'received_messages_in_cache() instead.',
+                      category=DeprecationWarning)
+        return self.received_messages_in_cache(start, limit, include_sensitive,
+                                               scrubber)
```
jerith, line 283: I'd rather find and fix all the places that call this now than deprecate it. There isn't any third-party code relying on it.

go/conversation/tests.py
smn: copy paste error, thanks, fixing.

go/conversation/tasks.py
jerith: `collect_messages()` isn't particularly useful for the UI because showing a hundred thousand messages isn't feasible in any case. I'm happy to have a ticket for removing it instead of doing it in this branch, but I think it's still worth factoring out the message processing part of it and using that with `load_all_bunches()` to simplify the logic here.

go/vumitools/conversation/utils.py
smn, line 283: Done & in a single, hopefully won't meddle with reviewability too much.

go/conversation/tasks.py
jerith: We make breaking changes in Vumi Go all the time, so I don't think it's worth doing the deprecation thing here unless we also decide to do it for all the other internal things. I think the cost of that is much too high when there's no evidence that anyone other than us is using the code.

jerith: I'm happy to keep the deprecation stuff here in the short term if celery needs it, though. I don't know how it handles code changing underneath scheduled tasks.

go/conversation/tasks.py
smn: It's what drives the loading of messages to show your messages in the paginated message list, we do this by only handing a subset of all keys to `collect_messages()` (only the keys for that page). It is probably worth factoring out but there's still going to be shared code between this & the UI as both need to allow for the scrubbing & skipping of sensitive messages.

go/conversation/tasks.py
smn: Ok, removed.

go/conversation/tasks.py
jerith: I meant factor it out into a thing on `ConversationWrapper` that takes a list of messages (or just one message, if that makes more sense) and processes it.

smn: Cool, can address that in a separate ticket.

go/conversation/tasks.py
jerith: Ah, good point.

jerith:
I'm happy with this branch once my commits have been reviewed and Travis has done its thing successfully.

PR #833, opened 2014-01-08
Split conversation and router config classes.

hodgestar:
Currently conversation and router config classes share too much infrastructure (and the infrastructure is rather hacky). This splits things up more and cleans things up a big. A further goal is to allow applications more control over how conversation config is translated into application config.

### DIFF
--- a/go/vumitools/app_worker.py
+++ b/go/vumitools/app_worker.py
@@ -6,7 +6,7 @@ from vumi import log
 from vumi.worker import BaseWorker
 from vumi.application import ApplicationWorker
 from vumi.blinkenlights.metrics import MetricManager
-from vumi.config import IConfigData, ConfigText, ConfigDict
+from vumi.config import IConfigData, ConfigText, ConfigDict, ConfigField
 from vumi.connectors import IgnoreMessage
 
 from go.config import get_conversation_definition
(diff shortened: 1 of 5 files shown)


go/vumitools/app_worker.py
```diff
@@ -289,13 +294,18 @@ def add_conv_to_msg_options(self, conv, msg_options):
 
 
 class GoApplicationMixin(GoWorkerMixin):
+    def get_config_data_for_conversation(self, conversation):
+        config = conversation.config.copy()
+        config["conversation"] = conversation
+        return GoWorkerConfigData(self.config, config)
```
hodgestar, line 300: This method was split out so that it can be overridden by sub-classes that want to construct their config differently. Happy to have a slightly different set of functionality split out here if anyone has a suggestion they like better.

go/vumitools/app_worker.py
```diff
@@ -429,7 +444,13 @@ def process_command_stop(self, user_account_key, router_key):
         yield router.save()
 
 
-class GoApplicationConfig(ApplicationWorker.CONFIG_CLASS, GoWorkerConfigMixin):
+class GoApplicationConfigMixin(GoWorkerConfigMixin):
+    conversation = ConfigConversation(
+        "Conversation instance for this message", required=False)
```
jerith, line 449: Why isn't this required?

go/vumitools/app_worker.py
```diff
@@ -495,7 +522,7 @@ def teardown_worker(self):
         self.pause_connectors()
         return self.teardown_router()
 
-    def get_config(self, msg):
+    def get_config(self, msg, ctxt=None):
```
jerith, line 525: What's `ctxt` for?

hodgestar, line 525: `ctxt` is part of the signature of `get_config()` -- it's for passing in additional config in situations where one needs dynamic config but the message isn't available (e.g. messages received via HTTP request). I put it in here so that this matches the signature of the base `get_config`.

go/vumitools/app_worker.py
hodgestar, line 449: There are situations where `get_config` is called without a msg -- e.g. in the http_api to check some things on incoming requests.

jerith:
:+1: once Travis is happy.

PR #757, opened 2014-01-10
Sort out requirements and setup.py

jerith:
We really should do something more like https://caremad.io/blog/setup-vs-requirement/ instead of trying to parse `requirements.pip` in `setup.py`.


hodgestar:
:+1:

PR #838, opened 2014-01-15
User profile inline admin form clashes with post-save hook

hodgestar:
Currently the inline user profile save box in the admin interface clashes with the post-save hook that automatically creates a user profile. If one sets a user profile value in the inline form on the user creation page, then on save the new user is saved and then Django attempts to create a second user profile, resulting in a uniqueness constraint error.

### DIFF
--- a/go/base/admin.py
+++ b/go/base/admin.py
@@ -14,6 +14,11 @@ class UserProfileInline(admin.StackedInline):
 
 
 class GoUserAdmin(UserAdmin):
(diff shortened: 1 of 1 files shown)


hodgestar:
Maybe implementing a custom https://docs.djangoproject.com/en/1.5/ref/contrib/admin/#django.contrib.admin.ModelAdmin.save_related can help?

smn:
:+1: because I'm unsure how you'd test this within the context of the admin

PR #841, opened 2014-01-23
Better test helper setup management

jerith:
This is required by some changes that are coming soon as part of praekelt/vumi#707.


hodgestar:
:+1:

PR #844, opened 2014-02-04
Application Multiplexer

vgeddes:
This is to track a special router which links various Go USSD apps to a single starcode.


go/config.py
```diff
@@ -136,6 +136,10 @@ def get_router_definition(router_type, router=None):
         'namespace': 'keyword',
         'display_name': 'Keyword',
     },
+    'go.routers.app_multiplexer': {
+        'namespace': 'application_multiplexer',
```
hodgestar: It's probably best if the namespace matches the module name, i.e. `app_multiplexer`. We don't do that in a few places but those are all places which we want to get rid of eventually.

go/routers/app_multiplexer/definition.py
```diff
@@ -0,0 +1,8 @@
+from go.vumitools.router.definition import RouterDefinitionBase
+
+
+class RouterDefinition(RouterDefinitionBase):
+    router_type = 'application_multiplexer'
```
hodgestar: This will need to be updated when we update the namespace name.

hodgestar:
General layout looks good. :)

go/router/templates/router/edit.html
```diff
@@ -24,6 +24,8 @@
                 {% for edit_form in edit_forms %}
                     <fieldset>
                         {{ edit_form|crispy }}
+                        <!-- TODO: style this correctly -->
+                        <strong>{{ edit_form.non_form_errors }}</strong>
```
hodgestar: I think this should be above the form and probably protected with a block that checks that `non_form_errors` exists (i.e. the form is a formset) and a note that this is a crispy forms bug (I've filed https://github.com/maraujop/django-crispy-forms/issues/298 for it).

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
+from django import forms
+
+from go.router.view_definition import RouterViewDefinitionBase, EditRouterView
+from go.routers.app_multiplexer.common import mkmenu
+
+
+class ApplicationMultiplexerTitleForm(forms.Form):
+    menu_title = forms.CharField(
+        label="Menu Title",
+        max_length=29
```
hodgestar: Let's set `max_length` to `100` or `255` or something. People configuring application multiplexers probably know what they're doing and if it looks bad they can fix it themselves easily enough. Or just leave out the max_length all together.

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
+from django import forms
+
+from go.router.view_definition import RouterViewDefinitionBase, EditRouterView
+from go.routers.app_multiplexer.common import mkmenu
+
+
+class ApplicationMultiplexerTitleForm(forms.Form):
+    menu_title = forms.CharField(
+        label="Menu Title",
```
hodgestar: Nitpick -- small `t` for `title` is probably more correct.

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
...
+        max_length=29
+    )
+
+
+class ApplicationMultiplexerForm(forms.Form):
+    application_title = forms.CharField(
+        label="Application Title",
```
hodgestar: I would call this `Application name` or `Application label`.

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
...
+            titles.append(title)
+            endpoints.append(endpoint)
+
+        # Menu size check for the benefit of USSD which is usually limited
+        # to 160 chars
+        if len(mkmenu(titles)) > self.MENU_CHAR_LENGTH:
+            raise forms.ValidationError(
+                "The generated menu is too large. Either reduce the length of "
+                "application titles or the number of applications."
+            )
```
hodgestar: Can we just drop this validation as discussed on IRC yesterday? There isn't actually a hard requirement that endpoint names are unique (although it'd be usual if they weren't), or that application names be unique (although that'd be unusual too) and the length of the menu that makes sense is really transport and application dependent.

hodgestar: Just because the Django docs have this, doesn't mean we have to. :)

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @staticmethod
+    def initial_from_config(data):
+        return [{'application_title': k, 'target_endpoint': v}
+                for k, v in sorted(data.items())]
+
+    def to_config(self):
+        mappings = {}
+        for form in self.forms:
+            if (not form.is_valid()) or form.cleaned_data['DELETE']:
```
hodgestar: Elsewhere we use `if not form.cleaned_data or form in self.deleted_forms` here which might be a better check (the form will already have been validated by the time `to_config` is called and `deleted_forms` will have been populated).

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
...
+                continue
+            application_title = form.cleaned_data['application_title']
+            target_endpoint = form.cleaned_data['target_endpoint']
+            mappings[application_title] = target_endpoint
```
hodgestar: We're going to need to maintain the ordering of the applications, so we probably want this to be a list of dicts, not a simple mapping.

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,94 @@
...
+        return mappings
+
+
+ApplicationMultiplexerFormSet = forms.formsets.formset_factory(
+    ApplicationMultiplexerForm,
+    can_delete=True,
+    extra=1,
+    can_order=True,
+    formset=BaseApplicationMultiplexerFormSet)
```
hodgestar: I suggest putting these forms in a `forms.py` module and keeping this module for the views (just to keep things tidy and a bit more like Django coders expect). Also allows gives a nice place (i.e. `tests/test_forms.py`) to put direct tests of the form methods.

go/routers/app_multiplexer/tests/test_views.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_get_edit_empty_config(self):
+        rtr_helper = self.router_helper.create_router_helper(started=True)
+        response = self.client.get(rtr_helper.get_view_url('edit'))
+        self.assertEqual(response.status_code, 200)
+
+    def test_get_edit_small_config(self):
+        rtr_helper = self.router_helper.create_router_helper(
+            started=True, config={
+                'endpoints': {
+                    'mykeyw[o0]rd': 'target_endpoint',
```
hodgestar: This should be updated with less confusing test data.

go/routers/app_multiplexer/tests/test_views.py
```diff
@@ -0,0 +1,132 @@
...
+            'endpoints-0-application_title': ['foo'],
+            'endpoints-0-target_endpoint': ['cat'],
+            'endpoints-0-DELETE': [''],
+            'endpoints-1-application_title': ['foo'],
+            'endpoints-1-target_endpoint': ['dog'],
+            'endpoints-1-DELETE': [''],
+        })
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, ('Application titles and endpoints '
+                                       'should be distinct.'))
```
hodgestar: I know the keyword router does it, but I wouldn't test this validation stuff here -- I would test the forms themselves directly.

go/routers/app_multiplexer/tests/test_views.py
```diff
@@ -0,0 +1,132 @@
...
+                'endpoints-%s-application_title' % i: ['foo-%s' % i],
+                'endpoints-%s-target_endpoint' % i: ['bar-%s' % i],
+                'endpoints-%s-DELETE' % i: [''],
+            })
+        response = self.client.post(rtr_helper.get_view_url('edit'), fields)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response,
+                            ("The generated menu is too large. Either reduce "
+                             "the length of application titles or the number "
+                             "of applications."))
```
hodgestar: Ditto for this test.

hodgestar:
Review done for now. I would work on `ApplicationMultiplexer` in `vumi_app.py` since that's were the core functionality lives.

go/router/templates/router/edit.html
```diff
@@ -23,6 +23,9 @@
             <form id="form-router" class="indent" method="post" action="">
                 {% for edit_form in edit_forms %}
                     <fieldset>
+                        {% if edit_form.non_non_field_errors %}
```
jerith: Is that supposed to be `non_non_` there?

go/routers/application_multiplexer/common.py
```diff
@@ -0,0 +1,22 @@
+# helpers lifted from GFM and Wikipedia
+
```
jerith: I think these eventually need to live in a more general module, but that doesn't need to happen now. Can you add a TODO to that effect?

go/routers/application_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,122 @@
...
+
+    router_class = ApplicationMultiplexer
+
+    @inlineCallbacks
+    def setUp(self):
+        self.router_helper = self.add_helper(
+            RouterWorkerHelper(ApplicationMultiplexer))
+
+        self.msg_helper = yield self.add_helper(MessageHelper())
+        self.persistence_helper = yield self.add_helper(PersistenceHelper())
```
jerith: A `RouterWorkerHelper` already has its own `PersistenceHelper` and `MessageHelper`. If it's hard to get at the things you need on it, we should fix the helper.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
+# Local Variables:
+# test-case-name: go.routers.application_multiplexer.tests.test_vumi_app
+# coding: utf-8
+# End:
```
jerith: I don't think Python will be able to find this `coding:` line. Do we actually need non-ASCII in this file? Generally that's only used in tests where we have unicode literals for test data.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
...
+class ApplicationMultiplexerConfig(GoRouterWorker.CONFIG_CLASS):
+
+    # Static configuration
+    session_expiry = ConfigInt(
+        "Maximum amount of time in seconds to keep session data around",
+        default=1800, static=True)
+
+    session_data_version = ConfigInt(
+        "Just in case we need to modify the schema of session storage",
+        default=1, static=True)
```
jerith: I don't think this should be a config. Session data is inherently transient, so we should just be able to recognise invalid sessions and discard them. Even if we decide to version the session format, the version we're expecting should be a constant in the code rather than a config variable.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
...
+    # Dynamic, per-message configuration
+    menu_title = ConfigDict(
+        "Content for the menu title",
+        default={'content': "Please select a choice."})
+    entries = ConfigList(
+        "A list of application endpoints and associated labels",
+        default=[])
+    keyword = ConfigText(
+        "Keyword to signal a request to return to the application menu",
+        default=':menu')
```
jerith: I'd rather default this to `None` and avoid possible unexpected hijacking of valid app messages. (It's less unexpected if this has explicitly been configured for a particular router instance.)

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
...
+
+    HANDLERS = {
+        STATE_START: 'handle_state_start',
+        STATE_SELECT: 'handle_state_select',
+        STATE_SELECTED: 'handle_state_selected',
+        STATE_BAD_INPUT: 'handle_state_bad_input'
+    }
+
+    def setup_router(self):
+        return super(ApplicationMultiplexer, self).setup_router()
```
jerith: Rather leave these out if all they're doing is calling `super()`.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
...
+        )
+
+    def valid_target_endpoint(self, config, active_endpoint):
+        """
+        Make sure the currently active endpoint is still valid.
+        """
+        endpoints = set([entry['endpoint'] for entry in config.entries])
+        if active_endpoint not in endpoints:
+            return False
+        return True
```
jerith: Maybe just `return active_endpoint in endpoints` here?

go/routers/application_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,122 @@
...
+        self.parent_redis = yield self.persistence_helper.get_redis_manager()
+        self.router_worker = yield self.router_helper.get_router_worker({
+            'worker_name': 'application_multiplexer',
+            'redis_manager': {
+                'FAKE_REDIS': self.parent_redis,
+                'key_prefix': self.parent_redis.get_key_prefix(),
+            }
+        })
```
jerith: Most of this is actually unnecessary. `RouterWorkerHelper` proxies everything you need from `MessageHelper` (so you can just replace the `self.msg_helper` with `self.router_helper` everywhere) and all the persistence config is already handled by `get_router_worker()`. (We do _something_ with `worker_name` in there as well, but I don't recall what offhand.) Basically all you need here is the `add_helper()` for `RouterWorkerHelper` and `self.router_worker = yield self.router_helper.get_router_worker({})`.

The helpers really need to be better documented, but I haven't had time to work on #824 yet.

go/routers/application_multiplexer/tests/test_views.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def test_initial_config(self):
+        rtr_helper = self.router_helper.create_router_helper(
+            started=True, config={
+                'menu_title': {'content': 'Please select an application'},
+                'entries': [
+                    {
+                        'label': 'Flappy Bird',
+                        'endpoint': 'flappy-bird',
+
```
hodgestar: Stray empty line.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
...
+    STATE_SELECT = "select"
+    STATE_SELECTED = "selected"
+    STATE_BAD_INPUT = "bad_input"
+
+    HANDLERS = {
+        STATE_START: 'handle_state_start',
+        STATE_SELECT: 'handle_state_select',
+        STATE_SELECTED: 'handle_state_selected',
+        STATE_BAD_INPUT: 'handle_state_bad_input'
+    }
```
hodgestar: I have a slight preference for constructing this sort of thing in `setup_router` since it means the worker fails straight away if one gets any of the names wrong and avoids a call to `getattr`. Not a big deal though.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,231 @@
...
+    @inlineCallbacks
+    def handle_state_select(self, config, session, msg):
+        """
+        NOTE: There is an edge case in which the user input no longer
+        matches an entry in the current config. The impact depends
+        on the number of users and how often the router config is modified.
+
+        One solution would be to compare hashes of the menu configs as
+        they existed at the START and SELECT states. If there is a mismatch,
+        invalidate the current session.
```
hodgestar: I think not breaking sessions if the endpoint name doesn't change is sensible. If the endpoint name changes, then people will have to redo their routing table anyway so all bets are off.

go/settings.py
```diff
@@ -1,4 +1,3 @@
-
 # Django settings for go project.
 import os
 import djcelery
```
hodgestar, line 4: We should undo this change for forms sake.

hodgestar:
Generally looking pretty good. Shout once you've addressed or responded to the current round of comments and made any other changes you want to make and we can look at doing a final review and landing.

go/routers/application_multiplexer/common.py
vgeddes: Good point. I'd wager that these routines belong in Vumi core, so I've opened https://github.com/praekelt/vumi/issues/727 there to track this task.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,237 @@
...
+        user_id = msg['from_addr']
+        session_manager = yield self.session_manager(config)
+        session_event = msg['session_event']
+
+        session = yield session_manager.load_session(user_id)
+        if not session or session_event == TransportUserMessage.SESSION_NEW:
+            log.msg("Creating session for user %s" % user_id)
+            session = {}
+            state = self.STATE_START
+            yield session_manager.create_session(user_id)
```
jerith: Rather do `session = yield session_manager.create_session(user_id)` instead of setting the session to an empty dict.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,237 @@
...
+        else:
+            log.msg("Loading session for user %s: %s" % (user_id, session,))
+            state = session['state']
+
+        try:
+            result = yield self.handlers[state](config, session, msg)
+            # update session with next state and updated session fields
+            if type(result) is tuple:
```
jerith: Firstly, `isinstance()` is generally better than `type(foo) is sometype` for various reasons. If you really need the latter instead for some reason (and you don't in this case), it should be commented.

Secondly, having two different kinds of return value is confusing. Either both parts should always be present (and then the `result = yield ...` becomes `next_state, session_update = yield ...`) or the handler should just take care of its own session data changes. I prefer the latter.

go/routers/application_multiplexer/vumi_app.py
jerith: You can store the mapping between menu item and endpoint in the session and grab it from there instead of the config. The only thing you really need to check is that the endpoint selected is still actually connected to the router, but that should probably happen for every message forwarded anyway.

jerith:
Any time we receive a "session end" message (in either direction), we should bypass all state management, forward the message, and clear the session.

go/routers/application_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,456 @@
...
+        A helper to validate routing behavior.
+
+        The state dict describes the messages which need to be sent, what
+        session data to initialize, and what data should be asserted
+        when the state handler completes execution.
+
+        Messages are represented by a tuple (content, {field=value, ...})
+
+        This could be made into a generic test helper one day.
+        """
```
jerith: I don't think this is the right tool for the kind of tests we need here. Our state management is very straightforward and we don't really have much session data to throw around. I'd be much happier with tests that just set up a router, sent some messages, and asserted that the right things happened.

vgeddes: I agree this helper is only useful for this router, and is perhaps a bit too clever and generalized. On the other hand, I think it still saves on fair amount of code duplication, and makes clearer the intent and expectations of each individual test. I can certainly make the helper less clever and more specific to this router though.

jerith: I find it makes the tests less readable by mixing setup, actions and assertions in one big dict. Also, in order to understand what's actually happening I need to read `check_state()` and the test and construct the flow of execution in my head. Doing all those things separately is a little more verbose (not all that much, really), but makes it possible to read and understand each test on its own.

go/routers/application_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def session_manager(self, config):
+        """
+        The implementation of SessionManager does the job of
+        appending ':session' to key names.
+        """
+        return SessionManager.from_redis_config(
+            config.redis_manager,
+            max_session_length=config.session_expiry
+        )
```
hodgestar: It's more common for us to have a single session manager based on static config, and I think that would be a good approach here too (i.e. set a session_manager attribute in `setup_router`).

jerith: We probably want the session store to have the router key in it somewhere, though, otherwise we risk session collisions across different routers/conversations.

vgeddes:
@jerith, regarding the handling of "session end" messages from the user, its not yet clear cut how we want to handle them. If the "session end" message is due to an MNO-initiated timeout, then users may be a bit unhappy if we clear their session, and they can't resume using the application they were using last. That's why I haven't added yet added a check for these close messages on the inbound path. I wanna play a bit with router once its deployed to get a feel for what sort of behavior is best.

For the outbound path, if a linked application sends a close message, then we want to revert back to the menu of endpoints, rather than delete the current session. This is still a minor TODO on my list.

jerith:
> @jerith, regarding the handling of "session end" messages from the user, its not yet clear cut how we want to handle them. If the "session end" message is due to an MNO-initiated timeout, then users may be a bit unhappy if we clear their session, and they can't resume using the application they were using last. That's why I haven't added yet added a check for these close messages on the inbound path. I wanna play a bit with router once its deployed to get a feel for what sort of behavior is best.

We're already clearing their session data when we get a start-of-session message anyway. In any case, I'd much rather have to choose my app from the menu again than not be able to switch apps because the router remembers what I was using in my last session.

> For the outbound path, if a linked application sends a close message, then we want to revert back to the menu of endpoints, rather than delete the current session. This is still a minor TODO on my list.

USSD session are aggressively time-limited and it's easy enough to dial in again and choose a different app if desired. Also, in order to implement this you'd have to modify the outbound message to wait for more input instead of closing the session. That would be really confusing if you don't also modify the text of that message, which is undesirable at best and risks breaking all sorts of things.

Basically, I don't believe this router should do anything more complex than present a menu at the start of every session and proxy the selected app for the remainder of that session. If we really want additional behaviour later we can add it in a separate branch or write a different router for it.

jerith:
This is a bit meh, but can you maybe rename `go.routers.application_multiplexer` to `go.routers.app_multiplexer` so we don't end up with long identifiers all over the place?

As well as making the `test-case-name` editor variable PEP-8 compliant, it'll reduce the size of various keys in various datastores.

vgeddes:
Heh, originally the module name was actually `go.routers.app_multiplexer`, but @hodgestar suggested I expand the name to its current form for the sake of consistency.

How about renaming the module to `go.routers.multiplexer`? That's short and avoids the awkward the `app` contraction.

jerith:
I think `go.routers.app_multiplexer` is better. All our application workers live in `go.apps` anyway.

hodgestar:
I don't mind much whether it's `app_...` or `application_...` --
originally it was both in different places so I justed suggested a
standardization. Happy either way.

go/config.py
jerith: The `namespace` value needs to change to match the module name.

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,274 @@
...
+            state = self.STATE_START
+            yield session_manager.create_session(user_id)
+        else:
+            log.msg("Loading session for user %s: %s" % (user_id, session,))
+            state = session['state']
+
+        try:
+            next_state, updated_session = yield self.handlers[state](
+                config, session, msg)
+            if next_state == self.STATE_ABORT_SESSION:
```
jerith: I'd rather use `None` to denote "we're done, clear the session" and avoid named states that don't have corresponding handlers.

go/routers/app_multiplexer/tests/test_views.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertContains(response, 'Please select an application')
+        self.assertContains(response, 'Flappy Bird')
+        self.assertContains(response, 'flappy-bird')
+
+    def test_initial_config_empty(self):
+        rtr_helper = self.router_helper.create_router_helper(started=True)
+        response = self.client.get(rtr_helper.get_view_url('edit'))
+        self.assertEqual(response.status_code, 200)
+
+    def test_user_input_good(self):
```
jerith, line 93: We should probably have a corresponding `test_user_input_bad()` or something.

go/routers/app_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,436 @@
...
+from go.routers.tests.helpers import RouterWorkerHelper
+
+
+def raise_error(*args, **kw):
+    raise RuntimeError("An anomaly has been detected")
+
+
+class TestApplicationMultiplexerRouter(VumiTestCase):
+
+    router_class = ApplicationMultiplexer
```
jerith: This attribute is unnecessary and should be removed.

go/routers/app_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,436 @@
...
+            user_id)
+        if 'created_at' in session:
+            del session['created_at']
+        self.assertEqual(session, expected_session,
+                         msg="Unexpected session data")
+
+    def dynamic_config(self, fields):
+        config = self.router_worker.config.copy()
+        config.update(fields)
+        return config
```
jerith: Given that this is only used in one place, can it be inlined instead?

go/routers/app_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,436 @@
...
+        self.assertEqual([], self.router_helper.ro.get_dispatched_events())
+
+        yield self.router_helper.ro.make_dispatch_outbound(
+            "foo", router=router)
+        self.assertEqual([], self.router_helper.ri.get_dispatched_outbound())
+        [nack] = self.router_helper.ro.get_dispatched_events()
+        self.assertEqual(nack['event_type'], 'nack')
+
+    @inlineCallbacks
+    def test_state_start_to_select(self):
```
jerith: Can you give these tests more descriptive names (`test_new_session` or `test_app_selection_menu` here, for example) and preferably docstrings as well?

We haven't been very good at docstrings in the past, but you can use https://github.com/praekelt/vumi/blob/develop/vumi/tests/test_test_helpers.py as an example.

go/routers/app_multiplexer/common.py
```diff
@@ -0,0 +1,25 @@
...
+def clean(content):
+    return (content or '').strip()
+
+
+def mkmenu(options, start=1, format='%s) %s'):
+    items = [format % (idx, opt) for idx, opt in enumerate(options, start)]
+    return '\n'.join(items)
+
+
+def smart_truncate(content, k, sfx='...'):
```
jerith: Nothing calls this and it's untested. Can you remove it please?

go/routers/app_multiplexer/tests/test_vumi_app.py
```diff
@@ -0,0 +1,436 @@
...
+        self.assertEqual(session, expected_session,
+                         msg="Unexpected session data")
+
+    def dynamic_config(self, fields):
+        config = self.router_worker.config.copy()
+        config.update(fields)
+        return config
+
+    @inlineCallbacks
+    def assert_routed_inbound(self, content, router, expected_endpoint):
```
jerith: Nothing calls this. Can it be removed?

go/routers/app_multiplexer/view_definition.py
```diff
@@ -0,0 +1,15 @@
+from go.router.view_definition import RouterViewDefinitionBase, EditRouterView
+
+from go.routers.app_multiplexer.forms import \
+    (ApplicationMultiplexerTitleForm, ApplicationMultiplexerFormSet)
```
jerith: Put the opening paren on the previous line and skip the trailing backslash. (Backslashes for line continuation are horrible for a variety of reasons and should basically never be used.)

hodgestar:
After discussion with @jerith I've been convinced that a dynamic session manager that uses the the name of the application worker combined with the router key as the key prefix is better than a static session manager. Apologies for the running around @vgeddes.

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,261 @@
...
+from go.vumitools.app_worker import GoRouterWorker
+from go.routers.app_multiplexer.common import mkmenu, clean
+
+
+class ApplicationMultiplexerConfig(GoRouterWorker.CONFIG_CLASS):
+
+    # Static configuration
+    session_expiry = ConfigInt(
+        "Maximum amount of time in seconds to keep session data around",
+        default=1800, static=True)
```
jerith: Half an hour is a very long default session expiry. Maybe change this to 5 minutes? We update the expiry every time we update the session, so it would expire 5 minutes after the last change instead of 5 minutes after the initial creation.

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,261 @@
...
+
+        try:
+            next_state, updated_session = yield self.handlers[state](
+                config, session, msg)
+            if next_state is None:
+                # Halt session immediately
+                log.msg(("Router configuration change forced session abort "
+                         "for user %s" % user_id))
+                yield self.session_manager.clear_session(user_id)
+                yield self.publish_error_reply(msg, config)
```
jerith: I don't think this is the right place for this logic. The handler should either raise an exception or it should log the error and send the response message itself.

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,261 @@
...
+        in the session data. Later, in the SELECT state, we load
+        these endpoints and retrieve the candidate endpoint based
+        on the user's menu choice.
+        """
+        reply_msg = msg.reply(self.create_menu(config))
+        yield self.publish_outbound(reply_msg)
+        endpoints = json.dumps(
+            [entry['endpoint'] for entry in config.entries]
+        )
+        returnValue((self.STATE_SELECT, dict(endpoints=endpoints)))
```
jerith: Why `dict(endpoints=endpoints)` instead of `{'endpoints': endpoints}`?

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,261 @@
...
+        session = yield self.session_manager.load_session(user_id)
+        if session and (session_event == TransportUserMessage.SESSION_CLOSE):
+            yield self.session_manager.clear_session(user_id)
+        yield self.publish_outbound(msg)
+
+    @inlineCallbacks
+    def handle_session_close(self, config, session, msg):
+        user_id = msg['from_addr']
+        if session.get('state', None) == self.STATE_SELECTED and \
+           session['active_endpoint'] in self.target_endpoints(config):
```
jerith: Backslash line continuation is bad. Rather use parens and an extra indent, like so:

``` python
    if (session.get('state', None) == self.STATE_SELECTED and
            session['active_endpoint'] in self.target_endpoints(config)):
```

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,261 @@
...
+            yield self.publish_inbound(
+                self.forwarded_message(
+                    msg, content=None,
+                    session_event=TransportUserMessage.SESSION_CLOSE),
```
jerith: I don't think we need to modify the inbound message here. It will already have `session_event=SESSION_CLOSE` and we don't actually care about the content.

jerith:
Finished this review pass. There are a whole pile of minor things (some of which probably slipped through previous review passes) and a couple of less minor ones.

go/routers/app_multiplexer/vumi_app.py
vgeddes: That form leads to slightly more readable code in this particular context, especially as I don't have to visually parse string literals, and other symbols required for dict literals. Its also easier to type for the same reasons.

jerith: A function call with a keyword parameter requires me to read and reason about the nature of the function being called, whereas the nature of a dict literal is immediately obvious. I find the latter more readable in most situations, especially when I'm expecting the expression to be a dict.

go/routers/app_multiplexer/vumi_app.py
vgeddes: I choose not to implement this control flow using an exception because its a path which is entirely expected under certain conditions and is not actually indicative of an error. The  `None` return value from state handlers indicates that there is no valid `next_state` to transition to and that the session should just be terminated immediately. 

As for whether this logic belongs here or not, I've aimed for the following separation of concerns - Where possible, state handler functions don't handle session management,  "error" messages, or cleanup. Thats the job of the `handle_inbound` function which delegates to the handlers.

I don't mind moving the `publish_error_reply()` call into the actual handlers though.

go/routers/app_multiplexer/vumi_app.py
vgeddes: Understood, though I think the nature of the `dict()` constructor is also pretty obvious and readable. While I would use dict literals in most cases, in this particular context, especially as we're mostly just creating a dict with 1 key, my subjective opinion is that the `dict()` form reads better.

go/routers/app_multiplexer/vumi_app.py
jerith: By putting this logic here, you're assuming that a state handler will never return `None` except when the config changes underneath them and the only way to examine that assumption is to examine all the handlers. More importantly, it's easy to invalidate that assumption by unknowingly modifying handler code.

A `next_state` of `None` should indicate that the session is complete and should be cleaned up, nothing more. Anything outside that belongs either in the state handler or an exception handler, both of which are better able to determine what to do.

go/routers/app_multiplexer/vumi_app.py
```diff
@@ -0,0 +1,260 @@
...
+            self.STATE_BAD_INPUT: self.handle_state_bad_input,
+        }
+        return d
+
+    def session_manager(self, config):
+        return SessionManager.from_redis_config(
+            config.redis_manager,
+            key_prefix=self.worker_name,
+            max_session_length=config.session_expiry
+        )
```
jerith, line 92: This needs to have the router key in the `key_prefix`. Something like `':'.join([self.worker_name, config.router.key])` should do the trick.

jerith:
Pointer to https://github.com/praekelt/vumi-go/pull/844#discussion_r10033326 which is hidden behind an "outdated diff" line.

go/routers/app_multiplexer/vumi_app.py
vgeddes, line 92: Good catch, I've pushed a commit which fixes that.

go/routers/app_multiplexer/vumi_app.py
vgeddes: Ok, I've modified the behavior as per your concerns. The inlined handler
for the `None` sentinel now just does nothing more than clear the
user's session data.

jerith:
:+1: :fireworks: \o/

PR #851, opened 2014-02-07
Message export should include session event flags.

smn:
Currently those aren't included but are useful when dealing with USSD exports.


go/conversation/tasks.py
```diff
@@ -19,12 +19,18 @@
...
     'in_reply_to',
+    'session_event',
+    'transport_type',
+    'direction',
 ]
 
 
-def write_messages(writer, messages):
+def write_messages(writer, messages, extra=None):
+    extra = extra or {}
```
hodgestar, line 29: Vague preference for `extra = extra if extra is not None else {}` since it seems more correct (i.e. `[]`, `0`, etc are not treated as the default).

hodgestar:
Other than the tiny bikeshed, :+1:.

Travis failure looks unrelated.

go/conversation/tasks.py
smn, line 29: This is all disappearing in #852

PR #856, opened 2014-02-10
Pass DiamondashApiErrors on to the client and log them rather than throwing an raw exception up to Django.

hodgestar:
Currently if Diamondash responds with an error, Vumi Go let's the exception bubble all the way up to Django's handlers rather than logging it and passing the error on to the client.

This is an issue when Diamondash is restarted and clients need to detect a need to refresh the page they're on.

A long term solution is probably to teach Diamondash about remote sources of dashboard config.

### DIFF
--- a/go/dashboard/client.py
+++ b/go/dashboard/client.py
@@ -9,6 +9,10 @@ class DiamondashApiError(Exception):
     Raised when we something goes wrong while trying to interact with
     diamondash api.
     """
+    def __init__(self, code, content, message):
+        super(DiamondashApiError, self).__init__(message)
+        self.code = code
+        self.content = content
 
 
 class DiamondashApiClient(object):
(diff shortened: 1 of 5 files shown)


hodgestar:
Ready for review.

go/dashboard/client.py
```diff
@@ -9,6 +9,10 @@ class DiamondashApiError(Exception):
     Raised when we something goes wrong while trying to interact with
     diamondash api.
     """
+    def __init__(self, code, content, message):
+        super(DiamondashApiError, self).__init__(message)
+        self.code = code
+        self.content = content
```
justinvdm, line 15: Should we maybe create `message` from the `code` and `content`?

justinvdm, line 15: As a default maybe?

justinvdm:
Looks good, just a small comment.

hodgestar:
Ready for re-review.

justinvdm:
:+1:

PR #857, opened 2014-02-10
message exports should include endpoints.

smn:


smn:
ready for review

hodgestar:
:+1:

vgeddes:
:+1: Looks good. That would be useful for BB where we do both USSD and SMS in the same conv.

PR #860, opened 2014-02-12
Point to the cache for metric collection

hodgestar:
and turn the metrics we had deactivated due to performance issues caused by not pointing to the cache.


justinvdm:
:+1:

smn:
:cake:

PR #863, opened 2014-02-12
Change credit definition and credit factor

hodgestar:
Currently a credit is defined as 4 (Note: in Vumi, billing costs are represented in US cents. So 4 cents is represented as 4, not .04)

Please change the credit definition to .1 (i.e. one tenth of one cent)

We'll also need to change the credit factor from 0.25 to 10 (i.e. such that a cost of 2.6 cents is reflected as 26 credits)


hodgestar:
@badlard Ready for review.

badlard:
:+1:

PR #864, opened 2014-02-13
Log messages should go to Sentry, not our inbox

justinvdm:
This may just be a configuration thing, but making a ticket anyways.

### DIFF
--- a/go/settings.py
+++ b/go/settings.py
@@ -212,6 +212,10 @@ LOGGING = {
         },
     },
     'handlers': {
+        'sentry': {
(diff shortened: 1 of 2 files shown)


justinvdm:
No idea if this will work, but ready for review. We should also maybe wait for travis, in case we are now stealing log messages tests are looking for, or something.

justinvdm:
Looks like travis is happy, probably can be reviewed now without the worry of new commits being added.

hodgestar:
:+1:

PR #865, opened 2014-02-14
500 error when searching message InBox

hodgestar:
I'm getting a 500 error when I conduct a search on the messages page in Vumi Go.

screen shot 2014-02-13 at 12 38 22 pm


badlard:
@hodgestar can we get this one over the finish line? I've got another project requiring keyword search across messages.

smn:
To do this properly we'd do a migration of the message store and change how we store messages. Currently it is impossible to use Riak's built-in full text search capabilities because we're using full stops in the JSON field names.

The current implementation does a map/reduce with a regex which has performance problems.

vgeddes:
Now I just need to sort out some JQuery / Ajax weirdness in the `message_list_table_load.html` template.

Particularly the following javascript error was logged in my FF developer console: `ReferenceError: $ is not defined`

smn:
Jquery probably isn't included in the page you're loading.

vgeddes:
The ajax is working now, but now I need to sort out https://github.com/praekelt/vumi/issues/748, which is blocking this PR.

badlard:
Cool - thx for the update and for working on it.

On 20 March 2014 17:06, Vincent Geddes notifications@github.com wrote:

> The ajax is working now, but now I need to sort out praekelt/vumi#748https://github.com/praekelt/vumi/issues/748,
> which is blocking this PR.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/praekelt/vumi-go/pull/865#issuecomment-38178041
> .

## 

_PRAEKELTFOUNDATION_.ORG

jerith:
I don't really like the patching in `MessageStoreClientHelper`. Given the scope of the interactions involved, I'd be much happier if the tests actually ran a real message store API and populated the store with appropriate messages.

If we're going to build a fake for this, it needs to live near the real thing and be tested alongside it. Also, it should operate at the HTTP level rather than replacing the client library, because we still need to test that we're calling the client library correctly.

hodgestar:
Closing this since the relevant functionality has been removed and we'll likely want to start a new implementation from scratch.

PR #870, opened 2014-02-18
Add management command for reconciling conversation message caches.

hodgestar:


go/base/management/commands/go_manage_message_cache.py
```diff
@@ -0,0 +1,68 @@
...
+            conv = user_api.get_conversation(self.options['conversation_key'])
+            batches.add(conv.batch.key)
+        if self.options.get('active_conversations'):
+            batches.update(
+                conv.batch.key for conv in user_api.active_conversations())
+        return list(batches)
+
+    def _apply_to_batches(self, func, dry_run=None):
+        if dry_run is None:
+            dry_run = self.options.get('dry_run')
```
jerith, line 46: This is only called in one place. Why not make `dry_run` a mandatory parameter?

go/base/management/commands/go_manage_message_cache.py
```diff
@@ -0,0 +1,68 @@
...
+
+        for user, user_api in self._get_user_apis():
+            batches = self._get_batches(user_api)
+            if not batches:
+                continue
+            self.stdout.write(
+                "Processing account %s ...\n" % user_details_as_string(user))
+            for batch_id in sorted(batches):
+                self.stdout.write(
+                    "  Preforming %s on batch %s ...\n"
```
jerith: Tyop: "Preforming" -> "Performing".

jerith:
The tests don't actually test that the reconcile happens, only that the expected output is generated.

go/base/management/commands/go_manage_message_cache.py
hodgestar, line 46: Mostly I was trying to avoid doing `self.options.get('dry_run')` everywhere that calls this. Maybe I should just remove the `dry_run` parameter entirely?

hodgestar:
Ready for re-review.

jerith:
:+1: when Travis is happy.

PR #876, opened 2014-02-21
npm post install hook exits unhappily when the symlinks already exist

justinvdm:
This means that a `fab js_setup_env` will stop after doing its thing for the first host.

### DIFF
--- a/utils/npm-post-install.sh
+++ b/utils/npm-post-install.sh
@@ -3,9 +3,13 @@
 BINS='grunt bower yuglify'
 PROJECT_ROOT=$(cd "`dirname $0`/.."; pwd)
 
-if [ -n $VIRTUAL_ENV ]; then
-  for bin in $BINS
-  do
-    ln -s $PROJECT_ROOT/node_modules/.bin/$bin $VIRTUAL_ENV/bin/$bin
-  done
-fi
+function make_links {
+    if [ -n $VIRTUAL_ENV ]; then
+        for bin in $BINS
+        do
+            ln -s $PROJECT_ROOT/node_modules/.bin/$bin $VIRTUAL_ENV/bin/$bin
+        done
+    fi
+}
+
+make_links || :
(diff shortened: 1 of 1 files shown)


utils/npm-post-install.sh
```diff
@@ -3,9 +3,13 @@
...
+function make_links {
+    if [ -n $VIRTUAL_ENV ]; then
+        for bin in $BINS
+        do
+            ln -s $PROJECT_ROOT/node_modules/.bin/$bin $VIRTUAL_ENV/bin/$bin
+        done
+    fi
+}
+
+make_links || :
```
justinvdm: Doing this instead of `exit 0` so we can add onto the script (we will probably want the script to exit early for those additions)

justinvdm:
Ready for review.

utils/npm-post-install.sh
```diff
@@ -3,9 +3,15 @@
...
-    ln -s $PROJECT_ROOT/node_modules/.bin/$bin $VIRTUAL_ENV/bin/$bin
-  done
-fi
+function make_links {
+    if [ -n $VIRTUAL_ENV ]; then
+        for bin in $BINS
+        do
+            ln -s $PROJECT_ROOT/node_modules/.bin/$bin $VIRTUAL_ENV/bin/$bin
+        done
+    fi
```
hodgestar: I'd like to suggest we structure this a little differently and write a function which just does:

```
ln -s $PROJECT_ROOT/node_modules/.bin/$bin $VIRTUAL_ENV/bin/$bin || echo "Skipping $bin, symlink already exists."
```

(then we can call that for each bin in $BINS).

justinvdm: +1, will do

justinvdm:
Ready for re-review.

utils/npm-post-install.sh
```diff
@@ -3,9 +3,16 @@
 BINS='grunt bower yuglify'
 PROJECT_ROOT=$(cd "`dirname $0`/.."; pwd)
 
+function link {
+    src="$PROJECT_ROOT/node_modules/.bin/$bin"
```
hodgestar: `$bin` -> `$1`?

justinvdm: Woops, thanks :)

hodgestar:
:+1:

PR #879, opened 2014-02-24
Expose Wikipedia config options in the UI.

jerith:
The Wikipedia app has a number of config options, some of those need to be exposed in the Go UI, especially `include_url_in_sms` and `mobi_url_host`.

### DIFF
--- a/go/apps/wikipedia/view_definition.py
+++ b/go/apps/wikipedia/view_definition.py
@@ -1,5 +1,32 @@
-from go.conversation.view_definition import ConversationViewDefinitionBase
+from django import forms
+
+from go.conversation.view_definition import (
+    ConversationViewDefinitionBase, EditConversationView)
+
+
+class ConfigForm(forms.Form):
+    api_url = forms.CharField(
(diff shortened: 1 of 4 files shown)


go/apps/wikipedia/view_definition.py
```diff
@@ -1,5 +1,43 @@
...
+        if data['api_url']:
+            config_dict['api_url'] = data['api_url']
+        if data['mobi_url_host']:
+            config_dict['mobi_url_host'] = data['mobi_url_host']
+        return config_dict
+
+
+class EditWikipediaView(EditConversationView):
+    edit_forms = (
+        (None, ConfigForm),
```
hodgestar, line 38: Cool trick. I don't think I was aware this existed.

jerith, line 38: It's supposed to be the default way to do things. I don't think it cooperates nicely with prefixed forms/formsets, though. We should probably fix that (if necessary) and standardise on non-prefixed config forms.

miltontony:
:+1:

PR #869, opened 2014-02-24
Add session cost option to Vumi billing

hodgestar:
We need to account for USSD costs that are billed on a per session basis.

Please add a 'session cost' to the 'message costs' section in /admin that will enable a site admin to apply a session cost to a tagpool (and account if required)

This change must also be reflected in the monthly statements that are generated. i.e. the statements must show the number of sessions and the associated cost, along with the per message cost.

Note: a tagpool (or account) can have both a session cost and a per message cost.

screen shot 2014-02-18 at 2 49 04 pm


hodgestar:
Our transaction table currently doesn't include the message id. I've created #883 for that.

go/billing/admin.py
```diff
@@ -99,16 +100,18 @@ def credit_amount(self, obj):
 
 class TransactionAdmin(admin.ModelAdmin):
     list_display = ('id', 'account_number', 'tag_pool_name', 'tag_name',
-                    'message_direction', 'message_cost', 'markup_percent',
-                    'credit_factor', 'credit_amount', 'status', 'created',
-                    'last_modified')
+                    'message_direction', 'message_cost',
+                    'session_created', 'session_cost',
+                    'markup_percent', 'credit_factor', 'credit_amount',
+                    'status', 'created', 'last_modified')
```
justinvdm, line 103: I think we maybe need a docstring for these terms somewhere. Seems like a lot of new terminology for everyone. Thoughts?

justinvdm, line 103: Not too relevant to the PR, though. Maybe we need a ticket or something for documenting the billing terminology?

hodgestar, line 103: What about documenting them on the model? We might be able to get those
displayed in the admin interface somehow too.

justinvdm, line 103: +1 to putting it on the model, and +1 for being able to re-use that in the UI somehow.

justinvdm:
+1 once we have a test for checking that `session_created` is set correctly (like the TODO says).

I think I'm missing some context to notice things that could be dangerous, but I do have a slightly better idea of the billing stuff now.

go/billing/admin.py
hodgestar, line 103: I created #884 for this. `help_text` is displayed in auto-generated forms so it should appear in the admin forms automatically.

hodgestar:
I created #885 for handling the test TODOs that aren't directly related to this PR.

hodgestar:
I also need to write a test for `calculate_credit_cost` and fix the `credit_cost` and `resulting_price` properties on the `MessageCost` model.

hodgestar:
Ready for re-review.

go/billing/forms.py
```diff
@@ -20,21 +26,26 @@ def clean(self):
...
-
             context = Context()
-            credit_cost = credit_cost.quantize(
-                app_settings.QUANTIZATION_EXPONENT, context=context)
-
-            if context.flags[Inexact] and credit_cost == Decimal('0.0'):
-                raise forms.ValidationError("The resulting credit cost is 0.")
+            credit_cost = MessageCost.calculate_credit_cost(
+                message_cost, markup_percent,
+                Decimal('0.0'), session_created=False, context=context)
```
justinvdm: I'm wondering if it would make sense to have the different `calculate_credit_cost` args default to `Decimal('0.0')`. Thoughts?

hodgestar: I've been trying to force people to call this function with exactly the
arguments they want to use because leaving out an argument and getting a
default is potentially bad. It is all a bit clunky right now though.

justinvdm: True, for things like costing, you want to know all the values.

justinvdm: Maybe we should put `session_cost=Decimal('0.0')`, and similarly for below (for clarity, so people know what is being put to 0)? Not too important, happy for it staying as is.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -18,11 +18,16 @@
...
+        self.transactions = []
+
+    def _record(self, items, vars):
+        del vars["self"]
+        items.append(vars)
 
     def create_transaction(self, account_number, tag_pool_name,
-                           tag_name, message_direction):
+                           tag_name, message_direction, session_created):
+        self._record(self.transactions, locals())
```
justinvdm, line 30: Where is `self.transactions` used? Or is it yet to be used?

justinvdm, line 30: Ah, sorry, I see where it is used.

justinvdm, line 30: Recording the locals feels a little weird. Any reason to not just explictly give it all the args?

justinvdm, line 30: Actually, I'm bikeshedding, happy with it staying as is.

go/billing/tests/test_api.py
```diff
@@ -25,6 +25,10 @@
     reason="Billing API requires PostGreSQL")
 
 
+# TODO: factor out common setup
+# TODO: factor out account setup method
+# TODO: factor out cost setup method
```
justinvdm, line 30: Sorry for not asking in the first review pass, but do we have tickets for all of these?

justinvdm:
+1 if we have tickets for the TODOs. Also left other minor comments, but not important enough to stop this from landing.

Not sure how well I reviewed since this stuff is new to me, but I'll get better at it soon :)

go/billing/tests/test_api.py
hodgestar, line 30: Yep -- all three are meant to be covered in #885.

justinvdm:
:+1:

hodgestar:
Ready for hopefully-final review. :)

justinvdm:
:+1:

PR #881, opened 2014-02-26
Set a default message cost on all tagpools

hodgestar:
Set a cost of 0.1 cents for all incoming and outgoing messages for all tagpools.

### DIFF
--- a/go/billing/api.py
+++ b/go/billing/api.py
@@ -386,8 +386,8 @@ class CostResource(BaseResource):
             message_cost = data.get('message_cost', None)
             session_cost = data.get('session_cost', None)
             markup_percent = data.get('markup_percent', None)
-            if (message_direction and message_cost
-                    and session_cost and markup_percent):
+            if all((message_direction, message_cost, session_cost,
+                    markup_percent, tag_pool_name or not account_number)):
                 d = self.create_cost(account_number, tag_pool_name,
                                      message_direction, message_cost,
                                      session_cost, markup_percent)
--- a/go/billing/api.py
+++ b/go/billing/api.py
@@ -386,8 +386,8 @@ class CostResource(BaseResource):
             message_cost = data.get('message_cost', None)
             session_cost = data.get('session_cost', None)
             markup_percent = data.get('markup_percent', None)
-            if message_direction and message_cost \
-                    and session_cost and markup_percent:
+            if (message_direction and message_cost
+                    and session_cost and markup_percent):
                 d = self.create_cost(account_number, tag_pool_name,
                                      message_direction, message_cost,
                                      session_cost, markup_percent)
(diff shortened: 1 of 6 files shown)


hodgestar:
Ready for review.

go/billing/api.py
```diff
@@ -386,7 +386,7 @@ def render_POST(self, request):
             message_cost = data.get('message_cost', None)
             session_cost = data.get('session_cost', None)
             markup_percent = data.get('markup_percent', None)
-            if tag_pool_name and message_direction and message_cost \
+            if message_direction and message_cost \
                     and session_cost and markup_percent:
```
jerith: Can you use parens rather than backslash line continuation here please?

hodgestar: Happy too. I initially left it as it was to minimize the diff and make it clear what was changing.

justinvdm:
Judging by the tests, things looking like they are working properly, so looks good. I know very little SQL though, so I can't really give a confident +1.

hodgestar:
I've realized we're not ignoring the case where a message cost has an account set but no tagpool (currently such a message cost row would never be used in a transaction). Since it's not clear whether `(account NULL, tagpool NOT NULL)` should win over `(account NOT NULL, tagpool NULL)`, I propose that we just make it illegal to add rows of the latter form?

@badlard ^^^

badlard:
+1

justinvdm:
@hodgestar so one can add costs where the account is null, but the tag poll isn't, but cannot add costs where the account is not null, but the tagpool is?

hodgestar:
@justinvdm Yep, that's the plan.

justinvdm:
Ok cool. Do we have a way of sending an api failure response if the account is not given by the tagpool is?

go/billing/api.py
```diff
@@ -621,7 +624,14 @@ def get_cost(self, account_number, tag_pool_name, message_direction,
...
                   AND t.name = %(tag_pool_name)s
-                  AND c.message_direction = %(message_direction)s) t
+                  AND c.message_direction = %(message_direction)s
+                  UNION
+                  SELECT NULL AS account_number, NULL AS tag_pool_name,
+                         c.message_direction, c.message_cost,
+                         c.session_cost, c.markup_percent
+                  FROM billing_messagecost c
+                  WHERE c.account_id IS NULL
+                  AND   c.tag_pool_id IS NULL) as t
```
jerith: The new query doesn't filter on `message_direction`. Is this intended?

(Thanks to some excellent people on the internet who pointed that out when I asked them to lend me their SQL-fu.)

hodgestar:
Ready for re-review.

justinvdm:
:+1: if people are happy with the sql query changes landing in its current state.

justinvdm:
:+1:

PR #885, opened 2014-02-27
Clean-up billing API tests

hodgestar:
Currently there's a lot of duplicated code and a lot of things being tested in individual tests.

### DIFF
--- a/go/billing/tests/test_api.py
+++ b/go/billing/tests/test_api.py
@@ -25,13 +25,10 @@ skipif_unsupported_db = pytest.mark.skipif(
     reason="Billing API requires PostGreSQL")
 
 
-# TODO: factor out common setup
-# TODO: factor out account setup method
-# TODO: factor out cost setup method
+@skipif_unsupported_db
+@pytest.mark.django_db
+class BillingApiTestCase(VumiTestCase):
 
-class TestUser(VumiTestCase):
-
-    @pytest.mark.django_db
     @defer.inlineCallbacks
     def setUp(self):
         connection_string = app_settings.get_connection_string()
@@ -42,14 +39,18 @@ class TestUser(VumiTestCase):
         root = api.Root(connection_pool)
         self.web = DummySite(root)
 
-    @pytest.mark.django_db
     def tearDown(self):
         self.connection_pool.close()
 
-    @skipif_unsupported_db
-    @pytest.mark.django_db
+
+# TODO: factor out account setup method
+# TODO: factor out cost setup method
+
+
+class TestUser(BillingApiTestCase):
+
     @defer.inlineCallbacks
-    def runTest(self):
+    def test_user(self):
         # Create a new user
         content = {
             'email': "test@example.com",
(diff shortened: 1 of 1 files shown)


hodgestar:
This needs to be done before #881 because that really needs to add more transaction creation API tests and those are a bit of a mess right now.

hodgestar:
Ready for review.

hodgestar:
Would really like to split these into separate tests, but that requires solving the issue of how to wrap the entire connection in a transaction and roll it back at the end of the test (essentially duplicating what Django does in its tests). That's probably best left for a separate PR -- things are already a lot cleaner.

hodgestar:
Alternatively we could wrap the tests in `django_db(transaction=True)` at the cost of making the tests **really** slow (because the database has to be deleted and recreated before each test).

hodgestar:
An acceptable hack might be to clear all the relevant tables at the end of each test?

smn:
I'm not sure Django creates & deletes the db for every test, I thought it just ran it in a single transaction which it then rolled back.

smn:
but that may be _entirely_ irrelevant to this PR so ignore me for now.

hodgestar:
Django doesn't by default, but one can ask it too. If we rely on Django's
transaction rollback that doesn't clean up the database because the changes
made by the billing api happen in their own connection.

smn:
thanks :cake:

go/billing/tests/test_api.py
```diff
@@ -25,218 +25,211 @@
...
-
-        self.assertEqual(response.responseCode, 200)
-        new_user = json.loads(response.value(), cls=JSONDecoder)
-        self.assertTrue('id' in new_user)
-
-        # Fetch the new user
-        url = 'users/%s' % (new_user.get('id'))
-        response = yield self.web.get(url)
+        http_method = getattr(self.web, method)
+        response = yield http_method(path, headers=headers, **kw)
```
justinvdm, line 48: Would there ever be a case where there were other headers besides `{'content-type': 'application/json'}`?

go/billing/tests/test_api.py
```diff
@@ -25,218 +25,211 @@
...
-        url = 'accounts/%s' % (new_account.get('account_number'))
-        response = yield self.web.get(url)
-        self.assertEqual(response.responseCode, 200)
-        account = json.loads(response.value(), cls=JSONDecoder)
-        self.assertEqual(account.get('account_number'),
-                         new_account.get('account_number'))
-        self.assertTrue(account.get('credit_balance', 0) == 100)
+        # test retrieving all users
+        user_list = yield self.get_api_user_list()
+        self.assertEqual(user_list, [new_user])
```
justinvdm, line 184: Testing retrieving all users feels like it should be in a separate test. A bit of a bikeshed though, so your call.

justinvdm:
It seem seem like most of the test cases are testing a few things inside a single test. Any reason to keep it that way? Sorry, I know this is a little bikesheddy. Happy for it to stay as is if it needs to.

go/billing/tests/test_api.py
hodgestar, line 48: Not with the API as it stands (it's meant to be all JSON over HTTP).

justinvdm, line 48: Ok cool.

go/billing/tests/test_api.py
hodgestar, line 184: Totally agree. I wanted to do that initially but it runs into problems with
how the database is managed (see discussion I had on this PR this with
@smn). Would like to solve those, but not in this PR (which unfortunately
is thus limited to a smaller set of cleaning up).

justinvdm, line 184: Ah, ok, np then. Its at least quite a bit cleaner and readable now, so that can wait.

hodgestar, line 184: I created #886 for splitting up the tests.

justinvdm:
:+1: I tried check if we are missing anything important from the assertions in the original tests, but can't seem to find anything. Much cleaner now :)

justinvdm:
:+1:

PR #887, opened 2014-02-27
Billing API doesn't list costs without accounts

hodgestar:


hodgestar:
Ready for review.

justinvdm:
Would it make sense to add a test for cases where we have a tag pool, but no account?

justinvdm:
Otherwise, looks good.

hodgestar:
@justinvdm It would. The `base_cost` defined at the start of the test is such a message cost entry (it's just not that evident from the diff).

justinvdm:
:+1:

PR #888, opened 2014-02-28
Create billing accounts when user accounts are created

hodgestar:
Billing people requires that they have a billing account. Currently no one has them. We need:
- To create them automatically when accounts are created.
- To add a management commands to create them for all the accounts that currently don't have them.

### DIFF
--- a/go/base/tests/test_go_manage_account.py
+++ b/go/base/tests/test_go_manage_account.py
@@ -91,6 +91,7 @@ class TestGoManageAccount(GoCommandTestCase):
         expected_output = "\n".join([
             u'Performing create_billing_account on account'
             u' Test User <user@domain.com> [test-0-user] ...',
+            u'  Billing account already exists.',
             u'done.',
             u''
         ])
--- a/go/base/tests/test_go_manage_account.py
+++ b/go/base/tests/test_go_manage_account.py
@@ -70,7 +70,8 @@ class TestGoManageAccount(GoCommandTestCase):
             'dummy')
 
     def test_billing_account_created(self):
-        self.assertEqual(self.get_billing_account(), None)
+        account = self.get_billing_account()
+        account.delete()
         expected_output = "\n".join([
             u'Performing create_billing_account on account'
             u' Test User <user@domain.com> [test-0-user] ...',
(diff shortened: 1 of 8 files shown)


hodgestar:
Ready for review.

jerith:
Do we unhook billing account creation in test helpers? Should we?

hodgestar:
We don't unhook it and I think we don't need to. We only disconnect
`create_user_profile` so we can catch and override the riak account
creation (and skip it if the test isn't using riak).

jerith:
> We don't unhook it and I think we don't need to. We only disconnect
> `create_user_profile` so we can catch and override the riak account
> creation (and skip it if the test isn't using riak).

That's reasonable. Thanks.

go/base/tests/test_go_manage_account.py
```diff
@@ -0,0 +1,103 @@
...
+        account = self.get_billing_account()
+        self.assertEqual(account.user.email, self.user_email)
+        self.assertEqual(account.account_number, self.user_helper.account_key)
+
+    def test_billing_account_already_exists(self):
+        initial_account = self.get_billing_account()
+        expected_output = "\n".join([
+            u'Performing create_billing_account on account'
+            u' Test User <user@domain.com> [test-0-user] ...',
+            u'done.',
```
justinvdm, line 95: Should we maybe say something about the account not being created since it already exists?

go/billing/models.py
```diff
@@ -38,6 +41,16 @@ def __unicode__(self):
...
+def create_billing_account(sender, instance, created, **kwargs):
+    if created:
+        Account.objects.create(user=instance.user,
+                               account_number=instance.user_account)
+
+
+post_save.connect(create_billing_account, sender=UserProfile,
+    dispatch_uid='go.billing.models.create_billing_account')
+
+
```
justinvdm, line 53: Do we have a test for this post-save hook?

justinvdm:
Looks good, just a few small comments.

hodgestar:
Ready for re-review.

justinvdm:
:+1:

PR #895, opened 2014-03-18
Test methods missing @inlineCallbacks decorator

jerith:
Based on praekelt/vumi#741 I found the following test methods that `yield` but aren't decorated with `@inlineCallbacks`:

TestGroupsResource.test_handle_count_members
TestApplicationMultiplexerRouter.test_create_menu
TestTxVumiRouterApi.test_get_router
TestVumiRouterApi.test_get_router


smn:
:+1:

PR #898, opened 2014-03-20
Support 'last' aggregation method in sandbox metrics resource

justinvdm:
The toolkit supports it, but not the actual resource. Supporting it should be simple.


justinvdm:
Ready for review.

smn:
:+1:

PR #899, opened 2014-03-26
TypeError when unroutable message error is logged

smn:
Patch:

``` diff
diff --git a/go/vumitools/routing.py b/go/vumitools/routing.py
index c5fd444..99b5c38 100644
--- a/go/vumitools/routing.py
+++ b/go/vumitools/routing.py
@@ -679,7 +679,7 @@ class AccountRoutingTableDispatcher(RoutingTableDispatcher, GoWorkerMixin):
         if target is None:
             raise UnroutableMessageError(
                 "No target found for outbound message from '%s': %s" % (
-                    connector_name, msg))
+                    connector_name, msg), msg)

         if self.billing_outbound_connector:
             target_conn = GoConnector.parse(target[0])
```


hodgestar:
:+1:

PR #901, opened 2014-03-26
Add support for wechat  & mxit fields on the contact model.

smn:


smn:
Ready for review.

smn:
Travis seems to be stuck.

jerith:
Don't we need a new model version and migration for the new fields on `Contact`?

smn:
no because they default to `null`.

jerith:
:+1:, but we _really_ need to find a better way to handle address types.

smn:
@jerith yeah, agreed, thanks.

PR #906, opened 2014-04-01
Update to backbone 1.1.2

justinvdm:
We need to upgrade to Backbone 1.1.2 to keep Backbone Relational happy.

It appears `this.options` isn't available in Backbone 1.1.2, so we need to fix that.

We could always postpone the upgrade, if people would prefer that, but I think we should try stay up to date.


justinvdm:
Ready for review.

hodgestar:
:+1: Yay for tests.

PR #909, opened 2014-04-03
Implement a group router

vgeddes:
As per the functional spec, the idea is route messages based on a user's membership or non-membership in a set of contact groups.


smn:
We may initially want to say this'll only work for static groups since detecting membership in smart groups is likely quite expensive.

jerith:
Ready for review.

hodgestar:
@justinvdm volunteered to review this. Tx.

go/routers/group/vumi_app.py
```diff
@@ -0,0 +1,52 @@
...
+
+    worker_name = 'group_router'
+
+    def endpoint_for_contact(self, config, contact):
+        if contact is None:
+            return 'default'
+        contact_groups = set(contact.groups.keys())
+        for rule in config.rules:
+            if rule['group'] in contact_groups:
+                return rule['endpoint']
```
justinvdm, line 32: I'm wondering if we should be validating the `rules` config field for cases where `group` or `endpoint` aren't given. Not very important though.

go/routers/group/tests/test_vumi_app.py
```diff
@@ -0,0 +1,99 @@
...
+        groups = [(yield self.router_helper.create_group(u"group%s" % i))
+                  for i in range(10)]
+        contact = yield self.router_helper.create_contact(
+            u"+27831234567", groups=groups[2:])
+        router = yield self.router_helper.create_router(started=True, config={
+            'rules': [
+                {'group': group.key, 'endpoint': '%s_ep' % group.name}
+                for group in groups
+            ]})
+        yield self.assert_routed_inbound(contact.msisdn, router, 'group2_ep')
```
justinvdm, line 99: I'm a bit confused that `group2_ep` matches, since the groups are sorted by name I'd expect `group1_ep` to match. Could you explain?

go/routers/group/tests/test_vumi_app.py
```diff
@@ -0,0 +1,99 @@
...
+    def test_inbound_contact_in_multiple_groups(self):
+        """
+        If the message is from a contact in multiple configured groups, it
+        should be dispatched to the endpoint in the first matching rule.
+        """
+        # We use lots of groups here to decrease the false positive chance.
+        groups = [(yield self.router_helper.create_group(u"group%s" % i))
+                  for i in range(10)]
+        contact = yield self.router_helper.create_contact(
+            u"+27831234567", groups=groups[2:])
```
justinvdm, line 93: I'm not getting why we create 10 groups if we only add the contact to the first two, could you explain?

go/routers/group/vumi_app.py
```diff
@@ -0,0 +1,52 @@
...
+        return 'default'
+
+    @inlineCallbacks
+    def handle_inbound(self, config, msg, conn_name):
```
justinvdm, line 36: It looks like `handle_inbound` and `handle_outbound` are given the connector name as a param, but `GoRouterWorker`'s signatures for these methods looks like `def handle_outbound(self, config, msg)`. Not sure if it makes sense to fix that in this PR, but we should maybe fix that while we are it?

justinvdm:
Looks good, just some questions.

justinvdm:
Just a reminder that this has been reviewed.

go/routers/group/vumi_app.py
jerith, line 32: We need better config validation for all apps/routers/whatever, but we don't yet have the tools. For now, we'll have to rely on the UI code building a valid config.

go/routers/group/tests/test_vumi_app.py
jerith, line 99: We skip the first two groups when we add the groups to the contact (line 93 above) so only groups 2 through 9 are added.

justinvdm, line 99: Ah, right.

go/routers/group/tests/test_vumi_app.py
jerith, line 93: We add all but the first two groups. There are eight matching rules to reduce the false positive rate of unordered rule selection to one in eight. Not fantastic, but more groups means more mucking about in Riak means slower tests.

go/routers/group/vumi_app.py
jerith, line 36: Good catch, thanks.

go/routers/group/tests/test_vumi_app.py
justinvdm, line 93: Sorry, read the slicing wrong here. Still confused about why we do the slicing though, do we do it to test that we don't use a group that the contact is not part of?

jerith, line 93: Correct. We're testing that we choose the _first_ rule that _matches_. Both parts are important. :-)

justinvdm, line 93: Thanks, makes sense :) A different question: what do you mean by 'unordered rule selection', the fact that the rules are a dict so they do not have an order?

jerith, line 93: An earlier version of the code put the rules in a dict keyed by group, so they were checked in arbitrary order. The current version leaves them in the list from the config, so the ordering is preserved.

justinvdm:
:+1: once travis is happy.

PR #907, opened 2014-04-03
Add option to trigger push of sandbox messages

hodgestar:


hodgestar:
@jerith, @justinvdm -- I think this is ready for review. It still needs an extra jsbox vumi_app.py test and some tests for the new outbound resource functionality, but all the logic is in place and tested via the dialogue tests.

hodgestar:
I'll be adding the remaining tests later today.

go/apps/jsbox/outbound.py
```diff
@@ -6,9 +6,49 @@
...
+    Construct a dummy inbound message used to trigger a push of
+    a new message from a sandbox application.
+    """
+    msg_options = {
+        'transport_name': None,
+        'transport_type': None,
+        'helper_metadata': {},
+        # mark this message as special so that it can be idenitified
+        # if it accidentally ends up elsewhere.
+        INBOUND_PUSH_TRIGGER: True,
```
justinvdm, line 27: Just wondering if this marker belongs in `helper_metadata`. Happy either way, just wanted to bring it up.

justinvdm:
Looks good. Happy with this landing once we have the missing jsbox vumi app test and a test for the outbound resource's handling of trigger messages.

go/apps/jsbox/outbound.py
hodgestar, line 27: I wondered about that too. The main advantage would be that `INBOUND_PUSH_TRIGGER` would end up in the helper_metadata too. The main disadvantage is that it looks more like a normal Vumi message (which it really isn't).

hodgestar:
Tests done, I think.

justinvdm:
:+1: Not sure if its worth waiting for travis.

hodgestar:
Worth waiting -- I'll land in the morning.

PR #910, opened 2014-04-04
Add name config field for dialogue apps

justinvdm:
We need to include a name config field for dialogue apps, since this is compulsory in v2.


justinvdm:
Ready for review.

justinvdm:
Ready for review.

hodgestar:
:+1:

PR #913, opened 2014-04-08
Fix poll loading that broke after migration to sandbox toolkit v2.

hodgestar:
The change needed is:

``` diff
diff --git a/go/apps/dialogue/vumi_app.js b/go/apps/dialogue/vumi_app.js
index 63c6ad3..f6c1295 100644
--- a/go/apps/dialogue/vumi_app.js
+++ b/go/apps/dialogue/vumi_app.js
@@ -32,10 +32,10 @@ var DialogueApp = App.extend(function(self) {

     self.get_poll = function() {
         return self
-            .im.sandbox_config.get('poll', {json: true})
+            .im.sandbox_config.get('poll', {json: false})
             .then(function(poll) {
                 return _.defaults(poll, self.poll_defaults);
-           });
+            });
     };

     self.add_state = function(desc) {
```

but we also need to fix tests.

### DIFF
--- a/go/apps/dialogue/tests/dummy_polls.py
+++ b/go/apps/dialogue/tests/dummy_polls.py
@@ -41,10 +41,6 @@ simple_poll = {
             # end specific
             "text": "Thank you for taking our survey",
         },
-        {
-            "uuid": "unknown-1",
-            "type": "unknown-or-garbled",
-        },
     ],
     "connections": [
         {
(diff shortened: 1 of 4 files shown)


hodgestar:
Ready for review.

justinvdm:
:+1: not sure if waiting for Travis is worth it? Your call.

Issue #914, opened 2014-04-07
Clean up API object layer


jerith:
This is a collection of work that needs to happen on the abstraction layer between various core Vumi Go functionality and the code that uses it. The idea is to open tickets individually for pieces of work referenced here (probably, but not necessarily, one branch per checklist item) and then close this one when everything's checked off.

`go.vumitools.api`:
- [ ] Rearrange `VumiUserApi` code to group related methods together.
- [ ] Split collections of methods into their own classes.
- [ ] Add metrics functionality. (Required by #822.)
- [ ] Break module up into package containing submodules.

`go.vumitools.conversation.utils`:
- [ ] Replace `ConversationWrapper` with API-style object.
- [ ] Split message store functionality into a new class.

misc:
- [ ] Move metrics manager setup from worker to API objects.

There's probably more, so we can reassess this ticket as necessary.

PR #822, opened 2014-04-03
Service Component infrastructure

jerith:
A Service Component is a collection of UI and backend code that provides some kind of service to other things within Vumi Go. It has a similar structure to a Conversation/Application, but has somewhat more freedom in both interface and implementation.

Some things that would work well as Service Components:
- airtime voucher management (using `airtime-service`)
- key-value store (potentially multiple variants, backed by different kinds of storage)
- scheduler (would need more hooks into other bits of Go)

Each component is backed by a Riak model in the same way a conversation is and probably has some kind of view definition along the same lines as a conversation. There probably needs to be some kind of loadable backend module (like a conversation definition, but with more stuff on it) to allow conversations, routers, jsbox apps, etc. to make use of the component's functionality.

PR #915, opened 2014-04-08
Move metrics manager setup from worker to API

jerith:
Part of #914, required by #822.


jerith:
This depends on praekelt/vumi#761 and builds will fail until that lands.

jerith:
This doesn't actually move the metric manager setup into an API object (yet), but it cleans up all the metrics infrastructure that makes it hard to do that.

go/base/middleware.py
```diff
@@ -60,5 +63,10 @@ def process_response(self, request, response):
 
         response_time = start_time - time.time()
         response['X-Response-Time'] = response_time
-        metric.oneshot(value=response_time)
+        # TODO: Better way to fire these metrics.
+        metrics = MetricManager(
+            get_go_metrics_prefix() + 'django.',
+            publisher=amqp.connection.get_metric_publisher())
```
hodgestar: Can we perhaps set up the MetricManager earlier?

go/dashboard/dashboard.py
```diff
@@ -156,19 +162,18 @@ def __init__(self, conv, entities=None):
     @classmethod
     def aggregator_from_target(cls, target):
         agg_name = target.get('aggregator')
-        return Aggregator.from_name(agg_name) if agg_name is not None else None
+        return Aggregator.from_name(agg_name) if agg_name is not None else AVG
```
justinvdm: I'm wondering if we should have a common place where we specify the default aggregation method for metrics in vumi-go, in case we end up changing the default.

go/vumitools/app_worker.py
```diff
@@ -269,20 +267,39 @@ def trigger_event(self, message, event_type, content):
...
+        # TODO: Move this to an API.
+        prefix = get_conversation_metric_prefix(conv)
+        return MetricManager(prefix, publisher=self.metric_publisher)
+
     def publish_account_metric(self, acc_key, store, name, value, agg=None):
-        metric = AccountMetric(acc_key, store, name, agg)
-        metric.oneshot(self.metrics, value)
+        # TODO: Collect the oneshot metrics and then publish all at once?
+        if agg is None:
+            agg = AVG
```
justinvdm: Same thing as [here](https://github.com/praekelt/vumi-go/pull/915/files#r11987534).

go/dashboard/dashboard.py
jerith: The default should probably come from `vumi.blinkenlights.metrics.Metric`.

justinvdm: Cool, lets maybe do that.

go/apps/tests/helpers.py
```diff
@@ -174,9 +174,13 @@ def dispatch_command(self, command, *args, **kw):
         yield self.dispatch_commands_to_app()
 
     def get_published_metrics(self, worker):
-        return [
-            (metric.name, value)
-            for metric, ((time, value),) in worker.metrics._oneshot_msgs]
+        metrics = []
+        for metric_msg in self.worker_helper.get_dispatched_metrics():
+            for name, _aggs, data in metric_msg:
+                [datapoint] = data
```
hodgestar: Perhaps we should just handle all datapoints? So `for _time, value in data: metrics.append((name, value))`?

go/vumitools/app_worker.py
```diff
@@ -269,20 +267,39 @@ def trigger_event(self, message, event_type, content):
     def publish_app_event(self, event):
         self.app_event_publisher.publish_message(event)
 
+    def get_account_metric_manager(self, account_key, store_name):
+        # TODO: Move this to an API.
+        prefix = get_account_metric_prefix(account_key, store_name)
+        return MetricManager(prefix, publisher=self.metric_publisher)
```
hodgestar: Could we perhaps add a helper method to go.vumitools.metrics and call that from here? I'd just like these bits of logic to be usable from outside app workers.

jerith: Those are going to move to API objects. I just haven't written that code yet.

go/base/middleware.py
hodgestar: I think we should also follow the `get_account_metric_manager` pattern here and do `get_django_metric_manager` here so that things are consistent?

hodgestar:
We need to think a bit what we want to happen with `.get_value` that currently sits on `ConversationMetric`. The metric worker periodically pings conversation workers to calculate metrics for their conversations and for this it's nice to have metric definitions that include how to calculate the value. I'm guessing that we'll soon need the same for account-wide metrics.

jerith:
The new `ConversationMetric` is nothing more than a wrapper around a conversation and a vumi `Metric` object so that it can have a `.get_value()` method. If we need more kinds of periodic metric collection we can change how that works or add more wrappers as appropriate.

go/dashboard/dashboard.py
```diff
@@ -148,27 +147,33 @@ def get_config(self):
...
-        agg_name = target.get('aggregator')
-        return Aggregator.from_name(agg_name) if agg_name is not None else None
+    def aggregator_from_target(self, target):
+        aggregator = target.get('aggregator')
+        # FIXME: We don't always get the aggregator in the target. In order to
+        #        handle this, we default to 'avg' if the aggregator is not
+        #        specified.
+        if aggregator is None:
+            aggregator = 'avg'
+        return aggregator
```
justinvdm, line 168: I think the only metrics being used in production are account metrics, so it'd be great if we could make the change the `FIXME` is talking about as part of this PR.

justinvdm, line 168: Oh, right, we don't make aggregator compulsory at the moment.

justinvdm, line 168: Could we maybe make use of  `vumi.blinkenlights.metrics.Metric` for the default aggregator here instead of hardcoding 'avg'?

go/vumitools/metrics.py
```diff
@@ -150,23 +54,13 @@ def get(self, name):
...
-class ConversationMetricSet(MetricSet):
-    def __init__(self, conv, metrics=None):
-        self.conv = conv
-        super(ConversationMetricSet, self).__init__(metrics)
-
-    def get(self, metric_name):
-        name = ConversationMetric.make_name(self.conv, metric_name)
-        return super(ConversationMetricSet, self).get(name)
-
-
```
justinvdm, line 169: Now that  `ConversationMetricSet` doesn't do anything special with the metric name for `.get`, I'm wondering if there's much use for it existing. Maybe we should remove it?

justinvdm, line 169: (and use a dict instead)

go/vumitools/tests/helpers.py
```diff
@@ -246,12 +246,9 @@ def get_metrics(self):
...
 
-    def publish_metric(self, metric_name, aggregators, value, timestamp=None):
-        metric_msg = MetricMessage()
-        metric_msg.append((metric_name,
-            tuple(sorted(agg.name for agg in aggregators)),
-            [(timestamp, value)]))
-        return self.publish_metric_message(metric_msg)
+    def get_metric_publisher(self):
+        from go.base.amqp import MetricPublisher
+        return MetricPublisher(self)
```
justinvdm, line 250: Why do we do an import here?

go/vumitools/metrics.py
jerith, line 169: I'd prefer not to use the dict directly because any changes to it need to be applied to the list as well.

go/dashboard/dashboard.py
jerith, line 168: We can do that.

go/vumitools/tests/helpers.py
jerith, line 250: To avoid pulling Django into non-Django tests. There are several other places we do this.

go/vumitools/metrics.py
justinvdm, line 169: Ah, why do we care about ordering again (I can't remember if/why we do)?

jerith, line 169: I don't remember myself, but it makes things more deterministic. The abstraction also makes it easier to modify `ConversationMetricSet` later if necessary.

justinvdm, line 169: Fair enough

go/vumitools/app_worker.py
```diff
@@ -265,20 +263,38 @@ def trigger_event(self, message, event_type, content):
...
 
         conv_type = conv.conversation_type
         conv_def = get_conversation_definition(conv_type, conv)
 
-        yield gatherResults([
-            m.oneshot(self.metrics, user_api=user_api)
-            for m in conv_def.get_metrics()])
+        for metric in conv_def.get_metrics():
+            value = yield metric.get_value(user_api)
+            metrics.oneshot(metric.metric, value)
```
justinvdm, line 295: Why was this changes to be synchronous?

jerith, line 295: Primarily because it's easier to do it that way now that the metric object no longer handles its own publishing. We can go back to collecting these concurrently if you prefer, but it'll be more complex.

go/apps/wikipedia/vumi_app.py
```diff
@@ -21,7 +21,7 @@ def validate_config(self):
 
     def _setup_metrics(self, metrics_prefix):
         # We don't want to use the underlying app's metrics.
-        pass
+        self.metrics = None
```
justinvdm, line 24: Where do we use `self.metrics`?

jerith, line 24: We don't, but the worker we're subclassing does.

go/apps/wikipedia/vumi_app.py
```diff
@@ -52,3 +53,8 @@ def send_sms_non_reply(self, msg, config, sms_content):
         return self.send_to(
             msg['from_addr'], sms_content, transport_type='sms',
             endpoint='sms_content', helper_metadata=helper_metadata)
+
+
+def fire_oneshot_metric(metrics, metric, value):
+    metrics.oneshot(metric, value)
+    metrics.publish_metrics()
```
justinvdm, line 60: Why do we publish after every oneshot here, but publish at the end of all the oneshots for conversation metrics?

go/vumitools/app_worker.py
justinvdm, line 295: Na, happy with this, just wondering.

go/vumitools/app_worker.py
```diff
@@ -265,20 +263,38 @@ def trigger_event(self, message, event_type, content):
     def publish_app_event(self, event):
         self.app_event_publisher.publish_message(event)
 
+    def get_account_metric_manager(self, account_key, store_name):
+        # TODO: Move this to an API.
+        prefix = get_account_metric_prefix(account_key, store_name)
+        return self.vumi_api.get_metric_manager(prefix)
+
+    def get_conversation_metric_manager(self, conv):
+        # TODO: Move this to an API.
```
justinvdm, line 272: Any reason why this can't be done now (and similarly for `get_account_metric_manager`)?

go/apps/wikipedia/vumi_app.py
jerith, line 60: It's a hack to make `.get_timer_metric()` work. We don't have anywhere to stash a metric manager for the duration of a request, so we have to make the timer self-sufficient.

go/vumitools/app_worker.py
jerith, line 272: One would have to go on `VumiUserApi` and the other would have to go on `ConversationWrapper`. Cleaning those up is a bigger job and moving these methods now is more likely to confuse than help.

go/apps/wikipedia/vumi_app.py
justinvdm, line 60: Ah, right.

go/vumitools/app_worker.py
justinvdm, line 272: Ok cool, maybe we should make a ticket for it if we don't have one already.

justinvdm:
Just the change for getting the default aggregator from `Metric`, otherwise :+1:

go/vumitools/app_worker.py
jerith, line 272: It's implicitly included in #914.

justinvdm:
:+1:

go/vumitools/app_worker.py
hodgestar, line 272: Explicit is better than implicit -- let's explicitly mention it in #914.

PR #894, opened 2014-04-09
Loading of cached messages loses order (was Message Store cache pagination display is broken)

smn:
Once the cache is full no new messages are displayed, what should happen is that new messages are inserted in the front & the oldest ones popped off.


smn:
This is probably a Vumi MessageStoreCache problem rather than a Go UI problem.

smn:
This is a Vumi issue, see https://github.com/praekelt/vumi/issues/766

smn:
Leaving this open because an other buglet came up that this will address.

jerith:
Can we add a test for this that asserts correct ordering?

smn:
ready for re-review

jerith:
:+1:

PR #917, opened 2014-04-15
multiple dial in boolean in v2 dialogue app seems to be ignored.

justinvdm:
Failing to dial in more than once, even with the checkbox checked.


justinvdm:
Ready for review. Probably worth waiting for Travis.

smn:
@justinvdm :+1: was it just a mismatch between where the bool is stored in the conversation config vs the test harness?

justinvdm:
Yup :)

PR #916, opened 2014-04-15
Get delivery class from jsbox config for push triggers

justinvdm:
At the moment, we get the delivery class off the conversation model: 
https://github.com/praekelt/vumi-go/blob/develop/go/apps/jsbox/definition.py#L25

The `Conversation` model's `delivery_class` attribute is no longer set now that we have routing table dispatchers, so we need to find the delivery class a different way.

The plan is to determine the delivery class from the jsbox conversation's config, since we rely on it on the toolkit side anyways. This will be easy enough to remove once we have proper address types.

### DIFF
--- a/go/apps/jsbox/tests/test_vumi_app.py
+++ b/go/apps/jsbox/tests/test_vumi_app.py
@@ -178,10 +178,11 @@ class TestJsBoxApplication(VumiTestCase):
 
         with LogCatcher() as lc:
             yield self.send_send_jsbox_command(conv)
+            print lc.errors
 
-            self.assertEqual(
-                [e['message'][0] for e in lc.errors],
-                "Bad jsbox js config: bad")
+            self.assertTrue(any(
+                "Bad jsbox js config: bad" in
+                e['message'][0] for e in lc.errors))
 
     @inlineCallbacks
     def test_user_message(self):
--- a/go/apps/jsbox/tests/test_vumi_app.py
+++ b/go/apps/jsbox/tests/test_vumi_app.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import json
 import pkg_resources
 
 import mock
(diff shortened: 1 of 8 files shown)


justinvdm:
@smn ready for review.

go/apps/jsbox/tests/test_vumi_app.py
```diff
@@ -77,28 +100,93 @@ def send_send_jsbox_command(self, conversation):
...
+            app=self.APPS['cmd'] % {'method': 'on_inbound_message'},
+            delivery_class='twitter')
+        config['jsbox_app_config']['config']['value'] = 'bad'
+
+        conv = yield self.setup_conversation(config=config, groups=[group])
+        yield self.app_helper.start_conversation(conv)
+
+        with LogCatcher() as lc:
+            yield self.send_send_jsbox_command(conv)
+            print lc.errors
```
smn: stray print?

go/apps/jsbox/vumi_app.py
```diff
@@ -124,8 +123,18 @@ def send_inbound_push_trigger(self, to_addr, conversation):
 
     @inlineCallbacks
     def process_command_send_jsbox(self, user_account_key, conversation_key,
-                                   batch_id, delivery_class):
+                                   batch_id):
         conv = yield self.get_conversation(user_account_key, conversation_key)
+
+        try:
+            js_config = jsbox_js_config(conv.config)
+            delivery_class = js_config.get('delivery_class')
```
smn, line 131: What does this do for old style conversations that didn't mandate a `delivery_class`?
Is that simply not supported for those?

justinvdm, line 131: `delivery_class` is optional in v02 jsbox conversations too. If `delivery_class` is not set in the config, we give `Contact.for_addr` a `None`, so it returns the contacts msisdn: https://github.com/praekelt/vumi-go/blob/develop/go/vumitools/contact/models.py#L77-L80

It does feel like the code here is relying on an obscure part of code living somewhere else, not really sure how to do this better though. I'm up for suggestions.

justinvdm, line 131: If it helps, here's the test that covers this case: https://github.com/praekelt/vumi-go/blob/feature/issue-916-get-delivery-class-from-jsbox-config-for-push-triggers/go/apps/jsbox/tests/test_vumi_app.py#L105-L134

justinvdm:
Ready for review, I think.

smn:
:+1:

PR #920, opened 2014-04-17
treq requirement missing in setup.py

smn:


smn:
@miltontony @justinvdm tiny!

smn:
Actually nevermind, this was due to Vumi's requirements not being installed properly.

PR #918, opened 2014-04-17
Include contact uuids in contact export

smn:
Makes it easier to identify uniques across multiple CSV downloads.

PR #919, opened 2014-04-17
Allow for updating existing contacts on import.

smn:
Currently we're only inserting new contacts on import, not allowing updating of existing ones. #918 adds support for exporting the UUID, if we find that exists we can give the option of allowing an update, the UI would need to show what needs to be treated as the  "Truth" though.

### DIFF
--- a/go/contacts/import_handlers.py
+++ b/go/contacts/import_handlers.py
@@ -2,7 +2,7 @@ from go.contacts import tasks, utils
 from go.contacts.parsers import ContactFileParser
 
 
-def handle_import_new_contacts(request, group):
+def dispatch_import_task(import_task, request, group):
     file_name, file_path = utils.get_file_hints_from_session(request)
     file_type, parser = ContactFileParser.get_parser(file_name)
     has_header, _, sample_row = parser.guess_headers_and_row(file_path)
@@ -15,17 +15,23 @@ def handle_import_new_contacts(request, group):
     normalizers = [request.POST.get('normalize-%s' % i, '')
                    for i in range(len(sample_row))]
     fields = zip(field_names, normalizers)
-
-    tasks.import_new_contacts_file.delay(
+    import_task.delay(
         request.user_api.user_account_key, group.key, file_name,
         file_path, fields, has_header)
 
     utils.clear_file_hints_from_session(request)
 
 
+def handle_import_new_contacts(request, group):
+    return dispatch_import_task(
+        tasks.import_new_contacts_file, request, group)
+
+
 def handle_import_upload_is_truth(request, group):
-    pass
+    return dispatch_import_task(
+        tasks.import_upload_is_truth_contacts_file, request, group)
 
 
 def handle_import_existing_is_truth(request, group):
-    pass
+    return dispatch_import_task(
+        tasks.import_existing_is_truth_contacts_file, request, group)
--- a/go/contacts/import_handlers.py
+++ b/go/contacts/import_handlers.py
@@ -1,0 +1,31 @@
+from go.contacts import tasks, utils
+from go.contacts.parsers import ContactFileParser
+
+
+def handle_import_new_contacts(request, group):
+    file_name, file_path = utils.get_file_hints_from_session(request)
+    file_type, parser = ContactFileParser.get_parser(file_name)
+    has_header, _, sample_row = parser.guess_headers_and_row(file_path)
+
+    # Grab the selected field names from the submitted form
+    # by looping over the expect n number of `column-n` keys being
+    # posted
+    field_names = [request.POST.get('column-%s' % i) for i in
+                   range(len(sample_row))]
+    normalizers = [request.POST.get('normalize-%s' % i, '')
+                   for i in range(len(sample_row))]
+    fields = zip(field_names, normalizers)
+
+    tasks.import_new_contacts_file.delay(
+        request.user_api.user_account_key, group.key, file_name,
+        file_path, fields, has_header)
+
+    utils.clear_file_hints_from_session(request)
+
+
+def handle_import_upload_is_truth(request, group):
+    pass
+
+
+def handle_import_existing_is_truth(request, group):
+    pass
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -230,8 +230,8 @@ def export_many_group_contacts(account_key, group_keys, include_extra=True):
 
 
 @task(ignore_result=True)
-def import_contacts_file(account_key, group_key, file_name, file_path,
-                         fields, has_header):
+def import_new_contacts_file(account_key, group_key, file_name, file_path,
+                             fields, has_header):
     api = VumiUserApi.from_config_sync(account_key, settings.VUMI_API_CONFIG)
     contact_store = api.contact_store
     group = contact_store.get_group(group_key)
@@ -263,7 +263,7 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
             }), settings.DEFAULT_FROM_EMAIL, [user_profile.user.email],
             fail_silently=False)
 
-    except:
+    except Exception:
         # Clean up if something went wrong, either everything is written
         # or nothing is written
         for contact in written_contacts:
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -12,6 +12,7 @@ from django.template.loader import render_to_string
 from django.utils.safestring import mark_safe
 
 from go.vumitools.api import VumiUserApi
+from go.vumitools.contact.models import ContactNotFoundError
 from go.base.models import UserProfile
 from go.base.utils import UnicodeCSVWriter
 from go.contacts.parsers import ContactFileParser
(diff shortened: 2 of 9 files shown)


smn:
When specifying a `key` in the upload:

vumi_go

When not specifying a `key` in the upload:

vumi_go

This is also for @mdegale.

mdegale:
wow that looks amazing!

On 17 April 2014 08:20, Simon de Haan notifications@github.com wrote:

> When specifying a key in the upload:
> 
> [image: vumi_go]https://cloud.githubusercontent.com/assets/1065/2728602/f87db34c-c5f7-11e3-95c8-53c7c762386b.png
> 
> When not specifying a key in the upload:
> 
> [image: vumi_go]https://cloud.githubusercontent.com/assets/1065/2728619/551abbd6-c5f8-11e3-92e9-67d90d343da0.png
> 
> This is also for @mdegale https://github.com/mdegale.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/praekelt/vumi-go/issues/919#issuecomment-40684716
> .

## 

*MATTHEW DE GALE *| HEAD OF PROJECTS: PRAEKELT FOUNDATION

_PRAEKELTFOUNDATION_.ORG | _PRAEKELT_.COM
_M_: +27 798910129 | _S_: mdegale | _Tw_: @mdegale, @praekeltfound

smn:
hah, this is the easy part so far. This is just UI work, not actual import work.

go/contacts/tasks.py
```diff
@@ -290,3 +292,83 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+@task(ignore_result=True)
+def import_upload_is_truth_contacts_file(account_key, group_key, file_name,
+                                         file_path, fields, has_header):
+
+    def merge_operation(contact, contact_dictionary):
+        # NOTE:     The order here is important, the new extra is
+        #           the truth which we want to maintain
+        new_extra = {}
+        new_extra.update(dict(contact.extra))
+        new_extra.update(contact_dictionary.pop('extra', {}))
```
justinvdm, line 350: Would this maybe cause an issue once we allow json extras?

go/contacts/tasks.py
```diff
@@ -290,3 +292,83 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+def import_existing_is_truth_contacts_file(account_key, group_key, file_name,
+                                           file_path, fields, has_header):
+
+    def merge_operation(contact, contact_dictionary):
+        # NOTE:     The order here is important, the existing extra is
+        #           the truth which we want to maintain
+        new_extra = {}
+        new_extra.update(contact_dictionary.pop('extra', {}))
+        new_extra.update(dict(contact.extra))
+        contact_dictionary['extra'] = new_extra
```
justinvdm: Should we do similar handling for contact's `subscriptions` field? Similarly for `import_upload_is_truth...` above.

go/contacts/import_handlers.py
```diff
@@ -0,0 +1,37 @@
...
+    file_type, parser = ContactFileParser.get_parser(file_name)
+    has_header, _, sample_row = parser.guess_headers_and_row(file_path)
+
+    # Grab the selected field names from the submitted form
+    # by looping over the expect n number of `column-n` keys being
+    # posted
+    field_names = [request.POST.get('column-%s' % i) for i in
+                   range(len(sample_row))]
+    normalizers = [request.POST.get('normalize-%s' % i, '')
+                   for i in range(len(sample_row))]
```
justinvdm, line 16: Why do we need the `''` default here but not for `column-%s` above?

go/contacts/tasks.py
```diff
@@ -290,3 +292,83 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+            contact = contact_store.get_contact_by_key(key)
+            contact_dictionary = contact_mangler(contact, contact_dictionary)
+            contact_store.update_contact(key, **contact_dictionary)
+            counter += 1
+        except KeyError, e:
+            errors.append((key, 'No key provided'))
+        except ContactNotFoundError, e:
+            errors.append((key, str(e)))
+        except Exception, e:
+            errors.append((key, str(e)))
```
justinvdm, line 324: I'm wondering if its okay that we are allowing some parts of the upload to succeed and some parts to fail instead of rolling back if a failure happens. Or isn't it a problem if that happens?

go/contacts/tasks.py
smn, line 350: Possibly? Our tests will tell us.

go/contacts/tasks.py
smn: thanks, adding those.

go/contacts/import_handlers.py
smn, line 16: The `normalize-%s` stuff aren't mandatory but the `column-%s` ones are. Removing the empty `'column-0': ''` stuff as that'll make it clearer.

go/contacts/tasks.py
smn, line 324: The reason I'm not comfortable doing a roll back here is because we don't have any transactional guarantees from Riak. If we roll back we don't know if the state we're rolling back to is still the same, someone may have accessed the content and updated it since.

justinvdm, line 324: Ah, ok cool.

smn:
@justinvdm ready for re-review

go/contacts/tasks.py
```diff
@@ -290,3 +292,102 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+    def merge_operation(contact, contact_dictionary):
+        # NOTE:     The order here is important, the new extra is
+        #           the truth which we want to maintain
+        new_extra = {}
+        new_extra.update(dict(contact.extra))
+        new_extra.update(contact_dictionary.pop('extra', {}))
+
+        new_subscription = {}
+        new_subscription.update(dict(contact.subscription))
+        new_subscription.update(contact_dictionary.pop('subscription', {}))
```
justinvdm, line 354: Don't we export extras and subscriptions with prefixes (`extras-` and `subscription-`), so we need to handle this on imports?

justinvdm, line 354: Ah, nvm, `ContactsFileParser` probably handles that already.

smn, line 354: It's an existing quirk unfortunately. We're not exporting the `subscription-` stuff only the `extra-` stuff. I can drop that `extra-` bit but only if we don't already have a column with that name. Wouldn't want to end up with two columns called `msisdn` if `msisdn` is a key in the extra dict.

justinvdm, line 354: Ok cool.

go/contacts/tasks.py
```diff
@@ -290,3 +292,102 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+
+        contact_dictionary['extra'] = new_extra
+        contact_dictionary['subscription'] = new_subscription
+        contact_dictionary['groups'] = [group_key]
```
justinvdm, line 358: Won't this overwrite the existing contact's groups? Or is that the point since the upload is the truth here?

smn, line 358: I thought so too earlier but it turns out `update_contact()` [handles that for us](https://github.com/praekelt/vumi-go/blob/develop/go/vumitools/contact/models.py#L137-L147)

go/contacts/tasks.py
smn, line 354: Pushed change that does this.

go/contacts/templates/contacts/import_upload_is_truth_completed_mail.txt
```diff
@@ -0,0 +1,17 @@
+Hi {{user.get_full_name|safe}}!
+{% if count > 0 %}
+We've successfully imported {{count}} of your contact(s).
+They're all stored in the group {{group.name|safe}}
+
+{% if errors %}Unfortunately there were also {{errors|length}} errors. These are listed below:{% endif %}{% else %}
+We were not able to import any of your contacts unfortunately.
```
justinvdm, line 7: Seems like it contradicts the `We've successfully imported ...` line. s/any/all maybe?

smn, line 7: There's an `{% else %}` at the end of the line above that matches the `{% if count > 0 %}` a few lines up. It'll only show if all failed.

justinvdm, line 7: Ah, right.

go/contacts/tasks.py
```diff
@@ -291,3 +295,102 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+        cloned_contact_dictionary['extra'] = new_extra
+
+        new_subscription = {}
+        new_subscription.update(contact_dictionary.pop('subscription', {}))
+        new_subscription.update(dict(contact.subscription))
+        cloned_contact_dictionary['subscription'] = new_subscription
+
+        for key, value in contact_dictionary.items():
+            if value:
+                cloned_contact_dictionary.pop(key)
```
justinvdm, line 394: Could you explain what's happening here? Can't seem to figure it out.

smn, line 394: hah, thanks, bug.

smn:
@justinvdm thanks, ready for re-review, that was a bad one.

go/contacts/tasks.py
```diff
@@ -291,3 +295,104 @@ def import_contacts_file(account_key, group_key, file_name, file_path,
...
+
+        new_subscription = {}
+        new_subscription.update(contact_dictionary.pop('subscription', {}))
+        new_subscription.update(dict(contact.subscription))
+        cloned_contact_dictionary['subscription'] = new_subscription
+
+        for key, value in contact_dictionary.items():
+            # If the contact already has any kind of value that resolves
+            # to `True` then skip it.
+            if getattr(contact, key, None):
```
justinvdm: Won't this not work for things like `0` and `False` that are actual, set contact attributes?

smn:
@justinvdm leaving it as is. I think the underlying issue is that if the user says their version is the source of Truth then we need to accept that. If that means it'll get overwritten by blanks because the user says that's the truth then that's what needs to be done.

Also our extras don't allow `0` and `False` values to be set, everything (currently) needs to be `Unicode` for extras.

For `msisdn` and `name` etc we'll always be passing in strings simply because we're dealing with CSV which doesn't know about types.

If we pass in `''` to `dob` that'll complain about it, similarly boolean type fields will.

I think this is ok?

justinvdm:
:+1:

PR #923, opened 2014-04-22
Add export-all option for contacts

hodgestar:


hodgestar:
Ready for review.

jerith:
I'd be more comfortable if we could impose a limit of one export at a time. I don't want people to ask for a large export several times because the first attempt hasn't finished yet.

smn:
:+1: on this and :+1: on limiting jobs per user however I don't see how we can do that realistically without a) ability to cancel an existing job and b) introspecting jobs per user. My vote is to cut a ticket to address that in a separate PR. Also, long term solution is probably to just have twistd handle these downloads directly over http instead of via celery (like the message store resource).

jerith:
I really don't want to open us up to a massive Riak DoS, so I'm :-1: until we can avoid that.

jerith:
Although I can live with it if the DoS is already possible and we put a warning next to the buttons.

PR #926, opened 2014-04-23
'Display this question in search results' isn't saved

justinvdm:
Dialogue states have a 'Display this question in search results' checkbox, but we don't seem to be saving the user's choice for this checkbox.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #928, opened 2014-04-23
Update to Django 1.6

jerith:
It's been around for a while and it's usually a good idea to keep up.


jerith:
Note: This removes `django.contrib.markup` (which was already deprecated in 1.5) and therefore our flatpage data won't be rendered.

PR #929, opened 2014-04-24
Exclude exhausted tagpools from new channel form

jerith:
Pools with no remaining tags in them are still listed in the channel creation form and cause explosions when we try to acquire the tags.


hodgestar:
It would be nice to show people which tag pools are exhausted or empty, but this is better than 500'ing so :+1: as soon as Travis finishes.

jerith:
> It would be nice to show people which tag pools are exhausted or empty

Agreed. I'll open an issue for that and work on it in the background.

PR #931, opened 2014-04-24
Compat fixes for Django 1.6

jerith:
This is a subset of #928 that gets us closer to Django 1.6 without actually switching to it.


go/templates/flatpages/default.html
```diff
@@ -1,9 +1,8 @@
 {% extends "app.html" %}
-{% load markup %}
 
 {% block page_title %}Vumi Go: {{flatpage.title}}{% endblock %}
 {% block content_title %}{{flatpage.title}}{% endblock %}
 
 {% block content_main %}
-{{flatpage.content|markdown}}
+{{flatpage.content}}
```
hodgestar, line 7: Are we okay with losing support for `markdown` in flatpages?

hodgestar, line 7: All of our production flatpages have HTML, not markdown, so I think it's okay to drop this if it is no longer in Django.

hodgestar:
:+1: once Travis is happy.

PR #932, opened 2014-04-25
Better VumiApi publisher setup

jerith:
This is a prerequisite of #915 but is sufficiently decoupled to be worth doing separately.


go/vumitools/api.py
```diff
@@ -529,20 +529,17 @@ def from_config_sync(cls, config, amqp_client=None):
...
-            sender = SyncMessageSender(amqp_client)
-        return cls(manager, redis, sender)
+        return cls(manager, redis, SyncMessageSender(amqp_client))
 
     @classmethod
     @inlineCallbacks
-    def from_config_async(cls, config, amqp_client=None):
+    def from_config_async(cls, config, command_publisher=None):
+        # Note: This takes a publisher rather than a client to avoid leaking
+        #       AMQP channels by making our own transient publishers.
```
justinvdm, line 538: I understand that `from_config_sync` and `SyncMessageSender` don't need to take in publisher instead of a client, but maybe they should for consistency with their async counterparts?

justinvdm, line 538: Ah, or are publishers async only?

go/vumitools/api.py
```diff
@@ -582,43 +577,40 @@ def __init__(self, amqp_client):
         self.amqp_client = amqp_client
 
     def send_command(self, command):
+        if self.amqp_client is None:
+            raise VumiError("No command message publisher available.")
```
justinvdm, line 581: Should this maybe say something about the client instead of publisher, or did we intentionally refer the client as a publisher here?

go/vumitools/app_worker.py
```diff
@@ -49,9 +51,6 @@ class GoWorkerConfigMixin(object):
     riak_manager = ConfigDict("Riak config.", static=True)
     redis_manager = ConfigDict("Redis config.", static=True)
 
-    api_routing = ConfigDict("AMQP config for API commands.", static=True)
-    app_event_routing = ConfigDict("AMQP config for app events.", static=True)
-
```
justinvdm, line 54: Are we sure nothing is using these in production?

justinvdm:
Looks good, just some questions.

go/vumitools/api.py
jerith, line 538: The sync AMQP stuff doesn't have publishers or consumers -- that's all handled by the client object.

We need equivalent functionality, but we don't need identical setup.

go/vumitools/api.py
jerith, line 581: This is for consistency with the async case. The sync AMQP client is also the publisher, so it's not entirely inaccurate.

go/vumitools/app_worker.py
jerith, line 54: I grepped the puppet repo. Also, anything that uses different config from the rest of the system will be unable to interact with it.

go/vumitools/api.py
justinvdm, line 538: Ok cool. From a programmer-using-api perspective, it feels a little surprising that the params differ, but I guess this isn't important.

go/vumitools/app_worker.py
jerith, line 54: To clarify: This kind of thing should be configureable, but not on a per-worker basis. We need a system-wide configuration thing and can put the config back when we have that.

justinvdm:
:+1: not sure if @hodgestar should take a look since this is a pretty core change. My thoughts are that this has been in the air long enough that it should land, up to you guys though.

go/vumitools/app_worker.py
justinvdm, line 54: Sounds good.

hodgestar:
:+1: from me landing -- I followed along a bit and we shouldn't rely on me
looking at everything in detail.

go/vumitools/api.py
jerith, line 538: In both the sync and async case, we're passing in the thing that publishes messages; it's just that those things are rather different. To make them look the same, we'd have to do a bunch of unnecessary wrapping.

PR #934, opened 2014-04-26
Make jsbox event handling optional

jerith:
Most (all?) jsbox apps don't actually need to process events, but we're currently spinning up the sandbox for them anyway. This should be optional and default to disabled.

### DIFF
--- a/go/apps/jsbox/vumi_app.py
+++ b/go/apps/jsbox/vumi_app.py
@@ -119,8 +119,10 @@ class JsBoxApplication(GoApplicationMixin, JsSandbox):
         We only want to process events in the sandbox if configured to do so.
         """
         config = yield self.get_config(event)
-        app_config = jsbox_js_config(config.conversation.config)
-        if app_config.get('process_events'):
+        js_config = self.get_jsbox_js_config(config.conversation)
+        if js_config is None:
+            return
+        if js_config.get('process_events'):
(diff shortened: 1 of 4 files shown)


smn:
:+1: but if possible it'd be great to have @hodgestar or @justinvdm confirm as well.

go/apps/jsbox/vumi_app.py
```diff
@@ -110,6 +113,22 @@ def process_message_in_sandbox(self, msg):
...
         yield super(JsBoxApplication, self).process_message_in_sandbox(msg)
 
+    @inlineCallbacks
+    def process_event_in_sandbox(self, event):
+        """
+        We only want to process events in the sandbox if configured to do so.
+        """
+        config = yield self.get_config(event)
+        app_config = config.jsbox_app_config.get('config', {}).get('value')
+        app_config = json.loads(app_config or '{}')
```
hodgestar: You should be able to use `jsbox_js_config` from https://github.com/praekelt/vumi-go/blob/develop/go/apps/jsbox/utils.py#L10-L12 with a `try: ... except: ...` instead so that the logic is only in one place.

hodgestar:
One minor comment, otherwise :+1:.

hodgestar:
:+1:

hodgestar:
:+1: again.

PR #936, opened 2014-04-27
Script to replace problematic send_jsbox action

jerith:


go/scripts/jsbox_send.py
```diff
@@ -0,0 +1,151 @@
...
+
+    worker_creator = WorkerCreator(options.vumi_options)
+    worker_creator.create_worker_by_class(
+        JsBoxSendWorker, options.get_vumigo_config())
+
+    in_file = sys.stdin
+    out_file = sys.stdout if not quiet else None
+
+    worker = yield JsBoxSendWorker.WORKER_QUEUE.get()
+    yield worker.process_file(in_file, out_file)
```
hodgestar: This is the only reference I can find to `.process_file` so I assume something else is supposed to be called instead?

jerith: Whoops, that's left over from the `inject_messages` script I pilfered the structure from.

hodgestar:
:+1:

smn:
Build failing because of npm issues, restarting.

smn:
:+1: from me trusting that there's no other thing we can wait on. Lip stick on pig feature would be to make it configurable, but it's still a pig after that.

Issue #875, opened 2014-02-21
Go Worker's *.control workers don't get QoS set


smn:
It consumes all messages available in the queue, the `amqp_prefetch_count` value isn't applied on those connectors.

This may apply to other connectors too.

PR #937, opened 2014-04-29
Worker control queues need to prefetch one message

jerith:
This is a special case of #875 that gets its own ticket because it doesn't address the general case.


hodgestar:
:+1:

PR #939, opened 2014-05-01
More jsbox send improvements

jerith:


go/scripts/jsbox_send.py
```diff
@@ -98,27 +100,48 @@ class JsBoxSendWorker(Worker):
...
         delivery_class = jsbox_js_config(conv.config).get('delivery_class')
-        to_addrs = yield self.get_contact_addrs_for_conv(conv, delivery_class)
+        excluded_addrs = self.get_excluded_addrs(addr_exclude_path)
+        to_addrs = yield self.get_contact_addrs_for_conv(
+            conv, delivery_class, excluded_addrs)
         ticker = Ticker(hz=hz)
-        for to_addr in to_addrs:
+        for i, to_addr in enumerate(to_addrs):
             yield self.send_inbound_push_trigger(to_addr, conv)
+            if (i + 1) % 100 == 0:
```
froztbyte, line 119: so we'll get a status print every 100 messages?

go/scripts/tests/test_jsbox_send.py
```diff
@@ -162,17 +167,48 @@ def test_send_inbound_push_trigger(self):
         worker_helper = self.vumi_helper.get_worker_helper('jsbox_transport')
 
         self.assertEqual(worker_helper.get_dispatched_inbound(), [])
+        self.assertEqual(worker.stdout.getvalue(), '')
         yield worker.send_inbound_push_trigger('+27831234567', conv)
+        self.assertEqual(
+            worker.stdout.getvalue(),
+            "Starting u'My Conversation' [%s] -> +27831234567\n" % (conv.key,))
```
froztbyte, line 174: I know it's a test and this might be bikeshedding, but shouldn't this number perhaps be a variable elsewhere, and referred to each time, incase it ever gets typo'd?

go/scripts/jsbox_send.py
jerith, line 119: Yes. This matches the batch size of the contact loading, so the address collection message also happens at multiples of 100.

go/scripts/tests/test_jsbox_send.py
jerith, line 174: It's better to have static text in the output assertion. The only reason the conversation key gets substituted in is that it changes on every test run.

smn:
:+1:

PR #940, opened 2014-05-03
Stash contact key on message in jsbox send script

jerith:


hodgestar:
I assume this would need a PR in praekelt/vumi-jssandbox-toolkit to become useful?

hodgestar:
Wondering whether this is still relevant or whether we should close it?

PR #943, opened 2014-05-05
Contact import stuff should recognise header when `key` column present

smn:
Currently it doesn't.


jerith:
:+1: if Travis is happy.

PR #944, opened 2014-05-05
Switch contact address lookup from search to index

jerith:
This will have to happen in two parts. This ticket is for the model updates and the next will be switching to using the indexes once the migrations have been done.


go/vumitools/contact/models.py
```diff
@@ -44,29 +43,31 @@ def __unicode__(self):
...
     created_at = Timestamp(default=datetime.utcnow)
     groups = ManyToMany(ContactGroup)
     extra = Dynamic(prefix='extras-')
     subscription = Dynamic(prefix='subscription-')
 
+    # Address fields
+    msisdn = Unicode(max_length=255, index=True)
+    gtalk_id = Unicode(null=True, index=True)
+    mxit_id = Unicode(null=True, index=True)
+    wechat_id = Unicode(null=True, index=True)
```
hodgestar, line 69: We should definitely move `twitter_handle` here, and maybe `facebook_id`, `bbm_pin` and `email_address`?

jerith, line 69: We don't currently have delivery classes for those, but we can move them and add indexes.

go/vumitools/contact/old_models.py
```diff
@@ -37,12 +36,12 @@ def __unicode__(self):
...
 
 class ContactVNone(Model):
+    """A contact"""
 
     bucket = "contact"
 
-    """A contact"""
     # key is UUID
-    user_account = ForeignKey(UserAccountV4)
+    user_account = ForeignKey(UserAccount)
```
hodgestar, line 41: Not sure why this changed? It's generally a bad idea to link to code that might evolve from migrators or to edit old versions of models.

go/vumitools/contact/old_models.py
```diff
@@ -157,6 +213,8 @@ def new_smart_group(self, name, query):
     def get_contact_by_key(self, key):
         contact = yield self.contacts.load(key)
         if contact is None:
+            # Local import to avoid potential circularity
+            from go.vumitools.contact.models import ContactNotFoundError
```
hodgestar: I guess `ContactNotFoundError` should be in an errors module somewhere.

go/vumitools/contact/old_models.py
```diff
@@ -337,40 +396,264 @@ def contact_for_addr(self, delivery_class, addr, create=True):
...
+            contact_id = uuid4().get_hex()
+            field.setdefault('msisdn', u'unknown')
+            returnValue(self.contacts(
+                contact_id, user_account=self.user_account_key, **field))
+
+        # Local import to avoid potential circularity
+        from go.vumitools.contact.models import ContactNotFoundError
+        raise ContactNotFoundError(
+            "Contact with address '%s' for delivery class '%s' not found."
+            % (addr, delivery_class))
```
hodgestar: In the past we seem to have tried to not copy all the methods into the old models and just copy the fields and a few selected methods?

go/vumitools/contact/old_models.py
jerith, line 41: Unless we're loading the object referenced by the the foreign key, all we care about is the key itself and that's independent of model version. Also, I don't want to import `old_models` anywhere except the tests that use them -- too much chance of accidental circularity.

go/vumitools/contact/old_models.py
jerith: Perhaps. Should I do that in this PR?

hodgestar: Happy with a separate PR or this one -- whichever seems easier.

go/vumitools/contact/old_models.py
jerith: I'd much rather copy the whole class and make only the changes necessary for the version difference. Referencing methods from other places leaves us open to unintended changes. On the other hand, I suppose it would be okay to just remove any unnecessary methods.

hodgestar: Are any of the methods necessary? We should be able to check migrations
just using the basics from Model?

go/vumitools/contact/old_models.py
hodgestar, line 41: I think importing `old_models` from other `old_models` is fine. Circularity can be worked around if needed (and it probably only exists if the original model packages were circular).

hodgestar, line 41: Not totally fussed by this, but I think I originally made @smn reference the old account models because I've been bitten by migration scripts importing changing code before.

go/vumitools/contact/migrations.py
```diff
@@ -20,3 +20,29 @@ def migrate_from_unversioned(self, mdata):
         mdata.set_value('wechat_id', None)
 
         return mdata
+
+    def migrate_from_1(self, mdata):
+
+        mdata.copy_values(
+            'name', 'surname', 'email_address', 'dob', 'created_at')
```
smn, line 27: Any specific reason we're not storing `created_at` as a index with a `_int` suffix so we can search on ranges?

go/vumitools/contact/migrations.py
```diff
@@ -20,3 +20,29 @@ def migrate_from_unversioned(self, mdata):
...
+            'msisdn', 'twitter_handle', 'facebook_id', 'bbm_pin', 'gtalk_id',
+            'mxit_id', 'wechat_id')
+
+        for field in new_index_fields:
+            value = mdata.old_data[field]
+            # Only set index if value is set.
+            if value is not None:
+                mdata.set_value(field, value, index=('%s_bin' % (field,)))
+            else:
+                mdata.set_value(field, value)
```
smn: Is this because of that `"None"` weirdness we had before?

go/vumitools/contact/models.py
```diff
@@ -304,23 +317,24 @@ def delivery_class_supported(self, delivery_class):
...
             raise ContactError(
                 "Unsupported transport_type %r" % delivery_class)
 
     def new_contact_for_addr(self, delivery_class, addr):
-        field = self._contact_field_for_addr(delivery_class, addr)
-        field.setdefault('msisdn', u'unknown')
-        return self.new_contact(**field)
+        field, value = self._contact_field_for_addr(delivery_class, addr)
+        field_dict = {field: value}
+        field_dict.setdefault('msisdn', u'unknown')
```
smn, line 336: Why do we need to maintain this? Perhaps time to see if we can kill it, if not in this PR then another? Making `msisdn` mandatory isn't needed anymore in current Vumi world.

go/vumitools/contact/old_models.py
smn, line 41: Lord help us. This is a hard problem.

hodgestar:
Is there a reason `test_migrations` got moved to / combined with `test_models`? I'd prefer us to keep migration tests in `test_migrations`?

hodgestar:
We're going to need something like the `go_migrate_conversations` management command to go with this (although a separate PR is fine).

hodgestar:
I think that's all the comments from my side for the moment.

go/vumitools/contact/old_models.py
jerith, line 41: If we're loading other objects from foreign keys on old models we're almost certainly doing something seriously wrong. I can put it back the way it was if you prefer, though.

go/vumitools/contact/migrations.py
jerith, line 27: Changing that is out of scope for this PR.

go/vumitools/contact/migrations.py
jerith: Yes. It probably needs to be fixed in the migration code in vumi, but I didn't want to chase down that rabbit hole today.

go/vumitools/contact/models.py
jerith, line 336: Out of scope for this PR. Tracking down everything that depends on that is a job for another day.

jerith:
> Is there a reason test_migrations got moved to / combined with test_models? I'd prefer us to keep migration tests in test_migrations?

Migrations are part of models. We really need a whole lot more tests around this stuff (and they need to be better organised), but I don't want to block this for the amount of time it'll take to write those.

jerith:
> We're going to need something like the go_migrate_conversations management command to go with this (although a separate PR is fine).

We already have multiple versions, so I assumed such a thing exists. We can add it in a new PR if it doesn't.

smn:
:+1: from me, @hodgestar ?

go/vumitools/contact/migrations.py
hodgestar, line 27: Might be worthwhile doing before we do an actual migration in production, so perhaps we should make a ticket for it if we want it?

hodgestar:
I'd really like tests for code `migrations.py` to live in `test_migration.py` and not `test_models.py`, otherwise I won't find it or know what to do if I have to add more migration tests.

jerith:
> I'd really like tests for code migrations.py to live in test_migration.py and not test_models.py, otherwise I won't find it or know what to do if I have to add more migration tests.

The migration tests also (currently) test a bunch of model object stuff, so if we split them we'll end up with a lot of model test code living in `test_migrations.py` unless we do the extra work to split that stuff up better. I don't want that rabbit hole to block this branch, especially since all the other Vumi Go models need similar work and doing them all at once will be more efficient.

jerith:
Also, `go.vumitools.account` has all its migration tests in `test_models` and no other model stuff has a `test_migrations`.

go/vumitools/contact/migrations.py
jerith, line 27: I don't want to block this PR on that. Maybe we can do it at the same time we fix the mandatory MSISDN field?

hodgestar, line 27: Fine by me.

hodgestar:
:+1:, but let's clean up our model / migration tests as soon as we have a moment.

go/scripts/migrate_contacts.py
```diff
@@ -0,0 +1,112 @@
...
+            raise usage.UsageError(
+                "Please provide the vumigo-config parameter.")
+        if not self['user-account-key']:
+            raise usage.UsageError(
+                "Please provide the user-account-key parameter.")
+
+    def get_vumigo_config(self):
+        with file(self['vumigo-config'], 'r') as stream:
+            return yaml.safe_load(stream)
+
```
hodgestar, line 39: I'm trying to figure out where we can put an example command-line so that --help is more helpful. Maybe we can use `longdesc`?

jerith, line 39: I want to build some better infrastructure for these scripts, probably in praekelt/vumi rather than here. Until then, I think we're okay.

hodgestar:
:+1:

PR #946, opened 2014-05-08
Support group state in dialogue app

justinvdm:
This is part of #941, but I think it makes sense to get this out first instead of having unusable but 'working' UI.

### DIFF
--- a/go/apps/dialogue/tests/dummy_polls.js
+++ b/go/apps/dialogue/tests/dummy_polls.js
@@ -21,14 +21,24 @@ var simple_poll = {
         ]
     },
     {
-        uuid: "freetext-1",
+        uuid: "group-1",
         name: "Message 2",
         store_as: "message-2",
-        type: "freetext",
+        type: "group",
+        group: {key: 'group-1'},
         store_on_contact: true,
         entry_endpoint: {uuid: "endpoint-3"},
+        exit_endpoint: {uuid: "endpoint-4"}
+    },
+    {
+        uuid: "freetext-1",
+        name: "Message 3",
+        store_as: "message-3",
+        type: "freetext",
+        store_on_contact: true,
+        entry_endpoint: {uuid: "endpoint-5"},
         // freetext specific
-        exit_endpoint: {uuid: "endpoint-4"},
+        exit_endpoint: {uuid: "endpoint-6"},
         text: "What is your name?"
     },
     {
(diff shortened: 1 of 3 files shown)


justinvdm:
Ready for review.

go/apps/dialogue/tests/dummy_polls.js
```diff
@@ -21,14 +21,24 @@ var simple_poll = {
...
     },
     {
-        uuid: "freetext-1",
+        uuid: "group-1",
         name: "Message 2",
         store_as: "message-2",
-        type: "freetext",
+        type: "group",
+        group: {key: 'group-1'},
         store_on_contact: true,
```
hodgestar: It's a little odd that we still have `store_as` and `store_on_contact` here but if it's hard to remove them I'm okay with that for now.

go/apps/dialogue/tests/test_vumi_app.js
```diff
@@ -175,5 +175,37 @@ describe("app", function() {
...
+                        api.contacts.add({
+                            msisdn: '+27123',
+                            groups: ['group-2']
+                        });
+                    })
+                    .setup.user.addr('+27123')
+                    .setup.user.state('choice-1')
+                    .input('1')
+                    .check(function(api) {
+                        assert.deepEqual(api.contacts.store[0].groups, [
```
hodgestar: I'm tempted to say we should sort the groups so that if the dummy contact store ever gets more advanced this test won't break.

go/apps/dialogue/tests/dummy_polls.js
justinvdm: woops, copy paste error

hodgestar:
Other than the minor comments, looks good.

go/apps/dialogue/vumi_app.js
```diff
@@ -116,6 +118,18 @@ var DialogueApp = App.extend(function(self) {
...
 
+    self.types.group = function(desc) {
+        return self
+            .states.create(self.next(desc.exit_endpoint))
+            .then(function(state) {
+                state.on('state:enter', function() {
+                    self.contact.groups.push(desc.group.key);
+                });
+
+                return state;
```
hodgestar: Sudden thought: Could we perhaps refactor this into an `self.create_action` method? Then we could re-use it later for other actions. E.g.

``` javascript
self.types.group = function(desc) {
    return self.create_action(desc, function() {
        return self.contact.groups.push(desc.group.key);
    });
}
```

hodgestar: I'm also wondering whether `'state:enter'` of the next state is the best place to put this. Maybe directly inside the creator is the best place because this creator can only be called once (since immediately after that we'll be in the new state) and the special `'state:enter'` hook will only get added if we're called via this creator anyway?

justinvdm:
Ready for review, I think.

hodgestar:
:+1:

PR #947, opened 2014-05-08
Vumi Go architecture diagram

hodgestar:


hodgestar:
For this to render in readthedocs.org I need to add a font to `_static`. Currently using DejavuSans-Bold which is 676K. Anyone have a better solution?

smn:
Just a note that `Pillow` doesn't build nicely on OS X Mavericks unless you set some env vars as per [this SO article](http://stackoverflow.com/questions/22334776/installing-pillow-pil-on-mavericks).

`Pillow` is installed as a dependency of `sphinxcontrib-blockdiag`. Making Pillow a dependency seems like a pretty heavy one unless we're convinced we'll be using it a lot? 

Anyway, :+1: from me because it's a good diagram and is better than what we currently have by miles.

hodgestar:
I do have a second diagram I want to draw already. :)

Note: It's only a doc build dependency currently, so it won't affect
running tests locally (although building docs locally is nice too).

requirements-docs.pip
```diff
@@ -2,6 +2,7 @@
 # only, used by readthedocs.org for publishing
 # docs on http://vumi-go.readthedocs.org/
 Sphinx
-Django==1.4.2
+sphinxcontrib-blockdiag
 -e git+git://github.com/praekelt/vumi.git@develop#egg=vumi-dev
+Django==1.5.5
```
jerith: Do we need Django and vumi for this? If so, can we not perhaps assume that we already have the normal dependencies installed?

hodgestar: I was wondering about that too. I think we can remove them now that our
setup.py is more sane. Maybe to try that in this branch.

jerith:
:+1: if you've tested that it builds correctly.

PR #867, opened 2014-05-09
Expose raw message exports via UI.

smn:
Combining the stuff that's landed in https://github.com/praekelt/vumi#723 (message store resource) and http://wiki.nginx.org/XSendfile allows us to provide access to raw message downloads.

Django could return the appropriate http header in the response, nginx then grabs the file from twistd using the batch_id and streams it to the end user.

### DIFF
--- a/go/conversation/templates/conversation/message_list.html
+++ b/go/conversation/templates/conversation/message_list.html
@@ -155,7 +155,7 @@
                     </span><br/></p>
                 </div>
                 <div class="modal-footer">
-                    <button type="submit" class="btn btn-primary" data-loading-text="uploading...">Schedule Export</button>
+                    <button type="submit" class="btn btn-primary">Schedule Export</button>
                 </div>
             </form>
         </div>
--- a/go/conversation/templates/conversation/message_list.html
+++ b/go/conversation/templates/conversation/message_list.html
@@ -24,7 +24,10 @@
         <a href="{{ conversation.get_absolute_url }}aggregates.csv?direction=inbound">Download Received Stats</a>
         <a href="{{ conversation.get_absolute_url }}aggregates.csv?direction=outbound">Download Sent Stats</a>
         -->
-        <a class="btn btn-default" data-toggle="modal" href="#expMessagesFrm">Export</a>
+        <a class="btn btn-default" data-toggle="modal" href="#expMessagesFrm">Export CSV via email</a>
+        <a class="btn btn-default" href="{% conversation_screen conversation "export_messages" %}?direction={{message_direction}}">
+          Download {% if message_direction == 'inbound' %}received{% else %}sent{% endif %} messages as JSON
+        </a>
     </div>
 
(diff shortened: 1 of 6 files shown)


smn:
vumi_go

smn:
vumi_go

changed wording slightly and added text differing between in- & outbound

smn:
ready for review

go/conversation/view_definition.py
```diff
@@ -194,6 +194,36 @@ def get(self, request, conversation):
         return self.render_to_response(params)
 
 
+class ExportMessageView(ConversationTemplateView):
```
jerith: If you're not using templates, rather subclass `ConversationApiView` here.

go/conversation/templates/conversation/message_list.html
```diff
@@ -152,7 +155,7 @@
                     </span><br/></p>
                 </div>
                 <div class="modal-footer">
-                    <button type="submit" name="_export_conversation_messages" class="btn btn-primary" data-loading-text="uploading...">Schedule Export</button>
+                    <button type="submit" class="btn btn-primary" data-loading-text="uploading...">Schedule Export</button>
```
hodgestar: Should this maybe say "downloading ..."?

hodgestar: And is this scheduling an export or just doing an export?

go/conversation/view_definition.py
```diff
@@ -194,6 +194,36 @@ def get(self, request, conversation):
...
+        response = HttpResponse()
+
+        url = '/message_store_exporter/%s/%s.json' % (conversation.batch.key,
+                                                      direction)
+
+        if settings.DEBUG:
+            response.write(url)
+
+        response['X-Accel-Redirect'] = url
+        return response
```
hodgestar: Could we perhaps put this in a utility method so that we can use it in other places where we want to do Nginx redirects?

go/conversation/templates/conversation/message_list.html
smn: it's silly old stuff, I'll remove it.

smn:
ready for re-review

go/conversation/view_definition.py
```diff
@@ -194,6 +194,34 @@ def get(self, request, conversation):
...
+        direction = request.GET.get('direction', 'inbound')
+        if direction not in ['inbound', 'outbound']:
+            raise Http404()
+
+        url = '/message_store_exporter/%s/%s.json' % (conversation.batch.key,
+                                                      direction)
+
+        response = sendfile(url)
+        if settings.DEBUG:
+            response.write(url)
```
hodgestar: Maybe this should be part of `sendfile`?

hodgestar: The we can just do `return sendfile(url)`.

smn:
ready for review again

hodgestar:
:+1:

smn:
\o/

PR #941, opened 2014-05-09
Contact group dialogue state

justinvdm:
We need to add a dialogue state for adding a recipient to a contact group. Check out the screen shots for the UI.

screen shot 2014-05-05 at 11 30 34 am

screen shot 2014-05-05 at 11 30 43 am


justinvdm:
The client side changes are pretty much done, still need to do the api side. That should probably happen before this lands, planning on doing that in a separate PR so these changes can get reviewed sooner. Basically, ready for review :)

go/base/static/js/src/apps/dialogue/models.js
```diff
@@ -179,6 +210,11 @@
       key: 'urls',
       relatedModel: Model
     }, {
+      type: Backbone.HasMany,
+      key: 'groups',
+      relatedModel: GroupModel,
+      collectionType: GroupCollection
+    }, {
```
justinvdm, line 217: The diff here is confusing, these are actually changes for `DialogueModel`.

I'm not sure if its a good idea for us to put the account's contact groups directly on the model, but the only alternative I can think of is having a separate model for the contact groups, with its own fetch (== ajax call) happening (bleh). If we had a consistent api for both dialogues and contacts, that would be a better option, but with the way things are at the moment, doing this kind of thing with Backbone will be ugly. Regardless, its probably not a great solution.

Up for other alternatives.

hodgestar, line 217: I don't think a separate query / search for groups to a method on the Go API is the worst idea. Maybe let's sort this out when the API side is written. My main objection to having it on the model is that the list of groups is rather open ended and perhaps a combo-boxy search is the way to go.

hodgestar:
Looks good once the API side is ready. The groups select/display issue is important (because there are accounts with lots of groups) but let's not let it hold up this PR or the API one. We can address it as a separate PR (just to keep things small).

go/base/static/js/src/apps/dialogue/models.js
justinvdm, line 217: Hmm ya, maybe its actually not a bad idea. Maybe lets discuss this in RL.

justinvdm:
The server side change was small enough, decided to do it in this PR. Ready for review.

justinvdm:
Ready for re-review.

hodgestar:
:+1: as soon as Travis is happy with the final commit.

PR #949, opened 2014-05-09
Messaging architecture drawing

hodgestar:


hodgestar:
Diagrams built successfully at http://vumi-go.readthedocs.org/en/feature-issue-949-messaging-architecture-drawing/dev/frontend-architecture.html and http://vumi-go.readthedocs.org/en/feature-issue-949-messaging-architecture-drawing/dev/messaging-architecture.html.

hodgestar:
Ready for review.

setup.py
```diff
@@ -31,7 +31,7 @@
         'raven>=2.0,<3.0',
         'django-debug-toolbar==0.9.4',
         'kombu>=2.5,<3.0',
-        'librabbitmq==1.0.1',
+        'librabbitmq==1.5.1',
```
hodgestar, line 34: This change is here because readthedocs had trouble installing `1.0.1`.

hodgestar, line 34: I think it's https://github.com/celery/librabbitmq/blob/master/Changelog#L37 issue that was encountered.

justinvdm:
:+1: we should be careful about deploying this because of the librabbitmq version change (as we discussed in RL).

PR #952, opened 2014-05-12
Use account's static groups for dialogue group state

justinvdm:
At the moment, the conversation's groups are used instead of the account's groups. This means the flow is broken when we set up a new dialogue conversation.


justinvdm:
Ready for review.

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #911, opened 2014-05-12
Dialogues won't work properly over channels other than sms and ussd

justinvdm:
Dialogue applications store each user's answers on their contact. For jsbox apps, we look up a contact based on the contact's address. At the moment, we need to explicitly set the delivery class for a jsbox conversation in order to figure out which type of contact address to use when looking up the contact, and default to `ussd` as the delivery class if none is given. Since `ussd` and `sms` delivery classes map to a contact's `msisdn`, dialogue apps will work for those two channel types. For other channels types (for e.g, twitter), this won't work, since we will still be looking up contacts by their msisdn.

The short way round would be to add a delivery class drop down to UI for dialogue apps.

The longer, better way round would be to change jsbox apps to figure out the type of contact address to look up based on the received message's address type. We will need praekelt/vumi#581 for this.

### DIFF
--- a/go/apps/dialogue/templates/dialogue/edit.html
+++ b/go/apps/dialogue/templates/dialogue/edit.html
@@ -12,14 +12,14 @@
 {% block content_actions_right %}
   <div class="pull-left form-inline">
       <div class="form-group">
-          <input type="checkbox" id="repeatable">
+          <input id="repeatable" type="checkbox">
           <label>Enable users to submit multiple surveys</label>
       </div>
 
       <div class="form-group">
-          <select class="form-control">
+          <select id="delivery-class" class="form-control">
               {% for delivery_class in delivery_classes %}
-              <option value="{{ delivery_class }}" {% if delivery_class == default_delivery_class %} selected {% endif %}>
+              <option value="{{ delivery_class }}" {% if delivery_class == current_delivery_class %} selected {% endif %}>
                {{ delivery_class|title }}
                </option>
               {% endfor %}
--- a/go/apps/dialogue/templates/dialogue/edit.html
+++ b/go/apps/dialogue/templates/dialogue/edit.html
@@ -10,9 +10,23 @@
 {% endblock %}
 
 {% block content_actions_right %}
-  <div class="pull-left">
-    <input type="checkbox" id="repeatable">
-    Enable users to submit multiple surveys
+  <div class="pull-left form-inline">
+      <div class="form-group">
+          <input type="checkbox" id="repeatable">
+          <label>Enable users to submit multiple surveys</label>
+      </div>
+
+      <div class="form-group">
+          <select class="form-control">
+              {% for delivery_class in delivery_classes %}
+              <option value="{{ delivery_class }}" {% if delivery_class == default_delivery_class %} selected {% endif %}>
+               {{ delivery_class|title }}
+               </option>
+              {% endfor %}
+          </select>
+
+          <label>Channel type</label>
+      </div>
   </div>
 
   <div class="pull-right">
--- a/go/apps/dialogue/templates/dialogue/edit.html
+++ b/go/apps/dialogue/templates/dialogue/edit.html
@@ -18,9 +18,9 @@
 
       <div class="form-group">
           <select id="delivery-class" class="form-control">
-              {% for delivery_class in delivery_classes %}
+              {% for delivery_class, label in delivery_classes %}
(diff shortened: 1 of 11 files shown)


justinvdm:
Ready for review.

Two problems:
- Since the editing on the dialogue screen is done on the client side, we need to save the delivery class to the poll config with the data set on the client side, then fish this out when fetching the dialogue's jsbox app config. Not sure of a better alternative.
- At the moment, our conversation edit screen templates are mostly on the server side. Since the dialogue views are client side, it is difficult to render things like the 'Enable users to submit multiple surveys' checkbox and delivery class select box in a non-hacky way. Simply telling the backbone view to use jquery to set the values of those input fields when the view renders means that for a second or so before the js on the page has loaded, the user sees the unrendered input field values, then sees them change to the actual values (bad UX). To avoid this, we need to do some of the rendering on the server side, some on the client side, where both need to dig into the poll config data to pull out the current delivery class (not ideal).

justinvdm:
@badlard here's how it looks
2014-05-13-101958_1440x900_scrot

badlard:
I think this is a good solution for now. Thanks.

On 13 May 2014 10:18, Justin van der Merwe notifications@github.com wrote:

> @badlard https://github.com/badlard here's how it looks
> [image: 2014-05-13-101958_1440x900_scrot]https://cloud.githubusercontent.com/assets/1731223/2954730/180b89d0-da77-11e3-929f-8fd7e40894ba.png
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/praekelt/vumi-go/pull/911#issuecomment-42928771
> .

## 

_PRAEKELTFOUNDATION_.ORG

hodgestar:
I think having the delivery_class part of the poll config is sensible.

go/apps/dialogue/templates/dialogue/edit.html
```diff
@@ -10,9 +10,23 @@
...
+      <div class="form-group">
+          <input id="repeatable" type="checkbox">
+          <label>Enable users to submit multiple surveys</label>
+      </div>
+
+      <div class="form-group">
+          <select id="delivery-class" class="form-control">
+              {% for delivery_class in delivery_classes %}
+              <option value="{{ delivery_class }}" {% if delivery_class == current_delivery_class %} selected {% endif %}>
+               {{ delivery_class|title }}
```
hodgestar: Title case is probably not the right thing -- I guess we need labels for the delivery classes so we can display them properly, e.g. `{% for delivery_class, label in delivery_classes %} ...`

justinvdm: Ya, title case is probably not the right thing. I feel like we're putting lipstick on a pig, but I guess its worth doing.

justinvdm: Thoughts on where these label mappings should be defined?

go/vumitools/contact/models.py
```diff
@@ -13,6 +13,19 @@
 from go.vumitools.opt_out import OptOutStore
 
 
+DELIVERY_CLASS_ADDR_TYPES = {
+    'sms': 'msisdn',
+    'ussd': 'msisdn',
+    'gtalk': 'gtalk_id',
+    'mxit': 'mxit_id',
+    'wechat': 'wechat_id',
+    'twitter': 'twitter_handle',
```
hodgestar: Maybe the labels should be defined here? Perhaps:

```
DELIVERY_CLASSES = {
    'sms': {'field': 'msisdn', 'label': 'SMS'},
    ...
}
```

hodgestar:
Other than the label names, looks good.

Agreed about painting the pig a bit, but getting names right seems important.

go/vumitools/contact/models.py
justinvdm: I'm -0.1 on putting it in `models.py`, labels are part of presentation, maybe in a view definition or something?

justinvdm: On second thought, maybe its not so bad, will do.

hodgestar: My thought was to have the delivery classes defined in one place? I'm not
sure these labels are entirely presentation -- they're like `
conversation.name`.

justinvdm:
Ready for re-review.

hodgestar:
:+1:

justinvdm:
thanks :) just waiting for travis.

PR #957, opened 2014-05-20
Delivery class isn't determined properly for dialogue push triggers

justinvdm:
Dialogue delivery classes are stored in the `poll` config field, not the `config` config field, so pushes don't get the delivery class they need.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #958, opened 2014-05-20
New groups aren't listed after a conversation is created

justinvdm:
The dialogue application uses `contact_store.list_static_groups()` to get the list of static groups on the account for use in the dialogue group state, but for some reason, groups created after the conversation is created don't show up in this list.

### DIFF
--- a/go/apps/dialogue/tests/test_views.py
+++ b/go/apps/dialogue/tests/test_views.py
@@ -107,6 +107,11 @@ class TestDialogueViews(GoDjangoTestCase):
         conv_helper = self.setup_conversation(
             with_group=False, started=False, name=u"myconv")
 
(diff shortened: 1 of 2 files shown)


justinvdm:
Turns out this is just a dialogue conversation bug, where we are saving the list of groups with the rest of the poll.

justinvdm:
Ready for review.

We could also exclude the groups on the client side when we sync with the api, or have the api remove the groups before saving the poll. Not sure how useful either is though.

hodgestar:
I think this is a safer way to construct `model_data`, so :+1:.

I think it makes sense to have the client side exclude the groups, so I think we should do that too (although perhaps in a separate PR).

PR #961, opened 2014-05-22
Prepare for new config implementation

jerith:
This is work that needs to be done to support praekelt/vumi#806 but which does not depend on it.


justinvdm:
:+1:

PR #964, opened 2014-05-28
Document the Go dashboard API

smn:
Specifically JSON & the HTTP API for publishing stats from 3rd party apps.

### DIFF
--- a/docs/dashboards.rst
+++ b/docs/dashboards.rst
@@ -91,8 +91,90 @@ Displays the last value of a given metric together with the last value of
 the same metric for the given time range in history. Makes it easy to
 compare a metric value over a specific period of time.
 
+
+.. code-block:: javascript
+
+    {
+      "type": "diamondash.widgets.lvalue.LValueWidget",
+      "time_range": "1d",
+      "name": "Last Ping Count",
+      "target": {
+        "metric_type": "account",
+        "store": "default",
+        "name": "total_pings",
+        "aggregator": "max"
+      }
+    }
+
 GraphWidget
 ^^^^^^^^^^^
 
 Displays a line graph. Multiple metrics can be rendered on the same graph.
 
+.. code-block:: javascript
+
+    {
+      "type": "diamondash.widgets.graph.GraphWidget",
+      "name": "Ping Counts",
+      "width": 6,
+      "time_range": "1d",
+      "bucket_size": "1h",
+      "metrics": [
+        {
+          "name": "Pings",
+          "target": {
+            "metric_type": "account",
+            "store": "default",
+            "name": "total_pings",
+            "aggregator": "max"
+          }
+        }
+      ]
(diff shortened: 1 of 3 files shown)


smn:
@justinvdm & @hodgestar I know there are two more widgets available, the histogram and the pie chart. Where can I find example definitions for those?

Anyway, not particularly proud of these docs but it's better than what we have now.

smn:
@imsickofmaps care to read over as well?

justinvdm:
@smn don't think we have examples of those yet, will gist something.

justinvdm:
@smn https://gist.github.com/justinvdm/82486f054c330c93eff2

smn:
thanks, ready for review.

docs/dashboards.rst
```diff
@@ -0,0 +1,180 @@
...
+Available Widgets
+~~~~~~~~~~~~~~~~~
+
+The following widgets are available:
+
+LValueWidget
+^^^^^^^^^^^^
+
+Displays the last value of a given metric together with the last value of
+the same metric for the given time range in history. Makes it easy to
```
hodgestar: I found the first sentence here rather confusing. It uses `last value` twice to mean different things and I'm not sure what `the given time range in history` means.

docs/http_api.rst
```diff
@@ -1,9 +1,18 @@
+.. _http_api:
+
 Vumi Go's HTTP API
 ==================
 
-The API allows for sending & receiving Vumi messages via HTTP. These messages are plain JSON strings. Two types of messages are available, TransportUserMessages and TransportEvents. These messages can be received via streaming HTTP or can be pushed to 3rd party URL via HTTP POST. Outbound messages & metrics can be pushed to Vumi Go via HTTP PUT.
+The API allows for sending & receiving Vumi messages via HTTP. These
+messages are plain JSON strings. Two types of messages are available,
+TransportUserMessages and TransportEvents. These messages can be received
+via streaming HTTP or can be pushed to 3rd party URL via HTTP POST.
```
hodgestar, line 9: Are we officially deprecating the streaming HTTP API?

hodgestar:
Otherwise looks good.

docs/dashboards.rst
smn: It is super confusing and my english is failing me. Can you suggest a better way to describe the LValueWidget?

hodgestar: @justinvdm Could you maybe suggestion something here? The implementation seems to return values at `now() - time_range` and `now() - 2 * time_range` with `now()` possibly floored to the last whole interval of `time_range`, which seems sort of odd and complicated?

smn: @justinvdm if you've got time to update this that'd be great, then I can ping Daniela to let the guys needing know that the docs have been published.

PR #967, opened 2014-05-30
Fix jsbox_sending to dialogues with delivery_class set.

hodgestar:
Currently the delivery_class is looked up incorrectly.

### DIFF
--- a/go/scripts/jsbox_send.py
+++ b/go/scripts/jsbox_send.py
@@ -8,6 +8,7 @@ from vumi.service import Worker, WorkerCreator
 from vumi.servicemaker import VumiOptions
 import yaml
 
+from go.apps.dialogue.vumi_app import dialogue_js_config
 from go.apps.jsbox.outbound import mk_inbound_push_trigger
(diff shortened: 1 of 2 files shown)


go/scripts/jsbox_send.py
```diff
@@ -96,7 +97,11 @@ class JsBoxSendWorker(Worker):
     stdout = sys.stdout
     stderr = sys.stderr
 
-    SUPPORTED_APPS = ('jsbox', 'dialogue')
+    JSBOX_CONFIG = {
+        'jsbox': lambda conv: jsbox_js_config(conv.config),
```
justinvdm, line 101: This will throw an error for bad json. It probably doesn't matter, since this is a script and this will happen before the sends, but thought I should mention it in case.

justinvdm:
:+1: not sure if its worth waiting for travis, the changes are pretty isolated, your call.

hodgestar:
Tx. Going to land now so I can deploy this fix before it gets even later on a Friday.

PR #974, opened 2014-06-12
Optionally list more tag and pool information when printing accounts in go_list_accounts

hodgestar:
Currently we only print out the basic account information and sometimes one wants more, specifically information about tags acquired and pools accounts have access to.


hodgestar:
Ready for review.

jerith:
:+1: when Travis is happy.

PR #963, opened 2014-06-17
Editing a group router forces you to define a new one

hodgestar:
screenshot 2014-05-27 11 12 16

Checking the "Delete" box on the new one lets you continue, but that's not very optimal :grin:


hodgestar:
@imsickofmaps @jerith -- ready for review. Feedback on the has_changed craziness appreciated.

go/routers/group/view_definition.py
```diff
@@ -8,6 +8,18 @@ class RoutingEntryForm(forms.Form):
         label="Endpoint"
     )
 
+    def has_changed(self):
+        # has_changed is only used to detect whether extra forms have
+        # been altered when saving. 'group' is set in extra forms
+        # because it is a choice widget without a null value. We
+        # remove 'group' here so that this doesn't result in the
+        # formset attempting to save extra groups that haven't been
+        # modified.
```
jerith: Maybe we should add a null value (as the default)? Then we can add the new entry if both group and endpoint are set, ignore it if both are unset, or return a validation error if one is set but not the other.

hodgestar: I don't really want to add a null value because it's never a valid choice.

jerith: Having an invalid default choice forces one to make a choice instead of accidentally getting the (valid) default. I'd rather get a validation failure if I forget to choose a group than silently get a group I didn't explicitly choose, especially since the first group in the list is unlikely to be the desired selection most of the time.

hodgestar: Good point. I can live with that. Want to maybe have a stab at implementing
that?

hodgestar:
:+1:

PR #882, opened 2014-06-17
Remove router set up from the wizard

hodgestar:
Remove the ability to set up routers in the wizard. Instead rely on the routing page.

Please check out the screen shots here for an explanation of what needs to change in the front end:
https://docs.google.com/a/praekeltconsulting.com/document/d/1y1JTMqC9FDjFJa4sXjaHUsMFGybIdPiN6TLjep4HfUk/edit


hodgestar:
Ready for review -- plan is to deploy to a test instance before landing.

hodgestar:
Looks good on staging now.

go/wizard/views.py
```diff
@@ -50,38 +44,23 @@ def get(self, request):
...
-        if request.POST.get('channel_kind') == 'new':
-            forms_to_validate.append(chan_form)
-        elif request.POST.get('channel_kind') == 'existing':
-            forms_to_validate.append(router_form)
 
-        if not all(form.is_valid() for form in forms_to_validate):
+        forms_to_validate = [conv_form, chan_form]
+        forms_valid = [form.is_valid() for form in forms_to_validate]
+        if not all(forms_valid):
             # TODO: Better validation.
```
jerith: I think we have better validation now. Maybe we should add some tests for it and remove the comment?

jerith:
Looks good, aside from the comment about validation. Happy to land this if @badlard is happy with it.

go/wizard/views.py
hodgestar: Happy for that to happen but preferably in a different PR (at least I don't
know off the top of my head what the comment is really about or what the
tests should look like). Happy for it to be this PR if it's small enough.

PR #977, opened 2014-06-25
Contact view can access ``smart_form_group`` before it's defined.

jerith:
```
Stacktrace (most recent call last):

(4 additional frame(s) were not displayed)
...
  File "django/utils/decorators.py", line 91, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "go/contacts/views.py", line 133, in _group
    return _smart_group(request, contact_store, group)
  File "django/utils/decorators.py", line 91, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/contrib/auth/decorators.py",
  line 25, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "go/contacts/views.py", line 315, in _smart_group
    'group_form': smart_group_form,
```

I think this is an unhandled error path in the `if` statement higher up in the code.

### DIFF
--- a/go/contacts/views.py
+++ b/go/contacts/views.py
@@ -69,6 +69,7 @@ def groups(request, type=None):
                                          group.key)
             messages.info(request, '%d groups will be deleted '
                                    'shortly.' % len(groups))
+            return redirect(reverse('contacts:groups'))
         elif '_export' in request.POST:
             tasks.export_many_group_contacts.delay(
                 request.user_api.user_account_key,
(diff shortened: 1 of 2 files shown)


go/contacts/views.py
```diff
@@ -209,6 +209,11 @@ def _static_group(request, contact_store, group):
                 utils.store_file_hints_in_session(
                     request, file_name, file_path)
                 return redirect(_group_url(group.key))
+            else:
+                # We didn't get any useful POST variables, so just redirect
+                # back to the group page without doing anything.
+                return redirect(_group_url(group.key))
```
hodgestar, line 218: I think the `_remove` and `_complete_contact_upload` paths can also potentially not set `group_form`.

hodgestar, line 218: Also not sure why `_remove` doesn't use `elif` like everything else.

hodgestar:
Other than that, looks good.

hodgestar:
:+1:

PR #979, opened 2014-06-25
Contact imports treat created_at as an extra

justinvdm:
Exporting a contact, then importing them results in an extra being created for the `created_at` attribute, when it should be updating the actual `created_at` contact model attribute.


justinvdm:
Ready for review.

jerith:
:+1: when Travis is happy.

PR #980, opened 2014-07-03
GoRouterWorker routes events even when there is no next hop.

hodgestar:
It probably shouldn't.

Also, it has no tests of its own. :/


hodgestar:
Ready for review (tomorrow).

jerith:
:+1:

PR #984, opened 2014-07-08
Event URL should be optional in nostream HTTP API

jerith:


jerith:
I've just dropped the logging from WARNING to INFO for the event message publishing. It's still worth noting in the logs, but shouldn't end up in Sentry.

hodgestar:
Sometimes people set up HTTP APIs just to send outbound messages or metrics (i.e. they might want to leave `push_message_url` out too). If we made that always optional then we could leave out the special casing on the streaming form.

Otherwise looks good.

jerith:
If we do that, I'd rather make the "don't send me these things" setting explicit so that people don't accidentally set up a conversation which drops all inbound messages. That's a bigger change to the UI than this is, and I'm not sufficiently comfortable with Django form stuff to be confident I can make it non-horrible.

hodgestar:
I think just leaving the field blank is already a pretty strong indication
in this case and the messages aren't entirely lost because they're still in
the message store?

jerith:
An empty field is the default, and it's easy to forget to fill something in. I've been bitten by this sort of thing often enough that I don't really want to make it easy for other people to be bitten by it. (The event case is a little different because events are usually less useful than messages.)

Having an explicit switch is better in all cases, but "thing A works and thing B doesn't" is easier to debug than "neither thing works".

hodgestar:
I'm really unclear on why all your logic about using blank `push_message_url` or `push_event_url` is a bad idea doesn't apply to the the streaming `http_api` case (especially since that is the one where there are multiple kinds of things that can happen), but :+1:

jerith:
> I'm really unclear on why all your logic about using blank push_message_url or push_event_url is a bad idea doesn't apply to the the streaming http_api case

It does (moreso, as you pointed out), but we're deprecating that app and I don't really want to put a bunch of effort into changing something that isn't going to be used; especially since doing it properly would require adding a push-vs-stream flag as well and that would mean we'd have to migrate existing conversations and rewrite more of the app worker.

PR #986, opened 2014-07-09
Log 'Replacing routing entry ...' at the INFO level

justinvdm:
This case is normal enough for us to log at a level that doesn't end up in sentry.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #987, opened 2014-07-10
appmux router leaks redis connections

jerith:
It calls `SessionManager.from_redis_config()` (which creates a redis connection) on every inbound message.


hodgestar:
:+1:

PR #991, opened 2014-07-16
settings.py swallows all ImportErrors from production_settings import

jerith:
This means things break horribly and silently if production_settings fails to import something.


jerith:
There's no test for this because it's really hard to test. I tested it manually on my local machine, though.

hodgestar:
:+1:

hodgestar:
P.S. This has bitten me before too.

smn:
:+1: much better thanks

PR #993, opened 2014-07-17
Update dependencies to use released versions

jerith:
We now have releases of our three previously-github-only dependencies, so we should use them.

### DIFF
--- a/requirements-docs.pip
+++ b/requirements-docs.pip
@@ -4,4 +4,4 @@
 Sphinx
 sphinxcontrib-blockdiag>=1.3.1
 protobuf==2.5.0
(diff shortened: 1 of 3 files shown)


justinvdm:
:+1:

Guessing this means we will need releases for vxpolls and vumi-wikipedia before we can release vumigo?

jerith:
We already have releases of vxpolls and vumi-wikipedia, otherwise this branch won't work.

justinvdm:
Ah, I see.

justinvdm:
(This probably goes without saying, but my +1 depends on travis)

hodgestar:
\o/

jerith:
Bleh. vumi-wikipedia needs to package its test data.

PR #999, opened 2014-08-27
Manage message cache for arbitrary batch keys

jerith:
The `go_manage_message_cache` command currently operates only on batches belonging to conversations. We need a way to operate on arbitrary batches by specifying batch keys.


hodgestar:
:+1:

PR #1002, opened 2014-09-12
Expose transliterate_unicode config option in Wikipedia app

jerith:


hodgestar:
Do we have a test that checks that setting the config option results in the text sent being ASCII-ified?

jerith:
> Do we have a test that checks that setting the config option results in the text sent being ASCII-ified?

We have a test in this PR that asserts the setting is passed to the underlying worker and we have tests in the vumi-wikipedia repo that assert correct behaviour based on the setting.

hodgestar:
:+1:

PR #1003, opened 2014-09-18
Update diamondash dependency

justinvdm:


justinvdm:
Ready for review. As usual, there were other client side dependency updates in the process. We can stop this by pinning our dependencies to a specific version, but I'm not sure if we want to do that.

hodgestar:
I think broader updates are fine in this case.

hodgestar:
:+1:

justinvdm:
Thanks for review :) Going to wait for Travis.

PR #1026, opened 2014-10-15
Added an asyncattr

SteveBarnett:


SteveBarnett:
Based on discussion IRL with @justinvdm, seems this is a dead file.
Also, the bigger issue is big, lumpy, JS files, esp. vendor.

PR #1031, opened 2014-10-21
Warn that conversation progress display may be inaccurate

jerith:
Due to praekelt/vumi#861, we may not be tracking events. We should note this in the UI.


hodgestar:
:+1:

SteveBarnett:
*puts on UX / content hat*

Could we change

> Updates may be temporarily disabled.

to

> Sorry, we're having trouble tracking message statistics at the moment, but we are working on it. That means the statistics might be out of date.

?

And for conversation/includes/charts could we add

> At the moment, these statistics are only generated for JavaScript apps.

?

---

@jerith Is this factually correct? ^^^
Want me to add commits to this branch with these changes?

Does this affect dashboard stats too?

jerith:
If you're happy with the way the longer content looks, please make the change. (I didn't inspect the result at all, merely changed the templates.)

Is the "[...] only generated for JavaScript apps" relevant to this issue? I don't know if it's accurate, but even if it is that text will only be visible for jsbox conversations.

jerith:
I'd rather not remove `charts.html` in this PR. It's not directly related to the issue and carries additional risk.

SteveBarnett:
@hodgestar, still happy?

jerith:
:+1: from me on the changes I didn't make, so we're good to go once Travis is happy.

hodgestar:
Still :+1: from me too.

PR #1034, opened 2014-10-22
Add billing statement table to admin site

hodgestar:


hodgestar:
Ready for review.

rudigiesler:
Is the default admin what we want for now? If so, :+1:

hodgestar:
I think it is -- we can always add things to it later.

PR #1039, opened 2014-10-24
Create basic billing statement PDF export

rudigiesler:
Create a basic PDF view for billing statements, with no fancy HTML/CSS.


rudigiesler:
Hmm problem.. Reportlab requires python >= 2.7

go/billing/views.py
```diff
@@ -1 +1,57 @@
...
+
+
+# Convert HTML URIs to absolute system paths so xhtml2pdf can access those
+# resources
+def link_callback(uri, rel):
+    # use short variable names
+    sUrl = settings.STATIC_URL    # Typically /static/
+    sRoot = settings.STATIC_ROOT  # Typically /home/userX/project_static/
+    mUrl = settings.MEDIA_URL     # Typically /static/media/
+    mRoot = settings.MEDIA_ROOT   # Typically /home/userX/project_static/media/
```
hodgestar: We should probably name these `static_url`, `static_root`, etc to fit with our variable naming convention.

hodgestar:
I'm tempted to try and land this as part of this sprint's billing work since it's so close -- essentially we just need to write the HTML template and we have something to show people and we don't need to leave the PR open? Thoughts? Anything else you think is missing from this PR?

rudigiesler:
I can make a simple HTML doc, just with the basic line data in a table, and we can add fancy formatting  in another PR. We will also need to add a button somewhere in the UI, but I'm not sure if that should be in another PR or not.

hodgestar:
I added a button, reshuffled `mk_transaction` and `mk_statement` into a helper module and changed the statement PDF urls to `/billing/statement/<id>`. Let me know what you think.

rudigiesler:
It looks good. Next is to create the table in HTML, and rename the variables that you commented on (that block of code was copied from the xhtml2pdf docs).

rudigiesler:
@hodgestar Very basic HTML invoice done. Ready for review.

SteveBarnett:
@rudigiesler, you happy with that?

Also (ping, @hodgestar): who's going to be looking at billing?
Are "Tag Name" and "Tag Pool Name" the right words for them (and not channel or something like that)?

rudigiesler:
@SteveBarnett Yes I'm happy with that.

hodgestar:
@SteveBarnett A channel is roughly the same thing as a `(tag pool, tag)` pair. I think `channel` is probably slightly more understandable word, but I'm in favour of landing this, showing people what the PDFs look like and then iterating from there (I suspect people might want things grouped slightly differently, e.g. by channel type or tag pool, and we're likely missing a bunch of other things like letter heads, dates, etc -- i.e. easily enough for another PR).

hodgestar:
:+1:

Issue #965, opened 2014-05-28
Not giving a `config` key for config 500 errors not warns


imsickofmaps:

hodgestar:
@imsickofmaps Do you remember what this issue was about? Was it jsbox related?

imsickofmaps:
@hodgestar yes - starting a new conversation of that type didn't give you any bootstrapped setup for you to fill and just broke with no error if you forgot config

hodgestar:
Cool, so this should be addressed by the new ticket. :)

imsickofmaps:
@hodgestar closing because addressed by #1032

PR #1032, opened 2014-10-28
Set a default key and value for jsbox config

justinvdm:
**key:**
config

**value:**
{
    "name": "<slugified-conversation-name>"
}

---

Friends with #965?

### DIFF
--- a/go/wizard/views.py
+++ b/go/wizard/views.py
@@ -52,6 +52,7 @@ class WizardCreateView(BaseWizardView):
         if not all(forms_valid):
             logger.info("Validation failed: %s" % (
                 [frm.errors for frm in forms_to_validate],))
+            print frm.errors
             return self._render(request, conversation_form=conv_form,
                                 channel_form=chan_form)
 
@@ -82,10 +83,14 @@ class WizardCreateView(BaseWizardView):
     def _create_conversation(self, request, conv_data):
         conversation_type = conv_data['conversation_type']
 
+        name = conv_data['name']
+        description = conv_data['description']
         view_def = get_conversation_view_definition(conversation_type)
+        config = view_def._conv_def.get_default_config(name, description)
+
         conversation = request.user_api.new_conversation(
-            conversation_type, name=conv_data['name'],
-            description=conv_data['description'], config={},
+            conversation_type, name=name,
+            description=description, config=config,
             extra_endpoints=list(view_def.extra_static_endpoints),
         )
         return view_def, conversation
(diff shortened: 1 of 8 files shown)


justinvdm:
@hodgestar ready for review.

There seems to be quite a bit of duplication in the code for creating wizards and the create conversation view, this PR just adds to that :/ not sure if cleaning that up is something for this PR (if it is even worth it at this point).

The tests were a similar case. The wizard create view's tests use actual conversation types while the create conversation view tests use dummy conversation types. I ended up getting the wizard create view's tests to know about dummy conversation types too in order to get this feature tested, but it wasn't pretty :/ Let me know if you think there are ways of cleaning this up that are worth it.

go/wizard/views.py
```diff
@@ -52,6 +52,7 @@ def post(self, request):
         if not all(forms_valid):
             logger.info("Validation failed: %s" % (
                 [frm.errors for frm in forms_to_validate],))
+            print frm.errors
```
hodgestar: Stray print.

hodgestar:
I notice we're not currently slugifying the conversation name to generate the default config (we're just using the static string `conv-name`. Is the slugifying something we want to do? Also happy for that to happen in a separate PR if we do want to do it.

justinvdm:
@hodgestar yup, noticed that too, should've probably mentioned something. That problem hurts us for user-defined names too, so I think that change should live in the toolkit.

justinvdm:
@hodgestar Ah, that should not be a static string, sorry :/ fixing

justinvdm:
@hodgestar point about the config 'name' field being parsed inside the toolkit still applies though (the static string was a different thing).

hodgestar:
I think I'd be happier to use `slugify` (`from django.template.defaultfilters import slugify`) here on the conversation name because this is meant to be developer facing config and I (with my sandbox user hat on) think I'd rather the sandbox didn't mangle the name unnecessarily. Ideally validation would provide early feedback if the config is invalid (although we don't have that yet).

justinvdm:
@hodgestar Cool, validation for the config sounds better. Would django's slugify work though? Can't remember if our metric names can contain `-`s.

hodgestar:
The regex is `^[a-zA-Z][a-zA-Z0-9._]{,100}$` but maybe we should just add
dashes to the allowed list of characters?

justinvdm:
+1, though that means keeping this open until we have graphite understanding that. Okay with that though. Will add the slugifying so long.

SteveBarnett:
:+1: Like @hodgestar [said](https://github.com/praekelt/vumi-go/pull/1032#issuecomment-60754275): I think our developer persona (WIP!) would prefer slugified with dashes. :)

justinvdm:
Ready for re-review, though I don't think we should land this until after the retreat. There's a chance graphite won't like our idea to use dashes.

hodgestar:
:+1: and agreed on not landing until we're back.

SteveBarnett:
\o/. Nice one, @justinvdm and @hodgestar.

PR #1042, opened 2014-11-04
Add dialogue action for sending over a different channel

justinvdm:
We need a way to choose to send a single message over a different channel (for example, SMS) than the main channel the dialogue conversation is being run over (for example, USSD).


justinvdm:
@SteveBarnett Ready for review.

The content in the template probably needs changing, would like your input there.

Also still need to put a 'You need to hook up this channel on the routing screen' message somewhere. Only showing the message when it hasn't been hooked up is a bit more complicated, so we either need to come up with a message that makes sense either way, or add a context-dependant message in a separate PR.

hodgestar:
Looks good to me. I think the Travis build failure is spurious (it only happened on Python 2.6 and it was a timeout during the PhantomJS tests) so I restarted the build.

justinvdm:
@hodgestar thanks for review :) I'm going to wait for @SteveBarnett to take a look at the content, html and css parts, because I'm pretty sure some improvement will be needed there.

SteveBarnett:
@justinvdm: content in the template looks good to me. It's consistent voice and tone with the other bits. I think https://github.com/praekelt/vumi-go/issues/1046 will do the most to make this bit less confusing. I will add the need to hook up msg.

SteveBarnett:
@justinvdm I was thinking about adding "You can set this in [Configure channels](https://go.vumi.org/routing/)," but was worried about people clicking that and either losing work (from not saving) or getting lost (if we had used `target="_blank"` on the link).

justinvdm:
@SteveBarnett thanks for the change :) ya, also thought about that, I think its okay as is though.

Issue #836, opened 2014-01-11
Misconfigured sms_tag in jsbox can create invalid routing table.


smn:
When saving a jsbox config with a correctly formatted `sms_tag` but with invalid values (not a valid pool / tag combination) can result in the accounts routing table being corrupted.

Steps to recreate:
1. create a jsbox config with `{"sms_tag": ["sms", "longcode"]}`
2. got to the channel screen and connect something to it.
3. change the jsbox's config with a valid tag value.
4. load the channel screen, connect a chanenl to the new valid endpoint and try and save.
5. fails with error in screenshot below

![screenshot_2014_01_11__2_39_pm](https://f.cloud.github.com/assets/1065/1894785/acb06392-7afc-11e3-9a45-80abe1d58ab6.png)

hodgestar:
This issue is valid but the fact the original tag was "invalid" is
irrelevant -- the "sms_tag" value's only role currently is to form the
"<pool>:<tag>" endpoint name (and send_to_tag does the same thing).

The underlying issue is that changing the value of sms_tag breaks the
routing table if something was connected to the old endpoint.

smn:
It's screwed up my routing table to the point where I cannot create new, nor save existing, jsbox conversations.

smn:
Managed to fix my routing table by de-activating & archiving the conversation.

PR #1045, opened 2014-11-05
Automatically configure dialogue endpoints

justinvdm:
Once #1042 lands, we will need a way of automatically determining which endpoints to use based on the chosen channel types.

### DIFF
--- a/go/apps/dialogue/tests/test_vumi_app.py
+++ b/go/apps/dialogue/tests/test_vumi_app.py
@@ -15,7 +15,7 @@ from vumi.tests.helpers import VumiTestCase
 from vumi.tests.utils import LogCatcher
 
 from go.apps.dialogue.vumi_app import (
-    DialogueApplication, PollConfigResource)
+    DialogueApplication, PollConfigResource, dialogue_js_config)
 from go.apps.dialogue.tests.dummy_polls import simple_poll
 from go.apps.tests.helpers import AppWorkerHelper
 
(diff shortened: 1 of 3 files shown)


justinvdm:
Ready for review. Probably safe to land before #1042, though once for #1042 lands, it could be potentially dangerous until #836 is fixed.

go/apps/dialogue/tests/test_vumi_app.py
```diff
@@ -257,17 +257,61 @@ def setUp(self):
...
+    def test_config_endpoints(self):
+        poll = simple_poll()
+
+        poll['channel_types'] = [{
+            'name': 'sms',
+            'label': 'SMS'
+        }, {
+            'name': 'twitter',
+            'label': 'Twitter'
+        }]
```
hodgestar, line 297: I realized that I don't really know what `channel_types` is for which is making me worried I missed something important in one of the previous PRs. :/

hodgestar, line 297: Q: So `channel_types` is a list of possible endpoints and labels for them? What if one wants two endpoints with the same delivery class? (possibly not the most important use case, just thinking through stuff).

justinvdm, line 297: `channel_types` is basically the same as `go.vumitools.contacts.models.DELIVERY_CLASSES`. The UI needs it to know what channels people can choose to send over in the new 'send' state. At some point it will change to being only the relevant delivery classes. Happy to use `go.vumitools.contacts.models.DELIVERY_CLASSES` instead.

I'm calling it `channel_types`, since I'm guessing it will stick around when we replace delivery classes with channel types.

justinvdm, line 297: A: Spoke to @badlard about this. This is an intentional restriction to avoid confusing people. It is limiting though, a bit +-0 on it.

hodgestar, line 297: Currently there is at most one endpoint per delivery class / channel type. My question is whether this is a good idea?

justinvdm, line 297: Yup, this is an intentional restriction to avoid confusing people. I'm not quite sure if its a good idea, I guess it depends if there is a common use case for dialogues with multiple endpoints that have the same delivery class. @badlard and @SteveBarnett seemed to prefer this approach, and apparently the project this feature is for won't need multiple endpoints with the same delivery class. I'm still a bit wary about this decision though.

hodgestar, line 297: Let's stick with this for now -- I'll try convince myself (and @SteveBarnett and @badlard) that there is a better solution over coffee or lunch sometime. :)

hodgestar:
:+1:

badlard:
I'm happy to discuss this more and am open to allowing multiple end points
per channel per conversation. I just thought it would be simpler with the
restriction, but open to being more flexible.

On 5 November 2014 15:29, Simon Cross notifications@github.com wrote:

> [image: :+1:]
> 
> ## 
> 
> Reply to this email directly or view it on GitHub
> https://github.com/praekelt/vumi-go/pull/1045#issuecomment-61806450.

## 

_PRAEKELTFOUNDATION_.ORG
Sign up to receive Vumi's newsletter http://eepurl.com/Yb5Xn

PR #1049, opened 2014-11-05
Add conversation definition helper for updating config

justinvdm:
The plan is to handle endpoints that were connected and have now been removed here by first removing the routing entries for them. This should solve #836.

### DIFF
--- a/go/vumitools/conversation/definition.py
+++ b/go/vumitools/conversation/definition.py
@@ -2,16 +2,15 @@ from go.vumitools.metrics import (
     ConversationMetricSet, MessagesSentMetric, MessagesReceivedMetric)
 
 
-def detach_removed_endpoints(conv, user_api, old, new):
+def detach_removed_endpoints(conv, user_account, old, new):
     conn = conv.get_connector()
-    user_account = user_api.get_user_account()
     rt = user_account.routing_table
 
     for endpoint in set(old) - set(new):
         rt.remove_endpoint(conn, endpoint)
 
     rt.validate_all_entries()
-    user_account.save()
+    return user_account.save()
 
 
 class ConversationDefinitionBase(object):
@@ -68,13 +67,15 @@ class ConversationDefinitionBase(object):
 
         return endpoints
 
-    def update_config(self, user_api, config):
+    def update_config(self, user_account, config):
         old_endpoints = self.get_endpoints(self.conv.config)
         endpoints = self.get_endpoints(config)
-        detach_removed_endpoints(self.conv, user_api, old_endpoints, endpoints)
 
         self.conv.set_config(config)
         self.conv.extra_endpoints = endpoints
+
+        return detach_removed_endpoints(
+            self.conv, user_account, old_endpoints, endpoints)
 
     def is_config_valid(self):
         raise NotImplementedError()
--- a/go/vumitools/conversation/definition.py
+++ b/go/vumitools/conversation/definition.py
@@ -72,7 +72,7 @@ class ConversationDefinitionBase(object):
         endpoints = self.get_endpoints(config)
 
         self.conv.set_config(config)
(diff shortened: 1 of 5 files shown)


justinvdm:
Ready for review.
It feels a bit weird doing this routing stuff in the conversation definition, maybe we should be handling detaching entries for removed endpoints somewhere else? Maybe I'm overthinking it.

go/vumitools/conversation/definition.py
```diff
@@ -2,6 +2,27 @@
     ConversationMetricSet, MessagesSentMetric, MessagesReceivedMetric)
 
 
+def detach_endpoint(rt, conn, endpoint):
+    src = rt.lookup_source(conn, endpoint)
+
+    if src is not None:
+        rt.remove_entry(*src)
+
+    rt.remove_entry(conn, endpoint)
```
hodgestar: I wonder if we shouldn't add this method to `RoutingTable` as `remove_endpoint`?

go/vumitools/conversation/definition.py
```diff
@@ -2,6 +2,27 @@
...
+def detach_removed_endpoints(conv, user_api, old, new):
+    conn = conv.get_connector()
+    user_account = user_api.get_user_account()
+    rt = user_account.routing_table
+
+    for endpoint in set(old) - set(new):
+        detach_endpoint(rt, conn, endpoint)
+
+    rt.validate_all_entries()
+    user_account.save()
```
hodgestar: I think the conversation definition is meant to be usable from Twisted land too, which means we need to be careful about using methods that might return deferreds there. I think we can dodge this bullet by:
- Passing through user_account instead of user_api to update_config.
- Returning the result of user_account.save() here and in update_config.

hodgestar:
The other option is to handle endpoint updating in the view definition rather? The same way we have `get_endpoints` there?

go/vumitools/conversation/definition.py
```diff
@@ -47,6 +68,23 @@ def get_metrics(self):
...
+
+        return endpoints
+
+    def update_config(self, user_api, config):
+        old_endpoints = self.get_endpoints(self.conv.config)
+        endpoints = self.get_endpoints(config)
+        detach_removed_endpoints(self.conv, user_api, old_endpoints, endpoints)
+
+        self.conv.set_config(config)
+        self.conv.extra_endpoints = endpoints
```
hodgestar: This needs to be `self.conv.c.extra_endpoints` is `self.conf` is a wrapped conversation (bleh). This is the cause of the two failing jsbox tests.

hodgestar:
Left a few comments. A bit clunky (not your fault) but I think that's likely the best we can expect here.

go/vumitools/conversation/definition.py
jerith: Yes, the conversation definition lives in both worlds. The view definition is Django-only.

justinvdm: Ah right, forgot we need to worry about the async cases, thanks will fix.

go/vumitools/conversation/definition.py
justinvdm: +1

go/vumitools/conversation/definition.py
justinvdm: Ah, I thought the conversation wrapper proxied to the conversation. Could you explain why that doesn't work?

hodgestar: It only proxies for getting attributes, not for setting them. Proxying for
setting introduces lots of craziness.

justinvdm: Ah, thanks, makes sense.

justinvdm:
Ready for re-review.

hodgestar:
:+1: once Travis is happy.

PR #1052, opened 2014-11-06
Fix permissions for viewing billing statement PDFs.

hodgestar:
Currently admins can't see other people's statements (which makes the link in the admin interface not particularly useful).


rudigiesler:
:+1:

PR #1054, opened 2014-11-10
HTTP API needs to respond with correct Content-Type in all cases

jerith:
Apparently there are still places where it uses `text/html` instead of `application/json`.


justinvdm:
:+1:

PR #1055, opened 2014-11-10
Extend metrics middleware

hodgestar:
This should allow us to track via Graphite, for each transport, network operator and tag or tagpool:
- number of sessions
- number of messages
- time taken for each session
- time taken for each session, rounded to the next `N` seconds (this is to allow very rough session costs calculations since some networks bill per `N` seconds or part there-of).

See also praekelt/vumi#856.

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -336,7 +336,7 @@ class MetricsMiddleware(BaseMiddleware):
                 '%s.tagpool.%s' % (name, pool), msg, session_dt)
         if config.get('track_all_tags') or tagname in config['tags']:
             self.fire_inbound_metrics(
-                'tag.%s.%s' % (pool, tagname), msg, session_dt)
+                '%s.tag.%s.%s' % (name, pool, tagname), msg, session_dt)
 
     def fire_outbound_metrics(self, prefix, msg, session_dt):
         self.increment_counter(prefix, 'outbound')
@@ -364,7 +364,7 @@ class MetricsMiddleware(BaseMiddleware):
                 '%s.tagpool.%s' % (name, pool), msg, session_dt)
         if config.get('track_all_tags') or tagname in config['tags']:
             self.fire_outbound_metrics(
-                'tag.%s.%s' % (pool, tagname), session_dt)
+                '%s.tag.%s.%s' % (name, pool, tagname), msg, session_dt)
 
     @inlineCallbacks
     def handle_inbound(self, message, endpoint):
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -5,6 +5,7 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 
 from vumi.middleware.base import TransportMiddleware, BaseMiddleware
 from vumi.middleware.message_storing import StoringMiddleware
+from vumi.middleware.tagger import TaggingMiddleware
 from vumi.utils import normalize_msisdn
 from vumi.blinkenlights.metrics import (
     MetricPublisher, Count, Metric, MetricManager)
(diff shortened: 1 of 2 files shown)


go/vumitools/middleware.py
```diff
@@ -78,28 +78,81 @@ def is_optout_message(message):
...
+    :param dict tagpools:
+        A dictionary defining which tag pools and tags should be tracked.
+        E.g.::
+
+            tagpools:
+                pool1:
+                    __track_pool__: true
+                    tagA: true
+                pool2:
+                    tagB: true
```
jerith: I'd rather structure this differently, perhaps something like:

```
tagpools:
    pool1:
        track_pool: true
        track_tags: ["tagA"]
    pool2:
        track_pool: false
        track_tags: ["tagB"]
    pool3:
        track_pool: true
        track_all_tags: true
```

jerith:
Despite my comment about the config format, I don't really think we should be trying to track tag-level things here. It adds additional ordering-dependence to our already-complicated middleware stack and potentially introduces exciting new failure modes.

Maybe we need something on the routing table dispatcher?

hodgestar:
I like the suggestion for the tagpools config format. Changing to that.

hodgestar:
I'd prefer all of this to live on the transports because:
- messages bounce through the routing table dispatcher multiple times
- we're trying to measure the length of sessions, which might be different at the routing table dispatcher and the transport (and the transport value is more correct).

go/vumitools/middleware.py
```diff
@@ -78,28 +78,82 @@ def is_optout_message(message):
...
+    * The number of messages sent and received.
+    * The number of sessions started.
+    * The length of each session.
+
+    The network operator is determined by examining each message. If the
+    network operator is not detected by the transport, consider using network
+    operator detecting middleware to provide it.
+
+    Network operator metrics must be enabled by setting ``operator_metrics`` to
+    ``true``.
```
jerith, line 103: Should we maybe rename this to `provider_metrics`? That's the field we're using in the message object and it's more inclusive of things like twitter and mxit than "network operator".

hodgestar, line 103: Happy to rename to `provider_metrics`.

hodgestar:
Ready for review.

jerith:
> I'd prefer all of this to live on the transports because:
> - messages bounce through the routing table dispatcher multiple times

Agreed. We can use some tricks to limit this to transport connectors, but that gets ugly.

> - we're trying to measure the length of sessions, which might be different at the routing table dispatcher and the transport (and the transport value is more correct).

I was thinking that we could use the timestamps in the message objects, but that doesn't work so well for outbound messages.

I withdraw my objection, but I still wish there were a better way.

jerith:
This middleware could really benefit from using confmodel instead of raw dicts, but we should probably make that change across all middlewares in a separate PR.

go/vumitools/middleware.py
```diff
@@ -178,35 +258,160 @@ def set_inbound_timestamp(self, transport_name, message):
...
+                '%s.tagpool.%s' % (name, pool), msg, session_dt)
+        if config.get('track_all_tags') or tagname in config['tags']:
+            slugname = self.slugify_tagname(tagname)
+            self.fire_inbound_metrics(
+                '%s.tag.%s.%s' % (name, pool, slugname), msg, session_dt)
+
+    def fire_outbound_metrics(self, prefix, msg, session_dt):
+        self.increment_counter(prefix, 'outbound')
+        if session_dt is not None:
+            self.fire_session_dt(prefix, session_dt)
```
jerith, line 356: Do we need to handle outbound messages that start sessions? I don't think we have any of those at present, but it's something that will probably happen eventually even if we don't.

jerith:
After reviewing the changes (for which the diff view is useful), can we perhaps make this a new middleware (temporarily, at least) and revert the changes to the existing middleware? That will let us apply it to the small number of things that urgently need it with an easier rollback plan and without the risk of breaking other transports.

hodgestar:
Without configuration changes the new middleware only publishes a new
session metric and it has tests that cover that. Not sure what the
advantage of making a new temporary middleware is.

go/vumitools/middleware.py
hodgestar, line 356: Thought about this and it seemed corner case, but happy to add support for
it.

jerith:
> Without configuration changes the new middleware only publishes a new session metric and it has tests that cover that. Not sure what the advantage of making a new temporary middleware is.

The advantage is that unexpected interactions with the new code are limited to places where we explicitly use the new version rather than potentially cropping up in exotic or low-volume transports.

I'd agree with you if the changes were purely additive, but this branch also touches a bunch of existing code and therefore has the potential to break existing functionality.

jerith:
+1 to the most recent commit.

PR #1053, opened 2014-11-11
Get dialogue app's endpoints to show up

justinvdm:
At the moment, we figure out which endpoints to use in a dialogue app for the vumi app side, but not the UI side. We need to get the dialogue app's conversation definition to know how to do that.

### DIFF
--- a/go/apps/dialogue/dialogue_api.py
+++ b/go/apps/dialogue/dialogue_api.py
@@ -1,14 +1,27 @@
 """Go API action dispatcher for dialogue conversations."""
 
+from twisted.internet.defer import inlineCallbacks, returnValue
 from go.api.go_api.action_dispatcher import ConversationActionDispatcher
+
+
+def definition(conv):
+    # to avoid problems caused by circular dependency
+    from go.apps.dialogue.definition import ConversationDefinition
+    return ConversationDefinition(conv)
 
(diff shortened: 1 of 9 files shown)


justinvdm:
Ready for review.

go/apps/dialogue/dialogue_api.py
```diff
@@ -1,14 +1,27 @@
...
-        conv.config["poll"] = poll
-        d = conv.save()
-        d.addCallback(lambda r: {"saved": True})
-        return d
+        dfn = definition(conv)
+
+        config = conv.config.copy()
+        config["poll"] = poll
+
+        dfn.update_config((yield user_api.get_user_account()), config)
```
justinvdm, line 24: Since this method obtains and sets the conversation's extra endpoints, we are relying on the fact that all values are unicode because of the json deserializing. If it were to somehow give a non-unicode value as an endpoint, an exception would get raised. Not sure if its something we need to worry about, problem not, but it all seems quite implicit at the moment.

hodgestar, line 24: There are some cases where simplejson returns byte strings but I think Python's json module always returns unicode ones.

hodgestar:
:+1:

PR #1056, opened 2014-11-11
Use the same endpoint name in dialogue app and UI for send action

justinvdm:
At the moment, the app uses the delivery class name and the UI uses its label.


justinvdm:
Ready for review.

hodgestar:
:+1:

Sorry for missing that the endpoints were set up incorrectly in earlier reviews.

PR #1057, opened 2014-11-12
Fire sums for response time and session time metrics in addition to averages.

hodgestar:
Currently we only fire averages, but for many calculations (e.g. billing) it's useful to also have the total time, for which we need to fire sums.

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -10,7 +10,7 @@ from vumi.middleware.message_storing import StoringMiddleware
 from vumi.middleware.tagger import TaggingMiddleware
 from vumi.utils import normalize_msisdn
 from vumi.blinkenlights.metrics import (
-    MetricPublisher, Count, Metric, MetricManager)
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

go/vumitools/middleware.py
```diff
@@ -78,6 +78,13 @@ def is_optout_message(message):
         return message['helper_metadata'].get('optout', {}).get('optout')
 
 
+class TimeMetric(Metric):
+    """
+    A time-based metric that fires both sums and averages.
+    """
+    DEFAULT_AGGREGATORS = [AVG, SUM]
```
justinvdm, line 85: I vaguely remember us only handling a single aggregator in some part of vumi-go, but probably not a part that affects this.

justinvdm, line 85: :+1:

hodgestar, line 85: Yep. That's in parts that go via conversation and account metrics.

PR #1058, opened 2014-11-12
Better stripping of strange characters in tagnames when firing metrics.

hodgestar:
Currently we just strip out odd characters but perhaps we should replace them with `_` or `.` if they're not at the start or end of the number?

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -181,7 +181,8 @@ class MetricsMiddleware(BaseMiddleware):
     """
 
     KNOWN_MODES = frozenset(['active', 'passive'])
-    TAG_SLUGIFIER_RE = re.compile(r"[^a-zA-Z0-9_-]*")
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

go/vumitools/middleware.py
```diff
@@ -181,7 +181,8 @@ class MetricsMiddleware(BaseMiddleware):
     """
 
     KNOWN_MODES = frozenset(['active', 'passive'])
-    TAG_SLUGIFIER_RE = re.compile(r"[^a-zA-Z0-9_-]*")
+    TAG_STRIP_RE = re.compile(r"(^[^a-zA-Z0-9_-]+)|([^a-zA-Z0-9_-]+$)")
```
rudigiesler, line 184: What about `\w-` instead of `a-zA-Z0-9_-`?

hodgestar, line 184: \w does tricky things with the LOCALE which I was hoping to avoid.

rudigiesler:
:+1: when Travis is happy.

PR #1008, opened 2014-11-18
Command messages raise IgnoreMessage but only real connectors handle this.

jerith:
Our command message processing appears to raise `vumi.connectors.IgnoreMessage` via `get_config_for_conversation` and `get_config_for_router` but only real connectors handle this and the custom connectors for command messages don't (so the errors escape unhandled and the consumer stops).

### DIFF
--- a/go/routers/tests/helpers.py
+++ b/go/routers/tests/helpers.py
@@ -193,7 +193,7 @@ class RouterWorkerHelper(object):
     def dispatch_command(self, command, *args, **kw):
         cmd = VumiApiCommand.command(
             self._worker_name(), command, *args, **kw)
-        yield self.dispatch_raw('vumi.api', cmd)
+        yield self.ri.dispatch_raw('vumi.api', cmd)
         yield self.dispatch_commands_to_router()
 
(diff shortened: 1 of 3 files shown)


go/routers/tests/helpers.py
```diff
@@ -193,7 +193,7 @@ def dispatch_commands_to_router(self):
     def dispatch_command(self, command, *args, **kw):
         cmd = VumiApiCommand.command(
             self._worker_name(), command, *args, **kw)
-        yield self.dispatch_raw('vumi.api', cmd)
+        yield self.ri.dispatch_raw('vumi.api', cmd)
```
hodgestar, line 196: What is this change for? Was the helper doing the wrong thing?

hodgestar:
One small question, looks good though.

go/routers/tests/helpers.py
jerith, line 196: Yes, it was broken before. (Apparently nothing was using it.)

hodgestar:
:+1:

PR #1060, opened 2014-11-19
Update statement generation

justinvdm:
It should generate statements that can be displayed like this:
statement

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -5,23 +5,84 @@ from dateutil.relativedelta import relativedelta
 from celery.task import task, group
 
 from django.db.models import Sum
+from django.conf import settings as go_settings
 
 from go.billing import settings
-from go.billing.models import Account, Transaction, Statement, LineItem
+from go.billing.models import (
+    Account, Transaction, MessageCost, Statement, LineItem)
+from go.vumitools.api import VumiUserApi
+
+
+def get_tagpools(account):
+    config = go_settings.VUMI_API_CONFIG
+    user_api = VumiUserApi.from_config_sync(account.account_number, config)
+    return user_api.tagpools()
+
+
+def get_message_transactions(account, from_date, to_date):
+    transactions = Transaction.objects.filter(
+        account_number=account.account_number,
+        created__gte=from_date,
+        created__lt=(to_date + relativedelta(days=1)))
+
+    transactions = transactions.values(
+        'tag_pool_name', 'tag_name', 'message_direction')
+
+    transactions = transactions.annotate(
+        cost=Sum('message_cost'), credits=Sum('credit_amount'))
+
+    return transactions
+
+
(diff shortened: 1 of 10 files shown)


justinvdm:
Ready for review. Styling and static parts of the template changes still need to be done, but this PR is large enough already, will add those in a new PR.

justinvdm:
Failing travis test is because of ordering issue, fixing.

justinvdm:
Ok, should be fixed and reviewable now.

justinvdm:
Hmm, looks like it isn't an ordering issue :/ looking into it.

justinvdm:
Ah, it was an ordering issue, just not where I was first looking. It was passing locally but failing on travis because postgres was returning a different order, and I wasn't ensuring the right order before doing a `groupby`. Should be reviewable for real real now, billing tests all seem to be passing using both sqlite and postgres.

go/billing/tasks.py
```diff
@@ -62,6 +117,6 @@ def generate_monthly_account_statements():
     task_list = []
     for account in account_list:
         task_list.append(
-            generate_monthly_statement.s(account.id, from_date, to_date))
+            generate_monthly_statement.s(account, from_date, to_date))
```
hodgestar: I think this might need to remain `account.id` because real Celery needs to be able to serialize the objects passed as parameters to tasks?

justinvdm: Right, thanks. We'll need to pass in both `account.id` and `account.account_number` then (so we can use the user api). Not ideal :/   I guess we could just pass in the account number and lookup accounts from it, but `account_number` probably isn't indexed the same as the id, so it'd be slower. Thoughts?

hodgestar: What we used to do is to pass the `account.id` and then retrieve the account object from Django inside the task (and get the `account_number` from that).

go/billing/tests/test_views.py
```diff
@@ -39,8 +55,68 @@ def check_statement_not_accessible_by(self, username, statement):
...
+                'items': list(items.filter(channel=u'Tag 2.1')),
+            }, {
+                'name': u'Tag 2.2',
+                'items': list(items.filter(channel=u'Tag 2.2')),
+            }],
+        }])
+
     def test_statement_accessable_by_owner(self):
-        statement = statement = self.mk_statement()
+        statement = self.mk_statement()
```
hodgestar, line 119: Woops. Tx for cleaning this up.

go/billing/tasks.py
justinvdm: Ah, or that :) sorry, brain wasn't working there

hodgestar:
Few minor comments. Looks good though.

Update: Where by "few" I appear to mean "one". :)

justinvdm:
Ready for re-review, I think.

hodgestar:
:+1:

PR #868, opened 2014-11-19
Add support for periodically archiving the transaction billing table

hodgestar:
The transaction billing table has an entry per message which means it'll get large very quickly. We should archive data regularly (probably at the end of each month we should archive data older than N months).

PR #1062, opened 2014-11-20
Update version of requests we depend on.

hodgestar:
We currently depend on requests 0.14.2. Since the current version on PyPI is 2.4.3, we should consider updating.

What triggered this is the Travis failure https://travis-ci.org/praekelt/vumi-go/jobs/41489295 from #868 which is caused by a bug in older versions of requests that overwrote `six.moves` with its own module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
         'django-registration==1.0',
         'lesscpy==0.9h',
         'xlrd==0.8.0',
-        'requests==0.14.2',
+        'requests',
         'mock==1.0.1',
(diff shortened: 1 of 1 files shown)


hodgestar:
Ready for review one Travis is done.

@smn, @jerith, @justinvdm: Anyone remember if there was a specific reason we pinned requests at this strange low version number?

jerith:
According to git history and `requests`'s changelog, that was the current version at the time we started using it. (`1.0.0` was released a month or so later.)

:+1: to the change.

PR #1064, opened 2014-11-20
Handle cost changes in billing statements

justinvdm:
At the moment, we are assuming the cost of all units are the same.

A similar, related issue that will probably get solved by the same fix: https://github.com/praekelt/vumi-go/issues/1063

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -67,7 +67,6 @@ def get_channel_type(transaction, tagpools):
 def get_message_credits(transaction):
     cost = get_message_cost(transaction)
     markup = transaction['markup_percent']
-    print markup
     return MessageCost.calculate_message_credit_cost(cost, markup)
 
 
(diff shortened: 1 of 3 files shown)


justinvdm:
Hmm, tests failing for postgres, investigating.

justinvdm:
Ok, tests should be fixed now.

go/billing/tasks.py
```diff
@@ -26,54 +20,81 @@ def get_message_transactions(account, from_date, to_date):
...
+    return transaction['total_message_cost']
+
+
+def get_count(transaction):
+    return transaction['count']
+
+
+def get_message_unit_cost(transaction):
+    count = get_count(transaction)
+    return get_message_cost(transaction) / count if count != 0 else count
```
hodgestar: Super bikeshed / minor semantic niggle -- I think `else count` should be `else 0` since although they're numerical idental, we're really returning a zero cost, not a zero count?

hodgestar:
Other than the tiny niggle, looks good.

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #1061, opened 2014-11-21
Update billing statement pdf static content and styling

justinvdm:


justinvdm:
Ready for review. I think it makes sense to add styling in a separate PR.

justinvdm:
Depends on #1064

justinvdm:
Made the change to render to html instead of pdf based on our decision to rely on browsers for printing the statements to pdf for now.

I decided to replace the pdf rendering part with an html part instead of trying to keep both around, because I was running into a problem that was starting to rabbit-hole: we need to resolve the link to the stylesheet to an absolute file path, which is difficult in debugging and testing because of the magic django pipelines is doing. Since it doesn't really make sense to keep broken functionality around, I removed the pdf rendering bits, but they will be easy enough to add back later.

justinvdm:
#1064 landed and merged in, should be reviewable now.

hodgestar:
I don't see a final merge from develop and it's not just github that is confused -- my local branches show the same issue and attempting to merge develop gives all sorts of conflicts. Maybe you forgot to push? Although the last commit is at 17:49 SAST, so there isn't much time for more commits to have happened.

_confused_

justinvdm:
[this](https://github.com/praekelt/vumi-go/commit/1104c195bcdde66a467a57c4106b2a925514f64d) was the final merge from develop, there seems to be a duplicate commit, and another weird commit afterwards. Not sure how that happened, looking at my zsh history, doesn't look like I did anything strange. Going to try fix this up.

justinvdm:
Ok, reset back to before the merge, merged in develop, re-fixed the conflicts, commited the conflict resolution and pushed it, and things look right now. Sorry about that, really not sure what happened :/

go/billing/views.py
```diff
@@ -1,42 +1,44 @@
...
 
 
+def groupby(values, fn):
+    return sorted([(k, list(g)) for k, g in _groupby(values, fn)])
+
+
+def totals_from_items(items):
+    return {
+        'cost': sum(item.cost for item in items),
+        'credits': sum(item.credits for item in items),
```
hodgestar: Do we need to worry about situations where `cost` or `credits` are null for some items?

go/billing/settings.py
```diff
@@ -38,6 +38,13 @@
     settings, 'BILLING_STATEMENTS_DEFAULT_ORDER_BY', '-from_date')
 
 
+STATEMENT_CONTACT_DETAILS = {
```
hodgestar: We should also do the `getattr` trick here? I wonder if we don't want to have dummy or blank data for defaults? If someone else installed Vumi Go it would probably not be good for their statements to claim to be from Praekelt by default?

go/billing/tests/test_views.py
```diff
@@ -83,13 +96,106 @@ def test_statement_billers(self):
...
+            'credits': 250,
+        }])
+
+        user = self.user_helper.get_django_user()
+        response = self.get_statement(user, statement)
+
+        self.assertEqual(response.context['totals'], {
+            'cost': Decimal('700.0'),
+            'credits': 900,
+        })
```
hodgestar, line 312: Now that we're generating HTML instead of PDFs, I'm wondering whether it would be good to assert something things about the HTML generated? Mostly I'd like to test template expressions a bit in case there are typos in variable names, etc.

hodgestar:
Left some small comments for Monday. Generally looks good. :)

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #878, opened 2014-11-21
Implement low credit alerts in billing

rudigiesler:
Having the billing API alert the account holder when there credits drop below a specified threshold.

Currently the billing API marks this is marked as TODO at https://github.com/praekelt/vumi-go/blob/e4320d12c81bc122cc9b70f96492c59ee1a58290/go/billing/api.py#L755-L757


SteveBarnett:
@rudigiesler Looking good! Here are some quick thoughts on the email text.

The length is great: short and succint.

What's the main Call To Action for this mail? Notification plus please top up, right? The mail is: 1. you are out of credits. 2. here's how to get more credits. How _do_ they do this? Mail @badlard?

I think the tone could be a little more "serious," since it's about money, but [we are generally quite chirpy in our comms](https://gist.github.com/rudigiesler/0fa71cb1f7fb1858629f).

How about [something like this](https://gist.github.com/SteveBarnett/6c5929ab1efd3c7eafdd)?

rudigiesler:
@hodgestar: Ready for review.

go/billing/api.py
```diff
@@ -795,7 +795,8 @@ def create_transaction_interaction(self, cursor, account_number,
         alert_credit_balance = result.get('alert_credit_balance')
         if (credit_balance < alert_credit_balance and
                 credit_balance + credit_amount > alert_credit_balance):
-            pass  # TODO: Raise a Low Credits alert; somehow
+            LowCreditNotification.objects.create_notification(
+                account_number, result.get('alert_threshold'), credit_balance)
```
hodgestar, line 817: I'm pretty sure we can't call this here because the API runs under Twisted and not Django so we need to have this fire off a celery task directly instead (and have that created the LowCreditNotification). I suspect tests only pass because we don't have an billing api test that covers this case.

go/billing/tasks.py
```diff
@@ -293,3 +298,49 @@ def generate_chunks(item_iter, items_per_chunk=10000, sep="\n"):
         archive.save()
 
     return archive
+
+
+@task()
+def low_credit_notification_confirm_sent(res, notification):
```
hodgestar: Just a note that if this is a task, it's arguments need to be serializable. I think Django model instances might be fine, but if not we'll need to send the notification id/pk and then retrieve it here.

go/billing/tasks.py
```diff
@@ -293,3 +298,49 @@ def generate_chunks(item_iter, items_per_chunk=10000, sep="\n"):
...
+def create_low_credit_notification(account_id, threshold, balance):
+    """
+    Sends a low credit notification. Returns (model instance, email_task).
+    """
+    account = Account.objects.get(pk=account_id)
+    notification = LowCreditNotification(
+        account=account, threshold=threshold, credit_balance=balance)
+    # Send email
+    subject = 'Vumi Go %s at %s%% of available credits' % (
+        account, threshold)
```
hodgestar: Perhaps we should use `account.user.email` here? The `__str__` of `account` contains the account ID which perhaps isn't right for the subject?

hodgestar: Or `.get_full_name`?

go/billing/tasks.py
jerith: Even if Django model instances are serializable, we probably don't want to be serializing them.

go/billing/templates/billing/low_credit_notification_email.txt
```diff
@@ -0,0 +1,14 @@
+Hi {{user.get_full_name}}!
+
+Please note that your account {{account}} has reached {{threshold_percent}}% of the total available credits. Your remaining balance is {{credit_balance}}.
+
+Once you have used all of your credits, your account will be capped and you will no longer be able to send any messages. Incoming messages will still be stored as usual.
+
+To top up your credits contact your Praekelt programme manager or reply back to this email with the number of messages/sessions you want loaded
```
hodgestar: Missing full stop at end of sentence?

go/billing/tests/test_models.py
```diff
@@ -171,3 +174,33 @@ def test_unicode(self):
         trans.save()
         self.assertNotEqual(trans.pk, None)
         self.assertEqual(unicode(trans), unicode(trans.pk))
+
+
+class TestLowCreditNotification(GoDjangoTestCase):
+    def setUp(self):
+        settings.EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
+        settings.CELERY_EMAIL_BACKEND = (
+            'django.core.mail.backends.locmem.EmailBackend')
```
hodgestar: We should use `self.vumi_helper.patch_settings` here instead so that the changes are undone after the test finishes.

go/billing/tests/test_tasks.py
```diff
@@ -416,3 +419,39 @@ def mk_transaction_set(n, created):
             transactions_before | transactions_after)
         self.assert_archive_in_s3(
             bucket, archive.filename, transactions_within)
+
+
+class TestLowCreditNotificationTask(GoDjangoTestCase):
+    def setUp(self):
+        settings.EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
+        settings.CELERY_EMAIL_BACKEND = (
+            'django.core.mail.backends.locmem.EmailBackend')
```
hodgestar: Same comment applies here. Might also be worth noting why we need to do this here (i.e. because Django overrides these before tests start so setting them in test settings doesn't help).

go/billing/tests/test_tasks.py
```diff
@@ -416,3 +419,39 @@ def mk_transaction_set(n, created):
...
+    def test_email_sent(self):
+        notification, res = self.mk_notification('70.1', '12.34')
+        self.assertTrue(res.get() is not None)
+        self.assertEqual(len(mail.outbox), 1)
+        [email] = mail.outbox
+
+        self.assertEqual(email.recipients(), [self.django_user.email])
+        self.assertTrue('Vumi Go' in email.subject)
+        self.assertTrue('70.1%' in email.subject)
+        self.assertTrue(str(self.acc) in email.subject)
```
hodgestar: Might be nicer to just assert on exactly what we expect the subject to be?

go/billing/api.py
```diff
@@ -795,7 +796,8 @@ def create_transaction_interaction(self, cursor, account_number,
         alert_credit_balance = result.get('alert_credit_balance')
         if (credit_balance < alert_credit_balance and
                 credit_balance + credit_amount > alert_credit_balance):
-            pass  # TODO: Raise a Low Credits alert; somehow
+            create_low_credit_notification.delay(
+                account_number, result.get('alert_threshold'), credit_balance)
```
hodgestar, line 817: This won't work because `go/billing/api.py` runs under Twisted -- i.e. we need to use something like deferToThread (http://twistedmatrix.com/documents/9.0.0/api/twisted.internet.threads.deferToThread.html) to call this. If you like, I can try write an example and add it to this branch?

hodgestar:
Mostly looks good, although we still need to figure out the calling-delay-from-Twisted issue. Left a few other comments.

go/billing/api.py
rudigiesler, line 817: I've pushed my commits for fixes for the other comments. Either add it to this branch or add a gist to a comment.

hodgestar:
Other things look good -- will try update the PR with deferToThread in a bit.

go/billing/api.py
```diff
@@ -795,7 +796,8 @@ def create_transaction_interaction(self, cursor, account_number,
         alert_credit_balance = result.get('alert_credit_balance')
         if (credit_balance < alert_credit_balance and
                 credit_balance + credit_amount > alert_credit_balance):
```
hodgestar: Not related to this PR, but it looks like we're mishandling an edge case here -- if at some point `credit_balance == alert_credit_balance`, we'll never alert? Probably we need to change one of these to a `<=`? Maybe `credit_balance <= alert_credit_balance < credit_balance + credit_amount`?

hodgestar: Or maybe `credit_balance - credit_amount <= alert_credit_balance < credit_balance`?

hodgestar:
@rudigiesler deferToThread added and edge-case in notifications fixed. I'm :+1: on your commits. If you could review mine then once we're happy we can add this PR to this of billing PRs to merge carefully.

I'm wondering whether we don't need a global config switch to turn notifications on and off so we can turn them off if we need to? And have it default to off so we can land this without suddenly sending notifications to everyone?

go/billing/api.py
```diff
@@ -793,9 +809,11 @@ def create_transaction_interaction(self, cursor, account_number,
 
         credit_balance = result.get('credit_balance')
         alert_credit_balance = result.get('alert_credit_balance')
-        if (credit_balance < alert_credit_balance and
-                credit_balance + credit_amount > alert_credit_balance):
-            pass  # TODO: Raise a Low Credits alert; somehow
+        if (credit_balance - credit_amount <= alert_credit_balance
+            < credit_amount):
```
rudigiesler: I'm not sure that this is equivalent, ie. 

```
>>> credit_balance = 97
>>> alert_credit_balance = 100
>>> credit_amount = 5
>>> credit_balance < alert_credit_balance and credit_balance + credit_amount > alert_credit_balance
True
>>> credit_balance <= alert_credit_balance < credit_balance + credit_amount
True
>>> credit_balance - credit_amount <= alert_credit_balance < credit_amount
False
```

rudigiesler:
@hodgestar : Just the one comment. We can have a global switch. It shouldn't send messages to everyone, since it only checks if it crosses the boundary, so only accounts that are sending messages and crossing the boundary will get messages. A global config switch would be nice though, so that we can try it out on QA first.

go/billing/settings.py
```diff
@@ -44,6 +44,8 @@
         'website': 'www.praekeltfoundation.org',
     })
 
+ENABLE_LOW_CREDIT_NOTIFICATION = False
```
hodgestar: We should do the same `getattr(settings, 'GO_ENABLE_LOW_CREDIT_NOTIFCATION', False)` trick that the other settings in this file use (so we can set this from Django settings).

hodgestar:
Other than the minor settings comment, looks good.

hodgestar:
:+1: !!

PR #1066, opened 2014-11-22
Add approximate total unique user counts to go_system_stats.

hodgestar:
Currently the stats only include total inbound and outbound uniques and don't provide anything that gives a reasonable estimate of total uniques. The idea here is to assume that inbound and outbound addresses that interact with a conversation overlap significantly and maintain a total of `max(inbound_uniques, outbound_uniqes)`.


hodgestar:
Ready for review.

smn:
:+1:

PR #1067, opened 2014-11-24
Disable fallback search for contact lookup

jerith:
We still have a contact-for-address fallback mechanism that uses search, for ancient and decrepit contacts that predate the appropriate indexes (which were added six months ago).

I'm reasonably sure we've migrated all the contacts since then, and the additional expense of a search feeding into a mapreduce every time we look for a contact for a new address is way more than we can manage at high load.

### DIFF
--- a/go/vumitools/contact/models.py
+++ b/go/vumitools/contact/models.py
@@ -158,7 +158,7 @@ class ContactStore(PerAccountStore):
     # but will result in false negatives if there are matching contacts that
     # have not yet been migrated to version 2.
     FIND_BY_INDEX = True
-    FIND_BY_INDEX_SEARCH_FALLBACK = True
+    FIND_BY_INDEX_SEARCH_FALLBACK = False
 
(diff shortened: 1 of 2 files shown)


rudigiesler:
Is there a way we can ensure that all the contacts are migrated?

jerith:
> Is there a way we can ensure that all the contacts are migrated?

Not easily, but any contact stored in the past six months will have the new indexes. I think we can reasonably not care about unmigrated contacts that haven't been touched since then.

We should probably rerun whatever "migrate all the contacts" tool we have, though.

rudigiesler:
+1 for migrating all contacts, if it's not something that will take forever. I think it is valuable just to make sure before making this change. The code looks fine to me, but I would prefer it if someone else would +1 it.

hodgestar:
I'm :+1:.

We can migrate the contacts later -- we have the tools and failing to find a contact is better than failing on a map reduce error.

PR #1068, opened 2014-11-25
Change email testing backend to celery

rudigiesler:
The current mail backend for testing is the django in-memory mail backend. This should be changed to use the celery mail backend, and the celery mail backend should be set to the django in-memory mail backend (for tests).


rudigiesler:
@hodgestar Ready for review. It seems like all tests are okay with this change. I don't know if you want to add any tests with this change?

hodgestar:
@rudigiesler I don't think we need new tests.

:+1:

PR #1069, opened 2014-11-25
Limit concurrent HTTP API send requests

jerith:
We currently accept as many concurrent "send message" requests as we're given. This branch adds a limit to avoid overwhelming the rest of the system when a single client pushes a high volume of messages.

### DIFF
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -164,10 +164,17 @@ class MessageResource(BaseResource):
             return
 
         in_reply_to = payload.get('in_reply_to')
-        if in_reply_to:
-            yield self.handle_PUT_in_reply_to(request, payload, in_reply_to)
-        else:
-            yield self.handle_PUT_send_to(request, payload)
+        user_account = request.getUser()
+        d = self.worker.concurrency_limiter.start(user_account)
+        try:
+            yield d  # Wait for our concurrency limiter to let us move on.
+            if in_reply_to:
+                yield self.handle_PUT_in_reply_to(
+                    request, payload, in_reply_to)
+            else:
+                yield self.handle_PUT_send_to(request, payload)
+        finally:
+            self.worker.concurrency_limiter.stop(user_account)
 
     @inlineCallbacks
     def handle_PUT_in_reply_to(self, request, payload, in_reply_to):
--- a/go/apps/http_api_nostream/tests/test_vumi_app.py
+++ b/go/apps/http_api_nostream/tests/test_vumi_app.py
@@ -15,60 +15,59 @@ from vumi.tests.utils import MockHttpServer, LogCatcher
 from vumi.tests.helpers import VumiTestCase
 
 from go.apps.http_api_nostream.vumi_app import (
-    ConcurrencyLimiter, NoStreamingHTTPWorker)
+    ConcurrencyLimitManager, NoStreamingHTTPWorker)
 from go.apps.http_api_nostream.resource import ConversationResource
 from go.apps.tests.helpers import AppWorkerHelper
 
 
-class TestConcurrencyLimiter(VumiTestCase):
+class TestConcurrencyLimitManager(VumiTestCase):
     def test_concurrency_limiter_no_limit(self):
         """
-        When given a negitive limit, ConcurrencyLimiter never blocks.
+        When given a negitive limit, ConcurrencyLimitManager never blocks.
         """
-        limiter = ConcurrencyLimiter(-1)
+        limiter = ConcurrencyLimitManager(-1)
         d1 = limiter.start("key")
         self.assertEqual(d1.called, True)
         d2 = limiter.start("key")
         self.assertEqual(d2.called, True)
 
         # Check that we aren't storing any state.
-        self.assertEqual(limiter._concurrents, {})
-        self.assertEqual(limiter._waiters, {})
+        self.assertEqual(limiter._concurrency_limiters, {})
 
         # Check that stopping doesn't explode.
         limiter.stop("key")
 
     def test_concurrency_limiter_zero_limit(self):
         """
-        When given a limit of zero, ConcurrencyLimiter always blocks forever.
+        When given a limit of zero, ConcurrencyLimitManager always blocks
+        forever.
         """
-        limiter = ConcurrencyLimiter(0)
+        limiter = ConcurrencyLimitManager(0)
         d1 = limiter.start("key")
         self.assertEqual(d1.called, False)
         d2 = limiter.start("key")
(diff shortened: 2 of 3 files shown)


go/apps/http_api_nostream/tests/test_vumi_app.py
```diff
@@ -187,7 +323,33 @@ def test_send_to(self):
...
+        without invasive changes to the app worker.
+        """
+        yield self.start_app_worker({'worker_concurrency_limit': 0})
+        msg = {
+            'to_addr': '+2345',
+            'content': 'foo',
+            'message_id': 'evil_id',
+        }
+
+        # TaggingMiddleware.add_tag_to_msg(msg, self.tag)
```
smn: stray?

jerith: That was in the test I copied. I'll clean it up in both places, thanks.

smn:
:+1: but what if someone just opens 500 requests, are we happy keeping those connections open? What will nginx do? Will it time them out? Then what?

jerith:
> but what if someone just opens 500 requests, are we happy keeping those connections open? What will nginx do? Will it time them out? Then what?

We'll process each request in turn until there are none left. We don't explicitly limit the number of concurrent open connections in the worker, although our load balancer has an overall connection limit per worker. The main reason we're doing this is to limit the potential flood of inbound messages that can interfere with other workers.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -33,11 +33,82 @@ class HTTPWorkerConfig(GoApplicationWorker.CONFIG_CLASS):
...
+        "zero disables the limit.",
         default=10)
     timeout = ConfigInt(
         "How long to wait for a response from a server when posting "
         "messages or events", default=5, static=True)
+    worker_concurrency_limit = ConfigInt(
+        "Maximum number of clients per account per worker. A value less than "
+        "zero disables the limit. (Unlike concurrency_limit, this queues "
+        "requests instead of rejecting them.)",
+        default=1, static=True)
```
hodgestar, line 45: I'd be a bit happier if we allowed `None` for this and had `None` mean no limit. Or if we allowed `0` to mean no limit. Just feel that configs that look like `worker_concurrency_limt: -1` are a bit strange.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -33,11 +33,82 @@ class HTTPWorkerConfig(GoApplicationWorker.CONFIG_CLASS):
...
+        if d is not None:
+            self._inc_concurrent(key)
+            d.callback(None)
+
+    def start(self, key):
+        if self._limit < 0:
+            # Special case for no limit, never block.
+            return succeed(None)
+        elif self._limit == 0:
+            # Special case for limit of zero, always block forever.
```
hodgestar, line 128: We might want to document this in the config docs.

go/apps/http_api_nostream/vumi_app.py
jerith, line 45: I'll have to see if `confmodel` allows that. I mostly chose this mechanism because it's what we use a few lines up for `concurrency_limit`.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -33,11 +33,82 @@ class HTTPWorkerConfig(GoApplicationWorker.CONFIG_CLASS):
...
+        self._limit = limit
+        self._concurrents = {}
+        self._waiters = {}
+
+    def _get_concurrent(self, key):
+        return self._concurrents.get(key, 0)
+
+    def _inc_concurrent(self, key):
+        concurrents = self._get_concurrent(key) + 1
+        self._concurrents[key] = concurrents
```
hodgestar: Could we just use `len(self._waiters[key])` instead of maintaining the `concurrents` counts? If feels like we do a lot of work to keep these two in sync.

go/apps/http_api_nostream/vumi_app.py
jerith, line 128: I'd rather just get rid of it, actually. It's only there to avoid collecting an ever-growing list of deferreds that will never fire if we don't allow any concurrents.

go/apps/http_api_nostream/vumi_app.py
hodgestar: I'm also wondering whether we can replace this thing with a suitable deferred queue?

hodgestar:
Conceptually I think this is worth trying. I left some questions about the implementation approach. I also have some questions about how this will affect people making requests in production, but I think we'll only learn this by trying things.

go/apps/http_api_nostream/vumi_app.py
jerith: They're not things to keep in sync. The concurrent count is the number of things that have been started but not stopped and the waiters list is for things that are waiting to start.

jerith: I should document this in the code, though.

hodgestar:
I'm wondering how much this will limit the rate in practice -- it only blocks while we're busy pushing things to Rabbit, right? So likely the rate will still be very high. We could also just limit this at the HAProxy level, maybe?

go/apps/http_api_nostream/vumi_app.py
hodgestar: Ah, cool. Let me re-read this bit of the code.

go/apps/http_api_nostream/resource.py
```diff
@@ -164,10 +164,17 @@ def handle_PUT(self, request):
...
 
         in_reply_to = payload.get('in_reply_to')
-        if in_reply_to:
-            yield self.handle_PUT_in_reply_to(request, payload, in_reply_to)
-        else:
-            yield self.handle_PUT_send_to(request, payload)
+        user_account = request.getUser()
+        d = self.worker.concurrency_limiter.start(user_account)
+        try:
+            yield d  # Wait for our concurrency limiter to let us move on.
```
hodgestar, line 170: Another question -- how much of an issue is that we accept a request and then pause holding the request open? Does this mean we'll potentially use up all our ports and fall over or run out of RAM, etc?

jerith, line 170: The load balancer will still limit the total number of concurrent requests each worker sees, so we're not going to get unbounded growth here. We're definitely going to have more connections open for longer in the cases where we see lots of concurrent connections for the same account, but we read the entire request body before we apply the concurrency limit so we're not going to be left with concurrency slots taken up by requests that die before they finish sending us their data.

hodgestar:
:+1:

PR #995, opened 2014-11-25
jsbox unicode stuff

jerith:
We're seeing a lot of unicode-related errors being logged from jsbox apps, which seems to indicate implicit bytes<->unicode conversions somewhere.

### DIFF
--- a/go/apps/jsbox/log.py
+++ b/go/apps/jsbox/log.py
@@ -79,9 +79,10 @@ class GoLoggingResource(LoggingResource):
         campaign_key = conv.user_account.key
         conversation_key = conv.key
 
-        internal_msg = "[Account: %s, Conversation: %s] %s" % (
+        # The keys may be unicode, so make everything unicode and then encode.
+        internal_msg = u"[Account: %s, Conversation: %s] %r" % (
             campaign_key, conversation_key, msg)
-        log.msg(internal_msg, logLevel=level)
+        log.msg(internal_msg.encode("ascii"), logLevel=level)
 
         yield self.log_manager.add_log(campaign_key, conversation_key,
                                        msg, level)
--- a/go/apps/jsbox/tests/test_log.py
+++ b/go/apps/jsbox/tests/test_log.py
@@ -152,6 +152,11 @@ class TestGoLoggingResource(ResourceTestCaseBase, LogCheckerMixin):
         ])
 
     @inlineCallbacks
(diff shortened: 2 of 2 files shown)


go/apps/jsbox/log.py
```diff
@@ -79,9 +79,10 @@ def log(self, api, msg, level):
         campaign_key = conv.user_account.key
         conversation_key = conv.key
 
-        internal_msg = "[Account: %s, Conversation: %s] %s" % (
+        # The keys may be unicode, so make everything unicode and then encode.
+        internal_msg = u"[Account: %s, Conversation: %s] %r" % (
             campaign_key, conversation_key, msg)
-        log.msg(internal_msg, logLevel=level)
+        log.msg(internal_msg.encode("ascii"), logLevel=level)
```
hodgestar, line 85: I'd be happier with encoding to UTF-8 here because it's the same for ASCII characters but won't trip up on non-ASCII characters if they sneak in somehow.

go/apps/jsbox/tests/test_log.py
```diff
@@ -152,6 +152,16 @@ def test_handle_info(self):
         ])
 
     @inlineCallbacks
-    def test_handle_info_failure(self):
-        yield self.assert_bad_command(
-            'info', u'Value expected for msg')
```
hodgestar, line 157: Shouldn't we keep this test?

hodgestar:
Two questions, but looks good.

go/apps/jsbox/log.py
hodgestar, line 85: This is not a particularly strong opinion.

go/apps/jsbox/tests/test_log.py
jerith, line 157: Whoops, I didn't mean to delete that. Thanks.

go/apps/jsbox/log.py
jerith, line 85: All three things in the message are supposed to be ASCII, so I'd prefer to get exceptions if they aren't. At worst, this gets us back to our current situation but without the implicit conversion silliness. The failure mode for UTF-8 encoding is that our logs fill up with non-ASCII bytes which can potentially break whatever we feed them to.

hodgestar:
:+1:

hodgestar:
We might want to wait for Travis to finish in case other tests are affected.

PR #1063, opened 2014-11-26
Billing statement message costs don't include session costs

justinvdm:
At the moment we assume they do.

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -85,7 +85,8 @@ def get_message_unit_cost(transaction):
 
 def get_session_unit_cost(transaction):
     count = get_count(transaction)
-    return get_session_cost(transaction) / count if count != 0 else count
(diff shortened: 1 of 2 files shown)


justinvdm:
Depends on #1064, otherwise ready for review.

hodgestar:
Looks good, although I'd like to quickly take a final look after #1064 lands.

One point which I think needs discussion is whether credits or cost should be considered fundamental in our statement generation. Unless the cost-per-credit changes (which is unlikely) the point is a bit moot, but current we treat costs as the basic unit (i.e. what we sum from the transactions) and then derive credits (by calculating) but the credits are really what we deduct when transactions happen, so perhaps it would be safer to sum those and then calculate the $ costs?

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #1072, opened 2014-11-26
Update requests usage to match later requests.

hodgestar:
```
File "/var/praekelt/vumi-go/ve/local/lib/python2.7/site-packages/django/views/generic/base.py", line 86, in dispatch
    return handler(request, *args, **kwargs)

  File "/var/praekelt/vumi-go/go/apps/dialogue/view_definition.py", line 40, in get
    model_data = r.json['result']['poll']

TypeError: 'instancemethod' object has no attribute '__getitem__'
```

### DIFF
--- a/go/api/go_api/tests/utils.py
+++ b/go/api/go_api/tests/utils.py
@@ -24,27 +24,27 @@ class MockRpc(object):
         return self.response
 
     @staticmethod
-    def make_response_data(result={}, id=None):
-        return {
+    def make_response_json_func(result={}, id=None, error=None):
+        response_data = {
             'id': id,
             'jsonrpc': '2.0',
             'result': result,
         }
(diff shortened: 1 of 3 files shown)


jerith:
This needs tests.

jerith:
I'd be happier if we had a test that used requests so we can avoid this sort of problem in the future.

justinvdm:
Looks good. I'm :+1: on this landing if we got tests to fail and then pass locally, don't think waiting for Travis is worth it if things are currently broken because of this.

hodgestar:
@jerith I'm happy to open a ticket for converting this to request sessions based tests now that I know how to do that? Would like to land this though so we can fix production.

jerith:
:+1: with a ticket for the tests that we work on today.

PR #1071, opened 2014-11-26
Add account fee to billing statements

justinvdm:
The plan is to do this with a simple config field for now, and maybe think about having this be settable via the admin UI if we end up needing that.

### DIFF
--- a/go/billing/models.py
+++ b/go/billing/models.py
@@ -332,7 +332,7 @@ class LineItem(models.Model):
         max_length=100, blank=True, null=True,
         help_text=_("Description of the item being billed"))
 
-    units = models.IntegerField('
+    units = models.IntegerField(
         default=0,
         help_text=_("Number of units associated to the item"))
 
--- a/go/billing/models.py
+++ b/go/billing/models.py
@@ -314,31 +314,32 @@ class LineItem(models.Model):
         help_text=_("The statement this line item is from."))
 
     billed_by = models.CharField(
-        max_length=100, blank=True, default='',
+        max_length=100, blank=True, null=True,
         help_text=_("Name of the entity the item is being billed for"))
 
     channel = models.CharField(
-        max_length=100, blank=True, null=True, default='',
+        max_length=100, blank=True, null=True,
         help_text=_("Name of the channel messages were sent/received over, "
                     "or null if there is no associated channel"))
 
     channel_type = models.CharField(
-        max_length=100, blank=True, null=True, default='',
+        max_length=100, blank=True, null=True,
         help_text=_("The type of channel messages were sent/received over "
                     "(e.g. SMS or USSD), or null if there is no associated"
                     "channel"))
 
     description = models.CharField(
-        max_length=100, blank=True, default='',
+        max_length=100, blank=True, null=True,
         help_text=_("Description of the item being billed"))
 
-    units = models.IntegerField(
+    units = models.IntegerField('
         default=0,
         help_text=_("Number of units associated to the item"))
 
     credits = models.IntegerField(
-        default=0,
-        help_text=_("Total cost of the item in credits"))
+        default=0, blank=True, null=True,
+        help_text=_("Total cost of the item in credits, or null if there is "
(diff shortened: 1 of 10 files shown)


justinvdm:
Depends on #1063, otherwise ready for review.

go/billing/models.py
```diff
@@ -337,8 +337,9 @@ class LineItem(models.Model):
         help_text=_("Number of units associated to the item"))
 
     credits = models.IntegerField(
-        default=0,
-        help_text=_("Total cost of the item in credits"))
+        default=0, blank=True, null=True,
+        help_text=_("Total cost of the item in credits, or null if there is "
+                    "no associated credit amount"))
```
hodgestar, line 342: I'm tempted to say there has to be a credit amount because those are the only things we can deduct? We should chat a bit about how deducting the account fee will actually work though (I'm assuming we'll deduct it from credits and create a transaction for it, but we should check with @badlard).

go/billing/tasks.py
```diff
@@ -94,10 +182,14 @@ def generate_monthly_statement(account_id, from_date, to_date):
...
     statement.save()
 
-    line_items = make_message_items(
-        account, statement, tagpools, from_date, to_date)
+    items = []
+    items.extend(make_message_items(
+        account, statement, tagpools, from_date, to_date))
+    items.extend(make_session_items(
+        account, statement, tagpools, from_date, to_date))
+    items.append(make_account_fee_item(account, statement))
```
hodgestar: We need to create the account line item from a transaction somehow, although we don't have anything that creates the transaction yet and it might also complicate summing the transactions?

hodgestar:
I uncovered a bit of a rabbit hole. Let's discuss in the morning when my brain is working better.

go/billing/models.py
hodgestar, line 342: After discussion with @badlard we won't be deducting credits for the monthly account fee, so we will need support for null credit values here.

go/billing/tasks.py
hodgestar: Also after discussion with @badlard, we will not include a transaction for the monthly account fee because it won't deduct credits.

hodgestar:
After even more discussion with @badlard, the plan is to not include the account fee line item in statements initially but to later add a per-account flag which indicates whether the account fee should be included.

justinvdm:
Ready for re-review.

justinvdm:
#1063 landed, should be reviewable now.

go/billing/tests/helpers.py
```diff
@@ -38,8 +46,31 @@ def mk_transaction(account, tag_pool_name='pool1',
...
+                 title=settings.MONTHLY_STATEMENT_TITLE,
+                 type=Statement.TYPE_MONTHLY,
+                 from_date=None,
+                 to_date=None,
+                 items=()):
+    if from_date is None:
+        from_date = start_of_month()
+
+    if to_date is None:
+        to_date = start_of_month()
```
hodgestar: Should this maybe be `end_of_month()`?

go/billing/views.py
```diff
@@ -34,9 +34,15 @@ def billers_from_items(all_items):
     billers = [{
         'name': billed_by,
         'channel_type': items[0].channel_type,
-        'channels': channels_from_items(items)
+        'sections': sections_from_items(items)
     } for billed_by, items in billers]
 
+    billed_by_vumi = next((b for b in billers if b['name'] == 'Vumi'), None)
```
hodgestar: I'm a little bit uncomfortable with us matching the string 'Vumi' here. Perhaps we should have a setting for who the account is billed by? Or maybe line items with credits set to None should be at the end instead?

hodgestar:
Just the comment about the way the account fee is moved to the end of the list. Looks good otherwise.

justinvdm:
Ready for re-review.

justinvdm:
Ready for another review.

hodgestar:
:+1:

PR #1075, opened 2014-11-27
Change dialogue jsbox app to record multiple answers

justinvdm:


justinvdm:
Ready for review. Going to make a separate PR for the [alternative solution](https://github.com/praekelt/vumi-go/issues/1076) since it is a quick enough change to make.

hodgestar:
:+1:

hodgestar:
Just recording for the record that no one is particularly wild about this change and that we're hoping to put a better solution in place in the not too distant future.

PR #1076, opened 2014-11-27
Change dialogue jsbox app to record multiple answers in a single extras field

justinvdm:
Similar to #1075, except using a single extras field for all answers. We might end up wanting to do this, but the change is quick enough to make that it is worth making a PR in the meantime.


justinvdm:
Ready for review, but the more I think about it, the more I think this one is a bad idea. Unless there is a good reason to use it, I think we should close this in favor of #1075.

justinvdm:
We're going with #1075, closing this PR.

PR #1082, opened 2014-12-01
App multiplexing router can create invalid sessions

jerith:
It creates an empty session and only adds the `state` entry (which it assumes will always be present) later. This leaves a window during which other workers can read an invalid session.


hodgestar:
:+1:

PR #1073, opened 2014-12-02
Cache routing table in dispatcher

jerith:
Routing tables don't change very frequently but need to be referenced on every hop a message takes through the routing table dispatcher. If we cache these for a few seconds, we can save a whole bunch of Riak lookups, especially on high volume conversations.

### DIFF
--- a/go/vumitools/routing.py
+++ b/go/vumitools/routing.py
@@ -211,6 +211,10 @@ class AccountRoutingTableCache(object):
         """
         Schedule the eviction of a cached routing table.
         """
+        if key in self._evictors:
+            # We already have an evictor for this routing table, so we don't
+            # need a new one.
+            return
         delayed_call = self._reactor.callLater(
             self._ttl, self.evict_routing_table_entry, key)
         self._evictors[key] = delayed_call
@@ -236,6 +240,11 @@ class AccountRoutingTableCache(object):
         """
         key = user_api.user_account_key
         if key not in self._routing_tables:
+            # Fetching the routing table returns control to the reactor and
+            # gives other things the opportunity to cache the routing table
+            # behind our back. If this happens, we replace the cached table
+            # (the one we fetched may be newer) and let schedule_eviction()
+            # worry about the existing evictor.
             routing_table = yield user_api.get_routing_table()
             if self._ttl <= 0:
                 # Special case for disabled cache.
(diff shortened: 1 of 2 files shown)


go/vumitools/routing.py
```diff
@@ -189,6 +190,61 @@ def unroutable_event_done(self):
...
+        """
+        Return the routing table for the provided user_api.
+
+        If the routing table is not cached, it will be fetched from Riak. If
+        caching is not disabled, it will also be added to the cache and
+        eviction scheduled.
+        """
+        key = user_api.user_account_key
+        if key not in self._routing_tables:
+            routing_table = yield user_api.get_routing_table()
```
hodgestar, line 248: After this yield the key could be in `_routing_tables` because another handler put it there? So we should only do the next bit if it's still not there since otherwise we might set and evict it twice?

hodgestar, line 248: And the second evict would raise an exception if the key was no longer there to evict?

hodgestar:
Other than the question about double evictions, looks good.

hodgestar:
Oh, one other thought that is maybe worth noting -- this will increase the window for potentially odd routing immediately after a router is saved. I'm not really awake enough to figure out all the possibilities but, e.g., one RTD might keep routing messages towards a conversation even though the inbound and outbound routes were removed in a single save and another RTD is complaining that the conversation sent a reply it doesn't know what to do with. I'm not quite sure to what extent this window existed before (it seems like it would depend on our values for N when reading and writing from Riak and the size of our cluster).

go/vumitools/routing.py
jerith, line 248: Good point, thanks.

jerith:
Ideally, we'd need a mechanism for telling all the dispatchers to clear a particular entry from their cache, but we don't have a way to broadcast a message to all workers of a particular type yet. The best we can do for now is keep the cache times low.

jerith:
Actually, maybe we can evict the cache entry on routing errors?

hodgestar:
In the scenario I imagined the worker with the newer routing table would be
the one throwing the error, so getting a new copy wouldn't help there.
Maybe let's keep this simple for now and see how things go?

hodgestar:
Do you think we need a test for the potential double-eviction case?

jerith:
> Do you think we need a test for the potential double-eviction case?

We do. I couldn't think of a reasonable way to test the race condition in `get_routing_table()`, but testing `schedule_eviction()` on its own is easy enough.

Thanks.

hodgestar:
:+1:

PR #1085, opened 2014-12-02
Add gzip support to s3utils upload.

hodgestar:


hodgestar:
Ready for review.

go/base/tests/test_s3utils.py
```diff
@@ -60,6 +70,41 @@ def test_no_config(self):
...
+        return gunzip("".join(parts))
+
+    def test_implements_IMultipartWriter(self):
+        self.assertTrue(IMultipartWriter.providedBy(GzipMultipartWriter()))
+
+    def test_push_chunk_and_done(self):
+        writer = GzipMultipartWriter(minimum_size=5)
+        parts = self._push_chunks(writer, ["ab" * 2, "c", "bcd"])
+        self.assertEqual(self._decode_parts(parts), "ababcbcd")
+        self.assertEqual(len(parts), 2)
```
justinvdm, line 111: Is there any way of us checking that we split the parts up properly (to `[ababc, bcd]`) like we did for the normal writer?

go/base/tests/test_s3utils.py
```diff
@@ -60,6 +70,41 @@ def test_no_config(self):
...
+class TestMultipartWriter(GoDjangoTestCase):
+    def _push_chunks(self, writer, chunks):
+        return [part.getvalue() for part in writer.push_chunks(chunks)]
+
+    def test_implements_IMultipartWriter(self):
+        self.assertTrue(IMultipartWriter.providedBy(MultipartWriter()))
+
+    def test_push_chunks(self):
+        writer = MultipartWriter(minimum_size=5)
+        parts = self._push_chunks(writer, ["abab", "c", "bcd"])
```
justinvdm, line 83: Wondering if we need to test more complicated cases, to make sure we are handling the minimum size check for arbitrary chunks, not only the first.

justinvdm:
Left some minor comments, but looks good.

go/base/tests/test_s3utils.py
justinvdm, line 111: Ah, or isn't it clear what that will be since it is gzipped?

hodgestar, line 111: We can't decode parts easily -- we could encode and then split that up and assert that the encoded parts match. The tricky part is that gzip files must contain a timestamp so we'd have to make sure those are the same. Python's GzipFile does provide a way to set that, but I'm not sure if it provides a way to read it.

justinvdm, line 111: Ok cool, I'm +1 on us keeping the tests as is then.

go/base/tests/test_s3utils.py
hodgestar, line 83: Test with multiple chunks added.

go/base/tests/test_s3utils.py
hodgestar, line 111: After some fiddling, I realized that I get the same assertion by checking that the entire result decompresses properly and that each part has the expected length.

hodgestar:
Ready for review again (in the morning).

hodgestar:
Now with much faster large data test (and the performance issue was inside boto/moto rather than in my code -- wheee).

justinvdm:
:+1:

PR #1083, opened 2014-12-02
Add management command for generating billing statements

justinvdm:

PR #1086, opened 2014-12-03
Add management command for archiving transactions

hodgestar:
Similar to the one in #1083 that handles statement generation.

This should support archiving to S3 but not deleting the transactions (for carefully trying things out in production).

### DIFF
--- a/go/billing/management/commands/go_archive_transactions.py
+++ b/go/billing/management/commands/go_archive_transactions.py
@@ -34,17 +34,14 @@ class Command(BaseGoCommand):
         account_number = user.get_profile().user_account
         account = Account.objects.get(account_number=account_number)
 
-        from_date = parse_date(opts['from_date'], default_timezone=None)
-        to_date = parse_date(opts['to_date'], default_timezone=None)
+        from_date = parse_date(opts['from_date']).date()
+        to_date = parse_date(opts['to_date']).date()
         delete = opts['delete']
 
         archive = archive_transactions(
             account.id, from_date, to_date, delete=delete)
 
         self.stdout.write(
-            "Transactions archived for account %s from %s to %s"
-            % (opts['email_address'], archive.from_date, archive.to_date))
-        self.stdout.write(
-            "Archived to S3 as %s." % (archive.filename,))
-        self.stdout.write(
-            "Archive status is: %s." % (archive.status,))
+            "Transactions archived for account %s." % (opts['email_address'],))
+        self.stdout.write("Archived to S3 as %s." % (archive.filename,))
+        self.stdout.write("Archive status is: %s." % (archive.status,))
--- a/go/billing/management/commands/go_archive_transactions.py
+++ b/go/billing/management/commands/go_archive_transactions.py
@@ -1,0 +1,50 @@
+from iso8601 import parse_date
+from optparse import make_option
+
+from go.billing.models import Account
+from go.billing.tasks import archive_transactions
+from go.base.command_utils import BaseGoCommand, get_user_by_email
+
+
+class Command(BaseGoCommand):
+    help = (
+        "Archive transactions for an account over a "
+        "specified time range to S3.")
+
+    option_list = BaseGoCommand.option_list + (
+        make_option(
+            '--email-address', dest='email_address',
+            help="Email address of the account to archive transactions for."),
+        make_option(
+            '--from', dest='from_date',
+            help=("Date to start archiving transactions from in iso8601"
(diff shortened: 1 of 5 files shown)


hodgestar:
Ready for review (in the morning).

go/billing/management/commands/go_archive_transactions.py
```diff
@@ -0,0 +1,50 @@
...
+                  " Default: FALSE.")),
+    )
+
+    def handle(self, *args, **opts):
+        user = get_user_by_email(opts['email_address'])
+        account_number = user.get_profile().user_account
+        account = Account.objects.get(account_number=account_number)
+
+        from_date = parse_date(opts['from_date'], default_timezone=None)
+        to_date = parse_date(opts['to_date'], default_timezone=None)
```
justinvdm: Could you explain why we are setting `default_timezone` to `None` here?

go/billing/tests/test_go_archive_transactions.py
```diff
@@ -0,0 +1,114 @@
...
+            delete=True)
+
+        self.assertEqual(cmd.stderr.getvalue(), "")
+        self.assertEqual(cmd.stdout.getvalue().splitlines(), [
+            'Transactions archived for account user@domain.com'
+            ' from 2014-12-01 00:00:00 to 2014-12-31 00:00:00',
+            'Archived to S3 as'
+            ' transactions-test-0-user-2014-12-01 00:00:00'
+            '-to-2014-12-31 00:00:00.json.',
+            'Archive status is: archive_completed.'
```
justinvdm, line 97: Really not important, but wondering if we should be using the readable labels for the statuses here?

go/billing/tests/test_go_archive_transactions.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assert_remaining_transactions([transaction_1, transaction_2])
+
+        archive = TransactionArchive.objects.get(account=self.account)
+        self.assertEqual(
+            archive.status, TransactionArchive.STATUS_TRANSACTIONS_UPLOADED)
+        self.assertEqual(archive.from_date, from_date)
+        self.assertEqual(archive.to_date, to_date)
+
+        s3_bucket = bucket.get_s3_bucket()
+        [s3_key] = list(s3_bucket.list())
```
justinvdm, line 72: heh, bucket list :)

justinvdm:
Looks good, left a question and a minor comment.

go/billing/management/commands/go_archive_transactions.py
hodgestar: Ah, thanks for reminding me about this. Tests fail for me on SQLite if I try use timezone-aware datetimes as `from_date` or `to_date`. This affects `go_gen_statement` too. Do the `go_gen_statement` tests pass for you locally on SQLite? I'm not entirely sure if using non-aware datetimes is acceptable here -- I think it is, because we construct non-aware datetimes elsewhere and I think Django will treat those as UTC given our settings, but I should double-check.

go/billing/tests/test_go_archive_transactions.py
hodgestar, line 97: I thought that developers running the management command would prefer to see the database values? I do want to fix up the filename though -- I don't think we really want it to include the "00:00:00" parts?

justinvdm, line 97: True, the database value makes more sense here. Re the times part, it might be useful to know for firefighting or auditing purposes, but I guess we have the model fields for that, so I'm happy either way.

go/billing/management/commands/go_archive_transactions.py
justinvdm: Ah, I actually have only tested `go_gen_statement` with postgres, so I'll check with sqlite. Given we are okay with non-aware datetimes, I can change `go_gen_statement` to do the same thing.

hodgestar: I've dug around a bit and we have `TIME_ZONE = 'UTC'` which according to https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-TIME_ZONE should mean that we store all datetimes as UTC regardless and thus that removing the TZ info here should be safe.

hodgestar: I found a better solution here which is to call `.date()` which is what `archive_transactions` was expecting anyway.

hodgestar:
Comments addressed and ready for re-review, I think.

justinvdm:
:+1:

PR #1091, opened 2014-12-05
Use message metadata instead of Redis for response time metrics

jerith:


go/vumitools/middleware.py
```diff
@@ -262,15 +262,15 @@ def key(self, transport_name, message_id):
 
     def set_inbound_timestamp(self, transport_name, message):
         key = self.key(transport_name, message['message_id'])
-        return self.redis.setex(
-            key, self.max_lifetime, repr(time.time()))
+        meta = message['helper_metadata'].setdefault('metrics_middleware', {})
+        meta[key] = repr(time.time())
```
hodgestar: The `key` seems a rather long thing to use as the key -- perhaps we could just use a fixed string like `inbound_message_timestamp`? I'm wondering whether we should also add something to mark `max_lifetime` as deprecated?

hodgestar: Might also be cool to have smaller helper functions for setting metrics_middleware helper_metadata in case we want to have others later? Not particularly fussed since there's only one use right now.

go/vumitools/middleware.py
```diff
@@ -262,15 +262,15 @@ def key(self, transport_name, message_id):
...
 
-    @inlineCallbacks
```
hodgestar, line 268: :)

hodgestar:
Small comments, looks good though.

go/vumitools/middleware.py
jerith: I'll do both those things.

jerith: Err, all three of those things.

smn:
:+1: from me

hodgestar:
:+1:

PR #1095, opened 2014-12-05
Add Content-Encoding: gzip for gzipped s3 uploads.

hodgestar:


hodgestar:
Ideally https://github.com/spulec/moto/issues/274 should be resolved before we land this so that we can test that the code works. Failing that, we can potentially work around the issuing by manually calling `set_metadata` on the key created by the upload (at the expense of some extra small S3 operations in production).

justinvdm:
I'm :+1: on this landing as is, and us uncommenting the relevant assertions when spulec/moto#274 lands.

PR #1100, opened 2014-12-08
Link to html billing statements from the UI

justinvdm:


justinvdm:
Ready for review.

go/account/templates/account/billing.html
```diff
@@ -26,11 +26,11 @@
                     <tr>
                         <td></td>
                         <td>{{ statement.from_date|date:"F Y" }}</td>
-                        <td><a href="{% url 'account:statement_view' statement.id %}">View</a></td>
+                        <td><a href="{% url 'billing:html_statement' statement.id %}">html</a></td>
                     </tr>
                     {% empty %}
                     <tr>
-                        <td colspan="2">
+                        <td colspan="3">
```
hodgestar: Was this colspan wrong?

go/account/tests.py
```diff
@@ -239,3 +256,112 @@ def test_send_scheduled_account_summary_task(self):
...
+    def test_billing_statements_pagination_default(self):
+        account = get_billing_account(self.user_helper.get_django_user())
+        mk_statement(account, from_date=datetime(2014, 1, 28))
+        mk_statement(account, from_date=datetime(2014, 2, 28))
+        mk_statement(account, from_date=datetime(2014, 3, 28))
+
+        resp = self.client.get(reverse('account:billing'), {'p': 1})
+        self.assertContains(resp, '>January 2014<')
+        self.assertContains(resp, '>February 2014<')
+        self.assertNotContains(resp, '>March 2014<')
```
hodgestar: If we're just testing the default page here, do we need this request and set of asserts?

go/account/tests.py
```diff
@@ -239,3 +256,112 @@ def test_send_scheduled_account_summary_task(self):
...
+
+    @mock.patch('go.billing.settings.STATEMENTS_DEFAULT_ORDER_BY', 'from_date')
+    def test_billing_statements_default_order(self):
+        account = get_billing_account(self.user_helper.get_django_user())
+        s1 = mk_statement(account, from_date=datetime(2014, 1, 28))
+        s2 = mk_statement(account, from_date=datetime(2014, 2, 28))
+        s3 = mk_statement(account, from_date=datetime(2014, 3, 28))
+        resp = self.client.get(reverse('account:billing'))
+
+        print resp.content
```
hodgestar: Stray print.

go/account/templates/account/billing.html
justinvdm: Err, sorry, no it wasn't, changing it back. We need a way of centering the text though, but not important or for right now.

go/account/tests.py
```diff
@@ -8,6 +10,21 @@
...
+        'billing:html_statement',
+        kwargs={'statement_id': statement_id})
+
+
+def contains_in_order(haystack, needles):
+    if not needles[0] in haystack:
+        return False
+
+    indices = [haystack.find(needle) for needle in needles]
+    return indices == sorted(indices)
```
hodgestar: I wonder if it would be clearer if we wrote something like:

``` python
indices = [haystack.find(needle) for needle in needles]
return (-1 not in indices) and (indices == sorted(indiced))
```

go/account/tests.py
justinvdm: What I'm testing here is that the default is the same as the first page. We don't need these assertions to test that, just thought it made it clearer, removing them.

go/account/tests.py
```diff
@@ -239,3 +256,112 @@ def test_send_scheduled_account_summary_task(self):
...
+            '>January 2014<',
+            '>February 2014<',
+            '>March 2014<',
+        ]))
+
+    def test_billing_statements_ascending_order(self):
+        account = get_billing_account(self.user_helper.get_django_user())
+        s1 = mk_statement(account, from_date=datetime(2014, 1, 28))
+        s2 = mk_statement(account, from_date=datetime(2014, 2, 28))
+        s3 = mk_statement(account, from_date=datetime(2014, 3, 28))
```
hodgestar: I wonder if we should create these in a slightly different order to avoid ordering by foreign key matching the ordering we test?

hodgestar:
Left a bunch of minor comments. Looks good though.

justinvdm:
Ready for re-review.

hodgestar:
:+1: once Travis is happy.

PR #1102, opened 2014-12-09
Move credits to own list item

SteveBarnett:
After a chat with @jerith (thanks!), moved `100,000,000 credits` to it's own bit.
He said that 

> Help | Account (100,000,000 credits) >

could be confusing since it looks like the drop down is for credits (rather than account). I agree. New version:

> Help | Account > | 100,000,000 credits

Looks like this: https://www.dropbox.com/s/zg3b2bm2swwoaa9/Screenshot%202014-12-09%2006.32.31.png?dl=0

Also made it a link to `/account/billing/`.

_/me edits for clarity_

### DIFF
--- a/go/templates/app.html
+++ b/go/templates/app.html
@@ -35,7 +35,7 @@
             {% if request.user.is_authenticated %}
             <li class="dropdown">
                 <a class="dropdown-toggle" data-toggle="dropdown" href="#">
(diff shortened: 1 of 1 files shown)


SteveBarnett:
I changed my mind about this one, so could you give it another quick eyeball, please @badlard ?

badlard:
:+1:

PR #1106, opened 2014-12-10
Write transport messages to message store in RTD

jerith:
The routing table dispatcher relies on stored outbound messages containing full routing information for its ability to route events back along the same path the outbound messages took. We currently rely on middleware to store these messages, but that's inefficient if the middleware is on the dispatchers (because we're storing messages unnecessarily) and a latency and reliability risk if the middleware is on the transport (because we lose the concurrency from the number of RTD workers and it's easy to misconfigure individual transports).

Given this, it seems reasonable for the routing table dispatcher to write each message to the message store before publishing it to a transport. (We probably also need a config option to disable the write in case we need to switch back to middleware for some reason.)

We can't do a similar write for messages to conversations and routers because the RTD doesn't load those objects and therefore can't include batch information.

### DIFF
--- a/go/vumitools/routing.py
+++ b/go/vumitools/routing.py
@@ -4,7 +4,7 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet import reactor
 
 from vumi.dispatchers.endpoint_dispatchers import RoutingTableDispatcher
-from vumi.config import ConfigDict, ConfigText, ConfigFloat
+from vumi.config import ConfigDict, ConfigText, ConfigFloat, ConfigBool
 from vumi.message import TransportEvent
 from vumi import log
 
@@ -288,6 +288,10 @@ class AccountRoutingTableDispatcherConfig(RoutingTableDispatcher.CONFIG_CLASS,
         "TTL (in seconds) for cached routing tables. If less than or equal to"
         " zero, routing tables will not be cached.",
         static=True, default=5)
+    store_messages_to_transports = ConfigBool(
+        "If true (the default), outbound messages to transports will be"
+        " written to the message store.",
+        static=True, default=True)
 
 
 class AccountRoutingTableDispatcher(RoutingTableDispatcher, GoWorkerMixin):
@@ -688,9 +692,16 @@ class AccountRoutingTableDispatcher(RoutingTableDispatcher, GoWorkerMixin):
 
         # We skip the store code in `process_outbound()` if billing is enabled,
         # so we need to duplicate it here.
-        yield self.vumi_api.mdb.add_outbound_message(msg)
+        yield self.store_outbound_transport_message(msg)
 
         yield self.publish_outbound(msg, dst_connector_name, dst_endpoint)
+
+    def store_outbound_transport_message(self, msg):
+        """
+        Add an outbound message to the message store if configured to do so.
+        """
+        if self.get_static_config().store_messages_to_transports:
+            return self.vumi_api.mdb.add_outbound_message(msg)
 
     @inlineCallbacks
     def handle_unroutable_inbound_message(self, f, msg, connector_name):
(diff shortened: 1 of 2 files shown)


justinvdm:
> We currently rely on middleware to store these messages, but that's inefficient if the middleware is on the dispatchers (because we're storing messages unnecessarily)

For my understanding, could you explain why putting the middleware on the dispatchers means we are storing messages unnecessarily? Is it because we will be storing messages for transport-bound, conversation-bound and router-bound messages, when we only need to store transport-bound messages?

Sorry if this is a noob question, still need a better idea of how all these things fit together (I probably should know better by now).

jerith:
Ideally, we want to store messages exactly twice: once when we first see them (so we have a record that they existed, even if they get lost somewhere along the way) and again when we have all the information we're ever going to have about them.

We can't quite get down to two stores just yet, though, because only application and router workers have enough information to handle the batch stuff. That means we need those workers to write the messages, so the best we can get to is once at each end (transport and app) and once for each router traversal.

Previously, we had middleware on transports and RTDs that stored messages on publish and middleware on apps and routers that stored on both publish and consume. Only the outbound-message-to-transport store was actually necessary here, because in all other cases the message was being forwarded to something that would store it again.

Currently, we have middleware on transports and apps that store on both publish and consume, middleware on routers that store on publish only, and no middleware on RTDs. After this change, we'll switch the transport middleware to store on publish only because the RTDs will be storing messages to transports internally.

justinvdm:
Ah, thanks, that answered my question and gave me a clearer picture of when and where we are storing things.

go/vumitools/routing.py
```diff
@@ -832,6 +847,11 @@ def process_outbound(self, config, msg, connector_name):
         dst_connector_name, dst_endpoint = yield self.set_destination(
             msg, target, self.OUTBOUND)
 
+        if target_conn.ctype == target_conn.TRANSPORT_TAG:
+            # This code is only reached if billing is disabled. There's a
+            # separate code path for messages from billing to transports.
+            yield self.store_outbound_transport_message(msg)
```
justinvdm, line 853: Can't we just have this call higher up in `process_outbound` (before it delegates to `publish_outbound_from_billing`) to avoid the duplication? Or is there a reason we can't do that?

justinvdm:
Can we have a test for checking that we store when billing is disabled? Or did I miss that test?

go/vumitools/routing.py
jerith, line 853: It needs to happen after we update the message with the routing information, which is done on the previous line.

jerith:
> Can we have a test for checking that we store when billing is disabled? Or did I miss that test?

There are equivalent tests in `TestRoutingTableDispatcher` (which has billing disabled) and `TestRoutingTableDispatcherWithBilling` (which has billing enabled).

justinvdm:
Ah, thanks.

go/vumitools/routing.py
justinvdm, line 853: Ok cool. A bit meh that we need to have that duplication, but can't think of a better idea.

go/vumitools/tests/test_routing.py
```diff
@@ -916,10 +918,85 @@ def test_outbound_message_gets_transport_fields(self):
         ])
         self.assertEqual([msg], self.get_dispatched_outbound('sphex'))
 
+    @inlineCallbacks
+    def test_outbound_message_gets_stored_before_transport(self):
+        """
+        Outbound messages going to transports are stored.
+        """
+        yield self.get_dispatcher()
```
justinvdm, line 926: Why do we get the dispatcher but not use it? Some side effect caused by `get_dispatcher`?

jerith, line 926: `get_dispatcher()` creates and starts the worker, but we interact with it entirely through the message broker in these tests.

justinvdm:
:+1: but wondering if others with more experience with the RTD should have a look.

PR #1107, opened 2014-12-10
Display statement costs in dollars

justinvdm:
At the moment, we display statement costs in cents, but the titles suggest the amounts are in dollars.


justinvdm:
Ready for review.

hodgestar:
I'm wondering whether three decimal points is a sensible default? Otherwise looks good.

justinvdm:
I would've thought two was ideal, but if we use 0.1 cents, that'd show up as '0.00'. We could also do four? Whatever you think is best.

hodgestar:
Let's go with three and see what people say? :)

:+1:

justinvdm:
Thanks, alternatively, we could have the statements display values in cents?

PR #1108, opened 2014-12-11
Better message storing for RTD

jerith:
There are a few places where we aren't correctly storing outbound messages to transports.


justinvdm:
Do we need a test for opt outs? Not sure how those fit in.

justinvdm:
Looks good otherwise.

jerith:
> Do we need a test for opt outs? Not sure how those fit in.

Those behave similarly to billing messages. Now that we're checking and storing messages in `.publish_outbound()` instead of various other places, the test isn't strictly necessary. I can write one if you'd prefer, though.

justinvdm:
Looking at the current tests, I think we've covered that already, happy with it as is. :+1:

PR #1109, opened 2014-12-11
Handle non-existent tagpools when generating statements

justinvdm:
At the moment, statement generation breaks when transactions with now non-existent tagpools are encountered. The plan is to default to the tag pool name for the display name, and use 'unknown' as the channel type name.


justinvdm:
On second thought, I think we should just set the channel type to `None` for this case and not display it (similary to how we do for `Vumi`, for example).

justinvdm:
Ready for review.

hodgestar:
:+1:

PR #1111, opened 2014-12-11
Don't allow S3 keys to be overwritten by default.

hodgestar:


hodgestar:
Ready for review.

justinvdm:
:+1:

PR #1113, opened 2014-12-12
Handle 'None' costs and markups when generating statements

justinvdm:
Transaction costs and markups can be `None`, we need to handle this when generating statements.

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -89,11 +89,13 @@ def get_channel_name(transaction, tagpools):
 
 
 def get_message_cost(transaction):
-    return transaction['total_message_cost']
+    cost = transaction['total_message_cost']
+    return cost if cost is not None else 0
 
 
 def get_session_cost(transaction):
-    return transaction['total_session_cost']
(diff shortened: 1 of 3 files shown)


justinvdm:
@hodgestar I'm wondering if we should just change the transaction model so it doesn't have any `None` costs or markups, and defaults to `0` instead. We'll still need this change for now, but thinking more about once we land all those PRs that change transactions.

Either we do that, or we will need to make sure we don't try calculate costs in the billing api if the cost or markup is `None`. Thoughts?

justinvdm:
Ready for review. This should work fine for generating statements before we archive, but after archiving, when we are doing all those transaction migrations, we either need to prevent the costs and markup from being `None` or handle these costs in the api.

If we go with plan B, we'll still need to check for `Nones` in the tasks when calculating unit costs (so we will need to add similar handling for storage costs). My vote is plan A.

hodgestar:
`None`s appear when people load credits so I think we need to continue to support those.

hodgestar:
:+1:

PR #1112, opened 2014-12-12
Add option to abort archiving transactions if the equivalent statment has not yet been generated

justinvdm:
If would be useful if the management command for archiving transacations refused to archive them unless the associated statement had already been generated.

I'm also sort of wondering whether it wouldn't have been better for archiving functions to take a statement as input instead of a date range (i.e. to always require a statement). Probably we just want a helper function that does this?

### DIFF
--- a/go/billing/management/commands/go_archive_transactions.py
+++ b/go/billing/management/commands/go_archive_transactions.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from optparse import make_option
 
-from go.billing.models import Account
+from go.billing.models import Statement, Account
 from go.billing.tasks import month_range, archive_transactions
 from go.base.command_utils import BaseGoCommand, get_user_by_email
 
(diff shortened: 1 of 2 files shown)


go/billing/management/commands/go_archive_transactions.py
```diff
@@ -22,30 +22,65 @@ class Command(BaseGoCommand):
             '--delete', dest='delete', action="store_true", default=False,
             help=("Delete the transactions after uploading the archive."
                   " Default: FALSE.")),
-    )
+        make_option(
+            '--no-statement',
+            dest='no_statement', action="store_true", default=False,
+            help=("Aborts an archive if no equivalent billing statement is "
+                  "found. Default: TRUE.")))
```
hodgestar: The help here seems a bit confusing to me. Maybe:

```
Allow transactions to be archived even if no equivalent billing statement is found. Default: FALSE.
```

go/billing/tests/test_go_archive_transactions.py
```diff
@@ -149,3 +169,33 @@ def test_archive_transactions_with_deletion(self):
...
+            delete=True)
+
+        self.assertEqual(cmd.stderr.getvalue().splitlines(), [
+            'Aborting archiving, no statements found for the following months:',
+            '2014-10',
+            '2014-12'
+        ])
+
+        self.assert_remaining_transactions(transactions)
+        self.assertEqual(list(bucket.get_s3_bucket().list()), [])
```
hodgestar, line 201: I think we need a test that shows that passing in `--no-statement` allows transactions to be archived even when there is no equivalent statement?

hodgestar:
Two comments. Otherwise looks good.

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #1114, opened 2014-12-15
create multiple low credit notification levels

rudigiesler:
Currently there is only one notification level. Add a list in the config of all the notification percentage levels that should be checked, and send the relevant notifications when the levels are reached.


go/billing/api.py
```diff
@@ -249,7 +249,7 @@ def render_POST(self, request):
             else:
                 self._handle_bad_request(request)
         elif len(params) == 2 and data is not None:
-            (account_number, path), rest = params[:2], params[2:]
+            (account_number, path) = params[:2]
```
rudigiesler, line 253: pep8 fix

go/billing/tests/test_admin.py
```diff
@@ -38,7 +38,8 @@ def test_link_to_html_view(self):
         response = client.get(reverse('admin:billing_statement_changelist'))
         self.assertContains(response, "1 statement")
         self.assertContains(
-            response, '<a href="/billing/statement/%s">html</a>' % statement.id)
+            response,
+            '<a href="/billing/statement/%s">html</a>' % statement.id)
```
rudigiesler: pep8 fix

go/billing/tests/test_api.py
```diff
@@ -137,7 +137,7 @@ def create_api_cost(self, account_number=None, tag_pool_name=None,
         return self.call_api('post', 'costs', content=content)
 
     def get_api_costs(self, account_number=None, tag_pool_name=None,
-                     message_direction=''):
+                      message_direction=''):
```
rudigiesler, line 140: pep8 fix

go/billing/tests/test_views.py
```diff
@@ -24,7 +24,9 @@ def mk_statement(self, **kw):
     def get_statement(self, username, statement):
         client = self.vumi_helper.get_client(username=username)
         return client.get(
-            reverse('billing:html_statement', kwargs={'statement_id': statement.id}))
+            reverse(
+                'billing:html_statement',
+                kwargs={'statement_id': statement.id}))
```
rudigiesler: pep8 fix

rudigiesler:
@hodgestar Ready for review.

go/billing/api.py
```diff
@@ -575,6 +573,8 @@ class TransactionResource(BaseResource):
     """Expose a REST interface for a transaction"""
 
     isLeaf = True
+    credit_notification_levels = map(lambda i: Decimal(str(i)), sorted(
+        app_settings.LOW_CREDIT_NOTIFICATION_LEVELS))
```
hodgestar: I think we should probably sort after converting to `Decimal`, becase e.g.

```
>>> sorted(Decimal(str(i)) for i in x)
[Decimal('-1.5'), Decimal('2.0'), Decimal('10.0')]
>>> [Decimal(str(i)) for i in sorted(x)]
[Decimal('-1.5'), Decimal('10.0'), Decimal('2.0')]
```

(I know we're expecting the config to be floats, but it won't complain about strings).

go/billing/migrations/0015_auto__del_field_account_alert_credit_balance__del_field_account_alert_.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def forwards(self, orm):
+        # Deleting field 'Account.alert_credit_balance'
+        db.delete_column(u'billing_account', 'alert_credit_balance')
+
+        # Deleting field 'Account.alert_threshold'
+        db.delete_column(u'billing_account', 'alert_threshold')
+
+        # Adding field 'Account.last_topup_balance'
+        db.add_column(u'billing_account', 'last_topup_balance',
```
hodgestar, line 18: If we run this migration, everyone's `last_topup_balance` will be zero. An alternative approach would be to break this migration into three parts -- add `last_topup_balance`, populate `last_topup_balance` from the `alert_` columns in a data migration, remove the `alert_` columns.

It's quite possible that it's okay for all the accounts to have zero `last_topup_balance` given that no one has billing notifications active yet and we'll need to load credits before we do that, but then we should be sure that we handle zero values for `last_topup_balance` well.

go/billing/settings.py
```diff
@@ -50,6 +50,9 @@
 ENABLE_LOW_CREDIT_NOTIFICATION = getattr(
     settings, 'GO_ENABLE_LOW_CREDIT_NOTIFICATION', False)
 
+LOW_CREDIT_NOTIFICATION_LEVELS = getattr(
+    settings, 'LOW_CREDIT_NOTIFICATION_LEVELS', [])
```
hodgestar: I think the `settings` name is missing a `GO_` at the front?

go/billing/tasks.py
```diff
@@ -346,7 +346,7 @@ def create_low_credit_notification(account_id, threshold, balance):
         {
             'user': account.user,
             'account': account,
-            'threshold_percent': threshold,
+            'threshold_percent': threshold * 100,
```
hodgestar: I was wondering whether we don't need to display `100 - threshold_percent` in the emails since people (by which I mostly mean 'I' currently) think in terms of "percent used" rather than "percent remaining"?

go/testsettings.py
```diff
@@ -58,6 +58,9 @@
 EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
 CELERY_EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'
 
+LOW_CREDIT_NOTIFICATION_LEVELS = [0.7, 0.9, 0.8]
+ENABLE_LOW_CREDIT_NOTIFICATION = True
```
hodgestar: If these are only used by one or two tests, perhaps we should just have those tests set them rather than including them here?

go/billing/api.py
```diff
@@ -575,6 +573,9 @@ class TransactionResource(BaseResource):
     """Expose a REST interface for a transaction"""
 
     isLeaf = True
+    credit_notification_levels = sorted(map(
+        lambda i: Decimal(str(i)), app_settings.LOW_CREDIT_NOTIFICATION_LEVELS
```
hodgestar: Bikeshed: We seem to prefer generator / list comprehensions over `map` usually (I think because a lot of people find them more readable). So,

``` python
sorted(Decimal(str(lvl)) for lvl in app_settings.LOW_CREDIT_NOTIFICATION_LEVELS)
```

Not really necessary to change -- just pointing it out in case that's useful.

hodgestar:
Left a couple of minor comments.

I'm going to see if I can figure out how to get the `mock` decorator to work. Besides those, looks good.

go/billing/tasks.py
rudigiesler: The text in the email is:

`Please note that your account {{account}} has reached {{threshold_percent}}% of the total available credits.`

Maybe we should change that to `{{threshold_percent}}% used of the total available credits` or `{{threshold_percent}}% remaining of the total available credits` depending on which way we choose?

go/testsettings.py
rudigiesler: We can have `ENABLE_LOW_CREDIT_NOTIFICATION` in each of the tests, but `LOW_CREDIT_NOTIFICATION_LEVELS` is read when the class is created, so it has to be here. 

I did notice that these should be `GO_...`, so I'll change that so long.

go/billing/migrations/0015_auto__del_field_account_alert_credit_balance__del_field_account_alert_.py
```diff
@@ -0,0 +1,167 @@
...
+                      self.gf('django.db.models.fields.DecimalField')(default='0.0', max_digits=20, decimal_places=6),
+                      keep_default=False)
+
+        # Populate the last_topup_balance fields with values
+        db.execute(
+            """
+            UPDATE billing_account SET last_topup_balance =
+            alert_credit_balance / alert_threshold
+            """)
```
rudigiesler: I'm not sure if this is 100% correct, @hodgestar if you could take a close look at it.

hodgestar:
I understood why `patch`  doesn't work -- the unwrapped function is a generator (and so returns straight away) and the function returned by `inlineCallbacks` returns a deferred straight away. So in both cases our patch is applied, function returns immediately, patch is removed, function does it's work without the patch.

go/billing/migrations/0015_auto__del_field_account_alert_credit_balance__del_field_account_alert_.py
hodgestar: Given that `alert_threshold` appears to be a percentage, I think we need:

```
UPDATE billing_account SET last_topup_balance = alert_credit_balance * 100 / alert_threshold
```

hodgestar: However, mixing data and schema migrations like this is probably a bad idea. The correct approach is described in http://south.readthedocs.org/en/latest/tutorial/part3.html (using the `orm` object).

I'm kind of in favour of just throwing the old values away though.

hodgestar:
@rudi could you review my commit (f958707) and let me know if you agree about reverting the data migration commit (aa60093)? After those two are addressed, we should be pretty close to landing.

rudigiesler:
@hodgestar :+1: on the patching of the deferred. Do you want to just throw away the data? Since from the few accounts I've checked the notification percentage is all at the default 0%, or should I do the multiple migrations?

hodgestar:
I'm for just throwing the old data away -- the notification system has
changed drastically so keeping the old data doesn't have much value.

rudigiesler:
Then I'm :+1: on reverting the migration commit

go/billing/api.py
```diff
@@ -808,16 +808,39 @@ def create_transaction_interaction(self, cursor, account_number,
...
+        Checks the current balance percentage against all those stored within
+        the settings. Sends the notification email if it is required. Returns
+        the alert percent if email was sent, or ``None`` if no email was sent.
+
+        :param credit_balance: The current balance (after the transaction)
+        :param credit_amount: The amount of credits used in the transaction
+        :param last_topup_balance: The account credit balance at the last topup
+        :param account_number: The account number of the associated account
+        """
+        for level in self.credit_notification_levels:
```
hodgestar: I'm thinking of how to remove this loop.

rudigiesler: The loop will break early if it can, but maybe we can move it to the celery task? Unless you can think of a way to remove it completely.

hodgestar:
@rudigiesler Non-loopy version ready for review.

hodgestar:
@rudigiesler Nice catch on the crossing-multiple-percentages at once case I missed. Improved version that handles this case ready for review.

P.S. Let's not land this tomorrow since it's a bad day to deploy, but hopefully we can have it ready to land first thing in the new year.

rudigiesler:
:+1: on the new `check_all_low_credit_thresholds` and the simplified patching.

hodgestar:
:+1: on all the other stuff.

Ready to land in the new year.

hodgestar:
Other than the two small comments. Looks good.

hodgestar:
:+1: once Travis is happy.

go/billing/api.py
```diff
@@ -808,28 +835,83 @@ def create_transaction_interaction(self, cursor, account_number,
...
+
+        :return:
+            A :class:`Decimal` percentage for the alert threshold crossed
+            or ``None`` if no threshold was crossed.
         """
-        return (
-            credit_balance <= alert_credit_balance <
-            credit_amount + credit_balance)
+        def ceil_percent(n):
+            return int(math.ceil(n * 100 / last_topup_balance))
```
hodgestar, line 907: We need to handle and test the case where `last_topup_balance` is `0` (or perhaps `None`) because once we deploy this there will be a lot of accounts where the topup balance hasn't been set yet.

hodgestar, line 907: Perhaps we should just skip all the checks entirely if `last_topup_balance` is `0` or  `None`?

hodgestar:
:+1: again once Travis is happy.

PR #1120, opened 2014-12-15
Feature/issue 1117 add less to grunt

SteveBarnett:
Ready for review. @justinvdm, if you would be so kind?

I've left in sourcemaps because they'll at least be useful locally, if not in prod.


justinvdm:
:+1: 

Separate from this PR, but wondering if we should:
- have the default grunt task do build stuff like the less compiling and then run the tests
- have a watch task than runs tests when js or css files change

SteveBarnett:
@justinvdm Hells yes.

Re 1: New task `build` that does `vendor`, `less`, `test`?

Re 2: Are you thinking `test:jsbox_apps` and `test:client`? Or just `test:client`? `test:client` takes about 20 seconds to run at the moment...

justinvdm:
@SteveBarnett 

Re Re 1: 
- +1 to a new `build` task
- `vendor` is a bit of a complicated animal. It calls out to the `bower` task, which is for copying over some of the bower components' files to `./go/base/static/vendor`, which is workaround, since django's collectstatic gets overwhelmed by having the entire `bower_components` dir there (collectstatic does some pretty crazy stuff). It isn't really a 'build' task, but maybe it is something we want to run as part of our typical build process. It does mean there will be random changes to `go/base/static/vendor` happening all the time. We should think a bit about that one, my vote is to leave it out of the `build` task for now.
- `test` isn't really a build step, I think `build` and `test` should be separate, and the default task can run `build` and then `test`
- I guess that only leaves `less`, but maybe oneday we will be able to move away from django pipelines and collectstatic crazyness, and we can have other tasks there (minification, concat, etc).

Re Re 2: I'm thinking we have multiple watch tasks, each watching their relevant source and test files. So a `jsbox_apps` one will watch the jsbox app source and test files, a `client` one will watch the client source and test files, and a one for when less files change to compile to css. We can then have a master `watch` task that delegates to all of those. That way, only the relevant tests are run when files change.

Thoughts?

SteveBarnett:
@justinvdm 

Re Re Re 1: oof, complicated. I would very much like to move to having minification and concat done via Grunt / Gulp, but I know the situation is more complex than I grok at the moment. I'm thinking the next spin-off thing we make (dialogues?) we try to lean away from django pipeline and collectstatic, and see what happens. If the wheels come off, we can always revert to them.

Re Re Re 2: sounds good. I'll try and sneak that in as part of the "tidy up the javascript" task I have. :)

justinvdm:
> Re Re Re 1: oof, complicated. I would very much like to move to having minification and concat done via Grunt / Gulp, but I know the situation is more complex than I grok at the moment. I'm thinking the next spin-off thing we make (dialogues?) we try to lean away from django pipeline and collectstatic, and see what happens. If the wheels come off, we can always revert to them.

+10

> Re Re Re 2: sounds good. I'll try and sneak that in as part of the "tidy up the javascript" task I have. :)

Sounds good :)

PR #1121, opened 2014-12-15
Configurable message length limit for HTTP API

jerith:
We currently accept all (valid) outbound messages in the HTTP API. Ideally, we'd like to reject messages that can't be sent for some reason. Since we don't (currently) have any way to get capability information from channels, the easiest thing to do here is add conversation config, starting with the length limit.

### DIFF
--- a/go/apps/http_api/view_definition.py
+++ b/go/apps/http_api/view_definition.py
@@ -24,6 +24,10 @@ class TokenForm(forms.Form):
     metric_store = forms.CharField(
         help_text='Which store to publish metrics to.',
         required=False)
+    content_length_limit = forms.IntegerField(
+        help_text=('Optional content length limit. If set, messages with'
+                   ' content longer than this will be rejected.'),
+        required=False)
 
     @staticmethod
     def initial_from_config(data):
@@ -34,6 +38,7 @@ class TokenForm(forms.Form):
             'push_message_url': data.get('push_message_url', None),
             'push_event_url': data.get('push_event_url', None),
             'metric_store': data.get('metric_store', DEFAULT_METRIC_STORE),
+            'content_length_limit': data.get('content_length_limit', None),
         }
 
     def to_config(self):
@@ -43,6 +48,7 @@ class TokenForm(forms.Form):
             'push_message_url': data['push_message_url'] or None,
             'push_event_url': data['push_event_url'] or None,
             'metric_store': data.get('metric_store') or DEFAULT_METRIC_STORE,
+            'content_length_limit': data.get('content_length_limit', None),
             # The worker code checks these, but we don't provide config UI for
             # them. They should always be False.
             'ignore_events': False,
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -124,7 +124,7 @@ class MsgCheckHelpers(object):
         Check that the message content is within the configured length limit.
         """
         length_limit = api_config.get("content_length_limit")
-        if length_limit is not None:
+        if (length_limit is not None) and (payload["content"] is not None):
             content_length = len(payload["content"])
             if content_length > length_limit:
                 return "Payload content too long: %s > %s" % (
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -127,7 +127,7 @@ class MsgCheckHelpers(object):
         if length_limit is not None:
             content_length = len(payload["content"])
             if content_length > length_limit:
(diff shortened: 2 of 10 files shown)


jerith:
This first part is reviewable on its own, but I haven't built the UI for the new config options yet.

go/apps/http_api_nostream/resource.py
```diff
@@ -109,6 +115,22 @@ def is_unicode_or_none(value):
...
+    @staticmethod
+    def is_within_content_length_limit(payload, api_config):
+        """
+        Check that the message content is within the configured length limit.
+        """
+        length_limit = api_config.get("content_length_limit")
+        if length_limit is not None:
+            content_length = len(payload["content"])
+            if content_length > length_limit:
+                return"Payload content too long: %s > %s" % (
```
justinvdm: Missing space after `return`. Surprised pep8 was fine with that.

jerith: I'm kind of surprised it worked at all. Good catch, thanks!

go/apps/http_api_nostream/resource.py
```diff
@@ -118,6 +140,10 @@ class SendToOptions(MsgOptions):
         'to_addr': MsgCheckHelpers.is_unicode_or_none,
     }
 
+    VALIDATION = (
+        MsgCheckHelpers.is_within_content_length_limit,
+    )
```
justinvdm, line 145: If I'm understanding this correctly, both send and reply messages uses the same config key for message limits. Would we ever want one to have a different limit to the other?

jerith, line 145: No. This really should come from the channel or something, but we don't have any way to do that (yet).

hodgestar:
Not reviewing, but just wanted to mention that I was thinking we should also specify a maximum age for replies in a later PR (e.g. if a reply comes one year after the message the conversation could respond with a 4xx error). Just mentioning it in case it affects implementation somehow.

jerith:
> Not reviewing, but just wanted to mention that I was thinking we should also specify a maximum age for replies in a later PR (e.g. if a reply comes one year after the message the conversation could respond with a 4xx error). Just mentioning it in case it affects implementation somehow.

That doesn't really affect anything this code does. It would have to be specific to the reply handler.

justinvdm:
:+1: when travis is happy.

jerith:
We still need the config UI side of this. :-)

go/apps/http_api_nostream/resource.py
```diff
@@ -75,8 +75,9 @@ class MsgOptions(object):
     """Helper for sanitizing msg options from clients."""
 
     WHITELIST = {}
+    VALIDATION = ()
 
-    def __init__(self, payload):
+    def __init__(self, payload, api_config):
```
hodgestar, line 80: I think it would be cleaner if the content length limit was passed in explicitly here, otherwise we're passing the config around everywhere and it gets hard to isolate things and track which bits of the config are used where.

hodgestar, line 80: Or if we really like this pattern, perhaps `__init__` should have the explicit options and there should be a `from_api_config` method that pulls the pieces out of the config?

hodgestar:
I think it would be cool to have unit tests for `SendToOptions` and `ReplyOptions`, even if it's just one test for the new functionality.

go/apps/http_api/view_definition.py
```diff
@@ -43,6 +48,7 @@ def to_config(self):
             'push_message_url': data['push_message_url'] or None,
             'push_event_url': data['push_event_url'] or None,
             'metric_store': data.get('metric_store') or DEFAULT_METRIC_STORE,
+            'content_length_limit': data.get('content_length_limit', None),
```
hodgestar, line 51: Just wondering what ensures that `content_length_limit` is not in data when a form is submitted with `content_length_limit` set to the empty string?

hodgestar:
Generally looks good. Left a few comments for discussion.

go/apps/http_api_nostream/resource.py
jerith, line 80: It's the validation checkers that need the config, and the base `MsgOptions` class is very generic and doesn't know anything about those. I'm fine with making it less generic, though.

go/apps/http_api/view_definition.py
jerith, line 51: This is a Django `IntegerField` which is documented to provide either an integer or `None`.

hodgestar, line 51: Woot.

hodgestar:
:+1:

PR #1126, opened 2014-12-17
Billing dispatcher discards messages when it's unhappy

jerith:
We catch `BillingError` and forward messages without billing for them, but any other exception results in the message being dropped.

### DIFF
--- a/go/vumitools/billing_worker.py
+++ b/go/vumitools/billing_worker.py
@@ -174,10 +174,14 @@ class BillingDispatcher(Dispatcher, GoWorkerMixin):
             yield self.create_transaction_for_inbound(msg)
             msg_mdh.set_paid()
         except BillingError:
-            log.warning("BillingError for inbound message: %r" % (msg,))
+            log.warning(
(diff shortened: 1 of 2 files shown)


jerith:
I've written some tests for the `BillingError` and non-`BillingError` cases. The latter tests fail.

hodgestar:
I'm a bit concerned that the 0.5s delay is much the same as throwing a message on the floor for USSD. Probably we should pick this up in monitoring elsewhere (e.g. response time graphs) though.

The code looks fine. I'm not super excited about this solution but I'm :+1: on landing this and trying it out rather than going round in circles discussing options.

:+1:

jerith:
> I'm a bit concerned that the 0.5s delay is much the same as throwing a message on the floor for USSD. Probably we should pick this up in monitoring elsewhere (e.g. response time graphs) though.

It's a compromise between "keep processing messages, even if that means not billing for them" and "block all messages when there's a problem". 0.5s means we'll process two messages per second per worker while there are problems, and it's configurable if we decide a different value (all the way down to none at all) is better.

PR #1129, opened 2014-12-24
Change routing table dispatcher to handle opt outs

justinvdm:
At the moment, we rely on an opt out middleware attached to the routing table dispatcher for this. The plan is to move this handling onto the routing table dispatcher. This will help make #1127 possible.

### DIFF
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -39,6 +39,45 @@ class NormalizeMsisdnMiddleware(TransportMiddleware):
         return message
 
 
+class OptOutMiddleware(BaseMiddleware):
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        self.vumi_api = yield VumiApi.from_config_async(self.config)
+
+        self.case_sensitive = self.config.get('case_sensitive', False)
+        keywords = self.config.get('optout_keywords', [])
+        self.optout_keywords = set([self.casing(word) for word in keywords])
+
+    def casing(self, word):
+        if not self.case_sensitive:
+            return word.lower()
+        return word
+
+    @inlineCallbacks
+    def handle_inbound(self, message, endpoint):
+        optout_disabled = False
+        msg_mdh = MessageMetadataHelper(self.vumi_api, message)
+        if msg_mdh.tag is not None:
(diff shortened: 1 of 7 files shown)


justinvdm:
I'm +1 on converting the opt out middleware to a utility class and using it from the rtd. Going with that approach if people are okay with that.

jerith:
I don't really like the thought of moving more of this stuff into the RTD, but I don't have any better ideas at the moment. Let's try keep it as isolated as possible so we can move it to channels or something if and when that becomes possible.

justinvdm:
Cool, +1 on keeping it isolated and being able to reuse it for channels. I think having it as a utility class will make that possible, or at the very least a start.

justinvdm:
Ready for review.

jerith:
Can you add some docstrings for the tests, please?

We need to get better about doing that, even if it's only for new code.

justinvdm:
Docstrings added for the opt out helper's tests.

go/vumitools/middleware.py
```diff
@@ -39,45 +39,6 @@ def handle_outbound(self, message, endpoint):
         return message
 
 
-class OptOutMiddleware(BaseMiddleware):
-
```
jerith, line 43: We should probably leave the middleware in for a while (although deprecate it and maybe rewrite it to use the new helper) to avoid breakage if the code gets deployed without the corresponding config changes.

It should be safe to apply it more than once, right?

justinvdm, line 43: Should be safe if we rewrite it to use the new helper. The helper doesn't reprocess a message if it already has opt out metadata.

justinvdm, line 43: The issue with this is that the opt out helper takes in an account when processing a message. We don't use it yet, but we will soon. I don't know how possible it is for us to get hold of accounts in the middleware in a way that does not perform badly (that is the reason for this PR's change in the first place).

justinvdm, line 43: Applying it more than once is probably fine though, so I guess we could leave the middleware around and just not try modify it to use the helper.

hodgestar:
:+1:

PR #1127, opened 2014-12-24
Support managing opt outs for accounts

justinvdm:
Currently, we support managing opt outs at the tagpool level, but we need to be able to do this at the account level.


justinvdm:
Depends on #1129 and #1132, otherwise ready for review.

justinvdm:
Ready for review.

jerith:
:+1:

PR #1132, opened 2014-12-24
Add a disable_optouts field to UserAccount model

justinvdm:

PR #1133, opened 2015-01-05
Add an account field for account types

justinvdm:
We need a way of setting whether an account is a developer account or not so that we can top up developer accounts with credits each month.

### DIFF
--- a/go/vumitools/account/migrations.py
+++ b/go/vumitools/account/migrations.py
@@ -109,3 +109,42 @@ class UserAccountMigrator(ModelMigrator):
         mdata.set_value('$VERSION', 4)
 
         return mdata
+
+    def migrate_from_5(self, mdata):
+        """
+        Add the account_types field and default it to `ACCOUNT_TYPE_NORMAL`
+        """
+
+        # Copy stuff that hasn't changed between versions
+        mdata.copy_values(
+            'username', 'created_at', 'msisdn', 'confirm_start_conversation',
+            'tags', 'event_handler_config', 'routing_table',
+            'can_manage_optouts', 'disable_optouts')
+        mdata.copy_indexes('tagpools_bin', 'applications_bin')
+
+        # import here to avoid circular import
+        from go.vumitools.account.models import UserAccount
+
+        # set the default `disable_optouts` value
+        mdata.set_value('account_type', UserAccount.ACCOUNT_TYPE_NORMAL)
+
+        # increment version counter
+        mdata.set_value('$VERSION', 6)
+
+        return mdata
+
+    def reverse_from_6(self, mdata):
+        """
+        Remove account_type field.
+        """
+        # Copy stuff that hasn't changed between versions
+        mdata.copy_values(
+            'username', 'created_at', 'msisdn', 'confirm_start_conversation',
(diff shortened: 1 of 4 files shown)


justinvdm:
The current plan is to have an account field like `account_type` and allow it to be one of a list of possible string values. `'normal'` and `'developer'` might work as the possible field values for now, though 'developer' might not be the right word to use, since that should also cover cases like internal, hackathon and entrepeneur accounts. Thoughts on a better name?

justinvdm:
Depends on #1132, otherwise ready for review.

go/vumitools/account/models.py
```diff
@@ -29,11 +29,19 @@ class UserAppPermission(Model):
 class UserAccount(Model):
     """A user account."""
 
-    VERSION = 4
+    VERSION = 6
     MIGRATOR = UserAccountMigrator
 
+    ACCOUNT_TYPE_NORMAL = u'normal'
+    ACCOUNT_TYPE_DEVELOPER = u'developer'
```
justinvdm, line 36: Really not sure if 'developer' is the right word to use here. Thoughts on a more fitting word?

justinvdm:
Hmm, wondering if this field should've been added to the billing account model instead of the user account model. For the reason this is being added, adding it to the billing account model might make more sense. Having an account type field does sound like something general enough that we might need it for other reasons later though.

go/vumitools/account/models.py
hodgestar, line 36: I think billing is associated with a campaign / project rather than an account. Since our user accounts are really campaigns currently, I think maybe this is the correct place?

justinvdm:
Ready for review.

go/vumitools/account/migrations.py
```diff
@@ -109,3 +109,42 @@ def reverse_from_5(self, mdata):
...
+
+        # Copy stuff that hasn't changed between versions
+        mdata.copy_values(
+            'username', 'created_at', 'msisdn', 'confirm_start_conversation',
+            'tags', 'event_handler_config', 'routing_table',
+            'can_manage_optouts', 'disable_optouts')
+        mdata.copy_indexes('tagpools_bin', 'applications_bin')
+
+        # import here to avoid circular import
+        from go.vumitools.account.models import UserAccount
```
hodgestar, line 126: There is potential for this to break if `UserAccount.ACCOUNT_TYPE_NORMAL` goes away or changes at some point, but I think that's okay since there are tests.

hodgestar:
:+1:

go/vumitools/account/migrations.py
justinvdm, line 126: True, that is a bit bleh. There is also the fact that adding account type class-level vars to the account model is bloating it up pretty quickly. Wondering if they should maybe be at the module level, or if it should just be strings.

hodgestar, line 126: Maybe we should just use a string here? Changing the constants would
probably require a new migration anyway?

justinvdm, line 126: +1

justinvdm:
Closing this in favour of #1137

PR #1136, opened 2015-01-06
Collection of smallish cleanups

jerith:
There are various PEP 8 violations and TODOs that can be cleaned up fairly easily. This is a ticket to collect those.

### DIFF
--- a/go/vumitools/api.py
+++ b/go/vumitools/api.py
@@ -20,9 +20,7 @@ from vumi.persist.txredis_manager import TxRedisManager
 from vumi.service import Publisher
 from vumi import log
 
-from go.config import (
-    configured_conversations,
-    configured_routers)
+from go.config import configured_conversations, configured_routers
 from go.vumitools.account import AccountStore
 from go.vumitools.channel import ChannelStore
 from go.vumitools.contact import ContactStore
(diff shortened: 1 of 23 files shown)


jerith:
It's probably easiest to review this one commit at a time, although the only change that is spread across multiple files is 1182d4c.

go/vumitools/api.py
```diff
@@ -181,9 +179,10 @@ def running_conversations(self):
 
     @Manager.calls_manager
     def draft_conversations(self):
-        # TODO: Get rid of this once the old UI finally goes away.
+        # TODO: This should probably be `stopped_conversations` instead, but we
+        #       still apparently use `draft` in the UI in places.
         conversations = yield self.active_conversations()
-        returnValue([c for c in conversations if c.is_draft()])
+        returnValue([c for c in conversations if c.stopped()])
```
justinvdm, line 185: Doesn't seem like the right PR for this change, but if the tests pass maybe its ok.

justinvdm, line 185: Ah, after reading the rest of the PR this makes a bit more sense.

justinvdm:
:+1: once Travis is happy.

go/vumitools/api.py
jerith, line 185: `is_draft` doesn't have much meaning anymore and it's functionally equivalent to `stopped` if archived conversations are excluded, which is the case here.

PR #1137, opened 2015-01-07
Support feature flags on account models

justinvdm:
We need to add an account field that will allow us to set feature flags, for example:
- whether an account should be topped up with credits each month
- whether an account can manage opt outs

### DIFF
--- a/go/vumitools/account/models.py
+++ b/go/vumitools/account/models.py
@@ -26,6 +26,19 @@ class UserAppPermission(Model):
     application = Unicode(max_length=255)
 
 
+def flag_method(name):
+    def fget(self):
+        return name in self.flags
+
+    def fset(self, value):
+        if value:
+            self.flags.append(name)
+        else:
+            del self.flags[list(self.flags).index(name)]
+
+    return property(fget, fset)
+
+
 class UserAccount(Model):
     """A user account."""
 
@@ -47,13 +60,8 @@ class UserAccount(Model):
     routing_table = RoutingTableField(default=RoutingTable({}))
     flags = ListOf(Unicode())
 
-    @property
-    def can_manage_optouts(self):
-        return u'can_manage_optouts' in self.flags
-
-    @property
-    def disable_optouts(self):
-        return u'disable_optouts' in self.flags
+    can_manage_optopts = flag_method(u'can_manage_optopts')
+    disable_optouts = flag_method(u'disable_optouts')
 
     @Manager.calls_manager
     def has_tagpool_permission(self, tagpool):
--- a/go/vumitools/account/models.py
+++ b/go/vumitools/account/models.py
@@ -26,7 +26,7 @@ class UserAppPermission(Model):
     application = Unicode(max_length=255)
 
 
-def flag_method(name):
+def flag_property(name):
     def fget(self):
         return name in self.flags
 
@@ -60,8 +60,8 @@ class UserAccount(Model):
     routing_table = RoutingTableField(default=RoutingTable({}))
     flags = SetOf(Unicode(), index=True)
 
-    can_manage_optouts = flag_method(u'can_manage_optouts')
-    disable_optouts = flag_method(u'disable_optouts')
+    can_manage_optouts = flag_property(u'can_manage_optouts')
+    disable_optouts = flag_property(u'disable_optouts')
 
     @Manager.calls_manager
     def has_tagpool_permission(self, tagpool):
--- a/go/vumitools/account/models.py
+++ b/go/vumitools/account/models.py
@@ -7,7 +7,7 @@ from twisted.internet.defer import returnValue
 
 from vumi.persist.model import Model, Manager
 from vumi.persist.fields import (
-    Integer, Unicode, Timestamp, ManyToMany, Json, Boolean)
+    Integer, Unicode, Timestamp, ManyToMany, Json, Boolean, ListOf)
 
 from go.vumitools.account.fields import RoutingTableField
 from go.vumitools.account.migrations import UserAccountMigrator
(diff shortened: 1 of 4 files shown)


go/vumitools/account/models.py
```diff
@@ -42,11 +42,18 @@ class UserAccount(Model):
...
-    disable_optouts = Boolean(default=False)
-    can_manage_optouts = Boolean(default=False)
     email_summary = Unicode(max_length=255, null=True)
     tags = Json(default=[])
     routing_table = RoutingTableField(default=RoutingTable({}))
+    flags = ListOf(Unicode())
+
+    @property
+    def can_manage_optouts(self):
+        return u'can_manage_optouts' in self.flags
```
justinvdm: Having these as properties does mean that anything that tries to set them manually in the current codebase won't work anymore. I don't that is the case, but we can check if tests fail.

This also means that setting flags in any new code needs to be done via the `flags` field. Should we add a method for each flag for doing the setting part? Not sure how necessary that is.

@jerith @hodgestar @rud thoughts?

justinvdm:
Ready for review, though it should probably wait for the to-be-implemented [`SetOf` field](https://github.com/praekelt/vumi/issues/892) before landing.

go/vumitools/account/models.py
jerith: I think methods or getter/setter properties for each flag are a good idea.

go/vumitools/account/models.py
```diff
@@ -26,10 +26,23 @@ class UserAppPermission(Model):
...
 
+def flag_method(name):
+    def fget(self):
+        return name in self.flags
+
+    def fset(self, value):
+        if value:
+            self.flags.append(name)
+        else:
+            del self.flags[list(self.flags).index(name)]
```
justinvdm: `ListProxy` doesn't have a `.remove`, thus the need for this weird way of removing the flag. This will go away when we add `SetOf`, which I'm assuming will happen before this PR lands.

justinvdm:
Added getter/setter properties for the flags.

go/vumitools/account/models.py
jerith: I'll add `.remove()` to `ListOf` in praekelt/vumi#892, since it makes sense to have it.

justinvdm:
Now using https://github.com/praekelt/vumi/pull/892, should be ready for review now.

go/vumitools/account/models.py
```diff
@@ -26,10 +26,23 @@ class UserAppPermission(Model):
...
+    def fget(self):
+        return name in self.flags
+
+    def fset(self, value):
+        if value:
+            self.flags.add(name)
+        else:
+            self.flags.discard(name)
+
+    return property(fget, fset)
```
jerith, line 39: Nice. Maybe call it `flag_property()`, though?

go/vumitools/account/old_models.py
```diff
@@ -235,6 +234,40 @@ def has_tagpool_permission(self, tagpool):
         returnValue(False)
 
 
+class UserAccountV6(Model):
+    """A user account."""
+
+    bucket = "useraccount"
+    VERSION = 6
+    MIGRATOR = UserAccountMigrator
```
jerith: Why do we need a copy of the current model version here? Also, this is slightly different from the other copy. (`ListOf` instead of `SetOf`, possibly other differences.)

go/vumitools/account/models.py
```diff
@@ -42,11 +55,13 @@ class UserAccount(Model):
...
     confirm_start_conversation = Boolean(default=False)
-    disable_optouts = Boolean(default=False)
-    can_manage_optouts = Boolean(default=False)
     email_summary = Unicode(max_length=255, null=True)
     tags = Json(default=[])
     routing_table = RoutingTableField(default=RoutingTable({}))
+    flags = SetOf(Unicode(), index=True)
+
+    can_manage_optouts = flag_method(u'can_manage_optouts')
+    disable_optouts = flag_method(u'disable_optouts')
```
jerith: I think it would be useful to have a comment at the start of the flag properties explaining that they aren't quite the same as normal fields and that we don't need to bump the model version to add new ones. Thoughts?

go/vumitools/account/old_models.py
justinvdm: Woops, forgot to update this copy. Will remove it, just did it because I noticed we were adding the current model version before I added v5 and v6 (unless I'm not remembering properly).

justinvdm:
Review comments addressed (unless the review is still ongoing).

go/vumitools/account/models.py
justinvdm: +1

justinvdm:
https://github.com/praekelt/vumi/pull/892 landed and released with vumi 0.5.7. I've restarted the last travis build, so it should be happy now. I also think the review comments have been addressed, so should be re-reviewable.

go/vumitools/account/old_models.py
```diff
@@ -5,8 +5,9 @@
 
 from vumi.persist.model import Model, Manager
 from vumi.persist.fields import (
-    Integer, Unicode, Timestamp, ManyToMany, Json, Boolean)
+    Integer, Unicode, Timestamp, ManyToMany, Json, Boolean, ListOf)
 
+from go.vumitools.account.models import flag_property
```
jerith: Unused imports?

jerith:
One minor comment, otherwise :+1:

justinvdm:
Ready for another review.

hodgestar:
:+1:

hodgestar:
Although I guess we should wait for Travis.

PR #1138, opened 2015-01-08
Add explicit config option for enabling / disabling billing.

hodgestar:
Currently we check the inbound and outbound billing connectors, but we need to be able to keep those set even when billing is disabled to support routing existing messages via billing.

### DIFF
--- a/go/vumitools/routing.py
+++ b/go/vumitools/routing.py
@@ -271,6 +271,12 @@ class AccountRoutingTableDispatcherConfig(RoutingTableDispatcher.CONFIG_CLASS,
     opt_out_connector = ConfigText(
         "Connector to publish opt-out messages on.",
         static=True, required=True)
+    billing_enabled = ConfigBool(
+        "If true (default is false) inbound and outbound messages are"
+        " dispatched to the billing connectors. Events may still be"
+        " sent to the outbound billing connector if the associated"
+        " inbound message was sent there before billing was disabled.",
+        static=True, default=False)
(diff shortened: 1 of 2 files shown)


go/vumitools/routing.py
```diff
@@ -790,10 +797,12 @@ def process_inbound(self, config, msg, connector_name):
...
-        if self.billing_inbound_connector:
+        if self.billing_enabled:
             if connector_type == self.TRANSPORT_TAG:
                 yield self.publish_inbound_to_billing(config, msg)
                 return
+
+        if self.billing_inbound_connector:
             if connector_type == self.BILLING:
                 # set the src_conn to the transport and keep routing
                 src_conn = str(GoConnector.for_transport_tag(*msg_mdh.tag))
```
jerith, line 808: Does this need to be wrapped in `if self.billing_inbound_connector`? There isn't any other way to receive a message with that connector type.

go/vumitools/routing.py
```diff
@@ -840,17 +849,17 @@ def process_outbound(self, config, msg, connector_name):
         if self.billing_outbound_connector:
```
jerith, line 849: Does this also need to be changed to `if self.billing_enabled`?

justinvdm:
Would it possible to test that this is doing what we want it to?

hodgestar:
This got too complex quickly and is adding more complexity to the RTD which is already mind bending. Closing in favour of #1139.

PR #1139, opened 2015-01-08
Add option to disable calling the billing API from the billing worker.

hodgestar:
This is an alternative to #1138.

### DIFF
--- a/go/vumitools/billing_worker.py
+++ b/go/vumitools/billing_worker.py
@@ -175,7 +175,10 @@ class BillingDispatcher(Dispatcher, GoWorkerMixin):
         log.debug("Processing inbound: %r" % (msg,))
         msg_mdh = self.get_metadata_helper(msg)
         try:
-            if not self.disable_billing:
+            if self.disable_billing:
+                log.info(
+                    "Not billing for inbound message: %r" % msg.to_json())
+            else:
                 yield self.create_transaction_for_inbound(msg)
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

justinvdm:
:+1: when Travis is happy.

go/vumitools/billing_worker.py
```diff
@@ -171,8 +175,9 @@ def process_inbound(self, config, msg, connector_name):
         log.debug("Processing inbound: %r" % (msg,))
         msg_mdh = self.get_metadata_helper(msg)
         try:
-            yield self.create_transaction_for_inbound(msg)
-            msg_mdh.set_paid()
+            if not self.disable_billing:
+                yield self.create_transaction_for_inbound(msg)
+                msg_mdh.set_paid()
```
jerith, line 183: Should we maybe log that we're not billing for a message because it's been disabled?

hodgestar, line 183: Log the whole message, or just that we're not billing?

jerith, line 183: Ideally, we want enough information that we can bill for these messages after the fact. That probably means logging the whole message.

hodgestar, line 183: Logging added.

hodgestar:
Ready for re-review.

jerith:
:+1:

PR #1141, opened 2015-01-09
fix low credit notification email subject

rudigiesler:


go/billing/tasks.py
```diff
@@ -384,8 +384,9 @@ def create_low_credit_notification(account_number, threshold, balance):
     # Send email
     threshold_percent = Decimal(100) - (threshold * Decimal(100))
     subject = 'Vumi Go account %s (%s) at %s%% left of available credits' % (
-        account.user.email, account.user.get_full_name(), threshold_percent)
-    email_from = settings.STATEMENT_CONTACT_DETAILS.get('email')
+        account.user.email, account.user.get_full_name(),
+        Decimal(100) - threshold_percent)
```
justinvdm, line 388: I think many of the current accounts are for projects rather than for individuals, so the name and surname are a bit misused. We usually do something like `name: Girlhub, surname: Ethiopia`, since those fields are compulsory. I guess this template works for that example (it would get displayed as `(Girlhub Ethiopia)` in the subject line for that case), so maybe this is the right thing to have. I guess lets see how it goes, and if people find it looks weird, we can change it.

justinvdm:
:+1:

go/billing/settings.py
```diff
@@ -43,7 +43,7 @@
 STATEMENT_CONTACT_DETAILS = getattr(
     settings, 'BILLING_STATEMENT_CONTACT_DETAILS', {
```
hodgestar, line 44: I realised that for consistency this should be `GO_STATEMENT_CONTACT_DETAILS`? I also wonder about having our own email addresses and web address set in the default settings? Maybe we should have dummy settings here and only have these values in our own production settings?

justinvdm, line 44: All our other billing settings use `BILLING_` as a prefix. It is only the new credit notification settings that use `GO_`. If we are to change the prefix to `GO_`, then the other settings should change too.

hodgestar, line 44: Ah. Maybe the new notification settings should change to `BILLING_`
instead?

justinvdm, line 44: +1, made the changes: https://github.com/praekelt/vumi-go/pull/1143

hodgestar, line 44: Tx. :)

PR #1142, opened 2015-01-09
Don't include billing contact details in repo

justinvdm:
At the moment, we include our actual contact details in the repo for generating statements, we should put dummy details in instead.

### DIFF
--- a/go/billing/settings.py
+++ b/go/billing/settings.py
@@ -42,9 +42,9 @@ STATEMENTS_DEFAULT_ORDER_BY = getattr(
 
 STATEMENT_CONTACT_DETAILS = getattr(
     settings, 'BILLING_STATEMENT_CONTACT_DETAILS', {
-        'tel': '27.11.482.8684',
-        'email': 'accounts@praekelt.com',
-        'website': 'www.praekeltfoundation.org',
+        'tel': '27.11.123.4567',
+        'email': 'foo@bar.com',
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review.

justinvdm:
@hodgestar turns out we don't actually need to modify tests because we stub out the contact details in tests (forgot that I did that).

hodgestar:
@justinvdm Woot (on the stubbing)

go/billing/settings.py
```diff
@@ -42,9 +42,9 @@
 
 STATEMENT_CONTACT_DETAILS = getattr(
     settings, 'BILLING_STATEMENT_CONTACT_DETAILS', {
-        'tel': '27.11.482.8684',
-        'email': 'accounts@praekelt.com',
-        'website': 'www.praekeltfoundation.org',
+        'tel': '27.11.123.4567',
+        'email': 'foo@bar.com',
+        'website': 'www.foo.org',
```
hodgestar: The officially mandated example domain name is `example.com` but `bar.com` has quite a fun web page and seems to be a de facto standard. `www.foo.org` returns a holding page so we probably shouldn't use it. Thoughts?

justinvdm: +1 on example.com

hodgestar:
:+1:

PR #1145, opened 2015-01-12
Update low credit notification email text

rudigiesler:
Since we're not quite enabling switching off accounts with negative balances, but will be soon, we should have the low credit notification email reflect this.

### DIFF
--- a/go/billing/templates/billing/low_credit_notification_email.txt
+++ b/go/billing/templates/billing/low_credit_notification_email.txt
@@ -2,8 +2,6 @@ Hi {{user.get_full_name}}!
 
 Please note that your Vumi Go account ({{account.user.email}}) has reached {{threshold_percent}}% used of the total available credits. Your remaining balance is {{credit_balance}}.
 
-In the near future, once you have used all of your credits, your account will be capped and you will no longer be able to send any messages. Incoming messages will still be stored as usual.
-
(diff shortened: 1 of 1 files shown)


rudigiesler:
@SteveBarnett I'm not quite sure how to phrase this, your input would be appreciated.

SteveBarnett:
@rudigiesler Yeah, it's tricky. My thinking is that this Sword of Damocles kind of wording isn't helpful: it just provokes questions (when is "near future"?).

So, either: 1) we don't say anything about the consequences of capping for now, and we update the email text later, or 2) we keep it vague and don't mention timing.

So 2) could be something like

> Once you have used all of your credits, your account **may** be capped and you **may** no longer be able to send any messages. Incoming messages will still be stored as usual.

Captain @badlard, what do you think?

badlard:
PMs are working hard enough getting their head around billing implications. Let's go with #1 please. i.e. let's take the line out entirely and add it back when conversation deactivation has been deployed.

rudigiesler:
@badlard @SteveBarnett Changes made. See if you are happy with them.

SteveBarnett:
@rudigiesler :+1: from me.

badlard:
:+1:

hodgestar:
:+1: (most reviewed PR ever :)

PR #1144, opened 2015-01-12
Add a task for generating statements then archiving

justinvdm:


justinvdm:
Ready for review.

go/billing/tasks.py
```diff
@@ -357,6 +357,36 @@ def generate_chunks(item_iter, items_per_chunk=1000, sep="\n"):
...
+def gen_statement_then_archive(account_id, from_date, to_date, delete=True):
+    """Task to generate a monthly statement for the given account,
+    then archive the account.
+    """
+    generate_monthly_statement(account_id, from_date, to_date)
+    archive_transactions(account_id, from_date, to_date, delete)
+
+
+@task()
+def gen_statement_then_archive_monthly(months_ago=1):
```
justinvdm, line 371: Any ideas on a better name?

hodgestar, line 371: Sounds okay to me. Bit long, but also descriptive.

go/billing/tasks.py
```diff
@@ -357,6 +357,36 @@ def generate_chunks(item_iter, items_per_chunk=1000, sep="\n"):
     return archive
 
 
+@task
+def gen_statement_then_archive(account_id, from_date, to_date, delete=True):
+    """Task to generate a monthly statement for the given account,
+    then archive the account.
+    """
+    generate_monthly_statement(account_id, from_date, to_date)
+    archive_transactions(account_id, from_date, to_date, delete)
```
hodgestar: Maybe we should chain these tasks as mentioned in http://celery.readthedocs.org/en/latest/userguide/tasks.html#avoid-launching-synchronous-subtasks ?

go/billing/tasks.py
```diff
@@ -357,6 +357,36 @@ def generate_chunks(item_iter, items_per_chunk=1000, sep="\n"):
...
+    from_date, to_date = month_range(months_ago=months_ago)
+
+    account_list = Account.objects.exclude(
+        statement__type=Statement.TYPE_MONTHLY,
+        statement__from_date=from_date,
+        statement__to_date=to_date)
+
+    account_list = account_list.exclude(
+        transactionarchive__from_date=from_date,
+        transactionarchive__to_date=to_date)
```
hodgestar, line 384: The idea behind these exclusions is that they make re-running the task easier. With just one operation happening, that was a bit simpler. Now we have two operations and we exclude all accounts that have had either of them happen. I think that makes sense, and will mean we'll only have to manually re-run things for accounts that were midway through when something broken? Mostly just noting this here so we think about it.

go/billing/tests/test_tasks.py
```diff
@@ -604,6 +604,150 @@ def mk_transaction_set(n, created):
...
+        tasks.generate_monthly_statement(self.account1.id, from_date, to_date)
+        tasks.generate_monthly_statement(self.account2.id, from_date, to_date)
+
+        statements = Statement.objects
+        self.assertEqual(len(statements.filter(account=self.account1)), 1)
+        self.assertEqual(len(statements.filter(account=self.account2)), 1)
+
+        tasks.gen_statement_then_archive_monthly()
+        self.assertEqual(len(statements.filter(account=self.account1)), 1)
+        self.assertEqual(len(statements.filter(account=self.account2)), 1)
```
hodgestar, line 657: We should perhaps also assert that no transaction archives have been created?

go/billing/tests/test_tasks.py
```diff
@@ -604,6 +604,150 @@ def mk_transaction_set(n, created):
...
+        tasks.archive_transactions(
+            self.account2.id, from_date, to_date, delete=False)
+
+        archives = TransactionArchive.objects
+        self.assertEqual(len(archives.filter(account=self.account1)), 1)
+        self.assertEqual(len(archives.filter(account=self.account2)), 1)
+
+        tasks.gen_statement_then_archive_monthly()
+        self.assertEqual(len(archives.filter(account=self.account1)), 1)
+        self.assertEqual(len(archives.filter(account=self.account2)), 1)
```
hodgestar, line 678: We should perhaps also assert that no statements have been generated?

hodgestar:
Hmm. Travis build failed?

Update: Looks like build froze during billing task tests. :/

hodgestar:
Left a few comments but generally looks good.

justinvdm:
@hodgestar ya, it looks like it keeps freezing for some reason. The entire test suit seems to pass locally (with the same test settings travis uses, I think), so pretty confusing. Will try figure it out.

go/billing/tasks.py
justinvdm, line 384: Yup, that was my thinking too. I've written tests to cover both cases (the case where the statement exists and the case where the transaction archive exists). This does however mean that if the task broke halfway, this task would exclude it if run again, even though either the statement model or transaction archive model are erroneous or non-existant.

hodgestar, line 384: I think that's okay? If the state gets too weird, we need to check things
manually anyway.

justinvdm:
Ready for re-review. Hopefully travis tests will pass now that the task isn't running its subtasks synchronously. Sorry, that was a bit of a mess up.

justinvdm:
We discovered that the travis build timeouts were actually happen when the task was using redis, and only when travis was using real redis. This was happening because the test was also using moto, which apparently intercepts all sockets, interfering with how we talk to redis.

Monkey patching `Bucket.upload` instead of using moto worked, we don't actually need moto for these tests, all we need is stop uploading from being attempted during tests.

Ready for review.

hodgestar:
:+1:

PR #1147, opened 2015-01-12
Fix typo in archiving task

justinvdm:


justinvdm:
Ready for review. Initially tried add a test for the fix, it doesn't seem moto offers anything for inspecting the headers. Tried mocking things out, but it was getting complicated, and I'm not sure if the returns are great  enough to go through that much trouble just to check that we have a content type header set to `'application/json; charset=utf-8'`. Happy to investigate further and try add a test if whoever reviews this thinks its necessary.

hodgestar:
Woot on catching this. I filed https://github.com/spulec/moto/issues/274 when I tried to test the headers before. It's been fixed but we'll need a new moto release before we can use it.

hodgestar:
:+1:

justinvdm:
> Woot on catching this. I filed spulec/moto#274 when I tried to test the headers before. It's been fixed but we'll need a new moto release before we can use it.

Ah right, thanks, remember you mentioning this before now.

PR #1148, opened 2015-01-13
add rounding to balance in low credit notification email

rudigiesler:


justinvdm:
I think we still need to add the change that makes it clear that remaining balance is in credits. Something like: `Your remaining credit balance is 123.45` or `Your remaining balance is 123.45 credits`

justinvdm:
Unless thats for a different PR?

go/billing/tasks.py
```diff
@@ -388,13 +390,14 @@ def create_low_credit_notification(account_number, threshold, balance):
         Decimal(100) - threshold_percent)
     email_from = settings.LOW_CREDIT_NOTIFICATION_EMAIL
     email_to = account.user.email
+    formatted_balance = str(floor(balance*100)/100)
```
hodgestar: PEP8 violations?

go/billing/tests/test_tasks.py
```diff
@@ -643,7 +643,7 @@ def test_email_sent(self):
                 '70.100'),
             email.subject)
         self.assertTrue('29.900%' in email.body)
-        self.assertTrue('12.34' in email.body)
+        self.assertTrue('12.34.' in email.body)
```
hodgestar: I think this now needs to be `12.34 credits`?

hodgestar:
:+1: (but we should wait for Travis)

justinvdm:
:+1: once travis is happy, looks great

PR #1149, opened 2015-01-13
Don't monkeypatch socket module

jerith:
`moto` does this, and it breaks all other network access in obscure ways.


jerith:
I still need direct tests for `S3Helper`, but everything else should be ready for an initial review.

hodgestar:
Looks good.

hodgestar:
Left one question about moto server clean up in the S3Helper tests, otherwise looks good.

hodgestar:
:+1:

PR #1151, opened 2015-01-14
Format numbers on billing numbers in the UI

justinvdm:
We have a function to format numbers better, we should use it for statements and the credit amount shown in the top-right hand corner.

### DIFF
--- a/go/billing/templatetags/billing_tags.py
+++ b/go/billing/templatetags/billing_tags.py
@@ -3,9 +3,29 @@ from decimal import Decimal
 from django import template
 from django.utils.translation import ungettext
 
+from go.base.utils import format_currency
 from go.billing import settings
 
 register = template.Library()
+
+
+@register.filter
+def format_cents(v):
+    """returns a formatted string representation of a number in dollars from a
+    decimal cents value (cents are the billing system's internal representation
+    for monetary amounts).
+    """
+    v = v / Decimal('100.0')
+    return format_currency(v, places=settings.DOLLAR_DECIMAL_PLACES)
+
+
+@register.filter
+def format_credits(v):
+    """Returns a formatted string representation of a number in dollars from a
+    decimal cents value (cents are the billing system's internal representation
+    for monetary amounts).
+    """
+    return format_currency(v)
 
 
 @register.simple_tag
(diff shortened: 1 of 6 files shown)


justinvdm:
Ready for review.

go/billing/templatetags/billing_tags.py
```diff
@@ -3,32 +3,42 @@
...
 register = template.Library()
 
 
+@register.filter
+def format_cents(v):
+    """returns a formatted string representation of a number in dollars from a
+    decimal cents value (cents are the billing system's internal representation
+    for monetary amounts).
+    """
+    v = v / Decimal('100.0')
```
rudigiesler, line 18: What if `v` is not a `Decimal`?

justinvdm, line 18: At the moment we are assuming any cents value given is a decimal. This fits our purposes at the moment. The behaviour we are handling at the moment for `format_credits` is actually wrong, we should also always expect decimals, changing it do that.

go/billing/templatetags/billing_tags.py
```diff
@@ -3,32 +3,42 @@
...
 
+from go.base.utils import format_currency
 from go.billing import settings
 
 register = template.Library()
 
 
+@register.filter
+def format_cents(v):
+    """returns a formatted string representation of a number in dollars from a
```
hodgestar: Capital R?

go/billing/templatetags/billing_tags.py
```diff
@@ -3,32 +3,42 @@
...
+    decimal cents value (cents are the billing system's internal representation
+    for monetary amounts).
+    """
+    v = v / Decimal('100.0')
+    return format_currency(v, places=settings.DOLLAR_DECIMAL_PLACES)
+
+
+@register.filter
+def format_credits(v):
+    """Returns a formatted string representation of a number in dollars from a
```
hodgestar: Docstring seems the same as the one for the previous function?

hodgestar:
Left a couple of minor comments. Looks good otherwise. :)

go/billing/templatetags/billing_tags.py
justinvdm, line 18: Hmm, it looks like the `LineItem` model's `credits` field is an integer field, weird :/ will have to fix that before this can land.

justinvdm:
Review comments addressed, though the travis build will fail until we have changed `LineItem.credits` to a decimal field. Will make a ticket once we decide if we should do that.

hodgestar:
`LineItem.credits` appears to be the only billing model where we use an `IntegerField` rather than a `DecimalField`, so +1 on changing it to a decimal field.

justinvdm:
Now depends on #1152, otherwise ready for review.

hodgestar:
:+1: after #1152 lands and Travis is happy.

PR #1152, opened 2015-01-14
Change LineItem credits field from an integer to a decimal field

justinvdm:
Currently, all of our other credits field in the billing world are decimal field, so we need to make `LineItem.credits` a decimal field too.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #1153, opened 2015-01-20
HTTP API support opt outs

rudigiesler:
In response to the card: https://trello.com/c/6rJ76cdx/1-24-ndoh-http-api-to-support-opt-outs

### DIFF
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -251,10 +251,13 @@ class MessageResource(BaseResource):
         if not account.disable_optouts:
             optout_store = self.worker.vumi_api.get_user_api(user_account) \
                 .optout_store
-            optout = yield optout_store.get_opt_out(
-                payload.get('to_addr_type', 'msisdn'), payload.get('to_addr'))
+            to_addr_type = payload.get('to_addr_type', 'msisdn')
+            to_addr = payload.get('to_addr')
+            optout = yield optout_store.get_opt_out(to_addr_type, to_addr)
             if optout:
-                self.client_error_response(request, "Recipient has opted out")
+                self.client_error_response(
+                    request, "Recipient with %s %s has opted out" % (
+                        to_addr_type, to_addr))
                 return
         conversation = yield self.get_conversation(user_account)
 
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -247,13 +247,15 @@ class MessageResource(BaseResource):
     @inlineCallbacks
     def handle_PUT_send_to(self, request, payload):
         user_account = request.getUser()
-        optout_store = self.worker.vumi_api.get_user_api(user_account) \
-            .optout_store
-        optout = yield optout_store.get_opt_out(
-            payload.get('to_addr_type', 'msisdn'), payload.get('to_addr'))
-        if optout:
-            self.client_error_response(request, "Recipient has opted out")
-            return
+        account = yield self.worker.vumi_api.get_user_account(user_account)
+        if not account.disable_optouts:
+            optout_store = self.worker.vumi_api.get_user_api(user_account) \
+                .optout_store
+            optout = yield optout_store.get_opt_out(
+                payload.get('to_addr_type', 'msisdn'), payload.get('to_addr'))
+            if optout:
+                self.client_error_response(request, "Recipient has opted out")
+                return
         conversation = yield self.get_conversation(user_account)
 
         msg_options = SendToOptions(
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -256,8 +256,8 @@ class MessageResource(BaseResource):
             optout = yield optout_store.get_opt_out(to_addr_type, to_addr)
             if optout:
                 self.client_error_response(
-                    request, "Recipient with %s %s has opted out" % (
-                        to_addr_type, to_addr))
+                    request, 'Recipient with %s %s has opted out' % (
+                        str(to_addr_type), str(to_addr)))
                 return
         conversation = yield self.get_conversation(user_account)
 
--- a/go/apps/http_api_nostream/resource.py
+++ b/go/apps/http_api_nostream/resource.py
@@ -247,6 +247,13 @@ class MessageResource(BaseResource):
     @inlineCallbacks
     def handle_PUT_send_to(self, request, payload):
         user_account = request.getUser()
+        optout_store = self.worker.vumi_api.get_user_api(user_account) \
+            .optout_store
+        optout = yield optout_store.get_opt_out(
+            payload.get('to_addr_type', 'msisdn'), payload.get('to_addr'))
+        if optout:
+            self.client_error_response(request, "Recipient has opted out")
+            return
         conversation = yield self.get_conversation(user_account)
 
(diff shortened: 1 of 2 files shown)


go/apps/http_api_nostream/resource.py
```diff
@@ -247,6 +247,13 @@ def handle_PUT_in_reply_to(self, request, payload, in_reply_to):
     @inlineCallbacks
     def handle_PUT_send_to(self, request, payload):
         user_account = request.getUser()
+        optout_store = self.worker.vumi_api.get_user_api(user_account) \
+            .optout_store
+        optout = yield optout_store.get_opt_out(
+            payload.get('to_addr_type', 'msisdn'), payload.get('to_addr'))
```
justinvdm: I'm wondering if defaulting the to addr type to `msisdn` should happen in the message class to avoid us having different defaults all over the codebase. Don't think its something that should block this PR from landing though.

justinvdm: Ah, after looking at https://github.com/praekelt/vumi/pull/895 it looks like we are doing that. Maybe we don't need this default here then?

justinvdm: Actually, if we are setting the default to `None` (just noticed the comments on that), then +1 on us setting the default here. Sorry for the noise.

go/apps/http_api_nostream/resource.py
```diff
@@ -247,6 +247,13 @@ def handle_PUT_in_reply_to(self, request, payload, in_reply_to):
...
+        if optout:
+            self.client_error_response(request, "Recipient has opted out")
```
justinvdm: Wondering if this case should be counted as an error case. I would guess people using the api would just expect us to send to non-opted out users and ignore opted out users. It does seem like we should be saying something though.

If we are going to send some kind of response for the opt out cases, maybe we should be including the address of the opted out person and mention more clearly what happened, something like: "Skipping message send to opted out user +2731234567".

Thoughts?

justinvdm:
Do we maybe want whether opt outs are handled here to be configurable somehow (and enabled by default) instead of always being enabled?

justinvdm:
Left minor comments, looks good.

go/apps/http_api_nostream/resource.py
rudigiesler: I think adding the address and address type to the message would be useful.

I'm not so sure that it shouldn't be an error though, the message that the user wanted to send isn't going to be sent. I think that BAD REQUEST is a good response to this, unless you have suggestions for another response code?

rudigiesler:
Would this be configurable just a global on/off config switch, or should it be per account?

go/apps/http_api_nostream/resource.py
justinvdm: Hmm, I'm a bit on the fence here :) Ignoring opted out users when sending messages seems like behaviour the client using the api would expect, so they didn't really make a bad api request. On the other hand, it does seem weird just silently not sending to some of the users that the client asked that messages get sent to. @hodgestar what are your thoughts?

justinvdm:
I was thinking a per-conversation on/off switch, but not at all set on that. @rudigiesler @hodgestar thoughts?

go/apps/http_api_nostream/resource.py
jerith: Our "send message" response doesn't really allow us to return a 2xx code with anything other than the JSON of the vumi message we're sending. I'd rather return a 4xx response than silently pretend we're sending a message we aren't sending, but a 2xx response with "message not sent because contact opted out" would probably be better.

justinvdm: Agreed, lets go with 400 for now then.

rudigiesler: There is `success_response` which returns `OK` by default, but any http response with a custom reason, vs `successful_send_response` which you can only send with the json of a message. So I think both are possible.

rudigiesler:
If we do a per-conversation or per-account on/off switch, then it would require a migration? Unless we keep a separate table with a list of all the conversations that have opt outs switched off for the HTTP API?

go/apps/http_api_nostream/resource.py
justinvdm: Ah, if both are possible, I think 202 might be a good fit.

justinvdm:
@rudigiesler I was thinking we look it up as a config field in the conversation's config. Sorry, should've been clearer in my previous message.

go/apps/http_api_nostream/resource.py
justinvdm: From the slack discussion, happy with us using an error code here, since there doesn't see to be a success code that makes sense.

justinvdm:
From the slack discussion, happy with us using the account level opt out flag here instead. We can worry about a per-conversation opt out flag when we need one.

go/apps/http_api_nostream/tests/test_vumi_app.py
```diff
@@ -423,6 +423,64 @@ def test_send_to_with_evil_to_addr(self):
...
+                                           self.auth_headers, method='PUT')
+
+        self.assertEqual(response.code, http.BAD_REQUEST)
+        self.assertEqual(
+            response.headers.getRawHeaders('content-type'),
+            ['application/json; charset=utf-8'])
+        put_msg = json.loads(response.delivered_body)
+        self.assertEqual(put_msg['success'], False)
+        self.assertEqual(
+            put_msg['reason'], 'Recipient with msisdn +5432 has opted out')
```
justinvdm, line 450: Could we use `.get_dispatched_outbound()` here to check that we haven't sent the message?

go/apps/http_api_nostream/tests/test_vumi_app.py
```diff
@@ -423,6 +423,64 @@ def test_send_to_with_evil_to_addr(self):
...
+        response = yield http_request_full(url, json.dumps(msg),
+                                           self.auth_headers, method='PUT')
+
+        self.assertEqual(response.code, http.OK)
+        self.assertEqual(
+            response.headers.getRawHeaders('content-type'),
+            ['application/json; charset=utf-8'])
+        put_msg = json.loads(response.delivered_body)
+        self.assertEqual(put_msg['to_addr'], msg['to_addr'])
+        self.assertEqual(put_msg['content'], msg['content'])
```
justinvdm, line 482: Same as [here](https://github.com/praekelt/vumi-go/pull/1153/files#r23290758)

justinvdm, line 482: Err, sorry, I should probably explain a bit clearer: could we use `.get_dispatched_outbound()` here to check that we actually do send the message.

go/apps/http_api_nostream/tests/test_vumi_app.py
rudigiesler, line 450: Added here: https://github.com/praekelt/vumi-go/pull/1153/files#diff-79b5b2aa90051d6460e534bca53a771fR451

go/apps/http_api_nostream/tests/test_vumi_app.py
rudigiesler, line 482: Added here: https://github.com/praekelt/vumi-go/pull/1153/files#diff-79b5b2aa90051d6460e534bca53a771fR485

justinvdm:
:+1: when travis is happy

PR #1156, opened 2015-01-21
Update docs with http optout info

rudigiesler:


hodgestar:
I wonder whether we should explain that `to_addr_type` and `from_addr_type` default to `msisdn` and list the other possible values?

hodgestar:
Otherwise looks good. :)

rudigiesler:
We could do that. I didn't see anywhere in the docs where it explains what all the fields are and what each field is though, but I'll add it for the address type so long.

hodgestar:
Tx.

docs/http_api.rst
```diff
@@ -95,6 +104,10 @@ The response to the PUT request is the complete Vumi Go user message
 and includes the generated Vumi ``message_id`` which should be stored
 if you wish to be able to associate events with the message later.
 
+If a message is sent to a recipient that has opted out, the response will be an
+HTTP 400 error, with the body detailing that the recipient has opted out.
+Messages sent as a reply will still go through to an opted out recipient.
```
justinvdm: I think we should maybe mention how the error looks, like mention that `success` will be `false` and that the `reason` field will detail that the recipient has opted out.

justinvdm: Separate point -- should we mention that this behaviour can be overridden by setting the account `disable_optouts` flag to `true`?

justinvdm:
Left two minor comments, looks good.

hodgestar:
:+1: from me, although someone should review the commit I added.

justinvdm:
:+1: :cake: for keeping the docs up to date

PR #1116, opened 2015-01-21
Delete the parts of the billing API that we're not using.

justinvdm:
There are large portions of the billing API that are not being used, are a maintenance burden and which probably shouldn't be used because the operations are better performed from the Django admin interface (or at least an API that uses Django's models rather than side-stepping them).


justinvdm:
Ready for review.

go/billing/forms.py
```diff
@@ -91,22 +92,8 @@ def __init__(self, *args, **kwargs):
...
-        # Update the selected account's credit balance
-        account.credit_balance += transaction.credit_amount
-        account.last_topup_balance = account.credit_balance
-
-        account.save()
-
-        # Update the transaction's status to Completed
-        transaction.status = Transaction.STATUS_COMPLETED
-        transaction.save()
+        return load_account_credits(
```
hodgestar: `load_account_credits` doesn't return anything so wondering whether we should not `return` here.

go/billing/tests/helpers.py
```diff
@@ -74,7 +93,7 @@ def mk_transaction(account, tag_pool_name='pool1',
 
 def mk_statement(account,
                  title=settings.MONTHLY_STATEMENT_TITLE,
-                 type=Statement.TYPE_MONTHLY,
+                 _type=Statement.TYPE_MONTHLY,
```
hodgestar: I always think of leading underscores as meaning that something is unused or an implementation detail. Perhaps `stmt_type` would be better? Or `statement_type`?

go/billing/tests/test_api.py
```diff
@@ -50,6 +53,8 @@ def setUp(self):
 
     @inlineCallbacks
     def tearDown(self):
+        yield super(BillingApiTestCase, self).tearDown()
+
         for table in ('billing_lowcreditnotification', 'billing_transaction',
                       'billing_messagecost', 'billing_tagpool',
                       'billing_account'):
```
hodgestar: I wonder whether we still need to manually clean up these tables if we're always using `transaction=True`?

hodgestar:
Left a few small comments. Looks good though. Woot on the tests for the helpers.

go/billing/tests/helpers.py
justinvdm: Trying to avoid using a reserved word here, and went with the convention that twisted uses (if I'm remembering correctly) since that seems to be what we are most familiar with. Will go with `statement_type` if you prefer that.

justinvdm:
Ready for re-review, I think.

hodgestar:
:+1: once Travis is happy. We should also check how long Travis builds take now.

PR #1157, opened 2015-01-22
Add a provider field to billing models

justinvdm:


justinvdm:
Ready for review.

go/billing/settings.py
```diff
@@ -56,6 +56,11 @@
 LOW_CREDIT_NOTIFICATION_EMAIL = getattr(
     settings, 'BILLING_LOW_CREDIT_NOTIFICATION_EMAIL', 'support@vumi.org')
 
+PROVIDERS = getattr(
+    settings, 'BILLING_PROVIDERS', (
+        ('mtn', 'MTN'),
+        ('vodacom', 'Vodacom')))
```
hodgestar, line 62: I was thinking of putting these in a separate database table so that admins can add new ones without bugging developers, but happy with that happening in a separate PR.

hodgestar, line 62: Hmm. Changing that might require a migration so perhaps we should add it in this PR?

hodgestar:
Other than the question about where the provider list is stored, looks good.

justinvdm:
@hodgestar now that we have discussed using a config-based list of provider key-display name mappings, do you want to give this PR another review?

hodgestar:
:+1:

PR #1158, opened 2015-01-23
Billing statements split up items from the same biller into separate sections

justinvdm:
This is caused by [`groupby`](https://docs.python.org/2/library/itertools.html#itertools.groupby) being used on an unorded list of items.


justinvdm:
Ready for review

rudigiesler:
:+1:

PR #1159, opened 2015-01-23
Add a type field to transaction table

justinvdm:
We need a way of differentiating between topup and message transactions

### DIFF
--- a/go/billing/models.py
+++ b/go/billing/models.py
@@ -216,9 +216,21 @@ class Transaction(models.Model):
         (STATUS_REVERSED, STATUS_REVERSED),
     )
 
+    TRANSACTION_TYPE_MESSAGE = 'Message'
+    TRANSACTION_TYPE_TOPUP = 'Top Up'
+    TRANSACTION_TYPE_CHOICES = (
+        (TRANSACTION_TYPE_TOPUP, TRANSACTION_TYPE_TOPUP),
+        (TRANSACTION_TYPE_MESSAGE, TRANSACTION_TYPE_MESSAGE),
+    )
+
     account_number = models.CharField(
         max_length=100,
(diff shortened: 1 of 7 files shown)


justinvdm:
Ready for review.

go/billing/models.py
```diff
@@ -216,10 +216,22 @@ class Transaction(models.Model):
         (STATUS_REVERSED, STATUS_REVERSED),
     )
 
+    TRANSACTION_TYPE_MESSAGE = 'Message'
+    TRANSACTION_TYPE_TOPUP = 'Top Up'
+    TRANSACTION_TYPE_CHOICES = (
+        (TRANSACTION_TYPE_TOPUP, TRANSACTION_TYPE_TOPUP),
+        (TRANSACTION_TYPE_MESSAGE, TRANSACTION_TYPE_MESSAGE),
```
hodgestar, line 223: I think I have a vague preference for the database values to be `"message"` and `"topup"` rather than the same as the labels (just feels more like what I'd expect and means less remembering capitalization and spacing).

hodgestar:
One very minor suggestion, otherwise looks good.

I wonder if we should leave deploying this for Monday?

go/billing/models.py
justinvdm, line 223: I originally had it like that, and also prefer it, but changed it to capitals to stay consistent with what we are doing for the status field on `Transaction` and the message direction field on `MessageCost`.

justinvdm:
Just changed the credit loading util to set the transaction type to topup, thought it is relevant enough for this PR. Sorry to add something after declaring ready for review.

go/billing/models.py
hodgestar, line 223: Cool.

hodgestar:
:+1:

PR #1161, opened 2015-01-26
Change statement generation to only use message transaction types

justinvdm:
We need to only use message transaction types, which we don't do at the moment.

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -48,6 +48,9 @@ def get_transactions(account, statement):
 
 
 def get_message_transactions(transactions):
+    transactions = transactions.filter(
+        transaction_type=Transaction.TRANSACTION_TYPE_MESSAGE)
+
     transactions = transactions.values(
         'tag_pool_name',
         'tag_name',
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review.

go/billing/tasks.py
```diff
@@ -166,8 +174,10 @@ def get_channel_type(transaction, tagpools):
 def get_message_description(transaction):
     if transaction['message_direction'] == MessageCost.DIRECTION_INBOUND:
         return 'Messages received'
-    else:
+    elif transaction['message_direction'] == MessageCost.DIRECTION_OUTBOUND:
         return 'Messages sent'
+    else:
+        return None
```
rudigiesler, line 180: Could you explain to me why this is needed?  When is a message neither outbound nor inbound?

justinvdm, line 180: This isn't strictly necessary, but did lead to some confusion when debugging. Before this PR, we weren't differentiating between topup transactions and message transactions. Topup transactions don't have a message direction, so they were being treated as 'Messages sent' line items, since they followed the `else` path.

I can't think of any way this could possibly happen again, but thought it might save some confusion when trying to debug stuff if this did for some reason happen again (possibly because of some future code change).

rudigiesler:
:+1:

hodgestar:
Do we need to wait for all message transactions to have a transaction_type set before this can be deployed?

justinvdm:
@hodgestar since this is a monthly task that won't run until a few more days, I think it would be fine to deploy this first.

hodgestar:
We will need to do a data migration (either by hand or using South), but let's sort that out in a separate PR if needed.

:+1:

PR #1162, opened 2015-01-27
Update billing api to include providers when determining message costs

justinvdm:


justinvdm:
Ready for review. The billing dispatcher still needs to send the api the providers, but it seems separate enough to be done in its own PR.

go/billing/api.py
```diff
@@ -141,47 +142,42 @@ def render_POST(self, request):
...
+                      (t.name = %(tag_pool_name)s OR c.tag_pool_id IS NULL)
+                      AND
+                      (c.provider = %(provider)s OR c.provider IS NULL)
+                      AND
+                      (c.message_direction = %(message_direction)s)
             ) as t
-            ORDER BY t.account_number
+            ORDER BY
+                t.account_number NULLS LAST,
+                t.tag_pool_name NULLS LAST
```
hodgestar: We need a `c.provider NULLS LAST` order clause here, I think.

hodgestar: Interesting that the test for this passed -- maybe it was passing because the MessageCost row with null provider was created last? Ordering in SQL results is tricky to test.

justinvdm: [Had the same thought](https://github.com/praekelt/vumi-go/commit/294d80eb00afd8628a8b80aba90a469b2e947815) :)

hodgestar:
:+1: when Travis is happy.

PR #1160, opened 2015-01-23
Change billing api to set transaction types

justinvdm:
This will allow us to only use message transactions when generating statements.

PR #1165, opened 2015-01-29
Add a migration for transaction type field

justinvdm:


justinvdm:
Ready for review. Should only get deployed after #1160 gets deployed.

hodgestar:
:+1:

PR #1170, opened 2015-02-02
add session length fields to models

rudigiesler:


go/billing/models.py
```diff
@@ -176,6 +176,14 @@ class Meta:
         max_digits=10, decimal_places=3, default=Decimal('0.0'),
         help_text=_("The base cost per session in cents."))
 
+    session_unit_time = models.DecimalField(
+        max_digits=10, decimal_places=3, default=Decimal('20.0'),
+        help_text=_("The time of one billed unit (in seconds)."))
```
justinvdm, line 181: I was under the impression we were going to keep this static. I've got a feeling we are going to need to assume a static unit time when we creating session length transactions from end-of-session messages, otherwise things could get pretty complicated. Maybe it isn't as complicated as I'm imagining though. @rudigiesler @hodgestar thoughts?

justinvdm:
Could we update the billing test helpers for making transactions and message costs with these new fields? Both helpers need to be able to parse float values to decimals for each decimal field.

It seems like that parsing-to-decimal behaviour in the helpers is making updating models more tedious, we should probably change the billing tests to just use decimals, but that's maybe not something for right now.

justinvdm:
I think we need to update one/some of the `MessageCost` helper methods to know about session length costs and add a new helper method for calculating session length credits from session length costs.

Edit: Ah, it looks like we have a separate checklist item for this on the trello card, guessing this'll happen there.

justinvdm:
Left some comments but looks good.

go/billing/models.py
rudigiesler, line 181: I'm not sure I understand the problem you have. The way I assumed it would work, is that when we process a message that is an end-of-session message, we would convert the real time to an amount of session units. The cost would then be worked out by multiplying the amount of session time units by the per time unit cost. The billing statement would show the amount of session time units, the cost per time units, and the total cost.

If the session time changed at some time, it wouldn't matter because we bill in terms of session time units, instead of actual time.

justinvdm, line 181: Ya, thinking about it a bit more, having the session unit time's be changeable per message cost maybe isn't as complicated as I thought.

@hodgestar unless I misunderstood you, I think your preference was to use a static unit time of 20 seconds, instead of allowing each a message cost to have its own configurable session unit time. If this is the case, could you explain why?

hodgestar, line 181: @justinvdm My preference was to have a unit time specified in the message
cost (what's currently in the PR).

justinvdm, line 181: Ok, sorry guys, this is just me misunderstanding then, +1 on this approach.

hodgestar:
:+1: from me, but would like @justinvdm to give a plus-one too.

justinvdm:
:+1:

PR #1171, opened 2015-02-02
update messagecost forms with session length

rudigiesler:


go/billing/models.py
```diff
@@ -120,6 +120,24 @@ def calculate_session_credit_cost(cls, session_cost, markup_percent,
...
+            session_length_cost, markup_percent, context=context)
+
+    @classmethod
+    def calculate_session_length_credit_cost(
+            cls, session_length_credits, markup_percent, context=None):
+        """
+        Return the cost of a session length (in credits).
+        """
+        return cls.apply_markup_and_convert_to_credits(
+            session_length_credits, markup_percent, context=context)
```
justinvdm: Not understanding what this method is for, could you explain?

go/billing/models.py
```diff
@@ -120,6 +120,24 @@ def calculate_session_credit_cost(cls, session_cost, markup_percent,
             session_cost, markup_percent, context=context)
 
     @classmethod
+    def calculate_session_unit_credit_cost(
+            cls, session_length_cost, markup_percent, context=None):
+        """
+        Return the cost per session length (in credits).
+        """
+        return cls.apply_markup_and_convert_to_credits(
+            session_length_cost, markup_percent, context=context)
```
justinvdm: I might be misunderstanding, but should the param maybe be named `session_unit_cost` rather than `session_length_cost`?

go/billing/models.py
justinvdm: Ah, maybe its the param name that is throwing me off. Should we maybe have `session_length_cost` instead of `session_length_credits`?

justinvdm:
Thoughts on adding a method to `MessageCost` for calculating the session length cost from the session unit cost and session unit time?

justinvdm:
Should we maybe add an `@property` for the session unit credits like we have for the other costs for consistency's sake?

justinvdm:
Left some minor comments, looks good.

go/billing/models.py
rudigiesler: Yes this is a bit ambiguous. The param is the cost of a session of one unit length, hence `session_length_cost`, but I think that `session_unit_cost` makes more sense.

rudigiesler:
w.r.t Adding a method to `MessageCost` for calculating session length cost from session unit cost, session unit length and session time, it's a separate item on the trello card, so I thought I would do it in a separate PR.

justinvdm:
@rudigiesler ah sorry, missed the trello card item, +1 on doing that in a separate PR.

justinvdm:
:+1: when travis is happy

PR #1172, opened 2015-02-03
Fix message aggregate computations

jerith:
There are a few places where we compute aggregate statistics from the message store. Some of these were written with the now-invalid assumption that the message store cache contains all message keys for a conversation.

### DIFF
--- a/go/conversation/view_definition.py
+++ b/go/conversation/view_definition.py
@@ -5,6 +5,8 @@ import functools
 import re
 import sys
 from StringIO import StringIO
+from collections import defaultdict
+from datetime import datetime
 
 from django.conf import settings
 from django.views.generic import View, TemplateView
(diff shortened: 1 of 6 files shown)


go/conversation/view_definition.py
```diff
@@ -583,9 +585,28 @@ def get(self, request, conversation):
...
         writer = csv.writer(sio)
         direction = request.GET.get('direction', 'inbound')
-        writer.writerows(conversation.get_aggregate_count(direction))
-        return HttpResponse(sio.getvalue(),
-            content_type='text/csv; charset=utf-8')
+        writer.writerows(self.get_aggregate_counts(conversation, direction))
+        return HttpResponse(
+            sio.getvalue(), content_type='text/csv; charset=utf-8')
+
+    def get_aggregate_counts(self, conv, direction):
```
hodgestar, line 592: Should we have a unit test for this new method?

hodgestar:
One comment, otherwise looks good. Woot on getting the stats working again.

go/conversation/view_definition.py
jerith, line 592: It's only used in one place and we have tests covering that. Separate tests covering `get_aggregate_counts()` would add extra time to the test runs without adding any real value.

hodgestar, line 592: I disagree a bit -- the point of unit tests is to test small bits of functionality separately -- but I'm okay with this not being tested separately for now.

hodgestar:
:+1:

PR #1173, opened 2015-02-04
Allow restricting which connectors the metric middleware fires metrics for.

hodgestar:
This is specifically for when the middleware is used for dispatchers such as load balancers, which see every message on multiple connectors (and thus publish multiple copies of each metric).


hodgestar:
Ready for review.

jerith:
:+1:

PR #1174, opened 2015-02-05
update transaction cost calculations

rudigiesler:
Update the transaction cost calculations to include calculating the cost of the time length of a session.

### DIFF
--- a/go/billing/models.py
+++ b/go/billing/models.py
@@ -151,7 +151,7 @@ class MessageCost(models.Model):
             base_cost += session_cost
         if session_length and session_unit_length and session_unit_cost:
             units = (
-                Decimal(session_length)/Decimal(session_unit_length)
+                session_length / session_unit_length
                 ).to_integral_exact(rounding=ROUND_CEILING)
             base_cost += units * session_unit_cost
         return cls.apply_markup_and_convert_to_credits(
--- a/go/billing/models.py
+++ b/go/billing/models.py
@@ -1,4 +1,4 @@
-from decimal import Decimal
+from decimal import Decimal, ROUND_CEILING
 
 from django.db import models
 from django.db.models.signals import post_save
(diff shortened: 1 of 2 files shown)


go/billing/models.py
```diff
@@ -147,6 +149,11 @@ def calculate_credit_cost(cls, message_cost, storage_cost, markup_percent,
         base_cost = message_cost + storage_cost
         if session_created:
             base_cost += session_cost
+        if session_length and session_unit_length and session_unit_cost:
+            units = (
+                Decimal(session_length)/Decimal(session_unit_length)
```
justinvdm: Sorry to bikeshed -- pep8 will probably complain about the `/` not having spaces around it.

justinvdm: Separate, more relevant comment: should we be doing the `Decimal()` wrapping we are doing here? We seem to assume the other params are decimals, so maybe we should assume these two are decimals also.

rudigiesler: Good catch on the pep8, I wonder why my editor didn't highlight it. All change to assume we're getting a Decimal.

go/billing/models.py
```diff
@@ -147,6 +149,11 @@ def calculate_credit_cost(cls, message_cost, storage_cost, markup_percent,
         base_cost = message_cost + storage_cost
         if session_created:
             base_cost += session_cost
+        if session_length and session_unit_length and session_unit_cost:
+            units = (
+                session_length / session_unit_length
+                ).to_integral_exact(rounding=ROUND_CEILING)
```
justinvdm, line 155: Based on the slack #vumi-mgmt discussion, we should ceil the value here, I think.

justinvdm:
One more minor comment, otherwise looks good.

go/billing/models.py
justinvdm, line 155: Doh, misread, we already are using ceil, sorry.

justinvdm:
:+1: when travis is happy.

hodgestar:
Belated :+1:. :)

PR #1178, opened 2015-02-05
Remove unnecessary outbound message look-up in HTTP API event processing.

hodgestar:
See https://github.com/praekelt/vumi-go/blob/develop/go/apps/http_api_nostream/vumi_app.py#L265-L268. Notice that the outbound message is only used for logging and that an event can't be routed to an application worker unless the routing table dispatcher has found an outbound message for it.


hodgestar:
Ready for review.

jerith:
:+1: when Travis is happy.

hodgestar:
Test failures are because Travis switched to Riak 2.0.2.

PR #1080, opened 2014-11-27
Add S3 bucket utilities

hodgestar:
Utilities for uploading files to S3 buckets. Part of #868.

PR #1177, opened 2015-02-06
Support session length cost in billing api

justinvdm:


justinvdm:
Depends on #1080, otherwise ready for review.

go/billing/api.py
```diff
@@ -118,44 +126,42 @@ def _create_notification_mapping(self):
...
+
+            d.addCallbacks(self._render_to_json, self._handle_error,
+                           callbackArgs=[request], errbackArgs=[request])
+
         return NOT_DONE_YET
 
+    def _parse_post(self, request):
+        data = self._parse_json(request) or {}
+        data = dict((k, data.get(k)) for k in self.FIELDS)
+        fields = set(self.FIELDS) - set(self.NULLABLE_FIELDS)
```
hodgestar: Perhaps we should put `NON_NULLABLE_FIELDS = tuple(sorted(set(FIELDS) - set(NULLABLE_FIELDS)))` on the class instead of calculating this each time?

go/billing/api.py
```diff
@@ -228,34 +244,49 @@ def create_transaction_interaction(self, cursor, account_number,
         session_credits = MessageCost.calculate_session_credit_cost(
             session_cost, markup_percent)
 
+        session_unit_credits = MessageCost.calculate_session_unit_credit_cost(
+            session_unit_cost, markup_percent)
```
hodgestar: It seems a bit weird to be storing `session_unit_credits` because there isn't really a transaction cost assosiate with this, right? There's just `session_credits` and `session_len_credits`?

This may affect #1171 too.

hodgestar:
Left one minor comment and one important question. Otherwise looks good.

go/billing/api.py
justinvdm: Ya, their isn't any transaction cost associated to it, since the unit cost is a preliminary step for working out the session length cost. We have a corresponding credit field for every other cost field, so I added it for the sake of completeness. It might also be useful for auditing purposes, though that's a bit of a guess. Not really against removing it, but wondering if there is any reason to remove it. I guess we'd be storing slightly less things, not sure if that's important. Any other reasons?

justinvdm: Ah, just realised I might've misunderstood your comment. There is an associated transaction cost field, `session_unit_cost`. Is that maybe what you meant?

justinvdm: From discussion in RL, plan is to remove session unit credit cost field: it is the only field in the transaction table which isn't relevant to how costs are calculated, so it doesn't really make sense to have it around. Will be easy enough to add later if we need it.

justinvdm:
#1180 landed, ready for review.

hodgestar:
:+1:

PR #1182, opened 2015-02-06
allow contacts to be looked up through field

rudigiesler:
Allow contacts to be looked up by specifying a field and field value, as well as the delivery class. This is to form part of the plan to replace delivery classes with address types. Full plan discussed here: https://github.com/praekelt/go-contacts-api/pull/29#issuecomment-73244656

### DIFF
--- a/go/vumitools/contact/models.py
+++ b/go/vumitools/contact/models.py
@@ -222,8 +222,9 @@ class ContactStore(PerAccountStore):
     @Manager.calls_manager
     def new_smart_group(self, name, query):
         group_id = uuid4().get_hex()
-        group = self.groups(group_id, name=name,
-            user_account=self.user_account_key, query=query)
+        group = self.groups(
+            group_id, name=name, user_account=self.user_account_key,
+            query=query)
         yield group.save()
         returnValue(group)
 
--- a/go/vumitools/contact/models.py
+++ b/go/vumitools/contact/models.py
@@ -56,6 +56,13 @@ class ContactNotFoundError(ContactError):
     """Raised when a contact is not found"""
 
 
+def normalize_addr(contact_field, addr):
+    if contact_field == 'msisdn':
+        addr = '+' + addr.lstrip('+')
+    elif contact_field == 'gtalk':
+        addr = addr.partition('/')[0]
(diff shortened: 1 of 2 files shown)


go/vumitools/contact/models.py
```diff
@@ -56,6 +56,13 @@ class ContactNotFoundError(ContactError):
     """Raised when a contact is not found"""
 
 
+def normalize_addr(contact_field, addr):
+    if contact_field == 'msisdn':
+        addr = '+' + addr.lstrip('+')
+    elif contact_field == 'gtalk':
+        addr = addr.partition('/')[0]
+    return (contact_field, addr)
```
hodgestar: It seems odd that `normalize_addr` returns the `contact_field` too. Perhaps it should just return the normalized address?

go/vumitools/contact/models.py
```diff
@@ -322,12 +329,12 @@ def new_contact_for_addr(self, delivery_class, addr):
...
     @Manager.calls_manager
-    def contact_for_addr(self, delivery_class, addr, create=True):
+    def contact_for_addr_field(self, field, value, create=True):
         """
-        Returns a contact from a delivery class and address, raising a
+        Returns a contact from a field (address type) and address, raising a
         ContactNotFoundError exception if the contact does not exist.
         """
-        field, value = contact_field_for_addr(delivery_class, addr)
+        field, value = normalize_addr(field, value)
```
hodgestar: I think I would like `contact_for_addr_field` to not call `normalize_addr` because it's meant to be a low-level utility method. I'm happy for particular higher-level things to call `normalize_addr` if they like. This would mean we'd have to call `normalize_addr` in `contact_for_addr` for backwards compatibility.

Woot on splitting out `normalize_addr` though.

hodgestar:
It looks like this branch needs develop merged in in order to pass Travis tests? Although I'm confused by the earlier passing build on the branch.

hodgestar:
Left some minor comments, otherwise looks good.

rudigiesler:
@hodgestar Ready for another review.

hodgestar:
:+1:

PR #1184, opened 2015-02-09
Improve bulk sending to many contacts

jerith:
Currently the bulk send loads all contacts into memory even when not deduping them. When not deduping, we should add them to the window manager while loading them rather than storing them in memory.

We also have some more advanced plans for a separate PR:
- Perform a paginated contact query and store the continuation key in Redis so that the send can be resumed even if a worker falls over.
- Use a Radix tree to determine uniqueness when deduping (this will drastically reduce the storage requirements), using, e.g. https://pypi.python.org/pypi/py-radix. [Edit: py-radix is IP address specific, but maybe we can find something else when we get there].

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -123,6 +123,36 @@ class TestBulkMessageApplication(VumiTestCase):
             (yield self.app.window_manager.count_in_flight(window_id)), 0)
 
     @inlineCallbacks
+    def test_bulk_send_command(self):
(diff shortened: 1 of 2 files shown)


hodgestar:
:+1: (although maybe we should wait for Travis)

hodgestar:
It turns out the the contact look-up does one giant index query, so we're not landing this for now. Keeping it open for a bit in case we decide we want it anyway.

hodgestar:
Closing this for now -- we can always resurrect the code if needed.

PR #1183, opened 2015-02-09
Support session length costs on statements

justinvdm:


justinvdm:
Ready for review.

go/billing/tests/test_tasks.py
```diff
@@ -425,6 +427,151 @@ def test_generate_monthly_statement_sessions_different_markups(self):
...
+
+        statement = tasks.generate_monthly_statement(
+            self.account.id, *this_month())
+        [item1, item2, item3] = get_line_items(statement).filter(
+            description='Sessions (billed per 20s)')
+
+        self.assertEqual(item1.credits, get_session_length_credits(100, 10))
+        self.assertEqual(item1.unit_cost, 100)
+        self.assertEqual(item1.cost, 100)
+        self.assertEqual(item1.description, 'Sessions (billed per 20s)')
```
hodgestar: After discussion in RL, different `session_length_costs` should be grouped together into one item if `session_unit_cost` is the same for all of them and split out if it is different.

go/billing/tasks.py
```diff
@@ -105,6 +105,27 @@ def get_session_transactions(transactions):
...
+def get_session_length_transactions(transactions):
+    transactions = transactions.filter(
+        transaction_type=Transaction.TRANSACTION_TYPE_MESSAGE)
+
+    transactions = transactions.values(
+        'tag_pool_name',
+        'tag_name',
+        'provider',
+        'session_unit_time',
+        'session_length_cost',
```
hodgestar: We should group on `session_unit_cost` here instead of `session_length_cost` (same issue as one described in comment in test for this).

go/billing/tests/test_tasks.py
```diff
@@ -438,7 +585,7 @@ def test_generate_monthly_statement_unaccessible_tags(self):
         statement = tasks.generate_monthly_statement(
             self.account.id, *this_month())
 
-        [item] = get_line_items(statement).filter(channel=u'tag2.1')
+        item = get_line_items(statement).filter(channel=u'tag2.1').latest('id')
```
hodgestar, line 613: Just wondering what this change is for?

hodgestar:
One major comment about grouping on `session_unit_cost` instead of `session_length_credits` and one minor comment but otherwise looks good.

go/billing/tests/test_tasks.py
justinvdm, line 613: We are now getting a line item for messages received and one for session length, this change is to only check one of them, since that seems to sufficient for this test.

justinvdm:
Ready for re-review. Sorry, messed this one up.

hodgestar:
:+1:

PR #1185, opened 2015-02-10
Paginated contact keys for group

jerith:
We currently only have unpaginated contact key lookup for groups.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -33,10 +33,13 @@ def delete_group(account_key, group_key):
     # We do this one at a time because we're already saving them one at a time
     # and the boilerplate for fetching batches without having them all sit in
     # memory is ugly.
-    for contact_key in group.backlinks.contacts():
-        contact = contact_store.get_contact_by_key(contact_key)
-        contact.groups.remove(group)
-        contact.save()
+    contacts_page = group.backlinks.contact_keys()
+    while contacts_page is not None:
+        for contact_key in contacts_page:
+            contact = contact_store.get_contact_by_key(contact_key)
+            contact.groups.remove(group)
+            contact.save()
+        contacts_page = contacts_page.next_page()
     group.delete()
 
 
@@ -45,11 +48,17 @@ def delete_group_contacts(account_key, group_key):
     api = VumiUserApi.from_config_sync(account_key, settings.VUMI_API_CONFIG)
     contact_store = api.contact_store
     group = contact_store.get_group(group_key)
-    contacts = contact_store.get_contacts_for_group(group)
+    contacts_page = contact_store.get_contact_keys_for_group(group)
+    # FIXME: We pull all keys into memory to avoid modifying search results if
+    #        we're deleting contacts that are part of a smart group.
+    contact_keys = []
+    while contacts_page is not None:
+        contact_keys.extend(list(contacts_page))
+        contacts_page = contacts_page.next_page()
     # We do this one at a time because we're already saving them one at a time
     # and the boilerplate for fetching batches without having them all sit in
     # memory is ugly.
-    for contact_key in contacts:
+    for contact_key in contact_keys:
         contact_store.get_contact_by_key(contact_key).delete()
 
 
@@ -114,10 +123,13 @@ def contacts_to_csv(contacts, include_extra=True):
 
 
 def get_group_contacts(contact_store, *groups):
+    # TODO: FIXME: Kill this thing. It keeps all contact objects in memory.
     contact_keys = []
     for group in groups:
-        contact_keys.extend(contact_store.get_contacts_for_group(group))
-
+        contacts_page = contact_store.get_contact_keys_for_group(group)
+        while contacts_page is not None:
+            contact_keys.extend(list(contacts_page))
+            contacts_page = contacts_page.next_page()
     return contacts_by_key(contact_store, *contact_keys)
 
 
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -53,7 +53,7 @@ def delete_group_contacts(account_key, group_key):
     #        we're deleting contacts that are part of a smart group.
     contact_keys = []
     while contacts_page is not None:
-        contact_keys.extend(list(contacts_page))
+        contact_keys.extend(contacts_page)
         contacts_page = contacts_page.next_page()
     # We do this one at a time because we're already saving them one at a time
     # and the boilerplate for fetching batches without having them all sit in
@@ -128,7 +128,7 @@ def get_group_contacts(contact_store, *groups):
     for group in groups:
         contacts_page = contact_store.get_contact_keys_for_group(group)
         while contacts_page is not None:
-            contact_keys.extend(list(contacts_page))
+            contact_keys.extend(contacts_page)
             contacts_page = contacts_page.next_page()
     return contacts_by_key(contact_store, *contact_keys)
 
--- a/go/contacts/tests.py
+++ b/go/contacts/tests.py
@@ -43,7 +43,7 @@ def get_all_contact_keys_for_group(contact_store, group):
     keys = []
     index_page = contact_store.get_contact_keys_for_group(group)
     while index_page is not None:
-        keys.extend(list(index_page))
+        keys.extend(index_page)
         index_page = index_page.next_page()
     return keys
 
--- a/go/contacts/tests.py
+++ b/go/contacts/tests.py
@@ -35,6 +35,19 @@ def group_url(group_key):
     return reverse('contacts:group', kwargs={'group_key': group_key})
 
 
+def get_all_contact_keys_for_group(contact_store, group):
+    """
+    Walk the index pages returned from get_contact_keys_for_group() and collect
+    all the keys.
+    """
+    keys = []
+    index_page = contact_store.get_contact_keys_for_group(group)
+    while index_page is not None:
+        keys.extend(list(index_page))
+        index_page = index_page.next_page()
+    return keys
+
+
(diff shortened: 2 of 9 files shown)


jerith:
Builds will fail until praekelt/vumi#906 has been merged and released.

go/contacts/tasks.py
```diff
@@ -114,10 +123,13 @@ def contacts_to_csv(contacts, include_extra=True):
 
 
 def get_group_contacts(contact_store, *groups):
+    # TODO: FIXME: Kill this thing. It keeps all contact objects in memory.
```
justinvdm, line 126: I'm sure I'm overlooking something, but couldn't we just turn this function into a generator to avoid keeping all contact objects in memory?

go/contacts/tasks.py
```diff
@@ -33,10 +33,13 @@ def delete_group(account_key, group_key):
...
-        contact = contact_store.get_contact_by_key(contact_key)
-        contact.groups.remove(group)
-        contact.save()
+    contacts_page = group.backlinks.contact_keys()
+    while contacts_page is not None:
+        for contact_key in contacts_page:
+            contact = contact_store.get_contact_by_key(contact_key)
+            contact.groups.remove(group)
+            contact.save()
+        contacts_page = contacts_page.next_page()
```
justinvdm, line 42: We seem to do this all over the place

``` python
while contacts_page is not None:
    for contact_key in contacts_page:
        # do stuff
     contacts_page = contacts_page.next_page()
```

Would it maybe simplify things if we had a way of iterating through the paginated contact keys as a normal iterable? For example, if the paginated result set had some `.all` method that did that:

``` python
for contact_key in group.backlinks.contact_keys().all():
    # do stuff here
```

go/contacts/tests.py
```diff
@@ -61,6 +74,18 @@ def mkcontact(self, name=None, surname=None, msisdn=u'+1234567890',
             surname=unicode(surname or TEST_CONTACT_SURNAME),
             msisdn=unicode(msisdn), **kwargs)
 
+    def get_group_contact_keys(self, group):
+        contact_keys = []
+        contact_keys_page = group.backlinks.contact_keys()
+        while contact_keys_page is not None:
+            contact_keys.extend(contact_keys_page)
```
justinvdm, line 81: Why do we sometimes wrap the page in a `list()` before doing a `.extend()` and sometimes just give it directly to `.extend()`?

go/contacts/tasks.py
jerith, line 126: The problem's actually in the callers, which take the huge list and sort it or whatever. We need to do something about those, but they're out of scope for this PR.

go/contacts/tasks.py
jerith, line 42: We'd have to use something like the `PausingDeferredQueue` in go-api-toolkit, but the extra layers of juggling there probably aren't worth it for the small reduction in boilerplate we'd get. Either way, it's out of scope for this PR.

go/contacts/tests.py
jerith, line 81: Because I sometimes forget that I can just pass an iterable to `.extend()`. I'll fix those, thanks.

go/vumitools/contact/models.py
```diff
@@ -381,3 +394,80 @@ def contact_for_addr(self, delivery_class, addr, create=True):
...
+    @Manager.calls_manager
+    def next_page(self):
+        pages = list(self._further_pages[:])
+        next_current_page = yield self._current_page.next_page()
+        if next_current_page is not None:
+            pages = [next_current_page] + pages
+        if pages:
+            returnValue(type(self)(self.manager, *pages))
+        else:
+            returnValue(None)
```
justinvdm, line 473: `ChainedIndexPages` and `PaginatedSearch` seem to belong more in `vumi.persists` than in a vumi-go contacts module. Any reason we aren't putting it there?

go/vumitools/contact/models.py
```diff
@@ -381,3 +394,80 @@ def contact_for_addr(self, delivery_class, addr, create=True):
...
+        return self._cursor is not None
+
+    @Manager.calls_manager
+    def next_page(self):
+        if self._cursor is None:
+            returnValue(None)
+        cursor, results = yield self._get_page_of_results()
+        returnValue(type(self)(
+            self._model_proxy, self._max_results, self._query, cursor,
+            results))
```
justinvdm, line 440: I'm not understanding the value this helper class provides us over just performing the search explicitly, then making use of `IndexPage.next_page` to iterate through the results. Could you explain why we need it?

go/vumitools/contact/models.py
```diff
@@ -240,42 +240,55 @@ def get_group(self, key):
...
+        # TODO: FIXME: Avoid building up the whole set in memory.
         # Grab all contacts we can find
         contacts = set([])
         for groups in conversation.groups.load_all_bunches():
             for group in (yield groups):
-                group_contacts = yield self.get_contacts_for_group(group)
-                contacts.update(group_contacts)
+                index_page = yield self.get_contact_keys_for_group(group)
+                while index_page is not None:
+                    contacts.update(list(index_page))
```
justinvdm: Necessary `list()`?

go/vumitools/contact/models.py
jerith, line 473: Possibly -- they were originally less generic. I think it's okay to leave them where they're used for now and refactor later if we need similar functionality elsewhere.

justinvdm, line 473: Unless there are reasons to leave it here for now, I'd prefer we moved it now.

go/vumitools/contact/models.py
jerith, line 440: Search results are paginated differently, so we don't get an `IndexPage`. `PaginatedSearch` gives us one and allows us to walk through search results the same way we walk through index results.

go/vumitools/contact/models.py
jerith: Same as `.extend()`, I'll remove the explicit `list()` calls. Thanks!

go/vumitools/contact/models.py
```diff
@@ -240,42 +240,55 @@ def get_group(self, key):
...
 
-    def get_dynamic_contacts_for_group(self, group):
+    def get_dynamic_contact_keys_for_group(self, group):
         """
         Use Riak search to find matching contacts.
         """
-        return self.contacts.raw_search(group.query).get_keys()
+        return self.search_contacts(group.query)
+
+    def search_contacts(self, query):
```
justinvdm, line 283: The other method names seem to be explicit about returning contact keys instead of contact objects. Maybe this method name should too? `search_contacts` could imply the objects are returned, I think.

go/vumitools/contact/models.py
jerith, line 473: Moving them will require another vumi PR and release, and there currently aren't any other things that could benefit. Also, we need to move away from Riak 1.x search before we can switch to Riak 2.x, so encouraging new code to use the old search will work against that.

go/contacts/tasks.py
justinvdm, line 42: If it is as complicated using something like `PausingDeferredQueue`, then agreed that it is out of scope. I'm not understanding why we'd have to use something like that instead of just moving this boilerplate into a method or function though. Wouldn't doing something like this be identical to what we are doing now, except without the repetition?

``` python
def all(page):
    while page is not None:
        for key in page:
            yield key
        page = page.next_page()    
```

jerith, line 42: No, because `.next_page()` returns a `Deferred` and Python's looping constructs don't understand those.

go/vumitools/contact/models.py
justinvdm, line 473: I would think that the way `PaginatedSearch` is used would look the same regardless of how the searching it does looks under the hood. I'm not convinced, but my request is bikesheddy enough to leave as is for now.

jerith, line 473: Until we have code that uses the new search stuff, I don't want to make any assumptions about how it will work.

go/vumitools/contact/models.py
jerith, line 283: I read it as "search for contacts matching the query" without a particular indication of the return value. `search_contact_keys` implies that only the keys will be searched (which is untrue) and `search_contacts_and_return_keys` is unwieldy. I'm open to suggestions on the naming, though.

justinvdm, line 283: Ya, isn't the easiest thing to name, also can't think of anything better. Happy with `search_contacts`.

go/vumitools/contact/tests/test_models.py
```diff
@@ -278,3 +279,341 @@ def test_new_contact_for_addr_gtalk(self):
...
+
+        first_page = yield store.get_contact_keys_for_group(group)
+        self.assertEqual(list(first_page), [matching_contact.key])
+        # This is the empty last page of the search results.
+        second_page = yield first_page.next_page()
+        self.assertEqual(list(second_page), [])
+        # This is the empty only page of the index results.
+        third_page = yield second_page.next_page()
+        self.assertEqual(list(third_page), [])
+        self.assertEqual(third_page.has_next_page(), False)
```
justinvdm, line 360: Would it be worth testing the case where we have a non-empty index page (if the smart group had static contacts)?

justinvdm:
[One minor comment](https://github.com/praekelt/vumi-go/pull/1185#discussion_r24491945) left, otherwise looks good.

go/vumitools/contact/tests/test_models.py
jerith, line 360: It is, thanks. I'll add the test.

justinvdm:
:+1:

PR #1186, opened 2015-02-12
Clean up message store cache rebuilding

jerith:
A bunch of message store cache rebuilding code is now obsolete and needs to be fixed or removed.


hodgestar:
:+1: once Travis is happy.

PR #1187, opened 2015-02-12
Use in-memory Riak backend instead of leveldb

jerith:


jerith:
The build time is at the low end of our usual time range, but less disk access might reduce the variance when Travis is busy.

hodgestar:
screenshot from 2015-02-12 11 02 38

hodgestar:
My slight concern is that this makes our Travis builds more different to production.

jerith:
> My slight concern is that this makes our Travis builds more different to production.

The set of features we use is identical between the `memory` and `eleveldb` backends. I'm specifically keeping vumi builds on `eleveldb` to catch any potential issues with our persistence layer there.

calston:
:+1:

hodgestar:
Happy for this to land and it's easy enough to switch back -- mostly just
wanted to register the concern.

PR #1191, opened 2015-02-13
Feature/http action

dbreedt:


go/apps/dialogue/vumi_app.js
```diff
@@ -21,6 +22,8 @@ var DialogueApp = App.extend(function(self) {
     };
 
     self.init = function() {
+        self.http = new JsonApi(self.im);
```
hodgestar: Might be worthwhile only constructing this in `self.types.http_json`.

go/base/static/js/src/apps/dialogue/models.js
```diff
@@ -182,6 +183,39 @@
...
+
+    defaults: function() {
+      return _({
+        text: '',
+        url: '',
+        http_method: '',
+        httpmeth: '',
+        entry_endpoint: {},
+        exit_endpoint: {}
+      }).defaults(GroupStateModel.__super__.defaults.call(this));
```
justinvdm, line 206: A few things here:
- Is `text` relevant for this state?
- Maybe `POST` would be a sensible default for `http_method`
- What does `httpmeth` do?
- Should we maybe make `url` not have a default? Not sure if a sensible default exists for it.

justinvdm, line 206: Ah, one last comment here: I think `http_method` should be `method` to match what we have in the relations above.

go/base/static/js/src/apps/dialogue/models.js
```diff
@@ -182,6 +183,39 @@
...
+      includeInJSON: 'method'
+    }, {
+      type: Backbone.HasOne,
+      key: 'entry_endpoint',
+      relatedModel: DialogueEndpointModel
+    }, {
+      type: Backbone.HasOne,
+      key: 'exit_endpoint',
+      relatedModel: DialogueEndpointModel
+    }],
```
justinvdm, line 198: I'm not sure if backbone-relational allows relations that are actually not backbone model types, but I think we should try stick to only having relations for backbone model types (so `url` and `method` would go).

justinvdm:
Minor style comment: could we try stick to one property per line in objects:

``` javascript
var foo = {
  bar: 23,
  baz: 32
};
```

instead of 

``` javascript
var foo = {bar: 23, baz: 32};
```

go/apps/dialogue/vumi_app.js
```diff
@@ -166,6 +167,27 @@ var DialogueApp = App.extend(function(self) {
...
+        var payload = {
+	        user: {
+		        answers: self.im.user.answers
+	        },
+	        contact: self.contact,
+            conversation_key: self.poll.conversation_key
+	    };
+
+	    return self
+	        .http.request(desc.method, desc.url, desc.method !== 'GET' ? { data: JSON.stringify(payload) } : {})
```
justinvdm: For the non-`GET` cases, we should give `.http.request` the unstringified data. For the `GET` case, we should give it `null` to tell it to not use a body for the request.

go/apps/dialogue/vumi_app.js
```diff
@@ -166,6 +167,27 @@ var DialogueApp = App.extend(function(self) {
...
+                .then(function(request) {
+		            if(request.body){
+			            return request.body;
+		            }
```
justinvdm: The promise is fulfilled with the response data, so we should rename `request` to `response`. Could you explain what we are doing here?

go/apps/dialogue/tests/test_vumi_app.js
```diff
@@ -318,5 +320,49 @@ describe("app", function() {
...
+                    .setup.user.addr('+27123')
+                    .setup.user.state('choice-1')
+                    .input('4')
+                    .check(function(api) {
+                        assert.equal(api.http.fixtures.fixtures[0].responses[0].code, 200);
+                    })
+                    .run();
+            });
+            
+            it("and it is a POST type the response should be 200", function() {
```
justinvdm: I'm not sure if this test description makes sense. I think this test is testing that we send the right data to the api. Could we change the test description to mention that? Similarly for the `GET` case.

justinvdm:
Other minor style comment: could we try stick to 4 spaces for indentation?

justinvdm:
Left some minor comments, looks good though :)

go/apps/dialogue/tests/dummy_polls.js
```diff
@@ -65,6 +67,24 @@ var simple_poll = {
...
+        exit_endpoint: {uuid: "endpoint-13"}
+    },
+    {
+        uuid: "http_json-2",
+        name: "json post 2",
+        type: "httpjson",
+        url: "www.foo.bar", 
+        method: "POST",
+        entry_endpoint: {uuid: "endpoint-15"},
+        exit_endpoint: {uuid: "endpoint-13"}
```
hodgestar: I think for sanity this should be `endpoint-16` and we should add `endpoint-16 -> endpoint-10` to the connection list since endpoint IDs are supposed to be unique.

go/apps/dialogue/tests/test_vumi_app.js
```diff
@@ -318,5 +320,68 @@ describe("app", function() {
...
+                                url: "www.foo.bar"
+                            }
+                        });
+                    })
+                    .setup.user.addr('+27123')
+                    .setup.user.state('choice-1')
+                    .input('4')
+                    .check(function(api) {
+                        assert.equal(api.http.fixtures.fixtures[0].responses[0].code, 200);
+                    })
```
hodgestar, line 343: I think we should also check that the end state was reached (i.e. that the person was successfully routed through to the next state).

go/apps/dialogue/tests/test_vumi_app.js
```diff
@@ -318,5 +320,68 @@ describe("app", function() {
...
+                                            'choice-1': 'value-5'
+                                        }
+                                    }, 
+                                    contact: api.contacts.store[0], 
+                                    conversation_key: poll.conversation_key
+                                })
+                            }
+                        });
+
+                        api.contacts.store[0].extra = {};
```
hodgestar, line 381: Why do we set `.extra` to an empty object here?

go/base/static/templates/apps/dialogue/states/httpjson/edit.jst
```diff
@@ -0,0 +1,6 @@
+<div class="form-group">
+  <label class="control-label">Method (GET/POST)</label>
+  <input class="text form-control httpjson-method" value="<%= method %>">
```
hodgestar: I wonder if perhaps this should be a dropdown? If not, I think we need to handle and test for strange values entered by people?

go/base/static/templates/apps/dialogue/states/httpjson/edit.jst
```diff
@@ -0,0 +1,6 @@
...
+  <label class="control-label">Url</label>
```
hodgestar, line 7: `Url` should be `URL`?

go/base/static/templates/apps/dialogue/states/httpjson/preview.jst
```diff
@@ -0,0 +1,6 @@
+<p class="text">
+Http Json Method: <span class="httpjson-method"><%= method %></span>
```
hodgestar: `Http Json Method` -> `HTTP JSON method` and `Http Json Url` -> `HTTP JSON URL`.

hodgestar:
Looking great! I left a bunch of small comments. Apologies if some of them are a bit nitpicky -- just trying to get the details right. Apologies also for the slight delay in reviewing.

hodgestar:
Further nitpick: I noticed that some of the lines have trailing whitespace. Would you mind removing that?

go/apps/dialogue/tests/dummy_polls.js
dbreedt: ok will do

go/apps/dialogue/tests/test_vumi_app.js
dbreedt, line 343: i will add the extra check

go/apps/dialogue/tests/test_vumi_app.js
dbreedt, line 381: extra is empty when the fixture is setup, if I pre-populate during the fixture and not empty it after the fixture has been created the choise adds two more entries to extra and then the test fails.

go/base/static/templates/apps/dialogue/states/httpjson/edit.jst
dbreedt: I'll make this a drop down.

go/apps/dialogue/tests/test_vumi_app.js
hodgestar, line 381: Ah, I see. Could you maybe add a comment like `// clear out extra so that the choice state can add them back`?

dbreedt:
The branch has been updated with the changes requested.

justinvdm:
:+1:

PR #1193, opened 2015-02-16
Add a Travis build using Riak 2.0 with legacy search enabled.

jerith:


jerith:
This is basically the same change as praekelt/vumi#911.

hodgestar:
:+1:

PR #1195, opened 2015-02-18
Add developer flag for accounts

rudigiesler:


go/billing/models.py
```diff
@@ -50,6 +50,11 @@ class Account(models.Model):
                     "in the calculation to determine when low-credits "
                     "notifications should be sent."))
 
+    is_developer = models.BooleanField(
+        default=False,
+        help_text=_("Determines whether or not the billing account is a "
+                    "developer account."))
```
hodgestar: I thought that perhaps we would do this as a feature flag on our `go.vumitools.account` objects?

rudigiesler: How is the billing account linked to `go.vumitools.account`?

hodgestar: Via `account_number`?

So something like: `vumi_api.get_user_account(account.account_number)`.

rudigiesler:
Ready for review again

go/billing/admin.py
```diff
@@ -36,6 +38,29 @@ def load_credits(self, request, queryset):
         return HttpResponseRedirect(redirect_url)
     load_credits.short_description = "Load credits for selected accounts"
 
+    def is_developer(self, obj):
+        vumi_api = vumi_api_for_user(obj.user)
+        account = vumi_api.get_user_account()
+        return account.is_developer
```
hodgestar, line 44: I'm a little concerned that this introduces a lot of riak look-ups on the admin pages, but I don't have a better solution so just noting this for now.

go/billing/tests/test_admin.py
```diff
@@ -72,3 +75,20 @@ def test_account_admin_view(self):
         self.assertContains(response, 'Description')
         self.assertContains(response, 'Credit balance')
         self.assertContains(response, 'Last topup balance')
+        self.assertContains(response, 'Is developer')
+
```
hodgestar, line 87: Could we perhaps have an additional separate test for `is_developer`? I know it's tested by the two new tests, but since it's used separately in the list view it feels like it could do with its own.

hodgestar, line 87: Ah, I see what is actually missing are tests for the `set_developer_flag` and `unset_developer_flag` actions?

hodgestar:
Just left a request for tests for the new admin actions -- otherwise looks good.

go/billing/tests/test_admin.py
rudigiesler, line 87: Yes, I wasn't sure exactly how to test this, so I tested the functions directly. Are there any examples of how to test Django admin actions?

justinvdm:
Other than the current comments, looks good to me.

go/billing/tests/test_admin.py
rudigiesler, line 87: Okay so I think I've figured out a good way of testing (writing this down for future reference). Call the admin function directly, passing a `queryset` that you get in the usual manner, and an instance of an empty mock object as `request`.

Example of the test here: https://github.com/praekelt/vumi-go/pull/1195/files#diff-c2e7d22d458eb41085f772dc23e9f732R104

rudigiesler:
Tests added. Ready for review again.

hodgestar:
:+1:

Perhaps let's land on Monday morning though?

PR #1198, opened 2015-02-20
HTTP conversation metrics resource looks up aggregates badly

hodgestar:
At https://github.com/praekelt/vumi-go/blob/develop/go/apps/http_api_nostream/resource.py#L288 it uses `getattr(metrics, agg_name)` instead of `metrics.Aggregator.from_name(agg_name)` which was specifically written for this purpose.

For backwards compatibility, we'll probably need to make this look up case insensitive.


hodgestar:
Ready for review.

jerith:
:+1: when Travis is happy.

PR #1201, opened 2015-02-20
Fix message store timestamp index parsing in go_account_stats

hodgestar:
Currently the Vumi message store formats timestamps in indexes using `str(msg['timestamp'])` which results in different behaviour depending on whether the microseconds are zero or not:

```
>>> str(n.replace(microsecond=0))
'2015-03-05 22:11:00'
>>> str(n)
'2015-03-05 22:11:00.124000'
```

Current go_account_stats only supports timestamps with microseconds set to 0 (unlikely in production).

### DIFF
--- a/go/base/management/commands/go_account_stats.py
+++ b/go/base/management/commands/go_account_stats.py
@@ -1,3 +1,5 @@
+import re
+
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

smn:
:+1:

PR #1202, opened 2015-02-20
Handle 0 session length unit costs when generating statements

justinvdm:
At the moment this is causing zero divisions.

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -172,7 +172,11 @@ def get_count(transaction):
 def get_session_length_count(transaction):
     length_cost = get_session_length_cost(transaction)
     unit_cost = get_session_length_unit_cost(transaction)
-    return length_cost / (unit_cost if unit_cost != 0 else 1)
+
+    if unit_cost == 0:
+        return Decimal(0)
+    else:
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review.

go/billing/tasks.py
```diff
@@ -171,7 +171,8 @@ def get_count(transaction):
 
 def get_session_length_count(transaction):
     length_cost = get_session_length_cost(transaction)
-    return length_cost / get_session_length_unit_cost(transaction)
+    unit_cost = get_session_length_unit_cost(transaction)
+    return length_cost / (unit_cost if unit_cost != 0 else 1)
```
hodgestar: Is it possible for `length_cost` to be non-zero when `unit_cost == 0`? If so, perhaps we should just explicitly return zero if the unit cost is zero?

hodgestar:
One small question, otherwise looks good.

go/billing/tasks.py
justinvdm: Not sure how useful the change is, but done.

hodgestar:
:+1:

PR #1206, opened 2015-02-24
Improve bulk sending to many contacts

jerith:
Same general idea as #1184, but a better implementation.

### DIFF
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -98,16 +98,19 @@ class TestBulkMessageApplication(VumiTestCase):
         self.clock = Clock()
         self.patch(WindowManager, 'get_clock', lambda _: self.clock)
 
-        self.app = yield self.app_helper.get_app_worker({})
-        self._setup_wait_for_window_monitor()
-
-    def _setup_wait_for_window_monitor(self):
         # Hackery to wait for the window manager on the app.
         self._wm_state = {
             'queue': DeferredQueue(),
             'expected': 0,
         }
-        orig = self.app.window_manager._monitor_windows
+        self.app = yield self.get_app_worker(extra_worker=False)
+
+    @inlineCallbacks
+    def get_app_worker(self, extra_worker=True):
+        app = yield self.app_helper.get_app_worker(
+            {}, extra_worker=extra_worker)
+
+        orig = app.window_manager._monitor_windows
 
         @inlineCallbacks
         def monitor_wrapper(*args, **kw):
@@ -115,8 +118,8 @@ class TestBulkMessageApplication(VumiTestCase):
             yield orig(*args, **kw)
             self._wm_state['queue'].put(object())
 
-        self.patch(
-            self.app.window_manager, '_monitor_windows', monitor_wrapper)
+        self.patch(app.window_manager, '_monitor_windows', monitor_wrapper)
+        returnValue(app)
 
     @inlineCallbacks
     def wait_for_window_monitor(self):
@@ -340,7 +343,7 @@ class TestBulkMessageApplication(VumiTestCase):
         self.app_helper.clear_dispatched_outbound()
         # Set up an unbroken worker to process the redelivered command and
         # redeliver the command.
-        yield self.app_helper.get_app_worker(extra_worker=True)
+        yield self.get_app_worker()
         yield self.app_helper.dispatch_command("bulk_send", **command_params)
         # Have the window manager deliver the messages.
         self.clock.advance(self.app.monitor_interval + 1)
@@ -407,7 +410,7 @@ class TestBulkMessageApplication(VumiTestCase):
         self.app_helper.clear_dispatched_outbound()
         # Set up an unbroken worker to process the redelivered command and
         # redeliver the command.
-        yield self.app_helper.get_app_worker(extra_worker=True)
+        yield self.get_app_worker()
         yield self.app_helper.dispatch_command("bulk_send", **command_params)
         # Have the window manager deliver the messages.
         self.clock.advance(self.app.monitor_interval + 1)
@@ -461,7 +464,7 @@ class TestBulkMessageApplication(VumiTestCase):
         self.assertEqual(["hello 1"] * 3, [m["content"] for m in msgs])
 
         # Set up a second worker to process the second command.
-        app2 = yield self.app_helper.get_app_worker(extra_worker=True)
+        app2 = yield self.get_app_worker()
         send_pauser2 = MessageSendPauser(app2, 3)
         send_pauser2.patch_app()
         second_d = self.app_helper.dispatch_command(
--- a/go/apps/bulk_message/tests/test_vumi_app.py
+++ b/go/apps/bulk_message/tests/test_vumi_app.py
@@ -152,6 +152,72 @@ class TestBulkMessageApplication(VumiTestCase):
         self.assertEqual(contact_addrs, msg_addrs)
 
     @inlineCallbacks
+    def test_bulk_send_command_with_duplicates(self):
+        """
+        If we send a bulk message to a number of contacts, we send a message to
+        the msisdn for each contact, even if we have duplicate msisdns.
+        """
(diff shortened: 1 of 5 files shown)


jerith:
The individual commits in this PR build logically on top of each other, so reviewing them one at a time might be easier than looking at the whole diff at once.

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -123,6 +122,268 @@ def test_consume_events(self):
...
+            "bulk_send",
+            user_account_key=conversation.user_account.key,
+            conversation_key=conversation.key,
+            batch_id=batch_id,
+            dedupe=False,
+            content="hello world",
+            delivery_class="sms",
+            msg_options={},
+        )
+        yield self.app_helper.kick_delivery()
```
hodgestar: Bit surprised we have to call `kick_delivery` here.

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -123,6 +122,268 @@ def test_consume_events(self):
...
+            msg_options={},
+        )
+        # We don't yield here, because this message will break the consumer and
+        # never be acked.
+        self.app_helper.dispatch_command("bulk_send", **command_params)
+        yield send_broken.get()
+        # Tell the fake broken that we've handled this message, because the
+        # now-broken consumer can't.
+        self.app.control_consumer._in_progress -= 1
+        self.app.control_consumer.channel.message_processed()
```
hodgestar: Could we avoid these if we put them into a `finally:` clause in `Consumer.consume`?

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -123,6 +122,268 @@ def test_consume_events(self):
...
+
+        send_progress = yield self.app.get_send_progress(conversation)
+        self.assertEqual(send_progress, sorted(c.key for c in contacts)[3])
+
+        self.app_helper.clear_dispatched_outbound()
+        # Now we undo the breaking patch, replace the broken command consumer,
+        # and redeliver the command message.
+        self.app.send_message_via_window = send_message_via_window
+        yield self.app._go_setup_command_consumer(self.app.get_static_config())
+        yield self.app_helper.dispatch_command("bulk_send", **command_params)
```
hodgestar, line 347: It would be more realistic create a new / second app_helper?

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -123,6 +122,268 @@ def test_consume_events(self):
...
+            dedupe=True,
+            content="hello world",
+            delivery_class="sms",
+            msg_options={},
+        )
+        yield self.app_helper.kick_delivery()
+        self.clock.advance(self.app.monitor_interval + 1)
+
+        contacts = yield self.get_opted_in_contacts(conversation)
+        msgs = yield self.app_helper.wait_for_dispatched_outbound(5)
```
hodgestar: Is there a reason we're calling `wait_for_dispatched_outbound(5)` rather than `get_dispatched_outbound()`?

go/apps/bulk_message/tests/test_vumi_app.py
```diff
@@ -123,6 +122,268 @@ def test_consume_events(self):
...
+        send_message_via_window = self.app.send_message_via_window
+        send_broken = DeferredQueue()
+        messages_sent = [0]
+
+        def breaking_send_message_via_window(*args, **kw):
+            if messages_sent[0] >= 4:
+                send_broken.put(None)
+                raise Exception("oops")
+            messages_sent[0] += 1
+            return send_message_via_window(*args, **kw)
```
hodgestar: Could we factor out the `breaking_send_message_via_window` into a function or callable class?

go/apps/bulk_message/vumi_app.py
```diff
@@ -55,6 +57,18 @@ def on_window_cleanup(self, window_id):
     def get_window_id(self, conversation_key, batch_id):
         return ':'.join([conversation_key, batch_id])
 
+    def _send_progress_key(self, conv):
+        return ':'.join([self.worker_name, conv.key, conv.batch.key])
```
hodgestar: Why do we include the batch key here?

hodgestar: What happens if someone initiates two bulk sends from the same conversation in a row with different text?

go/apps/bulk_message/vumi_app.py
```diff
@@ -70,26 +84,62 @@ def send_message_via_window(self, conv, window_id, batch_id, to_addr,
...
-        to_addresses = []
-        for contacts_batch in (
-                yield conv.get_opted_in_contact_bunches(delivery_class)):
-            for contact in (yield contacts_batch):
-                to_addresses.append(contact.addr_for(delivery_class))
-        if dedupe:
-            to_addresses = set(to_addresses)
+        # NOTE: We fetch all contact keys before we start. This is suboptimal,
+        #       but we need the keys sorted and deduplicated.
+        contact_keys = yield conv.get_contact_keys()
```
hodgestar, line 113: Just noting that we're still pulling all contact keys into memory at the start.

hodgestar:
Reviewed.

Thoughts:
- I think the question of how to handle multiple overlapping sends from the same conversation is important (especially since there is no easy way to see if a send has completed).
- We're still retrieving all contact keys up front into memory and then looking up all contacts, which isnt great.
- The big win is that we won't send messages multiple times if a send falls over half way.

go/apps/bulk_message/tests/test_vumi_app.py
jerith: Yes, but that may have other implications and I didn't really want to make a change to our central message processing loop just for a couple of tests.

go/apps/bulk_message/tests/test_vumi_app.py
jerith, line 347: I tried four or five different approaches here, and this is the one that ended up being the cleanest. We don't keep any state on the worker when we process these messages, so reusing the same worker is safe.

go/apps/bulk_message/tests/test_vumi_app.py
jerith: I copied this from an earlier test that did it this way. I think having the `kick_delivery()` before the `clock.advance()` means that the messages haven't necessarily been sent by the time we want them. I'll see if I can rewrite this to be a little more sensible.

go/apps/bulk_message/tests/test_vumi_app.py
jerith: Probably. I'll give it a try.

go/apps/bulk_message/vumi_app.py
jerith: I wanted to use the same key for both the window manager and the progress tracker. I'm open to suggestions for other ways to do this, though.

hodgestar: Sounds like a good enough reason.

go/apps/bulk_message/vumi_app.py
jerith, line 113: Yes. Changing that requires group contact listing code that returns sorted results, which is out of scope for this PR. (I can think of a way to do it, but it involves walking all relevant index/search result pages simultaneously (assuming search results are sorted by key) and that's a problem for search results if the result set is changing over the course of the query.)

We've seen in production that building a list of all contact keys before we start isn't as big a problem as loading all the contact objects to get the addresses, though.

jerith:
Unfortunately, we don't have any kind of unique command identifier we can use for differentiating between an overlapping send and an interrupted send. Maybe we can use a hash of the message content in place of the (redundant) batch identifier in the progress key? That will handle overlapping sends with the same content poorly, but will allow overlapping sends with different content.

jerith:
Ready for another review.

hodgestar:
:+1: once Travis is happy.

jerith:
This could use a quick review after the fix to the tests.

hodgestar:
:+1:

PR #1207, opened 2015-02-24
Filter line items with 0 units when creating statements

justinvdm:
Since session interval line items have 0 unit counts when the session unit cost is 0: We work out the unit count as `total_session_length_cost / session_unit_cost`, but use `0` if the total session length cost (and thus also the unit cost) is zero. These items are more confusing than they are useful.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #1208, opened 2015-02-25
Unique identifier for command messages

jerith:
Turns out it's useful to be able to differentiate between different commands that are otherwise identical.


justinvdm:
Looks good to me. I haven't taken a look at #1206 yet though, which I think is related to this PR, so might be best if @hodgestar also takes a look.

hodgestar:
:+1: from me too.

PR #1209, opened 2015-02-25
MessageCosts don't consider provider for uniqueness

justinvdm:
At the moment, `MessageCost`s are considered unique using the account, tagpool and message direction. In order for us to create a message cost with the same account, tagpool and message direction but a different provider, we need to add the provider field to this set of fields.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #1210, opened 2015-02-26
billing api to block accounts with insufficient credits

rudigiesler:
The billing API should return something in it's reply to notify the billing dispatcher that the account has run out of credits, which the billing dispatcher could then handle.

### DIFF
--- a/go/billing/api.py
+++ b/go/billing/api.py
@@ -267,12 +267,13 @@ class TransactionResource(BaseResource):
         last_topup_balance = result.get('last_topup_balance')
         credit_balance = result.get('credit_balance')
 
+        def credit_percent():
+            return self._ceil_percent(credit_balance, last_topup_balance)
+
         # If the message is outbound and limit is reached, don't charge
         if (app_settings.ENABLE_LOW_CREDIT_CUTOFF and last_topup_balance and
                 message_direction == MESSAGE_DIRECTION_OUTBOUND):
-            credit_percent = int(math.ceil(
-                (credit_balance) * 100 / last_topup_balance))
-            if credit_percent < self._notification_mapping[0]:
+            if credit_percent() < self._notification_mapping[0]:
                 defer.returnValue({'credit_cutoff_reached': True})
 
         # Create a new transaction
@@ -374,9 +375,7 @@ class TransactionResource(BaseResource):
                 account_number)
 
         if app_settings.ENABLE_LOW_CREDIT_CUTOFF and last_topup_balance:
-            credit_percent = int(math.ceil(
-                credit_balance * 100 / last_topup_balance))
-            if credit_percent < self._notification_mapping[0]:
+            if credit_percent() < self._notification_mapping[0]:
                 transaction['credit_cutoff_reached'] = True
 
         defer.returnValue(transaction)
@@ -421,6 +420,9 @@ class TransactionResource(BaseResource):
             return None
         return self._notification_mapping[percentage - minimum]
 
+    def _ceil_percent(self, num, den):
+        return int(math.ceil(num * 100 / den))
+
     def check_all_low_credit_thresholds(
             self, credit_balance, credit_amount, last_topup_balance):
         """
@@ -442,7 +444,7 @@ class TransactionResource(BaseResource):
             return None
 
         def ceil_percent(n):
-            return int(math.ceil(n * 100 / last_topup_balance))
+            return self._ceil_percent(n, last_topup_balance)
 
         current_percentage = ceil_percent(credit_balance)
         current_notification_level = self._get_notification_level(
--- a/go/billing/api.py
+++ b/go/billing/api.py
@@ -267,13 +267,11 @@ class TransactionResource(BaseResource):
         last_topup_balance = result.get('last_topup_balance')
         credit_balance = result.get('credit_balance')
 
-        def credit_percent():
-            return self._ceil_percent(credit_balance, last_topup_balance)
-
         # If the message is outbound and limit is reached, don't charge
         if (app_settings.ENABLE_LOW_CREDIT_CUTOFF and last_topup_balance and
                 message_direction == MESSAGE_DIRECTION_OUTBOUND):
-            if credit_percent() < self._notification_mapping[0]:
+            if (self._ceil_percent(credit_balance, last_topup_balance) <
+                    self._notification_mapping[0]):
                 defer.returnValue({
                     'credit_cutoff_reached': True,
                     'transaction': None,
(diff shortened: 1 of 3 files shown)


rudigiesler:
Ready for review.

go/billing/tests/test_api.py
```diff
@@ -259,6 +259,82 @@ def test_low_credit_notification_zero_last_topup_value(self):
...
+            tag_name='tag1',
+            message_direction=MessageCost.DIRECTION_INBOUND,
+            session_created=False,
+            transaction_type=Transaction.TRANSACTION_TYPE_MESSAGE)
+
+        self.assertTrue(transaction2.get('credit_cutoff_reached', False))
+        self.assertEqual(Transaction.objects.count(), 3)
+
+        @inlineCallbacks
+        def test_credit_cutoff_outbound(self):
```
hodgestar: This test appears to be indented too far?

go/billing/tests/test_api.py
```diff
@@ -259,6 +259,82 @@ def test_low_credit_notification_zero_last_topup_value(self):
...
+        transaction2 = yield self.create_api_transaction(
+            account_number=self.account.account_number,
+            message_id='msg-id-1',
+            tag_pool_name='pool1',
+            tag_name='tag1',
+            message_direction=MessageCost.DIRECTION_INBOUND,
+            session_created=False,
+            transaction_type=Transaction.TRANSACTION_TYPE_MESSAGE)
+
+        self.assertTrue(transaction2.get('credit_cutoff_reached', False))
```
hodgestar: Don't we want a transaction that causes the `credit_cutoff_reached` to become `True`? Perhaps we should separate this test into two tests? One for a transaction with `credit_cutoff_reached` when it is `False` and one for when it is `True`?

rudigiesler: This test does test both, this is an `assertTrue` with the `assertFalse` above. This could be split out into two tests if you think that it's better.

hodgestar: Ah! I see.

I think two tests would be better, but I misread the `.assertTrue` as an
`.assertEqual` because of the `, False` at the end (but I see now it's part
of the `.get()`) call.

go/billing/api.py
```diff
@@ -247,6 +250,32 @@ def create_transaction_interaction(self, cursor, account_number,
...
+
+        params = {'account_number': account_number}
+        cursor = yield cursor.execute(query, params)
+        result = yield cursor.fetchone()
+
+        if result is None:
+            raise BillingError(
+                "Unable to find billing account %s while checking"
+                " credit balance. Message was %s to/from tag pool %s." % (
+                    account_number, message_direction, tag_pool_name))
```
hodgestar, line 265: Should we add a test case for this?

go/billing/api.py
```diff
@@ -247,6 +250,32 @@ def create_transaction_interaction(self, cursor, account_number,
...
+
+        last_topup_balance = result.get('last_topup_balance')
+        credit_balance = result.get('credit_balance')
+
+        def credit_percent():
+            return self._ceil_percent(credit_balance, last_topup_balance)
```
hodgestar: Give that we update the value of `credit_balance`, perhaps we should rather not wrap this in a closure any more and instead just call `self._ceil_percent` directly?

go/billing/api.py
rudigiesler, line 265: This was moved up from below, so there is already a test case for this, buried in the huge `test_transaction` test: https://github.com/praekelt/vumi-go/blob/feature/issue-1210-billing-api-to-block-accounts-with-insufficient-credits/go/billing/tests/test_api.py#L536

go/billing/api.py
```diff
@@ -247,6 +250,32 @@ def create_transaction_interaction(self, cursor, account_number,
...
+
+        # If the message is outbound and limit is reached, don't charge
+        if (app_settings.ENABLE_LOW_CREDIT_CUTOFF and last_topup_balance and
+                message_direction == MESSAGE_DIRECTION_OUTBOUND):
+            if credit_percent() < self._notification_mapping[0]:
+                defer.returnValue({'credit_cutoff_reached': True})
```
hodgestar: Perhaps we should also explicitly record whether a transaction was created in the result returned?

I was vaguely thinking of changing the result to something like:

```
{
    transaction: { ... },  /* or transaction: null if there isn't one */
    credit_cutoff_reached: true
}
```

That would require adjusting the dispatcher at the same time though. If we think it's worth it, perhaps it's something for another PR?

hodgestar:
Left a bunch of small comments but otherwise looks good.

go/billing/api.py
rudigiesler: At the moment, the billing dispatcher does nothing with the result returned from the billing API, so this should be an easy enough change to add.

hodgestar: Woot. Maybe let's make the change now then?

rudigiesler:
Changes made. Ready for review again.

go/billing/api.py
hodgestar, line 265: Cool.

go/billing/tests/test_api.py
```diff
@@ -259,6 +259,110 @@ def test_low_credit_notification_zero_last_topup_value(self):
...
+            'message_id': 'msg-id-1',
+            'tag_pool_name': 'pool1',
+            'tag_name': 'tag1',
+            'message_direction': MessageCost.DIRECTION_INBOUND,
+            'session_created': False,
+            'transaction_type': Transaction.TRANSACTION_TYPE_MESSAGE
+        }
+
+        yield self.create_api_transaction(**transaction_kwargs)
+        transaction = yield self.create_api_transaction(**transaction_kwargs)
```
hodgestar, line 310: Could we load fewer credits instead of making two transactions?

hodgestar, line 310: We should perhaps also assert that the transaction is returned in the result (since this is a different case to the other tests).

go/billing/tests/test_api.py
```diff
@@ -259,6 +259,110 @@ def test_low_credit_notification_zero_last_topup_value(self):
...
+            'message_id': 'msg-id-1',
+            'tag_pool_name': 'pool1',
+            'tag_name': 'tag1',
+            'message_direction': MessageCost.DIRECTION_OUTBOUND,
+            'session_created': False,
+            'transaction_type': Transaction.TRANSACTION_TYPE_MESSAGE
+        }
+
+        yield self.create_api_transaction(**transaction_kwargs)
+        transaction = yield self.create_api_transaction(**transaction_kwargs)
```
hodgestar, line 365: Same suggestions about loading fewer credits and checking the returned transaction here.

hodgestar:
Left one more suggestion for the tests. Otherwise looks good.

go/billing/tests/test_api.py
rudigiesler, line 310: This was actually something I was wondering about that I didn't ask. The way it currently works is, it requires a message to take the balance below the threshold, and then it blocks all future messages. Should this be changer to block all messages that would take the balance below the threshold?

hodgestar, line 310: I think it's okay the way it is -- we can just start the test with the balance below the threshold?

hodgestar:
:+1:

PR #1213, opened 2015-02-27
don't credit developer accounts negatively

rudigiesler:
Make sure that the `set_developer_account_balances` function does not allow reducing the amount of credits of the account.

### DIFF
--- a/go/billing/tasks.py
+++ b/go/billing/tasks.py
@@ -561,6 +561,7 @@ def set_account_balance(account_number, balance):
     """
     account = Account.objects.get(account_number=account_number)
     credit_amount = Decimal(str(balance)) - account.credit_balance
+    credit_amount = max(Decimal('0.0'), credit_amount)
     load_account_credits(account, credit_amount)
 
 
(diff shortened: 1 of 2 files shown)


rudigiesler:
Change made, still needs tests.

rudigiesler:
Test added. Ready for review.

go/billing/tasks.py
```diff
@@ -561,6 +561,7 @@ def set_account_balance(account_number, balance):
     """
     account = Account.objects.get(account_number=account_number)
     credit_amount = Decimal(str(balance)) - account.credit_balance
+    credit_amount = max(Decimal('0.0'), credit_amount)
```
hodgestar, line 564: Should we skip the call to `load_account_credits` entirely if the credit_amount is zero?

hodgestar:
Otherwise looks good.

go/billing/tasks.py
rudigiesler, line 564: I thought it might be useful to see that the task actually ran, and credited with `0.0`, than not knowing if it was because the account had enough credits, or if the task failed. Happy to change it though.

hodgestar, line 564: Hmm. Maybe let's ask @badlard what would make more sense to him?

hodgestar:
:+1: on this as-is.

If @badlard makes the opposite choice we can always make a tiny new PR.

PR #1214, opened 2015-03-04
billing dispatcher to handle credit cutoff

rudigiesler:
The billing dispatcher should handle the the parameter set in https://github.com/praekelt/vumi-go/pull/1210 appropriately.

### DIFF
--- a/go/vumitools/billing_worker.py
+++ b/go/vumitools/billing_worker.py
@@ -232,8 +232,9 @@ class BillingDispatcher(Dispatcher, GoWorkerMixin):
                 log.info(
                     "Not billing for inbound message: %r" % msg.to_json())
             else:
-                yield self.create_transaction_for_inbound(msg)
-                msg_mdh.set_paid()
+                result = yield self.create_transaction_for_inbound(msg)
+                if result.get('transaction'):
+                    msg_mdh.set_paid()
 
         except BillingError:
             log.warning(
@@ -261,9 +262,10 @@ class BillingDispatcher(Dispatcher, GoWorkerMixin):
                 log.info(
                     "Not billing for outbound message: %r" % msg.to_json())
             else:
-                transaction = yield self.create_transaction_for_outbound(msg)
-                msg_mdh.set_paid()
-                if transaction.get('credit_cutoff_reached', False):
+                result = yield self.create_transaction_for_outbound(msg)
+                if result.get('transaction'):
+                    msg_mdh.set_paid()
+                if result.get('credit_cutoff_reached', False):
                     msg = self._handle_credit_cutoff(msg)
         except BillingError:
             log.warning(
--- a/go/vumitools/billing_worker.py
+++ b/go/vumitools/billing_worker.py
@@ -9,11 +9,14 @@ from twisted.internet import reactor
 from vumi import log
 from vumi.dispatchers.endpoint_dispatchers import Dispatcher
 from vumi.config import ConfigText, ConfigFloat, ConfigBool
+from vumi.message import TransportUserMessage
 from vumi.utils import http_request_full
 
 from go.vumitools.app_worker import GoWorkerMixin, GoWorkerConfigMixin
 
 from go.billing.utils import JSONEncoder, JSONDecoder, BillingError
+
+SESSION_CLOSE = TransportUserMessage.SESSION_CLOSE
 
 
 class BillingApi(object):
@@ -95,6 +98,9 @@ class BillingDispatcherConfig(Dispatcher.CONFIG_CLASS, GoWorkerConfigMixin):
         "Name of the session metadata field to look for in each message to "
         "calculate session length",
         static=True, default='session_metadata')
+    credit_limit_message = ConfigText(
+        "The message to send when terminating session based transports.",
+        static=True, default='Vumi Go account has run out of credits.')
 
     def post_validate(self):
         if len(self.receive_inbound_connectors) != 1:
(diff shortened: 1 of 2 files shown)


rudigiesler:
Ready for review.

go/vumitools/billing_worker.py
```diff
@@ -251,8 +261,10 @@ def process_outbound(self, config, msg, connector_name):
                 log.info(
                     "Not billing for outbound message: %r" % msg.to_json())
             else:
-                yield self.create_transaction_for_outbound(msg)
+                transaction = yield self.create_transaction_for_outbound(msg)
                 msg_mdh.set_paid()
```
hodgestar: Should we only call `msg.set_paid()` if transaction is not `None`?

hodgestar: `transaction` is probably a bad name for the result because the transaction data is not only one part of the result?

go/vumitools/billing_worker.py
```diff
@@ -263,7 +275,17 @@ def process_outbound(self, config, msg, connector_name):
                 "Error processing outbound message, sending without billing:"
                 " %r" % (msg,))
             log.err()
-        yield self.publish_outbound(msg, self.receive_inbound_connector, None)
+        if msg is not None:
+            yield self.publish_outbound(
+                msg, self.receive_inbound_connector, None)
+
+    def _handle_credit_cutoff(self, msg):
+        if msg.get('session_event') is not None:
```
hodgestar: I think we should return `None` when `session_event` is `new` too (i.e. not allow people to start sessions).

hodgestar: I'm wondering whether we should not replace the content when the session event is already `close`? I.e. allow conversations to immediately close sessions with a custom message.

@badlard Thoughts?

go/vumitools/tests/test_billing_worker.py
```diff
@@ -739,3 +744,67 @@ def test_outbound_message_session_length_custom_field(self):
...
+
+        yield self.make_dispatch_outbound(
+            "outbound",
+            user_account="12345",
+            tag=("pool1", "1234"),
+            helper_metadata={},
+            session_event='new'
+            )
+
+        published_msg = self.ri_helper.get_dispatched_outbound()[0]
```
hodgestar: We usually write this as: `[published_msg] = self.ri_helper.get_dispatched_outbound()` so that we raise an error if there isn't exactly one message.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -739,3 +744,67 @@ def test_outbound_message_session_length_custom_field(self):
...
+            helper_metadata={}
+            )
+
+        published_msgs = self.ri_helper.get_dispatched_outbound()
+        self.assertEqual(len(published_msgs), 0)
+
+    @inlineCallbacks
+    def test_inbound_message_credit_cutoff_session(self):
+        self.billing_api = BillingApiMock(credit_cutoff=True)
+        yield self.get_dispatcher
```
hodgestar: We seem to be missing `()` here? Bit confused about why this doesn't fail.

hodgestar:
Left a few comments but this turned out a lot cleaner than I feared it might. Woot!

go/vumitools/billing_worker.py
hodgestar: Not replacing the content when the session event is `close` does potentially allow someone using the HTTP API to set the session event to `close` on an SMS and thus have it sent anyway. Perhaps that's a big enough issue to decide this in favour of always replacing the content?

hodgestar: +1 from @badlard on always modifying the content so please ignore the `close` event part of this suggestion. :)

go/vumitools/tests/test_billing_worker.py
rudigiesler: Maybe `get_dispatcher()` isn't needed? I only put it in here because it was in all the other tests. I don't actually use the dispatcher in this test.

hodgestar: I think you might be asserting on the wrong message (i.e. you might be asserting that the message you sent into the worker exists, rather than the message the worker forwarding existing).

go/vumitools/tests/test_billing_worker.py
```diff
@@ -739,3 +746,83 @@ def test_outbound_message_session_length_custom_field(self):
...
+
+        msg = yield self.make_dispatch_inbound(
+            "inbound",
+            user_account="12345",
+            tag=("pool1", "1234"),
+            helper_metadata={},
+            session_event='new'
+            )
+
+        published_msg = self.ri_helper.get_dispatched_inbound()[0]
```
hodgestar: This should be `self.ro_helper.get...` (and the same for the other inbound test)`

rudigiesler:
Ready for re-review

hodgestar:
I've realised that we should send back `nacks` for messages that aren't sent because of the cutoff. Maybe that should happen in a separate PR though.

go/vumitools/tests/test_billing_worker.py
```diff
@@ -739,3 +746,85 @@ def test_outbound_message_session_length_custom_field(self):
...
+            "outbound",
+            user_account="12345",
+            tag=("pool1", "1234"),
+            helper_metadata={},
+            session_event='new'
+            )
+
+        self.assertEqual(len(self.ri_helper.get_dispatched_outbound()), 0)
+
+
```
hodgestar: Too many blank lines?

hodgestar:
:+1: once the extra blank line is removed.

Let's leave the `nack` for a separate PR. I'm happy with us deploying this today and leaving activating it for Monday if everyone else is.

rudigiesler:
https://github.com/praekelt/vumi-go/issues/1219 created for sending back `nacks`.

hodgestar:
:+1:

PR #1217, opened 2015-03-04
Fix answers sent by HTTP from dialogues

hodgestar:
When the dialogues send answers from HTTP they use the raw state IDs as the keys for the answer dictionary. We need to translate those to state names.

### DIFF
--- a/go/apps/dialogue/tests/test_vumi_app.js
+++ b/go/apps/dialogue/tests/test_vumi_app.js
@@ -322,70 +322,111 @@ describe("app", function() {
         });
 
         describe("when the user enters a send state which is a httpjson type", function() {
-            it("should successfully fetch the data from the url if the method is GET", function() {
+            beforeEach(function() {
                 return tester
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

Would also appreciate feedback on whether the idea implemented is sane.

dbreedt:
I'm happy with this change.

What will happen in the case where a user enters a session of 3 questions while answering the last questions a vumi user deactives the dialogue modifies it by removing one of the questions and activates it before the user submits the answer?

hodgestar:
@dbreedt If they delete the block, the UUID will no longer be present. I should update this to handle that.

hodgestar:
@justinvdm Would you mind looking at this when you have a chance?

hodgestar:
Support for ignoring answers to deleted states added. Ready for review.

justinvdm:
:+1:

justinvdm:
(when travis is happy)

PR #1218, opened 2015-03-06
different email for last low credit notification

rudigiesler:
The email that is sent for the lowest notification should be different than the rest, as the last notification is the credit cutoff, and no more messages can be sent after this. The email should reflect this.

### DIFF
--- a/go/billing/api.py
+++ b/go/billing/api.py
@@ -404,9 +404,10 @@ class TransactionResource(BaseResource):
         level = self.check_all_low_credit_thresholds(
             credit_balance, credit_amount, last_topup_balance)
         if level is not None:
+            cutoff_notification = level * 100 == self._notification_mapping[0]
             return spawn_celery_task_via_thread(
                 create_low_credit_notification, account_number,
-                level, credit_balance, level*100 == self._notification_mapping[0])
+                level, credit_balance, cutoff_notification)
 
     def _get_notification_level(self, percentage):
         """
--- a/go/billing/api.py
+++ b/go/billing/api.py
@@ -406,7 +406,7 @@ class TransactionResource(BaseResource):
         if level is not None:
             return spawn_celery_task_via_thread(
                 create_low_credit_notification, account_number,
-                level, credit_balance, level == self._notification_mapping[0])
+                level, credit_balance, level*100 == self._notification_mapping[0])
 
     def _get_notification_level(self, percentage):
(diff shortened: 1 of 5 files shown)


rudigiesler:
Ready for review.

go/billing/api.py
```diff
@@ -406,7 +406,7 @@ def check_and_notify_low_credit_threshold(
         if level is not None:
             return spawn_celery_task_via_thread(
                 create_low_credit_notification, account_number,
-                level, credit_balance)
+                level, credit_balance, level*100 == self._notification_mapping[0])
```
hodgestar: I think it would be good to do `cutoff_notification = (level * 100 == self._notification_mapping[0])` just to make it clear what the value being passed in is for.

hodgestar:
One minor comment, otherwise looks good to me.

rudigiesler:
Changes made. Ready for re-review.

go/billing/templates/billing/credit_cutoff_notification_email.txt
```diff
@@ -0,0 +1,13 @@
+
+Hi {{user.get_full_name}}!
+
+Please note that your Vumi Go account ({{account.user.email}}) has reached {{threshold_percent}}% used of the total available credits. Your remaining balance is {{credit_balance}} credits. This means that you will no longer be able to send any more messages, and all session based transport requests will be terminated with a message indicating that the account has run out of credits. All incoming messages will be received as normal, and you will still be charged for them.
```
SteveBarnett: How about something like this? Bit shorter, more spacier, less techy.

> Please note that your Vumi Go account ({{account.user.email}}) has reached {{threshold_percent}}% used of the total available credits. Your remaining balance is {{credit_balance}} credits.
> 
> You will no longer be able to send messages. Contacts on campaigns over USSD or Voice will receive a notification saying that your account has run out of credits. All incoming messages will be received as normal.

hodgestar:
:+1: from me. I guess we should wait to see if @SteveBarnett has any comments though.

rudigiesler:
@SteveBarnett I've implemented your suggestions. Have a look and see if you're happy with it.

SteveBarnett:
@rudigiesler Fixed my typos (sorry!). :+1:

hodgestar:
:+1: once Travis is happy.

PR #1223, opened 2015-03-09
HTTP API conversation should reject send requests without a to_addr.

hodgestar:


hodgestar:
Ready for review.

go/apps/http_api_nostream/tests/test_resource.py
```diff
@@ -81,39 +84,88 @@ def test_many_errors(self):
         self.assert_no_attribute(opts, "odd")
 
 
+class TestMsgCheckHelpers(TestCase):
+    def test_is_unicode(self):
+        self.assertTrue(MsgCheckHelpers.is_unicode(u"123"))
+        self.assertFalse(MsgCheckHelpers.is_unicode(None))
+        self.assertFalse(MsgCheckHelpers.is_unicode("123"))
```
jerith, line 91: `self.assertEqual(..., True)` is probably better here, because we are explicitly looking at the return value. (Also, we get much nicer failure messages.) `assertTrue()` and friends are harder to get decent output from, but are useful when an equality test isn't possible.

jerith, line 91: I see all the existing tests use `assertTrue()` and `assertFalse()`. If we do make the switch, we should probably just update the methods we're touching anyway.

jerith:
Looks good. I have a mild preference for the stylistic change I suggested above, but I'll leave the decision to you.

hodgestar:
I think I'm happy with `.assertTrue` / `.assertFalse` because when I'm reading them I immediately see which ones are expected to be valid and which invalid (rather than having to get to the end of the line). I agree generally about `.assertEqual` being nicer, but I'm not sure there's a big difference when the value on is checking is `True` or `False`.

jerith:
:+1:

hodgestar:
Thanks! Will land tomorrow.

PR #1221, opened 2015-03-09
Billing API server needs to get listening port from command line

jerith:
If we want to run multiple billing API servers behind a load balancer, we need to be able to specify the port they listen on from the command line rather than a settings file.


jerith:
The basic functionality's there (along with some bonus improvements to connection handling), but there's still a bunch of cleanup to do before it's ready to land.

go/billing/api.py
```diff
@@ -478,12 +482,25 @@ def create_transaction(self, account_number, message_id, tag_pool_name,
...
+        return NOT_DONE_YET
+
+
 class Root(BaseResource):
     """The root resource"""
 
     def __init__(self, connection_pool):
         BaseResource.__init__(self, connection_pool)
         self.putChild('transactions', TransactionResource(connection_pool))
+        self.putChild('ping', HealthResource(self.health_check))
```
hodgestar: I have no personal objections to `ping`, just mentioning that we've usually used `health` as the URL path.

go/billing/api.py
```diff
@@ -493,3 +510,32 @@ def getChild(self, name, request):
...
+        """
+        try:
+            yield self._connection_pool.runQuery("SELECT 1")
+        except (ConnectionDead, psycopg2.InterfaceError):
+            request.setResponseCode(503)
+            request.write("Database connection unavailable\n")
+        else:
+            request.setResponseCode(200)
+            request.write("OK\n")
+        request.finish()
```
hodgestar: I'd be happier if the request writing happened on the HealthResource (since it's meant to be handling the request) and this was a function that returned, e.g. `(200, "OK")` or `(503, "Database connection unavailable")`.

go/billing/api.py
```diff
@@ -493,3 +510,32 @@ def getChild(self, name, request):
...
+    """
+    Create and return a go.billing.api.Root resource for use with twistd.
+    """
+    connection_string = app_settings.get_connection_string()
+    connection_pool = DictRowConnectionPool(
+        None, connection_string, min=app_settings.API_MIN_CONNECTIONS)
+    resource = Root(connection_pool)
+    # Tests need to know when we're connected, so stash the deferred on the
+    # resource for them to look at.
+    resource._connection_pool_started = connection_pool.start()
```
hodgestar, line 544: It feels like the reason we're having to stash `_connection_pool_started` is because this method is really doing two completely different things (creating and starting a connection pool and creating an api Root resource). Perhaps it should only be doing one thing and we should have a separate method for creating the connection pool?

go/billing/tests/test_api.py
```diff
@@ -105,7 +164,8 @@ def create_api_transaction(self, **kwargs):
             'session_length': None,
         }
         content.update(kwargs)
-        return self.call_api('post', 'transactions', content=content)
+        self.called_api = True
```
hodgestar: Not following what `self.called_api = True` is for?

go/billing/api.py
hodgestar, line 544: I see this needs to be a single function that returns a resource for Twistd.

jerith, line 544: Yes. `twistd` needs a callable that takes no parameters and returns a `Resource` object.

go/billing/tests/test_api.py
jerith: Leftover debugging cruft, it can go. Thanks. :-)

go/billing/api.py
jerith: I put the request writing there to avoid another layer of indirection, but it doesn't buy much and it does give the request object to things that shouldn't really have it. Thanks.

hodgestar:
:+1:

PR #1225, opened 2015-03-17
Avoid new django-celery-email for now

jerith:
Version 1.1.0 of `django-celery-email` changes the parameters its `send_email` task takes, which breaks our billing email stuff: See https://github.com/pmclanahan/django-celery-email/pull/13 for more details.

Until that's resolved and we can either use a new version as-is or handle the difference ourselves somehow, we should stick to version 1.0.4 of `django-celery-email`.


hodgestar:
:+1:

PR #1226, opened 2015-03-17
Add support for setting whitelisted helper_metadata on outbound jsbox messages. 

hodgestar:
Applications that use voice transport potentially need to supply additional metadata describing, e.g. where to obtain audio files to send or what characters should terminate IVR input. It seems sensible to add slightly more generic support for setting helper_metadata to the jsbox outbound message resource.

Because some helper_metadata fields contain information that shouldn't be modifiable from the sandbox (e.g. which Vumi Go account and conversation a message is associated with) the helper_metadata fields that can be set need to be white listed.

### DIFF
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -82,9 +82,14 @@ class GoOutboundResource(SandboxResource):
     def _check_helper_metadata(self, helper_metadata):
         if any(key not in self._allowed_helper_metadata
                for key in helper_metadata.iterkeys()):
-            raise InvalidHelperMetadata(
-                "'helper_metadata' may only contain the following keys: %s" %
-                ', '.join(sorted(self._allowed_helper_metadata)))
+            if self._allowed_helper_metadata:
+                errmsg = (
+                    "'helper_metadata' may only contain the following keys: %s"
+                    % ', '.join(sorted(self._allowed_helper_metadata)))
+            else:
+                errmsg = (
+                    "'helper_metadata' is not allowed")
+            raise InvalidHelperMetadata(errmsg)
 
     def _handle_reply(self, api, command, reply_func):
         if not 'content' in command:
--- a/go/apps/jsbox/outbound.py
+++ b/go/apps/jsbox/outbound.py
@@ -79,17 +79,27 @@ class GoOutboundResource(SandboxResource):
     def _mkfaild(self, command, reason):
         return succeed(self._mkfail(command, reason))
 
-    def _check_helper_metadata(self, helper_metadata):
+    def _get_helper_metadata(self, command):
+        """
+        Get a legal helper_metadata dict from `command` or raise an
+        InvalidHelperMetadata exception with a suitable message.
+        """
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

I intentionally didn't update `send_to_tag` because that's deprecated. Perhaps I should explicitly add a TODO for that?

go/apps/jsbox/outbound.py
```diff
@@ -49,19 +49,48 @@ def is_inbound_push_trigger(msg):
...
+        if any(key not in self._allowed_helper_metadata
+               for key in helper_metadata.iterkeys()):
+            if self._allowed_helper_metadata:
+                errmsg = (
+                    "'helper_metadata' may only contain the following keys: %s"
+                    % ', '.join(sorted(self._allowed_helper_metadata)))
+            else:
+                errmsg = (
+                    "'helper_metadata' is not allowed")
+            raise InvalidHelperMetadata(errmsg)
```
jerith: I found this logic flow a little tricky. Would it make sense to rewrite it along the following lines?

``` python
if not helper_metadata:
    return  # Nothing to check
if not self._allowed_helper_metadata:
    raise InvalidHelperMetadata("...")
if any(...):
    raise InvalidHelperMetadata("...")
```

go/apps/jsbox/outbound.py
```diff
@@ -246,6 +290,14 @@ def handle_send_to_endpoint(self, api, command):
...
             return self._mkfaild(
                 command, reason=u"'to_addr' must be given in sends.")
+        if not isinstance(command.get('helper_metadata'), (dict, type(None))):
+            return self._mkfaild(
+                command, reason=u"'helper_metadata' must be object or null.")
+        cmd_helper_metadata = command.get('helper_metadata', {})
+        try:
+            self._check_helper_metadata(cmd_helper_metadata)
+        except InvalidHelperMetadata as err:
+            return self._mkfaild(command, reason=unicode(err))
```
jerith, line 302: If we rewrite `_check_helper_metadata(helper_metadata)` as `_get_helper_metadata(command)` or something, we could avoid some of this duplication. Is that a thing worth doing?

jerith:
Looks good, although I have a couple of (related) comments about possible minor improvements. I'm happy with this as-is if the builds are green, but I think it could be slightly cleaner.

go/apps/jsbox/outbound.py
hodgestar: Happy for you to restructure this if you'd like to.

go/apps/jsbox/outbound.py
hodgestar, line 302: Not sure what the _get_helper_metadata(command) idea would look like -- I
quite like the explicit step-by-step parameter checking and I'm not sure
how to roll that up nicely because of the need to constantly return errors.
I can imagine some nice helper objects but that seems like a whole separate
PR to use them consistently for all parameter checks and maybe too much
indirection.

hodgestar:
@jerith :+1: on your commits. Merging this branch.

PR #1230, opened 2015-03-26
Fix tests for unique address counts

jerith:
Now that the message store supports unique address counts again, we need to fix the tests that assert on it.


justinvdm:
Thanks for tackling this, :+1: when travis is happy.

PR #1231, opened 2015-03-26
Remove broken message search

jerith:
This has been broken for ages, but is still accessible in the UI and just causes sadness when anyone tries to use it. (See #865 for the (apparently abandoned) fix.)

We should remove it from the UI to stop people trying to use it.


justinvdm:
:+1: when travis is happy

Issue #1229, opened 2015-03-23
The task export_all_contacts collects all contacts in memory.


hodgestar:
For accounts with lots of contacts, this will run out of RAM and fail. We need to ensure it does not collect all contacts in memory and collects them on disk instead.

Alternatively, we need to add a streaming contacts export.

jerith:
There are a variety of problems with doing this sort of export by email, especially with large data volumes (which is the cause of the memory usage issue here). We should probably expose the streaming HTTP version and kill the email entirely.

smn:
I'm a bit -1 on streaming to end users. Even for developers it's a major undertaking to download the kinds of JSON/CSV files we produce in a streaming fashion, Internet dies, computers go to sleep, batteries run out etc... And worst of all, we don't support resuming streaming from where things left off.

I'd be +1 on a streaming API that is used internally to stream to some temporary location and then use nginx to serve it as a static & resumable download after having emailed the download link to the end user.

rudigiesler:
+1 on smn's idea. A paginated response does solve all the problems of resuming, but does require some sort of client.

PR #1236, opened 2015-04-07
Limit contact export size in celery task

jerith:
This is a short-term measure to avoid letting excessively large contact exports break other things. See #1229 for discussion around a longer-term solution.

### DIFF
--- a/go/contacts/tasks.py
+++ b/go/contacts/tasks.py
@@ -149,6 +149,16 @@ def export_contacts(account_key, contact_keys, include_extra=True):
 
     api = VumiUserApi.from_config_sync(account_key, settings.VUMI_API_CONFIG)
     contact_store = api.contact_store
+    all_key_count = len(contact_keys)
+
+    message_content_template = 'Please find the CSV data for %s contact(s)'
+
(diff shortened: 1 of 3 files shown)


go/contacts/tasks.py
```diff
@@ -149,6 +149,16 @@ def export_contacts(account_key, contact_keys, include_extra=True):
...
     contact_store = api.contact_store
+    all_key_count = len(contact_keys)
+
+    message_content_template = 'Please find the CSV data for %s contact(s)'
+
+    if all_key_count > settings.CONTACT_EXPORT_TASK_LIMIT:
+        contact_keys = contact_keys[:settings.CONTACT_EXPORT_TASK_LIMIT]
+        message_content_template = '\n'.join([
+            'NOTE: There are too many contacts to export.',
+            'Please find the CSV data for %%s (out of %s) contacts.' % (
```
rudigiesler, line 160: Won't this result in `Please find the CSV data for %s (out of xxxxxxx) contacts.`?

rudigiesler, line 160: Ah, I see what you're doing. Nevermind

rudigiesler:
:+1: from me

PR #1237, opened 2015-04-08
Feature/issue 1235 busy routers

SteveBarnett:


go/base/static/js/test/routing/views.test.js
```diff
@@ -118,53 +118,111 @@ describe("go.routing (views)", function() {
...
+
+    describe(".findMaxEndpoints", function() {
+      it("should return the max of two endpoint arrays", function() {
+        assert.equal(state.findMaxEndpoints(), 4);
+      });
+    });
+
+    describe(".tooManyEndpoints", function() {
+      it("should check if there are too many endpoints", function() {
+        assert.equal(state.tooManyEndpoints(), true);
```
justinvdm: `<bikeshed>`Could we use `assert()` here since its a boolean check?`</bikeshed>`

justinvdm:
Looks great :) :+1: when Travis is happy. I haven't tested out how this looks locally though, maybe we should double check that it looks right before landing?

SteveBarnett:
@justinvdm Sweet, thanks. I have done a local / visual check, and it looks good.

PR #1238, opened 2015-04-09
Billing transaction archiving task uses lots of memory

jerith:
The task fetches all transactions from the database before it processes them, which is a problem when there are lots of transactions.


hodgestar:
:+1: once Travis is happy.

Aside: I've very slightly concerned about the cost of ordering, but it can't be worse than holding everything in memory.

PR #1243, opened 2015-04-14
Fix jsPlumb connection detaching

justinvdm:
At the moment, destination endpoint connection detaches are broken. We've been using our own slightly modified jsPlumb fork with a workaround for this, but the workaround appears to not always work. After a bit of tinkering, it seems that jsPlumb is unhappy if `jsPlumb.detach()` is called on an already detached endpoint.  Fixing that is simple enough -- just infer whether the connection has already been detached before trying to detach it.

### DIFF
--- a/go/base/static/js/src/components/plumbing/connections.js
+++ b/go/base/static/js/src/components/plumbing/connections.js
@@ -55,13 +55,11 @@
     },
 
     destroy: function() {
-      var plumbConnection = this.plumbConnection;
-
-      if (plumbConnection) {
-        this.plumbConnection = null;
-        jsPlumb.detach(plumbConnection);
+      if (this.plumbConnection && this.plumbConnection.endpoints) {
+        jsPlumb.detach(this.plumbConnection);
       }
 
+      this.plumbConnection = null;
       return this;
     },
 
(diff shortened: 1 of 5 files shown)


justinvdm:
@SteveBarnett ready for review, a pretty small fix that we've needed for a very long time :)

SteveBarnett:
\o/, :+1: , once Travis is happy.

justinvdm:
@SteveBarnett thanks for the review, it seems this fix doesn't work jsPlumb 1.7.5 (the version I'm planning on upgrading us to), going to keep tinkering.

justinvdm:
Ah, so it seems like a race condition or something -- when setting breakpoints, `.plumbConnection.endpoints` are `null` by the time we check it in `.destroy()`. When breakpoints aren't set, `.plumbConnection.endpoints` aren't `null`. I'm going to go with a solution where we tell `.destroy()` whether the destroy is because of a connection detach done by the user, in which case it should not detach.

Related -- `jsPlumb.detach()` seems to cause endless recursion if we don't set `fireEvent` to `false`, though the docs say `false` is the default. Will make a comment.

justinvdm:
@SteveBarnett actually ready for review now :)

hodgestar:
:+1: from my side.

SteveBarnett:
:+1: from me too.

PR #1246, opened 2015-04-15
Add character count to dialogues

SteveBarnett:


SteveBarnett:
@justinvdm We should talk about how to limit this by channel?

justinvdm:
@SteveBarnett We can if you think it would help. I thought the first pass was going to just have a hardcoded char count though.

SteveBarnett:
@justinvdm Indeed. I was looking at this yesterday, though, and wondering about channels other than SMS and USSD: should we be displaying the character count there? Thoughts?

SteveBarnett:
Ready for review (@justinvdm ?). Would welcome some advice / comments on how to do this better, more gooder places to put stuff :).

SteveBarnett:
Re-ready for re-review, @justinvdm .

go/base/static/js/src/apps/dialogue/states/states.js
```diff
@@ -123,7 +125,11 @@
     detach: function() {
       this.$el.detach();
       return this;
-    }
+    },
+
+    // maxChars: function() {
+    //   return 140;
+    // }
```
justinvdm: Forgotten comment?

go/base/static/js/src/apps/dialogue/states/states.js
```diff
@@ -18,6 +18,8 @@
       ParametricEndpointView = endpoints.ParametricEndpointView,
       AligningEndpointCollection = endpoints.AligningEndpointCollection;
 
+  var maxChars = 140;
```
justinvdm, line 21: Are we sure we want to use `140` instead of `160`? `140` seems safer, since operator-that-will-not-be-named sometimes restricts USSD session messages to 140 instead of 160, and since other channels like twitter have 140 as a restriction. That said, 160 is quite common for USSD, in SA at least, not too sure on other countries. I just get the feeling most dialogues will have states exceeding 140 chars because of this, and the feature loses its usefulness.

go/base/static/js/test/apps/dialogue/states/choice.test.js
```diff
@@ -212,5 +212,58 @@ describe("go.apps.dialogue.states.choice", function() {
           editMode.$('[data-uuid="choice:new-endpoint"]')));
       });
     });
+
+    describe(".render", function() {
+      it("should display the char count", function() {
+        assert.equal(state.$('.char-count').text().trim(), '96 left of 140');
```
justinvdm: I think we should set the model's text, then call `.render()`, then assert that we rendering the right char count. The current test is probably giving us a true positive, but it will be easier for us to verify that we aren't getting a false positive if we explicitly set the model text and explicitly call `.render()`.

justinvdm:
I think[ this comment](https://github.com/praekelt/vumi-go/pull/1246/files#r28774884) applies to all the tests added in the PR. Sorry if its a little bikesheddy, but I think we are going to be at risk of false positives otherwise.

go/base/static/templates/apps/dialogue/states/choice/edit.jst
```diff
@@ -1,5 +1,6 @@
 <div class="form-group">
   <label class="control-label">Define the question</label>
+  <textarea class="text form-control"><%= model.get('text') %></textarea>
```
justinvdm, line 3: Could we do this for the end and freetext states too (and remove their `TextEditView` partial)?

justinvdm:
Choice states without choices seem to be showing '141 of 140': 2015-04-21-152225_1440x900_scrot

justinvdm:
Left a few comments, but looks good :)

go/base/static/js/src/apps/dialogue/states/states.js
SteveBarnett: Good spot. Removed.

justinvdm:
:+1:!

PR #1247, opened 2015-04-15
Handle non-detach connection detach events

justinvdm:
When a new connection created is in the UI by dragging a source endpoint, then dropped before it reaches a target endpoint, jsPlumb fires a connection detach event. We need to our connection detach event listeners to ignore this event to avoid us handling it as an ordinary detach.


justinvdm:

SteveBarnett:
:+1:

PR #1248, opened 2015-04-15
Better visual feedback for dialogue and routing screen

justinvdm:


justinvdm:
@SteveBarnett ready for review, will probably be easier to show you how it looks in person than to take screenshots.

justinvdm:
@SteveBarnett actually ready for review. The cursor changes we discussed are more complicated that I thought. Getting the cursor to change to `grab` when it is over a draggable is simple enough, getting it to change to `grabbing` during dragging is a bit more complicated. Planning on doing that in separate PR if we have time, but for now I'm using `move` in place of `grab` and `grabbing`.

SteveBarnett:
\o/, :+1:

PR #1249, opened 2015-04-16
Update routers to new button style

SteveBarnett:


SteveBarnett:
Ready for review.

hodgestar:
:+1:

jerith:
The builds are probably failing because of tests that make assertions on the old buttons.

SteveBarnett:
Ready for re-review. Happy to take suggestions for how to make the tests (like [this one](https://github.com/praekelt/vumi-go/commit/e68acc37eb83a979f7ddf83d0bbb098cdf2eb34c#diff-5d80514ad5ab883510ab83ba02b092f5R33)) look nicer!

go/routers/app_multiplexer/tests/test_views.py
```diff
@@ -30,7 +30,11 @@ def test_show_stopped(self):
         router = response.context[0].get('router')
         self.assertEqual(router.name, u"myrouter")
         self.assertContains(response, rtr_helper.get_view_url('start'))
-        self.assertNotContains(response, rtr_helper.get_view_url('stop'))
+        self.assertContains(
+            response,
+            '<button class="btn btn-danger action" data-action="stop" ' +
+            ' data-url="%s" disabled>Deactivate</button>'
+            % rtr_helper.get_view_url('stop'), html=True)
```
justinvdm, line 37: I think we usually just will do something like `self.assertContains(response, rtr_helper.get_view_url('stop'))` and `self.assertContains(response, '>Deactivate<')`, but I can't think of a reason for why testing these other things is necessarily bad. `disabled` is important here, so we definitely need that. Does seem a bit weird to be testing `button`, `class` and `data-action` though, since we are testing things hardcoded in the template, unless we want tests to fail if any of those change. Is that the case?

justinvdm:
Left a minor comment that applies to most of the test changes, but looks good :)

go/routers/app_multiplexer/tests/test_views.py
jerith, line 37: We should avoid assertions on things that can change for reasons unrelated to the functionality we're testing. That said, I don't really see a better way to do this. :-/

SteveBarnett, line 37: @jerith @justinvdm I'm not very happy with this either.

The old test said "the URL for the deactivate action shouldn't be in the response."
The new test is supposed to say something like "the deactivate btn (or URL) must be there, but must have the `disabled` attribute." Maybe this test should live with the karma stuff?

jerith, line 37: I don't think we can actually do much better without adding a whole lot of extra complexity that will probably be worse than the current fragility. Let's just go with this and worry about it next time we change this stuff.

justinvdm, line 37: Agreed

justinvdm:
:+1:

PR #1251, opened 2015-04-17
Feature/add 504 page

SteveBarnett:


SteveBarnett:
At the moment people see a standard server 504 page. This is not particularly helpful.
This PR adds a Vumi-branded 504 page so that users at least know that the problem is with Vumi, not with their internet connection, or something else.

---

Ready for review.

justinvdm:
Content looks good.

I'm not sure if we will be able to serve custom 504 pages with django, I think we are going to need to rely on nginx's [`error_page`](http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page) here, since that's the thing serving the 504 page (nginx fails to get a response in time from our web app, so it serves that page). That means we will need to have it be a static html page and we won't be able to make use of django's templating. I could be misunderstanding things though.

That might mean that this page should actually not be in this repo, but lets discuss this in RL on Monday.

SteveBarnett:
Discussed with @justinvdm: make a PR on the puppet repo instead, into `modules/vumigo_prod/files`, with static 504.html and tiny amount of CSS.

SteveBarnett:
Closing, in favour of https://github.com/praekelt/puppet/pull/623

PR #1253, opened 2015-04-22
Change to a freeform layout for dialogues

justinvdm:


justinvdm:

go/base/static/css/conversations.less
```diff
@@ -43,14 +46,25 @@
     }
   }
 
+  .info-add-state {
+    z-index: 3;
+    position: absolute;
+    width: 380px;
+    margin-left: -190px;
+    left: 50%;
```
justinvdm: Would appreciate input on a better way to do this.

go/base/static/js/src/apps/dialogue/layout.js
```diff
@@ -0,0 +1,143 @@
...
+    },
+
+    colOffset: function(state) {
+      var center = (this.cellWidth(state) - state.$el.outerWidth(true)) / 2;
+      return center + 35;
+    },
+
+    rowMargin: function() {
+      return 30;
+    },
```
justinvdm, line 40: The layout was initially using margins specified in css for figuring out cell widths, heights and offsets, but that actually makes things more complicated -- in order to retain the 3 column layout for existing dialogues, we'd need kind-of big margins, which complicates dragging states around, since they now have pretty big bounding boxes when colliding with the edges of the diagram, which makes moving states around annoying.

go/base/static/css/conversations.less
SteveBarnett: How about this instead?

``` css
.info-add-state {
  z-index: 3;
  position: relative;
  width: 30%;
  margin-left: auto;
  margin-right: auto;
}
```

go/base/static/css/conversations.less
```diff
@@ -1,3 +1,6 @@
+@import "./lesshat";
```
SteveBarnett, line 1: Not sure about this. We just use it for [.user-select(none);](.user-select%28none%29;), which just stops people from selecting the diagram area, right? I'm not sure how it helps.

justinvdm, line 1: It helps by making my editor stop complaining about unrecognised mixins :) is it pulling in the stylesheet more than once as a result?

go/base/static/css/conversations.less
justinvdm: \o/ slight modification: using `margin: 0 auto` instead to avoid the diagram jumping a bit when the alert is dismissed.

Dōmo arigatō sensei

go/base/static/js/test/apps/dialogue/testHelpers.js
```diff
@@ -6,96 +6,118 @@
...
-      entry_endpoint: {uuid: 'endpoint8'},
-      exit_endpoint: {uuid: 'endpoint9'}
-    }],
-    connections: [{
-     uuid: 'endpoint1-endpoint3',
-     source: {uuid: 'endpoint1'},
-     target: {uuid: 'endpoint3'}
-    }]
+   var modelData = function() {
+     return {
```
SteveBarnett, line 10: Can you explain why we're switching to a function here, please? Thanks!

justinvdm, line 10: Sure :)

Each time we were making changes to the models, we'd have irrelevant test failures, since the tests were basically expecting the models to look like some hardcoded data. To prevent these irrelevant test failures, I instead changed to comparing with a _new_ copy of the model data in those tests, so that those tests can only care about the changes relevant to the test. This was related to this PR because of the model changes we made in this PR, though thinking about it now, it probably should've been a separate PR :/

So, that's the reasoning behind trying to get a _new_ copy of the data. Having a function means we can ensure a new copy of the model data just by calling the function, as opposed to taking an existing data structure and trying to deep copy it. Deep copying in js is complicated, error prone, and not available in underscorejs ([_because_ it is complicated and error prone](https://github.com/jashkenas/underscore/issues/162)).

go/base/static/css/conversations.less
SteveBarnett, line 1: Ha! Okay, all good, then.

SteveBarnett:
@justinvdm This is awesome! \o/.

As discussed, let's do a final UX-y test in staging on some of the bigger dialogues, then we'll be good to go.

SteveBarnett:
Looks great. Awesome work! :+1:

PR #1252, opened 2015-04-23
Change the behaviour of the save button on dialogues

justinvdm:
At the moment, hitting the save button on a dialogue takes you back to the view conversation screen. We should:
1. Change that behaviour to be "save and stay on this page."
2. Add a “save and exit” (or something) button that does what the save button does now.


justinvdm:
@SteveBarnett ready for review

SteveBarnett:
Bueno! :+1:

PR #1254, opened 2015-04-23
Activate conversations when they are created

justinvdm:
It's pretty clear that this is the behaviour people are expecting, instead of needing to manually activate conversations.


hodgestar:
Agreed. Also conversations don't do anything until they're hooked up on the routing screen so there's little harm in activating them after they're created.

justinvdm:
@SteveBarnett &&|| a vumi person -- ready for review.

hodgestar:
:+1:

PR #1256, opened 2015-04-23
Feature/improve paths

SteveBarnett:
When there are lots of connections between dialogues, it can be hard to see what's going on. We should try and make the different connections more visually clear.


SteveBarnett:
Adding colours and dashes to the paths helps, but: it's still a bit of a mess; it's not really solving the problem. The big problem is **being able to follow lines to the next screen / state**.

Some thoughts on making that better: let people click on a path to highlight it (perhaps via CSS applied to a `:focus` state). Ideas:
- Highlight the path with colour, z-index, dashes.
- Dim out other paths.

@justinvdm I think I should close this PR and start a new one (at some point) for ^^^. Thoughts?

screenshot 2015-04-24 10 22 51

justinvdm:
@SteveBarnett +1 on new PR for highlighting focused paths and dimming the rest. How would focusing work though? Do you mean highlighting connections of the focused state?

SteveBarnett:
@justinvdm re focusing: ¯_(ツ)_/¯. We shall do some user testing :).

PR #1257, opened 2015-04-23
Upgrade to a newer jsPlumb

justinvdm:


justinvdm:

justinvdm:
(sorry about the overly large diff -- the PR isn't that big, its just the jsPlumb files skewing the diff size)

go/base/static/js/src/routing/views.js
```diff
@@ -171,21 +171,49 @@
...
 
-    repaint: function() { jsPlumb.repaintEverything(); },
+    onDrag: function() {
+      this.repaint();
+    },
 
-    render: function() {
-      this.states.each(function(s) { s.render(); });
+    repaintEl: function($el) {
+      // jsPlumb 1.7.5 doesn't seem to play nice with jquery ui's .sortable(),
```
SteveBarnett: @justinvdm Can we talk about this bit, please? I don't quite get what's happening with `.manage()`. Thanks!

justinvdm: @SteveBarnett tbh, I'm not super confident on my understanding of why jsPlumb needs `.manage()`  :) what I do know is that we need to tell jsPlumb offsets explicitly when using `.repaint()` (since we are using jsPlumb with jquery ui sortable, instead of the common use case -- freeform dragging), and doing that causes jsPlumb to sometimes get unhappy. I'm a bit unclear on the unhappiness, but what I have gathered is that jsPlumb keeps a lookup of the elements it is managing (the elements that connectors and anchors attach to), and that this lookup appear to _sometimes_ be empty. Calling `.manage()` puts the elements back in that lookup. I suspect the elements are being lost from that lookup either as a side effect of the `.repaint()`, or of some periodic task jsPlumb runs. If we simply ignored the unhappiness by doing something like `try { jsPlumb.repaint(...); } catch(e) {}`, jsPlumb still works as we want it to, which might suggest it is a bug in jsPlumb, but I'm not too sure. Ignoring errors is a terrible idea though, so calling `.manage()` seems a better idea.

The api docs, github issues and SO don't really point to a better solution (unless I'm not looking properly :)), which I'm guessing is because there aren't many people that have attempted to use jquery ui's `.sortable()` with newer jsPlumb, so this is really the best workaround I can think of without much knowledge of jsPlumb's codebase.

Hope that helps, happy to discuss this more on slack or something.

SteveBarnett: Thanks! That helps, and is plenty good enough for me.

SteveBarnett:
:+1: once Travis is happy.

hodgestar:
:+1: on the zero diff.

PR #1264, opened 2015-04-28
Feature/issue 1261 kill wizard

SteveBarnett:
Remove the wizard button and replace it with more helpful new channel, conversation, and router buttons.


SteveBarnett:
Ready for review, @justinvdm .

go/base/static/css/campaigns.less
```diff
@@ -93,8 +93,8 @@
     text-align:right; // cancel, save draft, next step
 
     // offsets the cancel button a bit.
-    a:first-child {
-      margin-right:20px;
+    #reset {
+      margin-left: 20px;
     }
```
justinvdm, line 98: Wondering if this is going to mess up other pages with elements using `reset` as an id, but sounds unlikely. I know we are already using id's in our css in some places ( /o\ ), but maybe we should try avoiding adding more, and rather add a class to the element or something?

SteveBarnett, line 98: You speak the sense. New commit pushed.

justinvdm:
:+1: once travis is happy.

I think we still need to remove `'go.wizard'` from `INSTALLED_APPS` in `settings.py`, think that could be while the travis build is erroring.

SteveBarnett:
Ready for re-review, please, @justinvdm .

justinvdm:
:+1: once travis is happy

PR #1267, opened 2015-04-29
Fix endpoint creating on routing screen

justinvdm:
At the moment creating bi-directional connections aren't possible, when the user drags from the endpoint it picks up the state instead.


justinvdm:
@SteveBarnett ready for review

SteveBarnett:
:+1:

PR #1271, opened 2015-04-30
Add Routing link to top nav

SteveBarnett:
Ready for (quick!) review, @justinvdm.

I was thinking about removing the Campaign Routing button from `.row.actions`, but I'm worried that might be confusing: "Where's routing gone?".


justinvdm:
:+1:

I think redundancy in UI is often a good thing, maybe not terribly effective in this case, but don't think it'll hurt keeping the current routing button around.

PR #1273, opened 2015-04-30
Show send status in outbound message view

jerith:
We also need a mechanism for seeing why sends failed. Delivery report handling would also be nice, but that's out of scope for now.

### DIFF
--- a/go/conversation/tests.py
+++ b/go/conversation/tests.py
@@ -846,9 +846,9 @@ class TestConversationViews(BaseConversationViewTestCase):
         r_out = self.client.get(
             self.get_view_url(conv, 'message_list'),
             {'direction': 'outbound'})
-        self.assertContains(r_out, "<td>pending", html=True)
-        self.assertNotContains(r_out, "<td>sent", html=True)
-        self.assertNotContains(r_out, "<td>failed", html=True)
+        self.assertContains(r_out, "<td>Sending", html=True)
+        self.assertNotContains(r_out, "<td>Accepted", html=True)
+        self.assertNotContains(r_out, "<td>Rejected", html=True)
 
     def test_message_list_outbound_status_failed(self):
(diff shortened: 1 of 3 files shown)


go/conversation/tests.py
```diff
@@ -839,6 +839,41 @@ def assert_messages(count):
...
+
+    def test_message_list_outbound_status_failed(self):
+        conv = self.user_helper.create_conversation(u'dummy', started=True)
+        msg = self.msg_helper.make_stored_outbound(conv, "hi")
+        self.msg_helper.make_stored_nack(conv, msg, nack_reason="no spoons")
+
+        r_out = self.client.get(
+            self.get_view_url(conv, 'message_list'),
+            {'direction': 'outbound'})
+        self.assertContains(r_out, "<td>failed: no spoons", html=True)
```
justinvdm: This might look weird in the UI if we have long nack messages. It's likely that we'll need some kind of way of dealing with long messages, but it shouldn't hold up this PR. Have we seen how it looks for the average nack message length?

justinvdm:
:+1:

go/conversation/tests.py
jerith: The message and status columns don't have widths specified, so they'll grow or shrink depending on the content. The message list page probably needs a more thorough UX overhaul, but that's out of scope for this PR.

SteveBarnett: @justinvdm: Jerith and I had this same discussion. Our conclusion was kind of the same as yours: we're not sure how long nack messages will be, so let's get it out there and see.

The table columns are squishy, at least, so the table will resize itself a bit for best fit.

justinvdm: Sounds good :)

justinvdm:
:+1:

PR #1275, opened 2015-04-30
Add highlight class for endpoints and connectors

SteveBarnett:


go/base/static/css/conversations.less
```diff
@@ -259,6 +260,18 @@
...
+    .highlighted {
+      &.endpoint {
+        background-color: @color-highlighted;
+      }
+      path, circle {
+        stroke: @color-highlighted;
+      }
+      circle, [pointer-events="all"] {
+        fill: @color-highlighted;
+      }
```
justinvdm, line 273: Could you explain why we need `[pointer-events="all"]` here?

justinvdm:
@SteveBarnett Sorry, just realised that this probably isn't ready for review yet.

go/base/static/css/conversations.less
SteveBarnett, line 273: I need to target the arrowheads, and it seems only they have `[pointer-events="all"]`.
¯\_(ツ)_/¯

justinvdm, line 273: Ah, I see, cool thanks :)

justinvdm:
@SteveBarnett I think we can probably close this now, right?

SteveBarnett:
@justinvdm Yeah. Not going to finish it off.

PR #1280, opened 2015-05-07
Persist routing screen positions using local storage

justinvdm:


justinvdm:
@SteveBarnett ready for review. Sorry for how long this one took.

go/base/static/js/test/local.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+      assert.strictEqual(go.local.get('foo'), null);
+      assert.notStrictEqual(go.local.get('bar'), null);
+    });
+
+    it("should support default values", function() {
+      assert(!('foo' in localStorage));
+      localStorage.bar = JSON.stringify({baz: 21});
+
+      assert.strictEqual(go.local.get('foo', 23), 23);
```
SteveBarnett, line 27: Can you clarify this one for me, please?
I'm not sure about this `assert`: it looks like the setting of `foo` happens in a different `describe` block?

justinvdm, line 27: We are asking `.get()` to give us the current local storage value for the key `'foo'`, but asking it to instead return `23` if `'foo'` does not exist. Does that help?

SteveBarnett, line 27: Yep. Sorry. I wasn't reading `go.local.get` properly. :-/

SteveBarnett:
Awesome work, @justinvdm! ^5 and :+1:

PR #1277, opened 2015-05-07
Fix channel releasing from channel show screen

justinvdm:
At the moment it 404s.


justinvdm:
Ready for review. I could add a test to check that we no longer reload the page, but it seems weird to add a test to check that we no longer have something that shouldn't have been there in the first place.

SteveBarnett:
@justinvdm :+1: . Agreed re no need to add test.

PR #1278, opened 2015-05-19
Allow group message conversations to be used with non-SMS transports.

hodgestar:
This used to be possible, but when conversations stopped honouring `conv.delivery_class`, group message was never updated.

We need to make `delivery_class` an option in the group message send message action (similar to how it is for dialogues) and then pass that option through to the management command.


hodgestar:
Ready for review.

jerith:
:+1:

PR #1284, opened 2015-05-19
Set default delivery class for bulk send to SMS (since USSD seldom makes sense).

hodgestar:
The default default delivery class is USSD, which isn't particularly useful for bulk send, although it happens to be the global default (which is really the default for dialogues).


hodgestar:
Ready for review.

rudigiesler:
:+1:

PR #1286, opened 2015-05-20
Add fallback edit view so that every conversation type has one.

hodgestar:
Currently the campaign routing table links to the edit view of conversations but some types (e.g. group message) don't have edit views. We should add a fallback edit view to avoid such 404s in this case and in potential other cases.


hodgestar:
Ready for review.

justinvdm:
:+1: once travis is happy

PR #1224, opened 2015-05-25
Assorted performance cleanups in views

jerith:
There are some Django views that make unnecessary slow or expensive calls, mostly because they predate paginated index lookups and such. This issue is for finding and fixing some (but not necessarily all) of them and cleaning up some related code to make it avoid doing expensive things accidentally.


jerith:
The only useful change I could find was to remove the unpaginated index query in the contacts views.

go/contacts/views.py
```diff
@@ -423,19 +423,27 @@ def _people(request):
...
-        keys = contact_store.list_contacts()
+        keys = contact_store.list_contacts_page(max_results=fetch_limit)
 
-    limit = min(int(request.GET.get('limit', 100)), len(keys))
-    messages.info(request, "Showing %s of %s contact(s)" % (limit, len(keys)))
+    key_count = len(list(keys))
+    limit = min(int(request.GET.get('limit', 100)), key_count)
+    # If we have a continuation token, it means there are more than 10k keys.
+    if getattr(keys, 'continuation', None) is not None:
+        key_count = "10k+"
```
hodgestar: This will be incorrect if `_fetch_limit` is set to something else. Perhaps we should do `"%dk+" % (fetch_limit / 1e3)`?

hodgestar:
Other than printing fetch_limit, looks good.

I'd like to see about bringing counts back at some point, especially since we use them for groups anyway and groups can easily be roughly the same size as the entire set of contacts, but let's get things working first and then we can go from there.

go/contacts/views.py
jerith: I'll just use `"%s+" % fetch_limit` instead of `"10k+"`.

jerith:
Ready for re-review.

hodgestar:
:+1:

PR #1287, opened 2015-05-28
Disable event statuses because they cause the message list to load too slowly.

hodgestar:
Displaying event status requires a riak index lookup for each message displayed in the message list (so 20 by default). These 20 index looks are slow and easily cause the page load to time out.

### DIFF
--- a/go/conversation/tests.py
+++ b/go/conversation/tests.py
@@ -13,6 +13,7 @@ from django.utils.unittest import skip
 from vumi.message import TransportUserMessage
 
 import go.base.utils
+import go.conversation.settings
 from go.base.tests.helpers import GoDjangoTestCase, DjangoVumiApiHelper
 from go.conversation.templatetags import conversation_tags
 from go.conversation.view_definition import (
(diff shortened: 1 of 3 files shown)


hodgestar:
Ready for review.

go/conversation/tests.py
```diff
@@ -353,6 +356,10 @@ def test_post_new_conversation_starting(self):
 
 
 class TestConversationViews(BaseConversationViewTestCase):
+
+    with_event_statuses = mock.patch(
+        'go.conversation.settings.ENABLE_EVENT_STATUSES_IN_MESSAGE_LIST', True)
+
```
jerith: Can we perhaps use `self.monkey_patch()` in the tests that need it instead of `mock`? Using different patching mechanisms in different places is confusing.

hodgestar:
Ready for re-review.

jerith:
:+1: after the build I restarted is happy. (Failure due to timeout in routing test.)

PR #1290, opened 2015-06-03
Add routers for celery tasks.

justinvdm:
Celery allows custom routers for tasks, see https://celery.readthedocs.org/en/latest/userguide/routing.html#routers. I propose adding two custom router classes:
- `CeleryRegexRouter` that directs tasks based on a regular expression for the task name (this is low credit notifications so that they don't need to set behind statement generation and archival tasks)
- `CeleryAppRouter` that directs tasks based on the first two parts of the task name (so all tasks that start with `go.billing.*` are directed to `go.billing`, all tasks that start with `go.conversation.*` go to `go.conversation`).

```
CELERY_ROUTES = (
    CeleryRegexRouter(r'^go\.billing\..*low_credit.*$', queue='go.billing.low_credit'),
    CeleryAppRouter(),
)
```

### DIFF
--- a/go/base/celery_routers.py
+++ b/go/base/celery_routers.py
@@ -1,0 +1,13 @@
+import re
+
+
+class CeleryRegexRouter(object):
+    def __init__(self, regex, queue):
+        self.regex = re.compile(regex)
+        self.queue = queue
+
+    def route_for_task(self, task, args=None, kwargs=None):
+        if self.regex.match(task):
+            return {'queue': self.queue}
+        else:
+            return None
--- a/go/base/celery_routers.py
+++ b/go/base/celery_routers.py
@@ -11,3 +11,8 @@ class CeleryRegexRouter(object):
             return {'queue': self.queue}
         else:
             return None
+
+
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review, I think.

go/base/celery_routers.py
```diff
@@ -0,0 +1,18 @@
...
+    def route_for_task(self, task, args=None, kwargs=None):
+        if self.regex.match(task):
+            return {'queue': self.queue}
+        else:
+            return None
+
+
+class CeleryAppRouter(object):
+    def route_for_task(self, task, args=None, kwargs=None):
+        return {'queue': '.'.join(task.split('.')[:2])}
```
hodgestar, line 18: I wonder if we should return `None` if `task.split('.')` doesn't have enough parts?

hodgestar, line 18: On second thoughts, I think this is fine.

smn:
ooh! :+1:

hodgestar:
:+1:

PR #1293, opened 2015-06-04
Warn if bulk send messages contain non-ASCII characters.

hodgestar:


hodgestar:
@justinvdm @steve Ready for review.

go/base/static/js/src/components/views.js
```diff
@@ -36,15 +36,30 @@
...
           $p = $('<p class="textarea-char-count"/>');
           this.$el.after($p);
       }
-      this.totalChars = this.$el.val().length;
+      this.containsNonAscii = this.contains_non_ascii(text);
+      this.totalChars = text.length * (1 + this.containsNonAscii);
       this.totalSMS = Math.ceil(this.totalChars / this.SMS_MAX_CHARS);
-      $p.html(this.totalChars + ' characters used<br>' + this.totalSMS + ' smses');
+      $p.html(
+          (this.containsNonAscii ? 'Non-ASCII characters detected<br>' : '') +
```
justinvdm: `Non-ASCII characters detected` is accurate, but people are likely to not understand the implications of this. Maybe we need to mention that this may cause problems for some channels?

justinvdm:
I think we probably want to add this feature to dialogues too. Probably not for this PR though.

go/base/static/js/src/components/views.js
```diff
@@ -36,15 +36,30 @@
...
     render: function() {
       var $p = this.$el.next();
+      var text = this.$el.val();
       if (!$p.hasClass('textarea-char-count')) {
           $p = $('<p class="textarea-char-count"/>');
           this.$el.after($p);
       }
-      this.totalChars = this.$el.val().length;
+      this.containsNonAscii = this.contains_non_ascii(text);
+      this.totalChars = text.length * (1 + this.containsNonAscii);
```
justinvdm: Struggling to understand this. Could you explain why we double the total chars if it contains ascii characters? Is it a channel-specific thing?

justinvdm:
Left some comments, looks good. Thanks for working on this.

SteveBarnett:
Thanks for doing this, @hodgestar!

If there's a problem with the content that the user has input we should tell them (exactly) what the problem is, and how to fix it. `Non-ASCII characters detected` is accurate, but maybe we can be more helpful?
Can we grab the offending characters in `contains_non_ascii` and use them in the message?

> Non-ASCII characters found in your message: `ó`. [description of what will go wrong here, in non-technical terms]

Should this be show as an error maybe? Stick the message in it's own `<div class="text-right text-danger"></div>` to make it red?

Also: we just want to warn, not stop, yes? Because only some channels with have a problem?
Can we check this for the user (and then add a `disabled` attribute to the Ok button so that people can't break their dialogue.)?

go/base/static/js/src/components/views.js
hodgestar: It's SMS specific but fairly generic. If there is a non-ascii character the
encoding changes to a 2-byte one and the number of bytes doubles.

justinvdm: Cool. I'm a bit worried by this. Depending on the channel used, the character count might or might not double. I guess most bulk sends are done over SMS, so its maybe okay to keep this behaviour, but still seems a bit weird for other cases. I guess this is more of a thing if we want to include this notification in dialogues.

@SteveBarnett thoughts?

hodgestar:
Agreed. I was wondering if there is somewhere we could put the non-ascii
check and character count function so that both can access those.

justinvdm:
> Agreed. I was wondering if there is somewhere we could put the non-ascii
> check and character count function so that both can access those.

+1, maybe `go.utils`? Probably something for a different PR.

justinvdm:
:+1: once [this](https://github.com/praekelt/vumi-go/pull/1293#issuecomment-109204700) is resolved.

go/base/static/js/src/components/views.js
hodgestar: Maybe we should keep the original count and do something like `80 chars (~160 bytes)` if there are non-ASCII characters present? Is "bytes" too confusing?

hodgestar:
Todo items:
- [x] Factor out non-ascii character check into a function in `go.utils`.
- [x] Better display of byte count (maybe `80 chars (~160 bytes)`).
- [x] Add `class="text-danger"` to make the text red when there are non-ASCII characters
- [x] Provide list of non-ASCII characters alongside warning.

@steve @justinvdm Did I miss anything?

go/base/static/js/src/components/views.js
justinvdm: I don't think "bytes" would be well understood. Maybe we just need to keep the original count and mention that non-ASCII characters means the character count will get doubled for SMS? I think that even for SMS, if the character count all of a sudden doubled and we didn't explain why, the user would probably be surprised and not understand the reason for it doubling. Thoughts?

justinvdm:
@hodgestar todo items looks good, thanks :)

go/base/static/js/src/components/views.js
SteveBarnett: Yeah, I think "bytes" is a technical term. 
I'm thinking something like: 

> Please note that Non-ASCII characters like `ó` take up two spaces instead of one.

SteveBarnett:
@hodgestar todo list looks great :)

hodgestar:
@justinvdm @SteveBarnett Ready for re-review, I think.

go/base/static/js/src/utils.js
```diff
@@ -118,6 +118,17 @@
     return s.charAt(0).toUpperCase() + s.slice(1);
   };
 
+  var non_ascii = function(s) {
```
hodgestar, line 121: @justinvdm Do you think this function will be useful for the dialogue case too? If not, perhaps we should modify what it does a bit.

justinvdm, line 121: It might be. Whatever we do to the bulk send screen, we should do similarly for dialogues, for the sake of consistency.

go/base/static/js/src/components/views.js
```diff
@@ -42,10 +42,30 @@
...
+      this.containsNonAscii = (non_ascii.length > 0);
+      this.totalChars = text.length;
+      this.totalBytes = text.length * (1 + this.containsNonAscii);
+      this.totalSMS = Math.ceil(this.totalBytes / this.SMS_MAX_CHARS);
+      var html;
+      if (this.containsNonAscii) {
+          html = [
+              'Non-ASCII characters: ' + non_ascii.join(', '),
+              this.totalChars + ' characters used (~' +
+                  this.totalBytes + ' bytes)',
```
justinvdm: I know this is getting very bikesheddy, but I'm wondering how useful showing the bytes is to people. I doubt most people will understand the implications of the bytes. I vote we either don't display the byte count, or we mention the implications of the byte count.

hodgestar: Happy for this to be bike shedded if people have concrete suggestions.

I'm personally okay with using the word bytes because:
- It's only there if you have non-ASCII characters.
- The SMS count is the most important part.
- If you're sending SMSes with non-ASCII characters you need to have some
  idea of what bytes are anyway, because they directly affect how much money
  you're going to spend (and spending $200,000 instead of $100,000 is enough
  reason to learn the meaning of "bytes").

hodgestar:
@justinvdm @SteveBarnett Ready for post bikeshedding review. :)

go/base/static/js/src/components/views.js
```diff
@@ -29,6 +29,14 @@
...
   var MessageTextView = Backbone.View.extend({
     SMS_MAX_CHARS: 160,
+    SMS_MAX_CHARS_NONASCII: 70,
+    NON_ASCII_WARNING: [
+        "These characters may not be understood correctly by some USSD and SMS",
+        " channels. For SMS channels, it will also reduce the number of",
+        " characters per SMS from 160 to 70 and possibly double the number of",
+        " SMSes billed for. For USSD channels, the maximum number of characters",
+        " will be similarly reduced."
+    ].join(""),
```
justinvdm, line 39: +1

justinvdm:
:+1:

hodgestar:
/me merges quickly before anyone else can comment. ;)

PR #1295, opened 2015-06-05
Remove zomby connections

justinvdm:
There appears to be an obscure edge case where an endpoint is removed, but the connection stays around. Finding and fixing this bug is one thing, but fixing existing dialogues broken as a result of this is another. This ticket is for fixing existing dialogues by removing these zomby connections.

### DIFF
--- a/go/base/static/js/src/components/plumbing/connections.js
+++ b/go/base/static/js/src/components/plumbing/connections.js
@@ -122,6 +122,8 @@
       jsPlumb.bind(
         'connectionDetached',
         _.bind(this.onPlumbDisconnect, this));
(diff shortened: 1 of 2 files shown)


justinvdm:
@SteveBarnett Ready for review.

Ideally, this "fix" should happen before views use the erroneous model data, but we currently don't have the tools we need to do this: we would need a way of searching through all endpoints in a diagram model, and at the moment, we only have the tools for doing this in views.

SteveBarnett:
:+1: for getting a quick fix out, but it would be great to go for a more complete solution, if we have the time.

PR #1296, opened 2015-06-17
Add help for HTTP API access.

hodgestar:
Currently we have no help and confused people email us on a regular basis asking now to use the HTTP API.

### DIFF
--- a/go/apps/http_api_nostream/templates/http_api_nostream/edit_help.html
+++ b/go/apps/http_api_nostream/templates/http_api_nostream/edit_help.html
@@ -32,7 +32,7 @@
 </p>
 
 
-<h1>Example send</h1>
+<h2>Example send</h2>
 
 <p>You can send messages via this conversation using an HTTP PUT request as follows:</p>
 
@@ -47,7 +47,7 @@
        -vvv
 </code></pre>
 
-<h1>Example metric firing</h1>
+<h2>Example metric firing</h2>
 
 <p>You can fire metrics via this conversation using an HTTP PUT request as follows:</p>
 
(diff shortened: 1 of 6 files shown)


hodgestar:
@steve @justinvdm @jerith Ready for first review.

Bikeshedding encouraged.

hodgestar:
http-api-help

Better formatting required.

SteveBarnett:
Yay to having help stuff! :)
I have a bunch of things of various levels of niggle, though (of course :)).
- Consistent capitalisation of API, URL
- For ignore messages and ignore events, use the `.help-block` text as the label instead.
- Change the right, example, column class to `.col-md-7` to give it a bit more space.
- In the notes bit:
  - we should say `user` instead of `username`, since that's what it's called in the example.
  - say where the user can find their Vumi Go account key: at the bottom of [Account Details](http://localhost:8000/account/details/)
  - Replace "The hexadecimal value at the end of the string is the unique identifier for this conversation." with "The value at the end of the string is the unique identifier for this conversation. This conversation is {{current convo uuid}}."

Chatting with @justinvdm, he suggest moving the whole notes section to be contextual, and I agree. Like: "If your account key is `foo` and your conversation's id is `bar`, then your HTTP PUT request would look like this:"

hodgestar:
http-api-help-2

hodgestar:
http-api-help-2

hodgestar:
@SteveBarnett @justinvdm I think I've addressed all of the comments except the one about changing `username` to `user`. A quirk of `curl` is that the argument to `--user` is both the username and the password so it seems less confusing to speak of `username` since it's only part of what is passed to `--user`.

hodgestar:
Ready for re-review.

SteveBarnett:
@hodgestar Looks great! :+1: from me.

justinvdm:
Agreed :) :+1:

PR #1297, opened 2015-06-18
Avoid creating a VumiApi when we only need Redis manager

jerith:
There are two places we do this, and it's a bunch of unnecessary work.

This is even more of a problem when we start needing to clean up Riak managers properly.

### DIFF
--- a/go/api/go_api/session.py
+++ b/go/api/go_api/session.py
@@ -1,10 +1,9 @@
 # -*- coding: utf-8 -*-
 # -*- test-case-name: go.api.go_api.tests.test_session -*-
 
-from django.conf import settings
 from django.contrib.sessions.backends.base import SessionBase, CreateError
-from vumi.persist.redis_manager import RedisManager
 
+from go.base.utils import get_redis_manager
 from go.api.go_api.session_manager import SessionManager
 
 
@@ -14,10 +13,8 @@ class SessionStore(SessionBase):
     """
     def __init__(self, session_key=None):
         super(SessionStore, self).__init__(session_key)
-        redis_config = settings.VUMI_API_CONFIG.get('redis_manager', {})
-        redis = RedisManager.from_config(redis_config)
         self.session_manager = SessionManager(
-            redis.sub_manager('session_manager'))
+            get_redis_manager().sub_manager('session_manager'))
 
     def encode(self, data):
         """Replace Django's pickle serialization with a null-op.
--- a/go/api/go_api/session.py
+++ b/go/api/go_api/session.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 # -*- test-case-name: go.api.go_api.tests.test_session -*-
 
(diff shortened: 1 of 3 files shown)


go/api/go_api/session.py
```diff
@@ -12,7 +13,8 @@ class SessionStore(SessionBase):
     """
     def __init__(self, session_key=None):
         super(SessionStore, self).__init__(session_key)
-        self.session_manager = vumi_api().session_manager
+        self.session_manager = SessionManager(
+            get_redis_manager().sub_manager('session_manager'))
```
hodgestar, line 17: Not wild about having to duplicate the `session_manager` and `token_manager` strings here and in the token view.

hodgestar:
One small question about approach to getting access to the session and token managers, but otherwise seems okay.

go/api/go_api/session.py
jerith, line 17: We can put them in class methods on `SessionManager` or something, I suppose. The real problem is that `VumiApi` and friends make all these things together and there's no easy way to get one without the others.

hodgestar, line 17: The idea of the `VumiApi` is that it should create everything somewhat lazily, but I think putting a classmethod on `SessionManager` and `TokenManager` that encapsulates this setup sounds like a good compromise to me.

jerith, line 17: There's currently no way to create a `VumiApi` without also creating all the Riak-related things. That's fine for the common case when we actually want all those things, but not so much for this case where we don't.

hodgestar, line 17: I'm suddenly tempted to make the components of the VumiApi properties that
are only created when they're accessed. E.g.

``` python

@property
def tpm(self):
    tpm = getattr(self, '_tpm', None)
    if tpm is None:
        tpm = self._tpm =
TagpoolManager(self.redis.sub_manager('tagpool_store'))
    return tpm
```

Not sure if this is a cool idea. We can probably avoid the boiler plate by
writing a `lazy_attribute` decorator that turns

``` python
@lazy_property
def tpm(self):
    return TagpoolManager(self.redis.sub_manager('tagpool_store'))
```

into the property described previously.

If you like but don't feel like writing these, I'm happy to give them a go
in a separate PR.

jerith, line 17: This doesn't actually solve the underlying problem, which is that we only need one small thing but the standard way of getting it is to create a big container full of all sorts of things.

In the case of this PR, we'd need to know that we can safely clean up the Riak manager that gets created but not the Redis manager. In the case of the `SessionStore` we also have nowhere to do any cleanup that might be necessary when we're finished with the session store -- we're already potentially leaking Redis clients with this code.

hodgestar, line 17: Not wild about these spreading around the code base, but perhaps cleaning it up can wait for another day. :)

hodgestar:
:+1:

PR #1298, opened 2015-06-19
Refactor management commands to use common base classes

jerith:
We have some base classes that a few of our management commands use. The others (with a couple of exceptions, probably) should be rewritten to use these.

### DIFF
--- a/go/base/management/commands/go_generate_export_conversations_urls.py
+++ b/go/base/management/commands/go_generate_export_conversations_urls.py
@@ -5,8 +5,7 @@ from optparse import make_option
 from django.core.management.base import CommandError
 from django.utils.text import slugify
 
-from go.base.utils import vumi_api_for_user
-from go.base.command_utils import BaseGoCommand, get_user_by_email
+from go.base.command_utils import BaseGoCommand
 
 
 class Command(BaseGoCommand):
@@ -38,8 +37,7 @@ class Command(BaseGoCommand):
             raise CommandError('--base-url is mandatory.')
         self.template = kwargs['template']
 
-        self.user = get_user_by_email(self.email)
-        self.user_api = vumi_api_for_user(self.user)
+        self.user, self.user_api = self.mk_user_api(self.email)
 
         conversation_store = self.user_api.conversation_store
         conversation_keys = conversation_store.list_conversations()
--- a/go/base/management/commands/go_generate_export_conversations_urls.py
+++ b/go/base/management/commands/go_generate_export_conversations_urls.py
@@ -8,6 +8,8 @@ from django.utils.text import slugify
 from go.base.command_utils import BaseGoCommand
 
 
+# We don't extend BaseGoAccountCommand since the '--email' option is used
+# instead of '--email-address'
 class Command(BaseGoCommand):
 
     help = "Dump URLs for use with cURL for downloading message data."
--- a/go/base/management/commands/go_list_opt_outs.py
+++ b/go/base/management/commands/go_list_opt_outs.py
@@ -1,24 +1,31 @@
 from optparse import make_option
 
-from go.base.command_utils import BaseGoCommand, make_email_option
+from django.core.management.base import BaseCommand
+
+from go.base.utils import vumi_api_for_user
+from go.base.command_utils import get_user_by_email
 from go.vumitools.opt_out import OptOutStore
 
 
-class Command(BaseGoCommand):
+class Command(BaseCommand):
     help = "List opt-outs from a particular account"
 
-    LOCAL_OPTIONS = [
-        make_email_option()
-    ]
-
-    option_list = BaseGoCommand.option_list + tuple(LOCAL_OPTIONS)
+    LOCAL_OPTIONS = (
+        make_option('--email-address',
+                    dest='email-address',
+                    help='Email address for the Vumi Go user'),
+    )
+    option_list = BaseCommand.option_list + LOCAL_OPTIONS
 
     def handle(self, *args, **options):
         options = options.copy()
         self.handle_validated(*args, **options)
 
     def handle_validated(self, *args, **options):
-        _, user_api = self.mk_user_api(options['email-address'])
+        email_address = options['email-address']
+
+        user = get_user_by_email(email_address)
+        user_api = vumi_api_for_user(user)
 
         self.show_opt_outs(user_api, email_address)
 
--- a/go/base/management/commands/go_list_opt_outs.py
+++ b/go/base/management/commands/go_list_opt_outs.py
@@ -1,22 +1,30 @@
 from optparse import make_option
 
-from go.base.command_utils import BaseGoAccountCommand
+from go.base.command_utils import BaseGoCommand, make_email_option
 from go.vumitools.opt_out import OptOutStore
 
 
 class Command(BaseGoCommand):
     help = "List opt-outs from a particular account"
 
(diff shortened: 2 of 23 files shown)


go/base/tests/test_go_start_conversation.py
```diff
@@ -18,8 +18,6 @@ def setUp(self):
         self.command.stderr = StringIO()
 
     def test_sanity_checks(self):
-        self.assertRaisesRegexp(CommandError, 'provide --email-address',
-            self.command.handle, email_address=None, conversation_key=None)
```
jerith, line 22: Can you put this assertion back and change the message it asserts on, please?

JayH5, line 22: I can but I removed the test because it would now test `BaseGoCommand#mk_user_api()`, not anything in `go_start_conversation`.

go/base/tests/test_go_manage_contact_group.py
```diff
@@ -14,13 +14,9 @@ def setUp(self):
 
     def invoke_command(self, command, **kw):
         options = {
-            'email-address': self.user_helper.get_django_user().email,
-            'list': False,
-            'create': False,
-            'create-smart': False,
-            'delete': False,
+            'email_address': self.user_helper.get_django_user().email,
+            'command': [command]
```
jerith: Please put trailing commas at the end of lists/dicts/whatever. It saves pain later when we reorder or add entries.

go/base/tests/test_go_assign_tagpool.py
```diff
@@ -19,9 +19,9 @@ def setUp(self):
 
     def handle_command(self, tagpool=None, max_keys=None, update=False):
         self.command.handle(**{
-            "email-address": self.user_helper.get_django_user().email,
+            "email_address": self.user_helper.get_django_user().email,
             "tagpool": tagpool,
-            "max-keys": max_keys,
+            "max_keys": max_keys,
             "update": update
```
jerith: Please add a trailing comma here as well, even though it was already missing before you touched it.

go/base/management/commands/go_manage_routing_table.py
```diff
@@ -51,48 +43,46 @@ def handle(self, *args, **options):
...
-        if len(command_options) != 1:
+        # Ensure a single option is chosen from the local options
+        cmd_options = [c.dest for c in self.LOCAL_OPTIONS if options[c.dest]]
+        if len(cmd_options) != 1:
             raise CommandError('Please provide exactly one of: %s' % (
-                ['--%s' % c for c in self.CONFLICTING_OPTIONS],))
-
-        user = get_user_by_email(options['email-address'])
-        user_api = vumi_api_for_user(user)
+                ['--%s' % c.dest for c in self.LOCAL_OPTIONS],))
```
jerith: Using `LOCAL_OPTIONS` here is a little less clear than the old `CONFLICTING_OPTIONS` mechanism, even with the explanatory comment. It's also problematic if we add a new non-conflicting option later. Any thoughts about a way to improve this?

go/base/management/commands/go_manage_application.py
```diff
@@ -1,21 +1,15 @@
...
+class Command(BaseGoAccountCommand):
     help = "Give a Vumi Go user access to a certain application"
 
     LOCAL_OPTIONS = [
-        make_option('--email-address',
-            dest='email-address',
-            help='Email address for the Vumi Go user'),
-        make_option('--application-module',
-            dest='application-module',
+        make_option('--application_module',
```
jerith: The option we accept from the command line should be `--application-module`, not `--application_module`. We only want to change the internal dict key, not the thing we have to type to get the behaviour we want.

go/base/management/commands/go_manage_routing_table.py
JayH5: It would be nice if these were actual commands (in the `BaseGoCommand` sense). But they don't quite fit that pattern. I guess `LOCAL_OPTIONS` is less clear but it just seems a bit confusing with `CONFLICTING_OPTIONS` when _all_ the options "conflict". Maybe we could call it something like `COMMANDS` and then leave a comment explaining there must only be 1 command?

go/base/tests/test_go_start_conversation.py
jerith, line 22: This tests the command from the outside. We know that it should complain if we don't give it an email address, but we don't know where the code that checks that lives. If we change this to not use `BaseGoCommand` for some reason, we could break the command's handling of the email address without any tests failing.

go/base/management/commands/go_manage_application.py
JayH5: oops.. did not mean to change that

go/base/management/commands/go_manage_routing_table.py
jerith: The problem with the old way of doing this is that `CONFLICTING_OPTIONS` was a separate list of things that has to match a subset of the things in `LOCAL_OPTIONS`. The problem with the new way is that there's no distinction between options that conflict and options that don't.

What about renaming `LOCAL_OPTIONS` to `COMMAND_OPTIONS` and having a separate (empty) `LOCAL_OPTIONS` as well?

go/base/tests/test_go_start_conversation.py
JayH5, line 22: Yay OO inheritance :disappointed:

jerith, line 22: This would still be a problem if we used composition instead of inheritance. ;-)

go/base/management/commands/go_generate_export_conversations_urls.py
```diff
@@ -5,8 +5,7 @@
 from django.core.management.base import CommandError
 from django.utils.text import slugify
 
-from go.base.utils import vumi_api_for_user
-from go.base.command_utils import BaseGoCommand, get_user_by_email
+from go.base.command_utils import BaseGoCommand
 
 
 class Command(BaseGoCommand):
```
jerith, line 13: This probably needs a comment explaining why it's not using `BaseGoAccountCommand`.

go/base/management/commands/go_list_opt_outs.py
```diff
@@ -1,37 +1,22 @@
...
-from django.core.management.base import BaseCommand
-
-from go.base.utils import vumi_api_for_user
-from go.base.command_utils import get_user_by_email
+from go.base.command_utils import BaseGoAccountCommand
 from go.vumitools.opt_out import OptOutStore
 
 
-class Command(BaseCommand):
+class Command(BaseGoCommand):
```
jerith: We import `BaseGoAccountCommand`, but use `BaseGoAccount`. Apparently we have no tests for this command.

Probably best to revert the changes to this module and create an issue to write tests and fix it separately.

JayH5: :cry: should have got my linter working sooner

jerith: If you had, we probably wouldn't have noticed the missing tests. :-/

jerith:
Looks good to merge after addressing the last two comments. Thanks!

JayH5:
Worth mentioning that...
- `VumiApi`s are now only created in `BaseGoCommand`
- `VumiUserApi`s are still created in quite a few commands, but there are far fewer than before

jerith:
:+1:

Great work, thanks.

PR #1299, opened 2015-06-23
Avoid performing many selects when archiving billing transactions

hodgestar:
Our current implementation performs an ordered select query over all transactions for the month for every 1000 transactions which heavily loads the database and is very slow.


hodgestar:
Old but good write-up of the underlying issues -- http://thebuild.com/blog/2010/12/13/very-large-result-sets-in-django-using-postgresql/

hodgestar:
The `djorm-ext-core` module can be found at https://github.com/niwinz/djorm-ext-core and supports both `Django<=1.6`  and `Django>=1.6`.

hodgestar:
Ready for review.

Any thoughts on how to test that the server side cursor is actually being used would be appreciated because I don't have any great ideas for that.

go/billing/django_utils.py
```diff
@@ -20,6 +22,28 @@ def to_json(self, transactions):
...
+    :param int items_per_chunk:
+        Number of objects to include in each chunk. Default 1000.
+    """
+    with server_side_cursors(itersize=items_per_chunk):
+        items = []
+        for item in queryset.iterator():
+            items.append(item)
+            if len(items) == items_per_chunk:
+                yield items
+                items = []
```
MrHamdulay: I like

``` python
iterator = queryset.iterator()
items = itertools.islice(iterator, itersize)
while items:
   yield items
   items = itertools.islice(iterator, itersize)
```

but islice isn't exactly the most obvious method :/

hodgestar:
Bleh. It looks like `djorm-core-ext` is thoroughly broken on Django 1.5.8. I will need to fix this up.

hodgestar:
Ah, it appears that things were not as dire as they first seemed -- we just needed a newer version of `psycopg2`.

hodgestar:
Ready for re-review.

hodgestar:
The new `pyscopg2` version was needed because the `withhold` argument to `cursor` was added in `2.4.3` -- see http://initd.org/psycopg/docs/cursor.html?highlight=withhold#cursor.withhold.

rudigiesler:
There's no real way that we can test memory usage during a test is there? We could use `resource`, but I'm not sure how reliable it would be.

hodgestar:
I was hoping I could inspect Django's model cache directly, but I couldn't find a way that didn't involve grubbing around in lots of private attributes on Django classes. We could use `gc.get_objects()` and look for instances of the model.

Update: I realised that `gc.get_objects()` and looking for model instances isn't really going to test what we want to because the memory might be used by a client side cursor before `iterator()` generates model instances. Plus we'd need to distinguish between instances of the model that we are keeping around for the test itself and ones being produced by the iterator.

rudigiesler:
Okay then :+1: from me. Wondering if we could could create a large table on solo using a Django shell, then running this and watching memory.

PR #1301, opened 2015-06-24
Allow unit costs to have a different number of decimal places to other costs.

hodgestar:
Unit costs can be much smaller than other dollar amounts and so if we format them with the same precision, they often end up appearing as `0.000`. We could allow the number of decimal places for unit costs to be configurable separately and default it to 5 decimal places.


hodgestar:
Ready for review.

rudigiesler:
:+1:

PR #1304, opened 2015-06-25
Improve Transaction Archive admin view.

hodgestar:
Currently the transaction archive list shows only the names of the archives which isn't very useful. We need to be able to see more of the details and be able to filter by status and account.


hodgestar:
Ready for review.

@badlard This is mostly so that we can easily check that transaction archiving was successful.

rudigiesler:
:+1:

badlard:
@hodgestar very helpful, thank you.

PR #1306, opened 2015-07-02
Clean up unnecessary get_tag_info juggling

jerith:


rudigiesler:
I'm not sure about the overall view what's happening. Why are we removing the caching? :+1: On the code, looks good to me.

jerith:
We're only getting it through the cache in one place (during routing), so the extra layer isn't buying us anything. The driving force behind the change is cleaning up unnecessary complexity around the message store so we can switch to the new implementation more easily.

rudigiesler:
:+1:

go/vumitools/utils.py
```diff
@@ -188,13 +188,6 @@ def set_tag(self, tag):
         TaggingMiddleware.add_tag_to_msg(self.message, tag)
         self.tag = TaggingMiddleware.map_msg_to_tag(self.message)
 
-    def get_tag_info(self):
-        if self.tag is None:
-            raise ValueError("No tag to look up metadata for.")
```
JayH5, line 193: So now we don't have this `None` check... Is it ok to pass `None` to `get_tag_info` or are we sure we won't do that?

jerith, line 193: We already have a `None` check at the call site. :-)

JayH5, line 193: oh lol. Long day.

JayH5:
:+1: :computer:

PR #1307, opened 2015-07-03
Use batch_ids instead of batch_id in message store calls

jerith:
The current message store methods accept both `batch_id` and `batch_ids` to specify batches. The new message store only accepts `batch_ids`. Switching everything to `batch_ids` now will simplify later work when we start using the new message store.


JayH5:
:+1: seems reasonable :smile:

PR #1308, opened 2015-07-06
Always use paginated message store index queries

jerith:
There are a few places where we don't.

### DIFF
--- a/go/base/management/commands/go_account_stats.py
+++ b/go/base/management/commands/go_account_stats.py
@@ -92,11 +92,20 @@ class Command(BaseGoCommand):
         self.do_batch_key(message_store, batch_key)
         self.do_batch_key_breakdown(message_store, batch_key)
 
+    def _count_results(self, index_page):
+        count = 0
+        while index_page is not None:
+            count += len(list(index_page))
+            index_page = index_page.next_page()
+        return count
+
     def do_batch_key(self, message_store, batch_key):
-        self.out(u'Total Received in batch %s: %s\n' % (
-            batch_key, message_store.batch_inbound_count(batch_key),))
-        self.out(u'Total Sent in batch %s: %s\n' % (
-            batch_key, message_store.batch_outbound_count(batch_key),))
+        in_count = self._count_results(
+            message_store.batch_inbound_keys_page(batch_key))
+        out_count = self._count_results(
+            message_store.batch_outbound_keys_page(batch_key))
+        self.out(u'Total Received in batch %s: %s\n' % (batch_key, in_count))
+        self.out(u'Total Sent in batch %s: %s\n' % (batch_key, out_count))
 
     def parse_timestamp_to_date(self, timestamp):
         return parse_vumi_date(timestamp).date()
--- a/go/base/management/commands/go_migrate_conversations.py
+++ b/go/base/management/commands/go_migrate_conversations.py
@@ -101,13 +101,19 @@ class FixBatches(Migration):
             return False
         return len(conv.batches.keys()) != 1
 
+    def _process_pages(self, index_page, batch_id, get_message, add_message):
+        while index_page is not None:
+            for key in index_page:
+                add_message(get_message(key), batch_ids=[batch_id])
+            index_page = index_page.next_page()
+
     def _copy_msgs(self, mdb, old_batch, new_batch):
-        for key in mdb.batch_outbound_keys(old_batch):
-            msg = mdb.get_outbound_message(key)
-            mdb.add_outbound_message(msg, batch_ids=[new_batch])
-        for key in mdb.batch_inbound_keys(old_batch):
-            msg = mdb.get_inbound_message(key)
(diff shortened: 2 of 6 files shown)


jerith:
This PR is probably best reviewed one commit at a time.

JayH5:
lgtm, I only worry that there are too many different (but similar) implementations of iterating through paginated results. Can there be One Way?

The code I'm going through atm has one of the more complicated implementations (see [here](https://github.com/praekelt/vumi/blob/develop/vumi/components/message_store_resource.py#L113-L125)).

And when I see code like [this](https://github.com/praekelt/vumi-go/blob/89f68241f6fc0180013f6886a0b69ab254df662a/go/base/management/commands/go_account_stats.py#L95-L100), I wonder if maybe the index pages could be smarter? Maybe implement `__len__()`?

This PR should be fine for now I guess but it got me wondering about new code in the new message store...

rudigiesler:
Just the one comment from me, looks good otherwise. I remember there being a good reason why we have to have the complex iterations, I just can't remember what that reason is.

hodgestar:
Could we not either:
- fix the message store methods (i.e. `batch_inbound_count`, `batch_inbound_keys` and friends) to use pagination and generators?
- add nicer iteration support to index objects?

JayH5:
@hodgestar: the new message store I'm working on uses pagination for all the APIs (at least where it makes sense). I believe that is the reason for this change.

hodgestar:
@jamie Woot. Does it include easy ways to get counts and iterator over keys?

JayH5:
@hodgestar still pretty similar to the existing pagination scheme afaik. The pages don't do anything fancy. See [here](https://github.com/praekelt/vumi-message-store/blob/develop/vumi_message_store/riak_backend.py#L349-L397) for the pagination implementation. And [here](https://github.com/praekelt/vumi-message-store/blob/develop/vumi_message_store/interfaces.py#L225-L342) for the interfaces to the message store.

hodgestar:
Any reason not to implement `__iter__` and `.count()` for index pages?

jerith:
> Could we not either:
> - fix the message store methods (i.e. `batch_inbound_count`, `batch_inbound_keys and friends`) to use pagination and generators?

I'm not touching the old message store, because we're trying to move to the new one. This PR is intended as a self-contained step in that transition.

> - add nicer iteration support to index objects?

Do you mean adding `__len__` and friends? If so, we can definitely do that -- I mostly didn't because I was sticking to the API of the underlying object from the client library.

If you mean having a way to iterate over index pages without the `page = page.next_page()` thing, then the best we can do is the queue-based setup we have in `go-contacts-api`. (We've already had that discussion and nothing's changed since then.)

jerith:
> Any reason not to implement `__iter__` and `.count()` for index pages?

Index pages already have `__iter__`, they just don't have the non-iterator sequence methods. We can add those, but I'd rather do that in the new message store and leave the old one as-is.

hodgestar:
:+1:

Apologies for the PR noise. My concern is largely that `walk_all_keys_from_index_page(...)` and `count_keys_from_index_page(...)` have separate implementations almost everywhere we use index pages, which is a bit of a mess. If that is being addressed in the new message store and this helps us get there, then I'm happy to wait for that. :)

jerith:
> I only worry that there are too many different (but similar) implementations of iterating through paginated results. Can there be One Way?
> 
> The code I'm going through atm has one of the more complicated implementations (see here).

Possibly, but my attempts have either been too limiting or no real improvement. I'm open to suggestions, though.

> And when I see code like this, I wonder if maybe the index pages could be smarter? Maybe implement `__len__()`?

See my previous comments. :-)

hodgestar:
P.S. by `__iter__` I meant an iterator over all results, not over the
results of a single page. Happy for that to be under `.results()` or
`.keys()` or something though instead.

P.P.S. I'm aware that such a thing is a bit tricky because of the issue of
getting deferreds back.

jerith:
> My concern is largely that `walk_all_keys_from_index_page(...)` and `count_keys_from_index_page(...)` have separate implementations almost everywhere we use index pages, which is a bit of a mess.

The former is a problem because of Python's poor async support at the language level. (I believe Python 3 addresses this somewhat, which doesn't help us at the moment.)

The latter is actually a deliberate choice on my part. Walking a potentially very large index to count the results can be extremely expensive (many hours for tens of millions of results) and having to walk the result set manually makes that cost explicit. The cheaper mechanisms for getting those counts (from the cache, for example) are single calls that are much easier to make.

jerith:
Whoops, I've been talking about the index pages as if they were a feature of the message store rather than the underlying Riak stuff. I'll look at adding `__len__` (and maybe some other things) this week.

JayH5:
:+1:
I would like to talk to you about the pagination stuff for the new message store sometime...

rudigiesler:
:+1: from me too.

PR #1309, opened 2015-07-08
Avoid listifying index pages

jerith:
Index pages now support `len()`, so we don't need `len(list())` anymore.


jerith:
The build failure is a problem with Travis. Please don't restart it, I'm reporting it to them and I need the logs to remain available.

JayH5:
yaaay :+1:

PR #1310, opened 2015-07-08
Switch to new message store

jerith:


go/apps/sequential_send/vumi_app.py
```diff
@@ -152,7 +152,9 @@ def send_scheduled_messages(self, conv):
     def send_message(self, batch_id, to_addr, content, msg_options):
         msg = yield self.send_to(
             to_addr, content, endpoint='default', **msg_options)
-        yield self.vumi_api.mdb.add_outbound_message(msg, batch_ids=[batch_id])
+        # XXX: Do we need to do this?
```
hodgestar: Probably not given that we now have middleware that knows how to store messages for conversations. The `batch_id` parameter can likely go too. Perhaps we should open an issue for this for later?

go/apps/http_api/tests/test_vumi_app.py
```diff
@@ -482,7 +482,8 @@ def test_invalid_in_reply_to_with_missing_conversation_key(self):
         # create a message with no conversation
         inbound_msg = self.app_helper.make_inbound('in 1', message_id='msg-1')
         vumi_api = self.app_helper.vumi_helper.get_vumi_api()
-        yield vumi_api.mdb.add_inbound_message(inbound_msg)
+        opms = vumi_api.get_operational_message_store()
```
hodgestar, line 485: What's with the `_operational_`? :)

jerith, line 485: The new message store is split into three pieces:
- The batch manager does stuff related to batches.
- The operational message store reads and writes messages by key only and is used by middleware and routing code.
- The query message store is read-only and is used by everything that needs to list, search, filter, etc.

hodgestar, line 485: Ah, it's the split interface and the other is `_query_`? /me keeps reading.

jerith, line 485: Yes. Ideally, most things only need to use one of the three. (Tests quite often need an extra operational store to create messages that the query store looks at.)

go/base/management/commands/go_migrate_conversations.py
```diff
@@ -133,11 +136,10 @@ class SplitBatches(Migration):
...
     def applies_to(self, user_api, conv):
-        mdb = user_api.api.mdb
-        tag_keys = mdb.current_tags.index_keys('current_batch', conv.batch.key)
-        if tag_keys:
-            return True
-        return False
+        batch_manager = user_api.api.get_batch_manager()
+        batch = batch_manager.get_batch(conv.batch.key)
+        current_tags = [batch_manager.get_tag_info(tag) for tag in batch.tags]
+        return any(ct.current_batch.key == batch.key for ct in current_tags)
```
hodgestar, line 142: I'm guessing we can consider deprecating this code because we no longer have any batches like this?

go/base/tests/test_go_migrate_conversations.py
```diff
@@ -133,9 +134,11 @@ def test_migrate_models(self):
             self.assertEqual(conv.name, loaded_conv.name)
 
     def setup_fix_batches(self, tags=(), num_batches=1):
-        mdb = self.user_api.api.mdb
+        batch_manager = self.user_api.api.get_batch_manager()
+        opms = self.user_api.api.get_operational_message_store()
         msg_helper = GoMessageHelper()  # We can't use .store_*(), so no mdb.
```
JayH5: "mdb" in the comment here...

go/base/management/commands/go_migrate_conversations.py
jerith, line 142: Yes. I don't want to make that kind of change in this PR, though.

go/vumitools/api.py
```diff
@@ -585,6 +584,21 @@ def from_config_async(cls, config, command_publisher=None,
         sender = AsyncMessageSender(command_publisher)
         returnValue(cls(manager, redis, sender, metric_publisher))
 
+    def get_batch_manager(self):
+        from vumi_message_store.message_store import MessageStoreBatchManager
```
JayH5, line 586: What's the advantage to doing the import here?

go/base/tests/test_go_migrate_conversations.py
jerith: Thanks.

go/vumitools/app_worker.py
```diff
@@ -211,53 +212,25 @@ def _find_outboundmessage_for_event(self, event):
...
     _EVENT_OUTBOUND_CACHE_KEY = "outbound_message_json"
 
-    def _get_outbound_from_event_cache(
-            self, event, key=_EVENT_OUTBOUND_CACHE_KEY):
-        """ Retrieve outbound message from the cache on an event.
-
-        :type event:
-            TransportEvent
-        :param event:
-            Event to look up the outbound message on.
```
hodgestar, line 221: We appear to have lost the param and returns part of the docstring for these two methods.

hodgestar, line 221: Also, these changes seem a bit unrelated -- just checking that that assumption is correct?

go/base/management/commands/go_migrate_conversations.py
hodgestar, line 142: Perhaps let's make a ticket? Always keen to delete code that we'd otherwise have to maintain.

go/vumitools/api.py
jerith, line 586: I did it that way because I wanted to limit the scope of things that touched the message store while I was juggling both versions, but that's not really relevant anymore.

Longer-term, I'm not sure access to the message store (especially the operational message store) really belongs in `VumiApi` and I'd like to avoid exposing it in places I don't have to.

hodgestar:
Left some comments but generally looks good to me.

JayH5:
lgtm too

go/vumitools/app_worker.py
jerith, line 221: Yes. They're internal methods that are each used exactly once and they don't really need pages of docstring. When I removed the `key` param I also removed the rest to reduce the amount of scrolling I had to do to read the code.

go/vumitools/api.py
```diff
@@ -354,9 +353,10 @@ def _update_tag_data_for_acquire(self, user_account, tag):
         # The batch we create here gets added to the tag_info and we can fish
         # it out later. When we replace this with proper channel objects we can
         # stash it there like we do with conversations and routers.
-        yield self.api.mdb.batch_start([tag], user_account=user_account.key)
+        yield self.api.get_batch_manager().batch_start(
```
JayH5: Minor thing... maybe put the batch manager in a variable here since `get_batch_manager()` creates a new manager (although they're pretty lightweight).

go/vumitools/api.py
```diff
@@ -424,7 +424,7 @@ def release_tag(self, tag):
         except ValueError, e:
             log.error("Tag not allocated to account: %s" % (tag,), e)
         else:
-            tag_info = yield self.api.mdb.get_tag_info(tag)
+            tag_info = yield self.api.get_batch_manager().get_tag_info(tag)
```
JayH5: And here?

go/vumitools/api.py
jerith: Good catch, thanks.

JayH5:
:+1: from me

JayH5:
:+1: from me although I think someone with a bit more vumi-go knowledge needs to read over 654cf8e

go/apps/sequential_send/vumi_app.py
hodgestar: I created #1322 for this.

go/base/management/commands/go_migrate_conversations.py
hodgestar, line 142: I created #1323 for this.

hodgestar:
I left some comments about the message rate counting, but otherwise looks good.

jerith:
> How do we end up with timestamps that are less than start if we specify start as the start of the range to retrieve messages for?

Old code that I forgot to remove. Thanks for picking that up.

(I was initially attempting to match the behaviour of the previous implementation, which started counting from the most recent message rather than the current time.)

> How big is the default index page currently? Perhaps we should specify an explicit value in this method? A bit concerned that any busy campaign will do well over one page per five minutes.

The default index page is 1k results. When we were using the batch info cache, we stored up to 2k keys. The effect of having more keys in the window than we can fit in the page is the same in both cases, but the threshold is a little different.

This information should really come from metrics rather than message store lookups, but that's out of scope for this PR.

hodgestar:
Cool. I think if we just document the 1k limit here that's good enough for
now.

hodgestar:
:+1: from me on what's here currently.

PR #1313, opened 2015-07-13
Remove Django dependency from vumitools

jerith:
We use `django.utils.datastructures.SortedDict` in `go.vumitools.api` which pulls Django into places it shouldn't be.

### DIFF
--- a/go/conversation/forms.py
+++ b/go/conversation/forms.py
@@ -10,8 +10,9 @@ class ConversationDetailForm(forms.Form):
 class NewConversationForm(ConversationDetailForm):
     def __init__(self, user_api, *args, **kwargs):
         super(NewConversationForm, self).__init__(*args, **kwargs)
+        apps = [v for k, v in sorted(user_api.applications().iteritems())]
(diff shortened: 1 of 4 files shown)


go/conversation/forms.py
```diff
@@ -10,8 +10,9 @@ class ConversationDetailForm(forms.Form):
 class NewConversationForm(ConversationDetailForm):
     def __init__(self, user_api, *args, **kwargs):
         super(NewConversationForm, self).__init__(*args, **kwargs)
+        apps = [v for k, v in sorted(user_api.applications().iteritems())]
```
rudigiesler: Can we not make this a generator instead of a list, since we're turning it into a list below?

rudigiesler:
Just the one comment (which would apply to all the instances where that pattern was used).

go/conversation/forms.py
jerith: We need the whole source list in-memory for sorting anyway, but making it a generator won't hurt.

rudigiesler:
:+1:

PR #1314, opened 2015-07-13
Allow Django app to run from installed package

jerith:
We currently make some assumptions that require a repository checkout at runtime instead of an installed package. This issue aims to fix that.


smn:
:+1:

PR #1311, opened 2015-07-13
Improve layout and readability of contact details page.

hodgestar:
Currently the contact details page seems really oddly and un-compactly laid out to me. Currently it looks like:
screenshot from 2015-07-09 09 20 45
...
screenshot from 2015-07-09 09 21 25
Perhaps we need to allow the non-editable information (extras and subcriptions) to re-flow to a second column?


SteveBarnett:
@hodgestar Agreed. I'll have a quick look today.

hodgestar:
Woot. Tx.

SteveBarnett:
We should:
- remove `col-md-4` from the `div`;
- add a `col-md-4` class to `section.profile`;
- add a `col-md-4` class and `col-md-push-1` class to `section.extras`;
- move the Save changes `button` to outside the div (so that it doesn't wrap funny).

@hodgestar writing this ^^^ rather than doing this because I should be doing numi stuff and I don't want to get told off by @badlard. Also worried about running into my usual getting-my-local-vumi-go-running rabbit hole. :D

hodgestar:
Not ready for review, but I started by cleaning up the structure to look more like idiomatic bootstrap and less special sauce.

PR #1315, opened 2015-07-14
Remove jsbox dependency on Django slugify()

jerith:


hodgestar:
What were the changes from the Django version?

jerith:
Adding a unicode check at the top, stripping out the `mark_safe` stuff that we don't need, and reformatting for PEP-8.

hodgestar:
:+1:

smn:
There's also https://pypi.python.org/pypi/slugify should people be interested in deferring this to a package instead.

hodgestar:
:+1: again from me. Would also be fine with using the slugify package if people prefer.

jerith:
I stuck with a copy of the one from Django because I didn't want to risk introducing subtle variations in slugification.

PR #1316, opened 2015-07-15
Move extract_auth_from_url to non-Django utils

jerith:
This is causing HTTP API workers to import Django stuff unnecessarily.


JayH5:
:+1:

PR #1318, opened 2015-07-15
Make Django dependencies optional

jerith:
This branch splits all Django and database dependencies into an optional `extra` and allows tests to be run on code that doesn't require them without needing to have them installed.


go/apps/dialogue/tests/test_definition.py
```diff
@@ -1,17 +1,19 @@
...
 from go.vumitools.tests.helpers import VumiApiHelper
-from go.apps.dialogue.definition import ConversationDefinition
 from go.apps.dialogue.tests.dummy_polls import simple_poll
+from go.apps.dialogue.definition import ConversationDefinition
 
 
 class TestConversationDefinition(VumiTestCase):
+    @inlineCallbacks
     def setUp(self):
-        self.vumi_helper = self.add_helper(VumiApiHelper(is_sync=True))
```
smn, line 11: Why was this here before?

smn, line 11: Are these now async tests because we've moved Django out?

jerith, line 11: These are now async tests because conversation definitions are used in both sync and async contexts and the sync helper code pulls in Django.

requirements.pip
```diff
@@ -1,3 +1,5 @@
 # Our dependencies are all specified in setup.py.
+# If we're using this requirements file, we almost certainly want Django extras
+# as well.
 
--e .
\ No newline at end of file
+-e .[django]
```
smn, line 5: Is this correct? Not at all familiar with this syntax, what does it do?

smn:
:+1: from me because I think it makes it easier to only run the non-django tests easier.

Do we want to update the matrix to run the django tests and the non-django tests separately? Would that speed things up as they potentially can then happen concurrently?

The longer I've been disconnected from this code base the more I'm starting to think that the maintenance burden of two separate ways of connecting to Redis would have been a cheaper trade off than the kind of testing setup we're finding ourselves in now.

@JayH5 what are your thoughts on that? Considering you're freshest to this code :smile:

requirements.pip
jerith, line 5: It's the syntax for specifying that a package should be installed with a particular set of extras: https://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

jerith:
> Do we want to update the matrix to run the django tests and the non-django tests separately? Would that speed things up as they potentially can then happen concurrently?

About half our test run time is JavaScript stuff and installing dependencies, so splitting the build like that would cost us time rather than saving it. Also, we still need build jobs with complete test coverage so we can report that accurately.

jerith:
> The longer I've been disconnected from this code base the more I'm starting to think that the maintenance burden of two separate ways of connecting to Redis would have been a cheaper trade off than the kind of testing setup we're finding ourselves in now.

This covers both Riak and Redis, and duplicating everything that uses those would be significantly worse than what we have now -- for starters, everything that operates on conversation, contact, or message objects would have to be duplicated.

A better way forward would be to move all the meaningful operations on those things into async-land services and replace the sync stuff with thin wrappers around calls to that. We've been making excruciatingly slow progress in that direction, but the new message store is an example of how we can do it.

smn:
cool, thanks, I'm :+1: anyway.

requirements.pip
smn, line 5: Hah, that's cool, thanks.

smn:
Out of interest, how did the Docker work lead to this work? Do the two relate? Did Docker need this?

jerith:
> Out of interest, how did the Docker work lead to this work? Do the two relate? Did Docker need this?

If our docker container is only running async processes, it has no need for the Django dependencies or access to Django's database, etc. If we can leave these out, we can have smaller container images that are faster to build and deploy.

Furthermore, `psycopg2` can only be used with cpython and there are some unresolved difficulties with `psycopg2cffi` (it randomly segfaults when running tests with pypy) which currently make it impossible to use pypy if that's a hard requirement for `vumi-go`. We already have everything we need to build pypy images, except we can't install `vumi-go` in them.

Unrelated to the docker stuff, Django is full of magic initialisation that happens at import time and our Django apps perform blocking I/O all over the place. In order to guarantee that we're not accidentally (or deliberately, as I discovered in a couple of places) pulling this into code that needs to be async and non-blocking, we need a way to run tests without Django installed.

JayH5:
:+1:

650dfb2 is pretty cool :stuck_out_tongue: 

@smn to be honest, I haven't really seen/done that much with Django so far so the implications of this code are somewhat lost on me. I learnt some interesting things about Python, though...

PR #1319, opened 2015-07-16
Container-based Travis builds

jerith:
They're faster to boot and support caching.


JayH5:
lgtm, my only concern is that I vaguely remember someone somewhere (can't remember exactly) telling me to copy `utils/app.config` to `/etc/riak/app.config` to get Riak working for Vumi/Vumi Go on my local machine. With that file deleted it may make setup harder for people just wanting to try out Vumi.

jerith:
Those configs still exist, but now they live in the `praekelt/riak-for-travis` repo and are better organised. :-)

JayH5:
:+1:
The documentation around setting up Vumi on your local machine could do with some work...

jerith:
Agreed, but that's out of scope for this PR.

PR #1320, opened 2015-07-27
Add command for removing tag pool permissions.

hodgestar:
Currently we don't have a command to do this.

### DIFF
--- a/go/base/management/commands/go_assign_tagpool.py
+++ b/go/base/management/commands/go_assign_tagpool.py
@@ -8,69 +8,100 @@ class Command(BaseGoAccountCommand):
     help = "Give a Vumi Go user access to a certain tagpool"
 
     LOCAL_OPTIONS = [
-        make_option('--tagpool',
+        make_option(
+            '--tagpool',
             dest='tagpool',
             help='The tagpool to give access to'),
-        make_option('--max-keys',
+        make_option(
+            '--max-keys',
             dest='max_keys',
+            default=None,
             help='Maximum number of keys that can be acquired '
-                    '(0 == Unlimited)'),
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

I ended up having to rework some bits of the command more than I might have liked, but the original structure was making things difficult.

go/base/management/commands/go_assign_tagpool.py
```diff
@@ -8,69 +8,100 @@ class Command(BaseGoAccountCommand):
...
+            raise CommandError(
+                'Could not remove permission. Tagpool permission not found.')
+        if options['max_keys'] is not None:
+            raise CommandError(
+                'The max-keys option is may not be set when when deleting a'
+                ' permission.')
+        self.remove_permission(perms[0])
+
+    def remove_permission(self, tagpool_permission):
+        tagpool_permission.delete()
```
rudigiesler, line 85: Why is this wrapped in a function?

rudigiesler:
Just the one comment, otherwise looks good

go/base/management/commands/go_assign_tagpool.py
jerith, line 85: My guess would be for symmetry with `update_permission()` above?

jerith:
:+1:

go/base/management/commands/go_assign_tagpool.py
hodgestar, line 85: Just to follow the pattern of function-for-processing-options and function-for-doing-work pattern that the other commands use.

rudigiesler:
:+1: from me too

PR #1324, opened 2015-07-31
Add support for using the HTTP Retry API to HTTP no-stream conversations.

hodgestar:


hodgestar:
Needs more tests (especially for retry failure case, but also for other paths that can trigger retries).

hodgestar:
Ready for review.

Travis error appears to have been transient.

go/apps/http_api_nostream/tests/test_vumi_app.py
```diff
@@ -272,6 +286,28 @@ def assert_metrics_published(self, metrics, prefix=None):
             [("%s.%s" % (prefix, name), value, agg)
              for name, value, agg in metrics])
 
+    def assert_retry(self, retry, url, method='POST', owner='test-0-user',
+                     intervals=(300, 1800, 3600)):
+        self.assertEqual(retry.method, "POST")
```
justinvdm, line 291: Should this not be `self.assertEqual(retry.method, method)`?

hodgestar, line 291: Perhaps I need to make it clearer somehow, but there are two different
requests that are relevant here -- the request to be retried and request to
the retry server. The `method` parameter refers to the method of the
request to be retried. The method of the call to the retry server is always
'POST.

go/apps/http_api_nostream/tests/test_vumi_app.py
```diff
@@ -818,12 +867,162 @@ def test_post_inbound_message_500_response(self):
...
+        retry.finish()
+
+        with LogCatcher(log_level=logging.WARNING) as lc:
+            yield msg_d
+
+        self.assertEqual(lc.messages(), [
+            "Failed to schedule retry request [account: u'test-0-user'"
+            ", request: %r, response: [500, 'Internal Server Error']]"
+            % (retry_data,),
+        ])
```
justinvdm, line 969: Based on RL discussion, the plan is to change failed retries to reraise the error so that the message doesn't get thrown away.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -271,12 +293,51 @@ def send_event_to_client(self, event, conversation, push_url):
...
+            "request": {
+                "url": url,
+                "method": method,
+                "body": data,
+                "headers": list_headers,
+            },
+        }).encode('utf-8')
+        try:
+            resp = yield http_request_full(
+                self.http_retry_api.encode("utf-8"), data=retry_data,
```
justinvdm: From the config field description, `http_retry_api` is the base url. If that's the case, we should add `/requests/` to the url here.

go/apps/http_api_nostream/tests/test_vumi_app.py
justinvdm, line 291: Oh, woops, sorry, misread.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -43,6 +45,18 @@ class HTTPWorkerConfig(GoApplicationWorker.CONFIG_CLASS):
...
+        " event requests are not retried.",
+        default=None, static=True)
+    http_retry_intervals = ConfigList(
+        "List of delays, in seconds relative to the initial failed request, at"
+        " which retries should be scheduled. Default is 5 minutes, 30 minutes"
+        " and 1 hour.",
+        default=(300, 1800, 3600), static=True)
+    http_retry_timeout = ConfigInt(
+        "How long to wait for a response from the retry API (in seconds).",
+        default=5, static=True)
```
justinvdm, line 59: Wondering if it would be worth adding a timeout test to check that we use this.

justinvdm:
Left a few comments, looks good :)

hodgestar:
We also need test for retrying of event messages (it uses all the same logic as message pushes, but we need to assert that events are actually retried).

go/apps/http_api_nostream/vumi_app.py
hodgestar, line 59: Tests added for default and custom timeouts.

hodgestar:
Ready for re-review.

hodgestar:
P.S. Just added the header construction refactoring we talked about (to make what's in the `try: ...` block in `def push(...):` more sane).

justinvdm:
:+1: once travis is happy. A bit worried that we break something with the header refactoring, though we did need some refactoring there.

hodgestar:
I'm not super worried about the refactoring, because we have tests and nothing in `extract_auth_from_url` or `_push_headers` looks like it should be allowed to throw exceptions into the existing `try ... except ..` handlers.

PR #1325, opened 2015-08-05
Discard pushed messages and events if the retrying is rate limited.

hodgestar:
Currently the HTTP API conversation worker falls over when retrying is rate limited. This makes some sense for other error conditions, but not for rate limiting.

### DIFF
--- a/go/apps/http_api_nostream/vumi_app.py
+++ b/go/apps/http_api_nostream/vumi_app.py
@@ -326,7 +326,12 @@ class NoStreamingHTTPWorker(GoApplicationWorker):
             resp = yield http_request_full(
                 retry_url, data=retry_data, headers=retry_headers,
                 timeout=self.http_retry_timeout)
-            if not (200 <= resp.code < 300):
+            if resp.code == 429:
+                log.warning(
+                    "Retrying is rate limited for account '%s'."
+                    " Discarding HTTP request." % (user_account_key,))
+                returnValue(None)
+            elif not (200 <= resp.code < 300):
                 raise HttpRetryApiError(
                     "HTTP retry failed: %s - %s" % (resp.code, resp.phrase))
         except Exception as err:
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

go/apps/http_api_nostream/vumi_app.py
```diff
@@ -326,7 +326,12 @@ def schedule_push_retry(self, user_account_key, url, method, data,
             resp = yield http_request_full(
                 retry_url, data=retry_data, headers=retry_headers,
                 timeout=self.http_retry_timeout)
-            if not (200 <= resp.code < 300):
+            if resp.code == 429:
+                log.warning(
+                    "Retrying is rate limited for account '%s'."
+                    " Discarding HTTP request." % (user_account_key,))
+                returnValue(None)
```
justinvdm, line 333: I think I might be reading this wrong, but do we maybe need a `return` below this `returnValue` to avoid us reaching the `Succesfully scheduled...` logging at the bottom of the method?

justinvdm:
Left one question, otherwise looks good.

go/apps/http_api_nostream/vumi_app.py
hodgestar, line 333: `returnValue` raises a special exception (which is then caught by the `inlineCallbacks` wrapper and turned into a result for the deferred). One can't have a `return` inside a generator.

hodgestar:
Travis failure seemed unrelated. Have restarted the one build that failed.

justinvdm:
:+1:

go/apps/http_api_nostream/vumi_app.py
justinvdm, line 333: One can have a `return` without an expression in a generator, and for some reason I thought we use that below `returnValue`s to short circuit, but I can't seem to find any place in the codebase where we do that, so I must be remembering wrong. Sorry.

hodgestar, line 333: Np. And yep to `return` without a value.

PR #839, opened 2014-01-17
pypy support

jerith:
This probably needs a little more thought before we decide if we want to land it, but we can update this branch periodically for Travis pypy builds or whatever.

PR #1327, opened 2015-08-24
pypy support (again)

jerith:
This is a successor to #839 which is very old and would be a nightmare to merge.


jerith:
Installing our own pypy is necessary until Travis has pypy 2.6.0 (or later), the `PYPY_GC_MAX_PINNED=0` workaround is necessary until Travis has a newer pypy than 2.6.0. (The bug it avoids has been fixed, but not yet released.)

JayH5:
A few questions:
- Have you tested the performance of `psycopg2cffi`?
- What is the effect of `VUMIGO_TEST_DB=postgres`? It seems to only not be used with the Python 2.7 build?
- So we can't run coverage reports with pypy?
- Do we still need #1318 now?

go/base/management/commands/go_manage_routing_table.py
```diff
@@ -115,8 +115,8 @@ def print_routing_table(self, routing_table):
             return
 
         self.stdout.write("Routing table:\n")
-        for source, values in routing_table._routing_table.iteritems():
+        for source, values in sorted(routing_table._routing_table.iteritems()):
```
JayH5, line 118: What is the reason for this?

jerith:
> - Have you tested the performance of `psycopg2cffi`?

No, but it's documented to be comparable to that of `psycopg2` and the maintainer seems to care about performance. I'm not really worried about that, and writing proper benchmarks to get useful measurements is nontrivial.

> - What is the effect of `VUMIGO_TEST_DB=postgres`? It seems to only not be used with the Python 2.7 build?

That tells the test runner to use postgres instead of sqlite for Django test and to run the tests for the billing API (which are postgres-specific). We leave it out for the `VUMIGO_SKIP_DJANGO=1` run because that skips everything that would use it anyway. I'll see if it makes sense to set it unconditionally in the same place as `VUMITEST_REDIS_DB`, because it serves a similar purpose.

> - Do we still need #1318 now?

Yes. Partial pypy support by avoiding `psycopg2` was only a side benefit of that. The main goals were to avoid contaminating our async code with Django and to avoid a large number of unnecessary dependencies when building containers and such.

go/base/management/commands/go_manage_routing_table.py
jerith, line 118: Dict entries are unordered, and `.iteritems()` returns them in arbitrary order. Historically, we've had a bunch of tests that implicitly rely on ordering and have "accidentally" worked because the ordering is deterministic and the complicated state which determines it hasn't changed between test runs. When we run the tests on pypy, that's no longer true and the tests or this code fail because the routing table entries are displayed in a different order.

Sorting the entries before displaying them gives us a single correct ordering that we can assert on, and also makes it easier to find particular entries in the output.

JayH5:
:+1:

jerith:
This seems to not actually be using the pyenv-installed pypy. I need to figure out why that is.

jerith:
There were more things to fix in here than I thought. :-(

.travis.yml
```diff
@@ -28,7 +31,11 @@ services:
   - redis-server
 
 before_install:
+  # If necessary, set up an appropriate version of pypy.
+  - if [ ! -z "$PYPY_VERSION" ]; then source utils/setup-pypy-travis.sh; fi
+  - if [ ! -z "$PYPY_VERSION" ]; then python --version 2>&1 | fgrep "PyPy $PYPY_VERSION"; fi
```
hodgestar, line 36: What is the `fgrep` here for?

setup.py
```diff
@@ -28,7 +28,6 @@
     'django-debug-toolbar==0.9.4',
     'kombu>=2.5.14,<3.0',
     'librabbitmq==1.5.1',
-    'hiredis==0.1.4',
```
hodgestar, line 31: Historically `hiredis` made quite a big difference to performance.

hodgestar:
Left a couple of questions but generally looks good.

.travis.yml
jerith, line 36: To make sure we're using the correct version. If the `fgrep` fails, Travis sees it as an error and aborts the build.

If we fail to set up pyenv correctly, it's possible for us to still be in the original virtualenv with the wrong Python.

setup.py
jerith, line 31: Historically we were using the blocking Redis client everywhere. As far as I can tell, `txredis` has a separate hiredis-backed client implementation (that we're not using) instead of magically pulling it in if it's available. For the kind of operations we're performing, I'd be very surprised if the protocol parsing overhead was a significant fraction of the overall call latency. That kind of thing makes more of a difference when the reactor is blocked while we make Redis calls to localhost.

.travis.yml
hodgestar, line 36: Woot.

setup.py
hodgestar, line 31: Cool.

hodgestar:
:+1:

jerith:
If Travis is happy with the merge from develop, I'll go ahead and land this. :-)

PR #1070, opened 2015-08-25
Split JS tests into a separate build

jerith:
They often fail spuriously (yay npm) and don't depend on Python version. Putting them in their own build should give us a speed boost and make it easier to rerun just the broken things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,17 +9,19 @@
     }
   },
   "dependencies": {
+    "lodash": "~2.4.1",
+    "q": "~0.9.7",
+    "vumigo_v01": "~0.1.22",
+    "vumigo_v02": "~0.2"
+  },
+  "optionalDependencies": {
     "go-jsbox-location": "^0.2.0",
     "go-jsbox-ona": "^0.1.0",
     "go-jsbox-metrics-helper": "^0.1.0",
     "go-jsbox-xform": "^0.0.1",
     "jed": "~0.5.4",
     "libxmljs": "~0.8.1",
-    "lodash": "~2.4.1",
-    "moment": "~2.6",
-    "q": "~0.9.7",
-    "vumigo_v01": "~0.1.22",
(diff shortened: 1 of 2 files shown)


jerith:
I'm not sure that putting a bunch of things in `optionalDependencies` in `package.json` is the right solution, but it does the job for now. I'm open to other suggestions.

package.json
```diff
@@ -9,17 +9,19 @@
     }
   },
   "dependencies": {
+    "lodash": "~2.4.1",
+    "q": "~0.9.7",
+    "vumigo_v01": "~0.1.22",
+    "vumigo_v02": "~0.2"
+  },
+  "optionalDependencies": {
```
hodgestar, line 17: Hmm. Failure to install optional dependencies doesn't fail the install process, which might be annoying -- https://docs.npmjs.com/files/package.json#optionaldependencies.

Just mentioning this in case others feel it's a show stopper.

justinvdm, line 17: Agreed, I don't think we can consider these deps optional.

jerith, line 17: The only alternatives I can think of are listing them in a file and running `cat extra-node-deps.txt | xargs npm install` or something, or having a second `package.json` with the extra deps it in that we can install separately. I really don't want to leave these in `dependencies`, because installing `libxmljs` takes rather a long time.

(FWIW, these deps _are_ optional in the sense that nothing in vumi-go depends on them -- they're only required by some jsbox apps.)

justinvdm, line 17: Ya, ideally those deps shouldn't be in the vumi-go repo, they're needed in production, not here. A better solution would be to have a separate `package.json` for production with these dependencies in that doesn't live in this repo. Maybe we should do something like that instead? In the mean time, I do think its going to be safer to not make these deps optional here. There's potential for deploys to go bad if dependencies that need to be deployed aren't deployed.

PR #1331, opened 2015-09-02
Documentation improvement: Leading plus in from_addr sample

hodgestar:
The documentation for the HTTP API provides a JSON sample for a message without a leading plus in the from_addr field. However the actual data sent appears to contain the leading plus.

https://github.com/praekelt/vumi-go/blob/develop/docs/http_api.rst


hodgestar:
@jpaidoussi Ready for review.

justinvdm:
:+1:

PR #1289, opened 2015-09-04
Add progress field to billing statement model

justinvdm:
This will allow us to tell whether statement generation for a particular statement.


justinvdm:
Ready for review.

go/billing/tasks.py
```diff
@@ -375,7 +375,8 @@ def generate_monthly_statement(account_id, from_date, to_date):
         title=settings.MONTHLY_STATEMENT_TITLE,
         type=Statement.TYPE_MONTHLY,
         from_date=from_date,
-        to_date=to_date)
+        to_date=to_date,
+        status=Statement.STATUS_STATEMENT_PENDING)
```
justinvdm, line 377: Can't think of a way to test this without hackily spying or mocking things.

jerith, line 377: Monkey-patch `get_transactions()` or `generate_statement_items()` with something that fails and assert that you end up with a pending statement?

justinvdm, line 377: Was hoping to avoid doing something like that, but I guess there is no other way.

justinvdm:
Actually ready for review.

go/billing/migrations/0024_auto__add_field_statement_status.py
```diff
@@ -0,0 +1,161 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding field 'Statement.status'
+        db.add_column(u'billing_statement', 'status',
+                      self.gf('django.db.models.fields.CharField')(default='statement_pending', max_length=20),
```
hodgestar: I wonder if the default here shouldn't be `statement_completed` since all existing statements are presumably complete?

hodgestar:
Left one question, otherwise looks good.

Build failures seem unrelated so I restarted the failed build.

justinvdm:
Ready for re-review, I think.

hodgestar:
:+1: once Travis is happy (I restarted the job).

PR #1337, opened 2015-10-05
Change billing statement generation to not perform heavy postgres operations

justinvdm:
At the moment, the queries needed for generating statements are heavy operations for accounts with many transactions. Since these queries lock the transaction table, this blocks our message processing (since inserting into the transaction table is a part of processing each message). It would help if we made smaller, chunked queries to postgres (similar to our archiving task) and aggregated on the client side.

### DIFF
--- a/go/billing/django_utils.py
+++ b/go/billing/django_utils.py
@@ -24,18 +24,34 @@ class TransactionSerializer(object):
 
 
 class Summaries(object):
+    """
+    Helper class to summarize incrementally given model data.
+
+    The constructor takes in a list of `select` fields and a list of `total`
+    fields. Each summary corresponds to models with the same `select` fields,
+    and results in a summation of those models' `total` fields.
+    """
+
     def __init__(self, select_fields, total_fields):
         self.select_fields = select_fields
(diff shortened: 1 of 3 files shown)


justinvdm:
Ready for review.

rudigiesler:
This is looking good to me. The Summary/Summaries classes were quite difficult to wrap my head around though. I'm wondering if there are some doc strings that we can place to make it a bit easier to understand?

justinvdm:
>  I'm wondering if there are some doc strings that we can place to make it a bit easier to understand?

Sure, will add some docstrings. `Summary` and `Summaries` was my second attempt, and I'm not convinced they are the best abstractions. Would've liked a more functional approach, but couldn't really come up with a nice solution that way. My first attempt made use of `groupby` a bunch of other `itertools` and was maybe more readable, but then I realised that a `groupby` would require us to first sort, which isn't something we can't do.

justinvdm:
Docstrings added, ready for re-review.

rudigiesler:
Docstrings look good, I think it's much clearer now. :+1: when travis is happy.

PR #1344, opened 2015-10-22
Fix routing key in tests

jerith:


hodgestar:
Could you explain a bit why the dummy transport name has to change?

jerith:
Part of my experimentation with simpler publisher machinery involves removing (unused) support for publishing to arbitrary routing keys, which in turn changes the routing key validation to happen once when the publisher is created instead of every message published.

That means the old fake routing key with caps in it will fail validation (which wasn't happening before) and the tests will break.

hodgestar:
:+1:

PR #1340, opened 2015-10-12
Include a button for the new download modal in messages view

justinvdm:

PR #1345, opened 2015-10-23
Hook up time range message downloads

hodgestar:


hodgestar:
Ready for review.

hodgestar:
Okay, maybe ready for review now. ;)

hodgestar:
I'm not sure how https://github.com/praekelt/vumi-go/pull/1340/files#diff-8b9eacc3cdf12b6761950e2895d5cd66R795 passed in #1340. I don't find the "unique people" in the conversation templates anymore?

justinvdm:
> I'm not sure how https://github.com/praekelt/vumi-go/pull/1340/files#diff-8b9eacc3cdf12b6761950e2895d5cd66R795 passed in #1340. I don't find the "unique people" in the conversation templates anymore?

Hrm, that is weird. I remember changing those tests to match the new way we are displaying the uniques, maybe I'm remembering wrong.

hodgestar:
> Hrm, that is weird. I remember changing those tests to match the new way we are displaying the 
> uniques, maybe I'm remembering wrong.

There is a new test -- it looks like the merging just got confused by the deletion of the old test and the creation of a new test above it.

go/conversation/forms.py
```diff
@@ -54,3 +54,35 @@ class ReplyToMessageForm(forms.Form):
...
+            ('csv', 'csv'), ('json', 'json'),
+        ])
+
+    direction = forms.ChoiceField(
+        required=True, initial='inbound',
+        choices=[
+            ('inbound', 'inbound'), ('outbound', 'outbound'),
+        ])
+
+    _LOCALS['date-preset'] = forms.ChoiceField(
```
hodgestar: I should probably just change the field names in the HTML.

hodgestar: @justinvdm Thoughts? How much much do you like the current `date-to`, `date-from`, `date-preset` dashes and how ugly would you find `date_to`, etc?

justinvdm: If it is just the `name` attribute that would be affected, then I think using `date_to`, etc makes sense, thanks.

hodgestar:
Ready for re-review.

Still need to:
- Open the modal if the form is bound.
- Restore the old data if the form is bound.

hodgestar:
Hopefully all done now.

go/base/decorators.py
```diff
@@ -0,0 +1,24 @@
+"""Decorator utilities."""
```
hodgestar: Happy to just delete this. It's a nice idea but it's unrelated to this PR and currently unused. Thoughts?

justinvdm: Hmm, if its unused I vote we remove it, since we'll always have it in our git history if we find we would like to use it.

go/conversation/templatetags/conversation_tags.py
```diff
@@ -56,3 +56,11 @@ def get_reply_form_for_message(message):
         })
     form.fields['to_addr'].widget.attrs['readonly'] = True
     return form
+
+
+@register.filter
+def checked(field, value):
+    #import pdb; pdb.set_trace()
```
justinvdm: Stray debug line.

go/conversation/view_definition.py
```diff
@@ -198,35 +199,40 @@ class ExportMessageView(ConversationApiView):
...
-    def get(self, request, conversation):
-        direction = request.GET.get('direction', 'inbound')
-        if direction not in ['inbound', 'outbound']:
-            raise Http404()
+    def post(self, request, conversation):
+        form = MessageDownloadForm(request.POST)
+        if not form.is_valid():
+            v = self.view_def.get_view(
+                'message_list/', message_download_form=form)
+            # import pdb; pdb.set_trace()
```
justinvdm: Stray debug line.

go/conversation/forms.py
```diff
@@ -54,3 +57,79 @@ class ReplyToMessageForm(forms.Form):
...
+        date_preset = self.cleaned_data.get('date_preset')
+        if date_preset:
+            start_date, end_date = self._parse_date_preset(date_preset)
+            filename_date = date_preset
+        else:
+            start_date = self._date_to_utc(self.cleaned_data.get('date_from'))
+            end_date = self._date_to_utc(self.cleaned_data.get('date_to'))
+            filename_date = self._format_custom_date_filename(
+                start_date, end_date)
+        return start_date, end_date, filename_date
```
justinvdm, line 141: If I'm understanding things properly, we would get filenames like `conv1-outbound-1d.csv`, `conv-outbound-19000101-now` and `conv-outbound-until-now.csv`. If the user was to be doing downloads quite often, it is likely they would end up with multiple files with the same name, and it might be difficult for them to figure out which is which. I think it would be useful if we could include the actual dates in the filenames to avoid this (e.g, instead of `conv1-outbound-1d.csv`, we have `conv-outbound-19000101-19000102`).

The `until-` case is a bit more difficult I guess, since we don't know the start date, but I think filenames like `conv-outbound-until-19000101` are still useful.

I don't think the same holds for the `-now` case, since 'now' is relative. I think it would useful if we could use the current date instead of 'now' (e.g. `conv-outbound-until-19000101` if 'now' is `19000101`). It would mean we would be giving the message export api an end time instead of having it use the actual time the export is started as the end time, but I think people might find it confusing to find messages that arrived after the time they requested the download to start to appear in their message download. If you're strongly against this, I'm okay with us keeping it at 'now' though.

Thoughts?

go/conversation/tests.py
```diff
@@ -641,84 +724,146 @@ def test_aggregates(self):
...
+            '%(conv)s-outbound-7d.csv')
+
+    @mock.patch('go.conversation.forms.datetime.datetime',
+                Summer1969)
+    def test_download_messages_date_preset_30d(self):
+        self.check_download_messages(
+            {'format': 'csv', 'direction': 'outbound', 'date_preset': '30d'},
+            '/message_store_exporter/%(batch)s/outbound.csv?'
+            'start=1969-11-01T12%%3A00%%3A00%%2B00%%3A00',
+            '%(conv)s-outbound-30d.csv')
```
justinvdm: Do we need tests for the custom date range cases?

go/conversation/templates/conversation/message_list.html
```diff
@@ -243,17 +219,32 @@ <h3 class="text-center">Send statistics</h3>
...
+                            </div>
+                            {% endif %}
+
                             <div class="well collapse date-custom-container">
                                 <div class="input-daterange input-group" data-provide="datepicker" data-date-format="dd/mm/yyyy">
-                                    <input type="text" class="input-sm form-control" value="{% now "d/m/Y" %}" name="data-from">
+                                    <input type="text" class="input-sm form-control" value="{{ form.date_from.value|date:"d/m/Y" }}" name="date_from">
                                     <span class="input-group-addon">to</span>
-                                    <input type="text" class="input-sm form-control" value="{% now "d/m/Y" %}" name="date-to">
+                                    <input type="text" class="input-sm form-control" value="{{ form.date_to.value|date:"d/m/Y" }}" name="date_to">
```
justinvdm, line 216: Will the to and from form date fields default to the current date?

justinvdm:
Other than the non-input-validation error handling stuff that we discussed over slack and the filenames comment, just a few minor things. Looks good.

justinvdm:
:+1:! (when travis is happy)

hodgestar:
Oops. Missed a file in the last commit.

PR #1346, opened 2015-10-29
Post message export form to the correct URL

hodgestar:
Current we post it to the message list display URL, instead of the message export URL.


hodgestar:
Ready for review.

justinvdm:
:+1:, sorry for not catching this.

PR #1348, opened 2015-11-05
Add account task scheduler

hodgestar:
The immediate use case is scheduling conversation actions such as a bulk send or dialogue send.

This issue is for the models and task executor.


rudigiesler:
I'm not sure if I'm going the best way about this, but I've committed my first attempt. Ready for review.

hodgestar:
Looks good to me.

I think we might need to guard against tasks being called multiple times by recording when a pending task starts being processed. I'll add that now.

go/scheduler/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+
+@task()
+def perform_task(pending_id):
+    """ Perform a task. """
+    pending = PendingTask.objects.get(id=pending_id)
+    if pending.started_timestamp is None:
+        pending.started_timestamp = datetime.datetime.utcnow()
+        pending.save()
+    else:
+        return
```
rudigiesler, line 23: I'm not sure here if we should delete the task or not.

If we don't delete it, then we could get into the situation where every time we try to run the task, we load it from the database and then just return.

If we do delete it, then we could run into the trace condition of getting an error because we delete it before the original task runner finishes, and then the original task runner tries to delete it, but it doesn't exist.

rudigiesler:
@hodgestar Just one thing that I'm not sure about, but ready for review.

go/scheduler/tasks.py
hodgestar, line 23: What if we moved the `started_timestamp` onto the `Task` model instead of the `PendingTask` model? Then we would retain the information on when the task was run even after the pending task was deleted.

We probably need to handle the `pending_id` no longer existing either way (and treat it as if the task had already run).

We should probably also log a message saying that the task had already been executed so we can see what happened (using celery's task logging helper).

hodgestar:
We'll also need to add a migration for the new field once we're happy with which model it's on.

rudigiesler:
Okay I think it's ready for another review. I'll create the migrations once we're happy with the models.

hodgestar:
Looks good!

rudigiesler:
@hodgestar Okay I've (hopefully) added the migrations successfully. Ready for another review.

hodgestar:
:+1:

Thanks for catching the South migration field registration thing. :)

PR #1349, opened 2015-11-09
Log download form error data directly in log message

hodgestar:
Currently we include it in `extra`, but that doesn't get logged to the console (maybe Django's error logger is suboptimal?).


hodgestar:
Ready for review.

justinvdm:
:+1:

justinvdm:
(when Travis is happy)

PR #1350, opened 2015-11-10
Update bootstrap

justinvdm:
We are seeing bootstrap#10206 with the message download modal's message direction form field.


go/base/static/css/vumigo.css
```diff
@@ -14,8 +14,8 @@ button {
   pointer-events: auto;
 }
 .main-container {
+  width: auto;
   min-width: 650px;
-  max-width: none;
```
justinvdm, line 18: Our max width rule was overriding a bootstrap rule that must've changed. Since it was no longer overriding anything, this rule became a no-op and we were left with a narrower width. `width: auto` brings back that overriding.

justinvdm:
Ready for review.

hodgestar:
:+1:

justinvdm:
@hodgestar thanks, just noticed another broken thing, fixing.

justinvdm:
Ready for re-review.

`.list-group-item.active` seemed to have stopped working for the 3.0 releases (or have been removed intentionally?). 3.3.5 (the most recent release in the ~3.0 range) seems to bring it back, and doesn't seem to break anything else, though it is likely that there are earlier versions that fix it.

justinvdm:
Ok, actually ready for review now.

bower.json
```diff
@@ -17,10 +17,10 @@
     "jsplumb": "~1.7.5",
     "lodash": "~2.0.0",
     "backbone-relational": "0.8.6",
-    "bootstrap": "3.0.0",
     "bootbox": "~4.0.0",
     "diamondash": "~0.4",
-    "bootstrap-datepicker": "~1.4.1"
+    "bootstrap-datepicker": "~1.4.1",
+    "bootstrap": "3.0.1"
```
hodgestar: Your PR comment mentioned bootstrap `3.3.5` but this says `3.0.1`. Just checking that this is correct?

hodgestar:
Left one question and restarted some Travis builds that failed for unrelated reasons. Otherwise looks good.

bower.json
justinvdm: Woops, this isn't correct, thanks. I might not have used `--save`, so it is using bootstrap 3.3.5, but isn't being represented in `bower.json`. Fixing.

hodgestar: The version of `bootstrap` in the diff is `3.3.5`, so I'm guessing just this number is wrong?

justinvdm: Fixed, so we were using 3.3.5, `bower.json` just wasn't up to date.

hodgestar:
:+1:

PR #1351, opened 2015-11-16
Add auth bouncer support to Go API.

hodgestar:
Currently the Go API only supports Django session authentication. It would be nice if it also supported our authentication bouncer.


hodgestar:
Ready for review (assuming Travis builds pass).

hodgestar:
/me berates past self for using txjsonrpc.

go/api/go_api/auth.py
```diff
@@ -47,11 +52,76 @@ def requestAvatarId(self, credentials):
         raise error.UnauthorizedLogin()
 
 
+class GoAuthBouncerCredentialFactory(BasicCredentialFactory):
+    implements(ICredentialFactory)
```
jerith: Rather use the `@implementer(ICredentialFactory)` decorator. It's more readable and less magic. :-)

go/api/go_api/tests/utils.py
```diff
@@ -1,5 +1,27 @@
 from mock import Mock, patch
 
+from vumi.tests.utils import MockHttpServer
```
jerith, line 3: Could we use `vumi.tests.fake_connection.FakeHttpServer` instead? It's not quite a drop-in replacement, but it's more deterministic and less prone to random network-related failures.

jerith:
Could you add explanatory docstrings to the new test methods, please? It's much easier to review tests when there's a description of exactly what is being tested.

go/api/go_api/tests/utils.py
jerith, line 3: I can take a stab at this if you like, although it would be nice to spread the knowledge around a bit. Either way, I'm happy to help.

go/api/go_api/auth.py
hodgestar: I'm trying to be consistent with the other uses in the same file. We can change all of them, but I'd rather do that separately if that's what we want to do.

go/api/go_api/tests/utils.py
hodgestar, line 3: Practical question -- how would we make `treq` access the fake agent provided by `FakeHttpServer.get_agent()`? I'm sure it's possible, but not sure it's possible without making a bit of a mess.

Philosophical disagreement -- if a TCP packet randomly fails to traverse the localhost network, your system is very broken. I have yet to see it happen. Spinning up a temporary HTTP server is simple and provides a clean API that matches how the code is used outside of tests. `FakeHttpServer`on the other hand is 300 lines of overriding Twisted agents and protocols, which doesn't fill me with confidence.

hodgestar:
> Could you add explanatory docstrings to the new test methods, please? It's much easier to review tests when there's a description of exactly what is being tested.

Are there particular tests that were confusing? A lot of the tests are very short and it feels like adding a docstring wouldn't add much that the name of the test method and class don't already cover.

jerith:
> Are there particular tests that were confusing? A lot of the tests are very short and it feels like adding a docstring wouldn't add much that the name of the test method and class don't already cover.
- `test_scheme` -- What should the scheme be?
- `test_decode` -- What is being decoded and how do we know if it worked?
- `test_request_avatar_id_no_auth_header` -- What should happen if there is no header?

A test docstring should give me a way to determine whether the test code is actually testing what it's supposed to be testing, which I can't do from just the test names in most cases. For example, a reasonable docstring for `test_request_avatar_id_no_auth_header` could be "When no auth header is present, authentication fails."

go/api/go_api/tests/utils.py
jerith, line 3: I don't have an answer for the practical question, so that's probably "no, we can't easily use it".

As for the philosophy:
- We regularly (I'd guess 1-5% of builds based on having to restart things, I haven't measured properly) have timeouts and breakages in tests that use the localhost network.
- Making sure cleanup happens correctly for the real network case is nontrivial and we have to jump through hoops to pass ports around. I don't consider that to be a clean API at all.
- Almost all the code in `fake_connection` is machinery to manage and hook into connection state in tests. There is very little overriding of existing Twisted code, and everything that does modify existing behaviour checks that it needs to before doing so.
- `FakeServer` and `FakeHttpServer` allow testing of things that cannot be tested using real network connection without replacing quite a lot of Twisted's internals. You can simulate a connection timeout, for example. Or accept a connection and then not do anything with it.
- `fake_connection` is entirely deterministic, so it's possible (although perhaps not as easy as I'd like) to write tests that fail if something doesn't happen instead of waiting for a timeout.

Maybe we should discuss this somewhere other than this PR, though?

go/api/go_api/auth.py
jerith: It's not a very big change. I'd be happy with a single commit that did it across the files this PR touches already, but I'm also okay with not doing it here.

hodgestar:
Test docstring added and switch to `implementer` made.

In the absence of a concrete plan for changing testing away from the MockHttpServer, I'm inclined to leave that as in.

Ready for re-review.

hodgestar:
Travis builds have passed.

go/api/go_api/tests/test_auth.py
```diff
@@ -84,22 +88,129 @@ def test_request_avatar_id_bad_username(self):
...
+        self.add_cleanup(self.auth.stop)
+        yield self.auth.start()
+        self.checker = GoAuthBouncerAccessChecker(self.auth.url)
+
+    def mk_request(self, path='', token=None):
+        request = DummyRequest(path.split('/'))
+        request.path = path
+        if token is not None:
+            request.headers["authorization"] = "Bearer %s" % (token,)
+        return request
```
jerith, line 150: Nothing passes the `path` arg to `mk_request()`. Is it necessary?

jerith:
Are the test docstrings `pep8`-compliant? If so, there's just the `mk_request()` question remaining.

hodgestar:
I think I left the `path` variable in because it was more understandable to
me than:

```
request = DummyRequst([''])
request.path = ''
```

but I can change that if the above is okay?

jerith:
We already have that sort of request creation in other `mk_request()` methods, and the presence of the arg implies that it's being used somewhere.

hodgestar:
Docstrings update to be more Django-like and unused path argument removed.

jerith:
:+1: when Travis is happy.

PR #1353, opened 2015-11-20
Update celery and friends

jerith:
In order to run multiple celery workers on the same machine, we need to set the worker names explicitly. In order to include the hostname in the worker name, we need `celery>=3.1`. In order to use `celery>=3.1`, we need newer `django-celery` and `kombu`.


rudigiesler:
:+1:

PR #1354, opened 2015-11-23
Fix race condition in jsbox contacts tests

jerith:
There are a bunch of places where we're not correctly `yield`ing on a helper method.


hodgestar:
:+1:

rudigiesler:
:+1:

PR #1356, opened 2015-11-24
Session state race condition in app multiplexer

jerith:
Currently, the app multiplexer sends its reply message before it updates its state. In addition to the annoying random build failures this causes, there's also a risk of the session state being inconsistent when the next message arrives.


hodgestar:
:+1: if the Travis errors are unrelated.

smn:
fwiw I've restarted this job 2 times now and it's been failing every time.

jerith:
This failure is in JS tests, which cannot be affected in any way by the changes in this branch unless they're somehow depending on the implementation details of random Python code.

PR #1357, opened 2016-01-07
Fix billing API bootstrap

rudigiesler:
Currently, if you use the script for bootstrapping vumi go, the billing API will fail to start. This is because it's still trying to use an old django admin command, instead of `twistd`. The plan is to remove the old command, and to change the supervisor command to use twistd. Also, we will need to update the readme to have postgres setup steps, as the billing api does not support sqlite.


rudigiesler:
Ready for review.

go/base/management/commands/go_setup_env.py
```diff
@@ -582,7 +582,13 @@ def create_billing_api_supervisord_conf(self):
...
             cp = ConfigParser()
             cp.add_section(section)
-            cp.set(section, "command", "./go-admin.sh runbillingserver")
+            cp.set(
+                section, "command",
+                "twistd -n web --class=go.billing.api.billing_api_resource "
+                "--port=910%(process_num)s")
+            cp.set(
+                section, "environment",
+                "DJANGO_SETTINGS_MODULE='go.settings',PYTHONPATH='.'")
```
hodgestar, line 591: Just checking that the comma between the two environment variables is correct?

hodgestar:
One small question -- otherwise looks good!

go/base/management/commands/go_setup_env.py
rudigiesler, line 591: Yes, according to the docs it's correct: http://supervisord.org/configuration.html#supervisord-section-example

hodgestar, line 591: Woot.

hodgestar:
:+1:

PR #1358, opened 2016-01-11
add support@vumi.org to low credit notifications

KaitCrawford:


KaitCrawford:
@rudigiesler @hodgestar ready for review

go/billing/tasks.py
```diff
@@ -547,7 +548,9 @@ def create_low_credit_notification(
             'reference': notification.id,
         })
 
-    email = EmailMessage(subject, message, email_from, [email_to])
+    # remove duplicate emails from the recipient list
+    recipient_emails = list(set([email_to, support_email]))
+    email = EmailMessage(subject, message, email_from, recipient_emails)
```
rudigiesler: As discussed on slack, we probably want the `support_email` to be a `bcc`. Should be as simple as adding `bcc=support_email` to the list of args, and fixing the relevant test.

KaitCrawford: done :)
The test doesn't need to be fixed as it checks recipients():
"recipients() returns a list of all the recipients of the message, whether they’re recorded in the to, cc or bcc attributes"

hodgestar:
Other than the bcc thing @rudigiesler spotted, looks good to me too.

rudigiesler:
:+1: when travis is happy.

go/billing/tests/test_tasks.py
```diff
@@ -1286,7 +1287,8 @@ def test_credit_cutoff_email_sent(self):
         self.assertEqual(len(mail.outbox), 1)
         [email] = mail.outbox
 
-        self.assertEqual(email.recipients(), [self.django_user.email])
+        self.assertEqual(email.recipients(),
+                         [self.django_user.email, 'support@vumi.org'])
```
hodgestar, line 1292: `email.bcc` contains the list of bcc'ed addresses so we could add an assert like:

``` python
self.assertEqual(email.bcc, ['support@vumi.org'])
```

in addition to the current assert.

KaitCrawford:
@hodgestar are you happy for me to merge?

hodgestar:
:+1:

PR #1359, opened 2016-01-14
Add bulk send schedule frontend

rudigiesler:


rudigiesler:
@hodgestar Ready for review. Also, @justinvdm if you want to take a look too?

My front-end-foo is not that strong, so please take a close and critical look at what I've done and see that everything is okay.

rudigiesler:
Screenshot (on my machine) of the new button:
image

And of the datetime picker:
image

go/base/static/vendor/eonasdan-bootstrap-datetimepicker/css/bootstrap-datetimepicker.css
```diff
@@ -0,0 +1,373 @@
+/*!
```
hodgestar: Do we need this file in addition to `bootstrap-datetimepicker.min.css`?

go/conversation/templates/conversation/bulk_send.html
```diff
@@ -0,0 +1,102 @@
...
+                </div>
+                <div class="row">
+                    <div class="col-md-12">
+                        <div id="timezone" class="pull-right"></div>
+                    </div>
+                </div>
+            </div>
+
+            <div class="modal-footer">
+                <button class="btn btn-primary" type="submit" id="scheduled_submit">Save</button>
```
hodgestar: Make this label should read `Schedule` rather than `Save`?

go/conversation/templates/conversation/bulk_send.html
```diff
@@ -0,0 +1,102 @@
+{% extends "app.html" %}
+{% load crispy_forms_tags %}
+
+{% block content_title %}{{ action.action_display_name }}{% endblock %}
+
+{% block content_actions_right %}
+<div class="pull-right">
+    <button name="schedule" class="btn btn-primary" value="Schedule" data-toggle="modal" data-target="#schedule-modal">{{ action.action_schedule_verb|default:"Schedule" }}</button>
```
hodgestar: Looking at this, I wonder if we could put this modal into the base action template and just wrap it with `{% if action.action_schedule_verb %}`? It would make it easier to add basic scheduling to other actions (e.g. dialogue push) but harder to customize scheduling if needed and we'd need to document that forms that set `action_schedule_verb` need to have a `scheduled_datetime`.

Thoughts?

go/base/static/vendor/eonasdan-bootstrap-datetimepicker/css/bootstrap-datetimepicker.css
rudigiesler: No we probably don't. You can see here the line that causes that file to be copied across: https://github.com/praekelt/vumi-go/pull/1359/files#diff-0a08a7565aba4405282251491979bb6bR50 . Let me change that to only take the min version.

go/conversation/templates/conversation/bulk_send.html
rudigiesler: I was following the pdf that Ben created, but happy to change it to `Schedule`, since it seems to make more sense.

go/conversation/templates/conversation/bulk_send.html
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    $('#submit').on('click', function(e) {
+        $('#conversation_action').submit();
+    });
+
+    $('#scheduled_submit').on('click', function(e) {
+        if ($('#id_scheduled_datetime').val() === '') {
+            $('#id_scheduled_datetime').val(
+                moment().utc().format("YYYY-MM-DD HH:mm")
```
hodgestar: Does this mean that if they don't pick a date, it runs the action right now? Seems like that might be a bit dangerous?

go/apps/bulk_message/definition.py
```diff
@@ -19,11 +24,29 @@ def check_disabled(self):
...
+                    'action_kwargs': {
+                        'batch_id': self._conv.batch.key,
+                        'msg_options': {},
+                        'content': action_data['message'],
+                        'delivery_class': action_data['delivery_class'],
+                        'dedupe': action_data['dedupe'],
+                    },
+                },
+                scheduled_for=action_data['scheduled_datetime'])
+            task.save()
```
hodgestar: If we do move support for scheduling into the base action, we can add a `schedule_command` method that is similar to `send_command` and that actions will need to call?

go/conversation/templates/conversation/bulk_send.html
hodgestar: Woot. Let's change it to `Schedule` and @badlard can complain if needed. :)

hodgestar:
I think we might currently show a Django message saying something like "Messages sent" when we return to the `show` view. We should probably fix that but it could happen in a separate PR if its invasive.

hodgestar:
Left some comments but looks really good to me.

go/conversation/templates/conversation/bulk_send.html
rudigiesler: Well, by default the date that is selected on the popup is the current datetime, so they would have to look at the date, and then click the button to confirm. I'm not sure if there's a better way of doing it? Maybe we should even take out setting the scheduled datetime, since that will run the task immediately instead of scheduling it.

go/apps/bulk_message/definition.py
rudigiesler: Yes that does sound good.

go/conversation/templates/conversation/bulk_send.html
rudigiesler: Yes, we could put it in a base template. I didn't do it here because I remember us discussing to do it as a custom thing here, and to break it out into a reusable module later. But we can also do it in this PR?

hodgestar: I'm vaguely in favour of this PR if we're agreed on the direction.

go/conversation/templates/conversation/bulk_send.html
hodgestar: Ah, if the current date and time is selected by default, then this looks good.

hodgestar:
Last two commits look good.

rudigiesler:
@hodgestar Ready for review again. I factored out the scheduling things into the base template.

hodgestar:
:+1:

PR #1363, opened 2016-01-20
Better testing for scheduled actions

rudigiesler:
At the moment we just test by making sure that the database endtry is correct.
However, this is prone to errors. We change testing to actually run the
task and call the vumi api, in order to catch errors with the `action_kwargs`
field, which has been error prone.

### DIFF
--- a/go/apps/bulk_message/definition.py
+++ b/go/apps/bulk_message/definition.py
@@ -39,9 +39,7 @@ class BulkSendAction(ConversationAction):
                 'conversation_key': self._conv.key,
                 'action_name': 'bulk_send',
                 'action_kwargs': {
-                    'batch_id': self._conv.batch.key,
-                    'msg_options': {},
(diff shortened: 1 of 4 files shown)


rudigiesler:
Ready for review.

hodgestar:
:+1:

I'm wondering whether we should assert that the created task has the correct `scheduled_time` and `status`? Happy for that to happen in this PR or another one (or not happen if we're happy it's checked enough elsewhere).

And we should perhaps also check that the associate PendingTask is created too?

rudigiesler:
@hodgestar We do still have the tests for `scheduled_for` and `status`: https://github.com/praekelt/vumi-go/pull/1363/files#diff-f54726dc1cc3fc7cb9cf429a223419cfL249 , I didn't remove those. I just removed the `task_data` test because I thought it very unnecessary with the new check, and it's very error prone.

PR #1364, opened 2016-01-20
Move scheduling actions from the Action to the View

rudigiesler:


rudigiesler:
Ready for review.

go/conversation/view_definition.py
```diff
@@ -574,11 +577,32 @@ def post(self, request, conversation):
...
+                self.action.action_display_name,))
+        else:
+            self.schedule_action(action_data, conversation)
+            messages.info(request, 'Action successfully scheduled: %s!' % (
+                self.action.action_display_name,))
         return self._action_done(request, conversation)
 
+    def schedule_action(self, action_data, conversation):
+        scheduled_for = action_data.pop('scheduled_datetime')
+        action_data.pop('display')
```
hodgestar, line 594: What is `action_data['display']`?

hodgestar:
One question, but otherwise looks good!

go/conversation/view_definition.py
rudigiesler, line 594: It comes from here: https://github.com/praekelt/vumi-go/pull/1364/files#diff-a6cac4bf3e6adc22b863f06ebdc14232R565 . I have no idea what it's for though.

hodgestar, line 594: Ah, it's for displaying the details of the action to the user during action confirmation. That means that ignoring it here is the correct thing to do.

hodgestar:
:+1:

PR #1366, opened 2016-01-27
Add actions to scheduled task view

rudigiesler:


rudigiesler:
Ready for review. The javascript isn't in a module yet. I've noticed when doing this view that we might want to slightly change the way that we create to module, so I think it's best to do that in another PR where we change both views to use the same module. Preview of how it looks running locally: 
image

go/scheduler/templates/scheduler/task_list.html
```diff
@@ -5,14 +5,68 @@
...
+});
+
+$('#timezone').text(function() {
+    return 'Timezone: ' + moment.tz.guess();
+});
+
+$('#schedule-modal').on('show.bs.modal', function(event) {
+    var button = $(event.relatedTarget);
+    $('#datechangeform').attr("action", button.data('task-url'));
+    $('#id_scheduled_datetime').val(moment.utc().format("YYYY-MM-DD HH:mm"));
```
hodgestar, line 69: Is there a reason to only do this `on('show.bs.modal')` and not as soon as the page loads?

hodgestar, line 69: Oh, because it depends on which rows button was pressed. :)

hodgestar:
:+1: when Travis is happy.

P.S. I restarted a build failure that looked unrelated.

PR #1368, opened 2016-02-17
Channels with the same tag name are difficult to distinguish

hodgestar:
On the campaign routing screen, channels only display their tag name, not their tag pool. This makes distinguishing channels with the same tag name difficult:

pasted image at 2016_02_17 04_35 pm

### DIFF
--- a/go/api/go_api/api_types.py
+++ b/go/api/go_api/api_types.py
@@ -127,7 +127,8 @@ class ChannelType(Dict):
         uuid = channel.key
         conn = channel.get_connector()
         return {
-            'uuid': uuid, 'tag': (pool, tagname), 'name': tagname,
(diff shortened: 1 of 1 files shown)


hodgestar:
@justinvdm I've taken a very simple approach here (i.e. changed the label returned by the API) and arguably that's useful to all API users, but perhaps I should fix this in the UI rather. Thoughts?

justinvdm:
@hodgestar A little worried about cases where the name gets too long. I think those cases are already quite common, so if we add the tagpool name quite a few names will go outside of their boxes. Maybe we should do your suggestion of fixing this in the UI rather? It seems like the better approach anyways, since this is a presentation thing rather than an API thing.

PR #1369, opened 2016-02-19
Conversation Metrics Middleware

KaitCrawford:


KaitCrawford:
@hodgestar here's what I have so far. tbh I'm not really sure where I'm going with this. I'm not sure how the middleware fits into things.

hodgestar:
What you have so far looks good. Let's discuss what to do next on Monday. :)

KaitCrawford:
@hodgestar, for interest, here's what I was busy with prior to our discussion yesterday

go/vumitools/middleware.py
```diff
@@ -603,6 +604,151 @@ def handle_failure(self, failure, connector_name):
...
+                names are always used for those since those message types are
+                not guaranteed to have a `transport_name` value.
+        """,
+        default="passive", static=True)
+
+    metric_connectors = ConfigList(
+        "List of connector names to fire metrics for. Useful for when"
+        " wrapping dispatchers with many connectors, only a subset of which"
+        " should generate metrics. Defaults to all connectors.",
+        required=False, static=True)
```
hodgestar: We probably don't need `op_mode` (because we publish based on conversation name or router name) or `metric_connectors` (because we're just using the messages to trigger publishing of counts).

go/vumitools/middleware.py
```diff
@@ -603,6 +604,151 @@ def handle_failure(self, failure, connector_name):
...
+
+    def increment_counter(self, prefix, message_type):
+        metric = self.get_counter_metric('%s.%s' % (prefix, message_type))
+        metric.inc()
+
+    def get_acc_key(self, msg):
+        mdh = MessageMetadataHelper(
+            self.vumi_api, msg, conversation_cache=self._conversation_cache)
+        acc_key = yield mdh.get_account_key()
+        returnValue(acc_key)
```
hodgestar: Functions that use `yield` to yield on deferreds need the `@inlineCallbaks` decorator.

go/vumitools/middleware.py
```diff
@@ -603,6 +604,151 @@ def handle_failure(self, failure, connector_name):
...
+        self.increment_counter(prefix, 'messages_sent')
+
+    def fire_outbound_conversation_metrics(self, name, msg):
+        self.fire_outbound_metrics(name, msg)
+
+    @inlineCallbacks
+    def handle_inbound(self, message, connector_name):
+        name = self.get_prefix(message)
+        self.fire_inbound_metrics(name, message)
+        returnValue(message)
```
hodgestar, line 657: One needs to `yield` in an `inlineCallbacks` method (otherwise the function isn't a generator, and `inlineCallbacks` only makes sense with generators).

KaitCrawford:
@rudigiesler @justinvdm If you're bored please take a look at the new middleware and provide feedback.
@hodgestar I know you're busy today. just tagging you to keep you informed

go/vumitools/middleware.py
```diff
@@ -603,6 +604,54 @@ def handle_failure(self, failure, connector_name):
...
+    """
+
+    CONFIG_CLASS = ConversationMetricsMiddlewareConfig
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        # We don't use a VumiApi here because we don't have a Riak config for
+        # it.
+        self.redis = yield TxRedisManager.from_config(
+            self.config.redis_manager)
```
hodgestar: We will need to set up the `conversation_cache` here, with, e.g.
```python
self._conversation_cache = ModelObjectCache(
            reactor, self.config.conversation_cache_ttl
```

go/vumitools/middleware.py
```diff
@@ -603,6 +604,54 @@ def handle_failure(self, failure, connector_name):
...
+
+    def teardown_middleware(self):
+        return self.redis.close_manager()
+
+    @inlineCallbacks
+    def get_conv_key(self, msg):
+        mdh = MessageMetadataHelper(
+            self.vumi_api, msg, conversation_cache=self._conversation_cache)
```
hodgestar: We don't retrieve the conversation, we just use its key, so we can remove `conversation_cache=self._conversation_cache` here.

go/vumitools/middleware.py
```diff
@@ -603,6 +604,54 @@ def handle_failure(self, failure, connector_name):
...
+        conv_key = yield mdh.get_conversation_key()
```
hodgestar: `get_conversation_key()` doesn't return a deferred, so we don't need to yield here (although if we remove the yield, we'll also need to remove the inlineCallbacks decorator and the returnValue).

hodgestar:
The direction is looking good. Next steps are:
1. Write some tests for the new middleware.
2. Write the celery task that will fire the metrics.
3. Write tests for the celery task.

go/vumitools/middleware.py
```diff
@@ -603,6 +604,54 @@ def handle_failure(self, failure, connector_name):
...
+        returnValue(conv_key)
+
+    def record_conv_seen(self, msg):
+        conv_key = self.get_conv_key(msg)
```
rudigiesler: `get_conv_key` will return a deferred, so we need to yield on it before using it.

hodgestar: Yep, except that get_conv_key doesn't need to return a deferred, so it
could also not return one. Welcome to the condundrum of async programming
-- when to return a deferred. :)

KaitCrawford:
@hodgestar @rudigiesler @justinvdm ready for another review please. will start writing tests now

go/vumitools/middleware.py
```diff
@@ -603,6 +604,64 @@ def handle_failure(self, failure, connector_name):
...
+    """
+
+    CONFIG_CLASS = ConversationMetricsMiddlewareConfig
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        # We don't use a VumiApi here because we don't have a Riak config for
+        # it.
+        self.redis = yield TxRedisManager.from_config(
+            self.config.redis_manager)
```
KaitCrawford, line 634: @justinvdm you suggested I use a sub_manager here and in the task but I need an instance of a manager to do that and I'm not sure what instance to use.

go/vumitools/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+    for conv_details in conversation_keys:
+        details = json.loads(conv_details)
+        user_api = get_user_api(details.account_key)
+        conv = get_conversation(details.account_key, details.conv_key)
+        publish_conversation_metrics(user_api, conv)
+
+
+def get_and_reset_recent_conversations():
+    # I'm not sure how to get the correct redis manager from here
+    redis = TxRedisManager.from_config()
```
KaitCrawford: perhaps using sub_managers (as mentioned previously) will help my confusion here?

go/vumitools/middleware.py
```diff
@@ -603,6 +604,73 @@ def handle_failure(self, failure, connector_name):
...
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        # We don't use a VumiApi here because we don't have a Riak config for
+        # it.
+        self.vumi_api = yield VumiApi.from_config_async({
+            "riak_manager": self.config.riak_manager,
+            "redis_manager": self.config.redis_manager,
+        })
+        self.subredis = self.vumi_api.redis.sub_manager("conversationMiddleware")
```
KaitCrawford: I'm not convinced I'm creating or using the submanager correctly

KaitCrawford:
@hodgestar, @justinvdm, @rudigiesler finally ready for review \o/
If someone could point out how to fix the "Unclosed Riak managers found during cleanup" errors I'd appreciate it

go/settings.py
```diff
@@ -314,6 +314,10 @@ def static_paths(paths):
         'task': 'go.scheduler.tasks.poll_tasks',
         'schedule': crontab(minute='*/5'),
     },
+    'update-daily-conversation-metrics': {
+        'task': 'go.vumitools.tasks.send_recent_conversation_metrics',
```
hodgestar: I have a feeling that we should put `tasks` into `go.base` instead of `go.vumitools` because `go.vumitools` is meant to be things that don't depend on Django modules and `tasks` need Django and Celery. Likely it'll work as is and there is some advantage to it being closer to `go.vumitools.middleware`, but I think it's probably best not to break our no-django-in-go-vumitools rule.

go/vumitools/middleware.py
```diff
@@ -603,6 +603,67 @@ def handle_failure(self, failure, connector_name):
...
+    CONFIG_CLASS = ConversationMetricsMiddlewareConfig
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        # We don't use a VumiApi here because we don't have a Riak config for
+        # it.
+        self.vumi_api = yield VumiApi.from_config_async({
+            "riak_manager": self.config.riak_manager,
+            "redis_manager": self.config.redis_manager,
+        })
```
hodgestar: We should probably actually avoid using VumiApi here so that we don't need to set up a RiakManager:
- You can get a base redis object with `redis = yield TxRedisManager.from_config(self.config.redis_manager)`.
- Call `redis.close_manager()` in `teardown_middleware`.
- When constructing the `mdh`, pass `None` for the `vumi_api` (since `.get_conversation_key` and `.get_account_key` don't need a `vumi_api`.

hodgestar: Hopefully this will also solve the riak manager not closed during shutdown issues. :)

go/vumitools/middleware.py
```diff
@@ -603,6 +603,67 @@ def handle_failure(self, failure, connector_name):
...
+    def teardown_middleware(self):
+        return self.vumi_api.close()
+
+    def record_conv_seen(self, msg):
+        mdh = MessageMetadataHelper(
+            self.vumi_api, msg)
+        conv_key = mdh.get_conversation_key()
+        acc_key = mdh.get_account_key()
+
+        # this string should be valid json
```
hodgestar: Might be an idea to explain here why you're manually constructing the JSON (i.e. because we need the values added to the redis set to have exactly the value each time).

go/vumitools/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+
+    for conv_details in conversation_details:
+        details = json.loads(conv_details)
+        user_api = get_user_api(details["account_key"])
+        conv = get_conversation(details["account_key"], details["conv_key"])
+        publish_conversation_metrics(user_api, conv)
+
+
+def get_and_reset_recent_conversations():
+    redis = vumi_api().redis.sub_manager("conversation.metrics.middleware")
```
rudigiesler: This is where you're getting cleanup errors. You're creating a new `VumiApi`, which has its own riak manager, which you're not cleaning up. You can call `api.close()` (where `api` is your api instance) to cleanup the riak and redis managers. (You will need to do this everywhere you call `vumi_api()` and `vumi_api_for_user()`

go/vumitools/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+
+from go.base.command_utils import get_user_by_account_key
+from go.base.utils import vumi_api, vumi_api_for_user
+from go.config import get_conversation_definition
+from go.vumitools.metrics import get_conversation_metric_prefix
+
+
+@task(ignore_result=True)
+def send_recent_conversation_metrics():
+    conversation_details = get_and_reset_recent_conversations()
```
hodgestar: We probably want to restructure this method to just get one VumiApi at the start, and then close it at the end:

``` python
api = vumi_api()
conversation_details = get_and_reset_recent_conversations(api)
for ...:
     ...
     user_api = api.get_user_api(details["account_key"])
     conv = user_api.get_conversation(details["conv_key"])
     ...
api.close()
```

Maybe with a `try: ... finally: ...` block to ensure the `api.close()` call happens.

hodgestar: P.S. This avoids looking up the user and profile in Django (which happens when one calls `get_user_by_account_key`).

go/vumitools/tasks.py
hodgestar: I'm wondering whether we should define `"conversation.metrics.middleware"` as a constant on the ConversationMetricsMiddleware class so we can reference it here (and thus the dependency on the prefixes being the same explicit).

go/vumitools/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+    user = get_user_by_account_key(user_account_key)
+    return vumi_api_for_user(user)
+
+
+def get_conversation(user_account_key, conversation_key):
+    user_api = get_user_api(user_account_key)
+    return user_api.get_conversation(conversation_key)
+
+
+def publish_conversation_metrics(user_api, conversation):
```
hodgestar: We probably need to pass in the `vumi_api` here to ensure that we don't have unnecessary VumiApis flying around (and to avoid having to clean the extra ones up at the end too).

hodgestar: Ideally we need to add a method to the VumiApi that deals with all of this craziness, but probably best left for another day. :|

go/vumitools/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+        publish_conversation_metrics(user_api, conv)
+
+
+def get_and_reset_recent_conversations():
+    redis = vumi_api().redis.sub_manager("conversation.metrics.middleware")
+
+    # makes use of redis atomic functions to ensure nothing is added to the set
+    # before it is deleted
+    redis.rename("recent_coversations", "old_recent_conversations")
+    conversation_details = redis.get("old_recent_conversations")
```
hodgestar: I'd prefer to use `smembers` here, just so its explicit that we expect this to be a set.

go/vumitools/tests/test_tasks.py
```diff
@@ -0,0 +1,136 @@
...
+        subredis = self.redis.sub_manager("conversation.metrics.middleware")
+        subredis.sadd("recent_coversations", conv_details)
+
+        [details] = get_and_reset_recent_conversations()
+
+        # Check response and that redis sets are emtpy
+        self.assertEqual(details, conv_details)
+        self.assertIsNone(subredis.get("recent_conversations"))
+        self.assertIsNone(subredis.get("old_recent_conversations"))
+
```
hodgestar: Could we add a test for the case when the `recent_conversations` set doesn't exist?

hodgestar:
Left some comments, but generally looks awesome!

KaitCrawford:
@hodgestar, @rudigiesler, @justinvdm ready for re-review

go/base/tasks.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_and_reset_recent_conversations(vumi_api):
+    redis = vumi_api.redis.sub_manager("conversation.metrics.middleware")
+
+    # makes use of redis atomic functions to ensure nothing is added to the set
+    # before it is deleted
+    try:
+        redis.rename("recent_coversations", "old_recent_conversations")
+    except:
+        return []
```
KaitCrawford, line 34: My exception-handling-foo is pretty poor so please pay special attention here.
I'm not 100% happy catching all exceptions but it seems fake redis throws a different ResponseError exception to real redis. Should I leave it like this or catch both types of exceptions?

KaitCrawford:
@hodgestar did you remove a comment? I have an email for one about defining the submanager prefix as a constant on the middleware class but I don't see it here.
(just asking to ensure I haven't overlooked anything)

hodgestar:
@KaitCrawford I think the prefix-as-constant comment likely got hidden during the other changes. I didn't delete it. :)

go/base/tasks.py
hodgestar, line 34: Bleh. This is annoying. The best solution I can think of is to sticking the appropriate exception class on the redis_manager so we can do `except redis.RESPONSE_ERROR`, but that requires more changes to core Vumi. For now, I suggest one of:
- Check whether `recent_conversations` exists first (since we shouldn't be calling the task twice simultaneously and only the task should delete it).
- Something messy like:

``` python
except Exception as err:
    if type(err).__name__ != "ResponseError":
        raise
    return []
```

I'm leaning towards the first option, because relying on class names is meh (even though its how the Vumi redis tests check things).

go/base/tests/test_tasks.py
```diff
@@ -0,0 +1,141 @@
+import go.base.utils
+from go.conversation.view_definition import ConversationViewDefinitionBase
+from go.vumitools.conversation.definition import ConversationDefinitionBase
+from go.vumitools.metrics import ConversationMetric
+from go.vumitools.tests.helpers import djangotest_imports
+from go.base.tasks import (
```
hodgestar: You'll need to move the `from go.base.tasks` and `import go.base.utils` imports into the `with djangotest_imports` block below so that the `VUMIGO_SKIP_DJANGO` test run doesn't try to import them.

go/base/tasks.py
KaitCrawford, line 34: I think this might also work:

``` python
from redis.exceptions import ResponseError as redis_response_error
from vumi.persist.fake_redis import ResponseError as fake_response_error
```

and then

``` python
except (redis_response_error, fake_response_error):
```

I also prefer the first method but there is this line in the ticket "The idea will be to only fire the metrics once a day (or hour, or every 100 messages)." If they want to change the regularity at some point do you think the atomic-ness will become a requirement?

KaitCrawford, line 34: actually, the first method is probably fine. I'll do that.

hodgestar, line 34: I'm busy doing the proper thing and adding the `redis.RESPONSE_ERROR` to
Vumi. :)

hodgestar:
The current Travis build error is a genuine error:

``` python
go/base/tests/test_tasks.py:28: in <module>
    'dummy': (DummyConversationDefinition, ConversationViewDefinitionBase),
E   NameError: name 'ConversationViewDefinitionBase' is not defined
============= 1015 passed, 836 skipped, 1 error in 270.93 seconds ==============
```

It's related to the `djangotest_imports`. The body of the `width djangotest_imports(...):` block is not executed when running only the non-Django tests, so you need to change the magic to:

``` python
dummy_classes = [
    'ConversationViewDefinitionBase',
]
with djangotest_imports(globals(), dummy_classes=dummy_classes):
```

which will define a dummy `ConversationViewDefinitionBase` even if the Django is not being imported.

hodgestar:
:+1: once Travis is happy!

hodgestar:
Still :+1: when Travis is happy. :)

PR #1371, opened 2016-03-04
Remove unneeded riak_manager config from ConversationMetricsMiddleware

hodgestar:
The middleware asks for a riak_manager in its config by doesn't use it.


hodgestar:
@KaitCrawford Ready for review.

KaitCrawford:
:+1: annoyed that I missed these. I'm surprised the tests passed considering the riak_manager is flagged as required by the config but not supplied by the tests

hodgestar:
@KaitCrawford The reason the tests pass is a bit complicated. `mk_config(...)` adds in `redis_manager`and `riak_manager` configs, because they're needed for creating the `riak_manager` for the VumiApi, even though our middleware doesn't need it.

PR #1373, opened 2016-03-07
Add unique inbound and outbound address count metrics to default conversation reports.

hodgestar:
The new metrics added in #1369 should be added to the default reports so people can see them easily.


hodgestar:
@KaitCrawford @justinvdm Ready for review in the morning.

Asking for two reviewers because changing this didn't break any tests. Eep.

KaitCrawford:
:+1:
Looking at the tests it seems like they don't check what metrics were displayed which would explain why the tests don't break (I think?)

PR #1375, opened 2016-03-15
Improve conversation middleware

KaitCrawford:
Prevent the conversation middleware from hitting redis for every message.

### DIFF
--- a/go/base/tasks.py
+++ b/go/base/tasks.py
@@ -26,6 +26,9 @@ def get_and_reset_recent_conversations(vumi_api):
     redis = vumi_api.redis.sub_manager(
         ConversationMetricsMiddleware.SUBMANAGER_PREFIX)
 
+    # Reset the local cache for recent conversations
+    ConversationMetricsMiddleware.LOCAL_RECENT_CONVS = set()
+
     # makes use of redis atomic functions to ensure nothing is added to the set
     # before it is deleted
     try:
--- a/go/base/tasks.py
+++ b/go/base/tasks.py
@@ -26,9 +26,6 @@ def get_and_reset_recent_conversations(vumi_api):
     redis = vumi_api.redis.sub_manager(
         ConversationMetricsMiddleware.SUBMANAGER_PREFIX)
 
-    # Reset the local cache for recent conversations
-    ConversationMetricsMiddleware.LOCAL_RECENT_CONVS = set()
-
     # makes use of redis atomic functions to ensure nothing is added to the set
     # before it is deleted
     try:
--- a/go/vumitools/middleware.py
+++ b/go/vumitools/middleware.py
@@ -633,7 +633,7 @@ class ConversationMetricsMiddleware(BaseMiddleware):
             self.config.redis_manager)
         self.redis = self.redis_manager.sub_manager(self.SUBMANAGER_PREFIX)
         self.local_recent_convs = set()
-        self._looper = LoopingCall(self.update_redis_recent_convs)
+        self._looper = LoopingCall(self.reset_local_recent_convs)
         self._looper.start(1800)
 
     def teardown_middleware(self):
@@ -641,11 +641,8 @@ class ConversationMetricsMiddleware(BaseMiddleware):
             self._looper.stop()
         return self.redis_manager.close_manager()
 
-    def update_redis_recent_convs(self):
-        # Note: This set will be emptied by a celery task that publishes
-        # the metrics for conversations we have seen
-        self.redis.sadd(self.RECENT_CONV_KEY, *self.local_recent_convs)
-        self.local_recent_convs = set()
+    def reset_local_recent_convs(self):
+        self.local_recent_convs.clear()
 
     def record_conv_seen(self, msg):
         mdh = MessageMetadataHelper(None, msg)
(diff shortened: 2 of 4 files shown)


KaitCrawford:
@hodgestar first pass ready for feedback
I'm sure I should ping someone else for review but I'm not sure who

go/base/tasks.py
```diff
@@ -26,6 +26,9 @@ def get_and_reset_recent_conversations(vumi_api):
     redis = vumi_api.redis.sub_manager(
         ConversationMetricsMiddleware.SUBMANAGER_PREFIX)
 
+    # Reset the local cache for recent conversations
+    ConversationMetricsMiddleware.LOCAL_RECENT_CONVS = set()
```
hodgestar: This won't actually do anything useful because the middleware is running in an entire different process, probably on another machine. :)

go/vumitools/middleware.py
```diff
@@ -623,6 +623,7 @@ class ConversationMetricsMiddleware(BaseMiddleware):
     SUBMANAGER_PREFIX = "conversation.metrics.middleware"
     RECENT_CONV_KEY = "recent_conversations"
     OLD_RECENT_CONV_KEY = "old_recent_conversations"
+    LOCAL_RECENT_CONVS = set()
```
hodgestar: We probably want to create this on each instance (maybe in `.setup_middleware`) so that each instance has its own set (mostly I'm just very wary of having mutable variables on classes, because people largely expect each instance to be a separate thing).

go/base/tasks.py
hodgestar: Since we can't clear the set here, we'll need to clear it somewhere else. I suggest that each middleware clear its own set. Twisted has a nice helper for running a function periodically -- https://twistedmatrix.com/documents/16.0.0/api/twisted.internet.task.LoopingCall.html.

KaitCrawford:
@hodgestar ready for re-review. I'm not sure if it's the right move keeping the conv keys in memory and only adding them to redis later, like I've done now, but I thought it would reduce the likelihood of something weird happening due to the redis set and the local set being cleared at different times.

hodgestar:
@KaitCrawford Because Twisted is really single threaded it's actually safe to do:

``` python
# in looping call
self.local_recent_convs.clear()
...
# later in record_conv_seen
if conv_details not in self.local_recent_convs:
    self.local_recent_convs.add(conv_details)
    return self.redis.sadd(self.RECENT_CONV_KEY, conv_details)
```

KaitCrawford:
@hodgestar ready for review again.
I'm not sure why the tests are failing since Dummy Request does have an attribute called 'headers'?

go/vumitools/middleware.py
```diff
@@ -631,10 +632,18 @@ def setup_middleware(self):
         self.redis_manager = yield TxRedisManager.from_config(
             self.config.redis_manager)
         self.redis = self.redis_manager.sub_manager(self.SUBMANAGER_PREFIX)
+        self.local_recent_convs = set()
+        self._looper = LoopingCall(self.reset_local_recent_convs)
+        self._looper.start(1800)
```
hodgestar: We should perhaps make this configurable? 1800 seems like a good default value though.

go/vumitools/tests/test_middleware.py
```diff
@@ -1028,37 +1028,47 @@ def setUp(self):
         self.conv = yield self.user_helper.create_conversation(
             u'bulk_message', name=u'Test Conversation', started=True)
 
-    @inlineCallbacks
-    def assert_conv_in_redis(self, mw, msg):
+    def assert_conv_key_stored(self, mw, msg):
```
hodgestar, line 1032: You still need the `@inlineCallbacks`.

go/vumitools/tests/test_middleware.py
```diff
@@ -1028,37 +1028,47 @@ def setUp(self):
...
         value = yield mw.redis.smembers(
             ConversationMetricsMiddleware.RECENT_CONV_KEY)
         self.assertSetEqual(value, set([]))
+        self.assertSetEqual(mw.local_recent_convs, set([]))
 
     @inlineCallbacks
     def test_inbound_message(self):
         mw = yield self.mw_helper.create_middleware()
         msg_helper = GoMessageHelper(vumi_helper=self.mw_helper)
+        self.assert_conv_key_not_stored(mw)
```
hodgestar: Needs a yield.

go/vumitools/tests/test_middleware.py
```diff
@@ -1028,37 +1028,47 @@ def setUp(self):
...
 
-        yield self.assert_conv_not_in_redis(mw)
         [msg] = yield msg_helper.add_inbound_to_conv(self.conv, 1)
         yield mw.handle_inbound(msg, "conn_1")
-        yield self.assert_conv_in_redis(mw, msg)
+        self.assert_conv_key_stored(mw, msg)
```
hodgestar: And here and a few more places.

hodgestar:
Left a few more comments. Looks good though.

KaitCrawford:
@hodgestar comments handled

go/vumitools/middleware.py
```diff
@@ -607,6 +608,9 @@ class ConversationMetricsMiddlewareConfig(BaseMiddleware.CONFIG_CLASS):
 
     redis_manager = ConfigDict(
         "Redis configuration parameters", default={}, static=True)
+    local_store_reset_timer = ConfigInt(
+        "Time in seconds to store conversations keys for locally.",
+        default=1800, static=True)
```
hodgestar: 30 minutes is probably too long given our current "publish metrics every hour", so I think we should change this default to 5 minutes.

hodgestar:
Left one last minor comment. We still need to understand what the test failures are about before this can land though (I'm looking at that a bit now).

hodgestar:
Test failures are the result of a change in Twisted's DummyRequest object in the new Twisted release (16.0). They're being addressed in #1376.

hodgestar:
:+1:

(and I've checked that the build failures are unrelated and addressed by #1376)

PR #1376, opened 2016-03-17
Fix DummyRequest headers retrieval

hodgestar:
Twisted 16.0 changed the way DummyRequest store headers (to match real requests more closely). We need to update our tests to match.


hodgestar:
Ready for review.

Note that the tests no longer work with Twisted < 16.0 but we only test with the latest Twisted in Travis. Not entirely sure this is okay.

hodgestar:
Ready for re-review.

KaitCrawford:
:+1:

PR #1377, opened 2016-03-17
Wrap messages so that action column isn't off screen

hodgestar:
Currently on the message display page the message column isn't wrapped which can push the action column off screen if there are long messages without spaces in the response column:

selection_038

We need to add some CSS like:

``` css
max-width: 30em;
word-wrap: break-word;
```

to cause responses without spaces to reflow nicely (the `30em` is also just a good width for columns of text).


hodgestar:
Ready for review.

@peterp @justinvdm ^^^

hodgestar:
We also need to remove the `Type` column which contains only the conversation `delivery_class` which is deprecated. Leaving that for a separate PR though.

